Issue #23, opened 2017-12-04
crash when uploading large files in CentOS 6.8


seguler:
Some older versions of libgcrypt (used by GnuTLS) seems to crash the application when uploading large files (in multiple threads). Here is the symptom:

$ cp ../largedata/myfile .
cp: closing `./myfile': Software caused connection abort

In blobfuse logs, following error is observed before the segfault:
ath.c:193: _gcry_ath_mutex_lock: Assertion `*lock == ((ath_mutex_t) 0)' failed.

Here is the backtrace on the coredump:
#0  0x00002b3c00293495 in raise (sig=6) at ../nptl/sysdeps/unix/sysv/linux/raise.c:64
#1  0x00002b3c00294c75 in abort () at abort.c:92
#2  0x00002b3c0028c60e in __assert_fail_base (fmt=<value optimized out>,
    assertion=0x2b3c02634761 "*lock == ((ath_mutex_t) 0)", file=0x2b3c02634740 "ath.c", line=<value optimized out>,
    function=<value optimized out>) at assert.c:96
#3  0x00002b3c0028c6d0 in __assert_fail (assertion=0x2b3c02634761 "*lock == ((ath_mutex_t) 0)", file=0x2b3c02634740 "ath.c",
    line=193, function=0x2b3c026347a0 "_gcry_ath_mutex_lock") at assert.c:105
#4  0x00002b3c025f33da in ?? () from /lib64/libgcrypt.so.11
#5  0x00002b3c025f08e0 in ?? () from /lib64/libgcrypt.so.11
#6  0x00002b3c025ec220 in ?? () from /lib64/libgcrypt.so.11
#7  0x00002b3c025ff062 in ?? () from /lib64/libgcrypt.so.11
#8  0x00002b3bff5fab71 in ?? () from /usr/lib64/libgnutls.so.26
#9  0x00002b3bff5fae93 in _gnutls_hmac_fast () from /usr/lib64/libgnutls.so.26
#10 0x000000000044e7b9 in microsoft_azure::storage::hash (to_sign=Unhandled dwarf expression opcode 0xf3
)

seguler:
PR #27 addresses this issue. For now it is available in dev branch. End of December, it will be released as part of the binary.

PR #24, opened 2017-12-06
list attribute cache & fix for the gcrypt multithread issue on older libgcrypt versions

seguler:
Fixing multi-thread issue with older versions of gcrypt that crashes blobfuse:

#3  0x00002b3c0028c6d0 in __assert_fail (assertion=0x2b3c02634761 "*lock == ((ath_mutex_t) 0)", file=0x2b3c02634740 "ath.c",
    line=193, function=0x2b3c026347a0 "_gcry_ath_mutex_lock") at assert.c:105

http://gnu.ist.utl.pt/software/gnutls/manual/html_node/Multi_002dthreaded-applications.html
Fix is to use pthread as the thread mechanism for the mutex locks.


seguler:
Fixes #23

PR #29, opened 2017-12-08
List attribute cache to improve small file perf

seguler:


blobfuse/fileapis.cpp
```diff
@@ -84,10 +84,10 @@ int azs_open(const char *path, struct fuse_file_info *fi)
     auto fmutex = file_lock_map::get_instance()->get_mutex(path);
     std::lock_guard<std::mutex> lock(*fmutex);
 
-    // If the file/blob being opened does not exist in the cache, or the version in the cache is too old, we need to download / refresh the data from the service.
+    // If the file/blob being opened does not exist in the cache, or the version in the cache is too old, or it occupies zero blocks with a size (List attribute cache), we need to download / refresh the data from the service.
```
AlexanderYukhanov: are you sure that's true that ftruncated file doesn't occupy blocks on every filesystem?

blobfuse/directoryapis.cpp
```diff
@@ -94,19 +94,35 @@ int azs_readdir(const char *path, void *buf, fuse_fill_dir_t filler, off_t, stru
...
+                fprintf(stdout, "azs_readdir list blobs failed with error = %d\n", errno);
+            }
+            return 0 - map_errno(errno);
         }
-        return 0 - map_errno(errno);
-    }
 
-    filler(buf, ".", NULL, 0);
-    filler(buf, "..", NULL, 0);
+        // List attriute cache: touch directorySignifier file to note down the last listing time
```
AlexanderYukhanov: s/sttriute/attribute/

blobfuse/directoryapis.cpp
```diff
@@ -142,6 +158,22 @@ int azs_readdir(const char *path, void *buf, fuse_fill_dir_t filler, off_t, stru
                         stbuf.st_nlink = 1;
                         stbuf.st_size = listResults[i].content_length;
                         fillerResult = filler(buf, prev_token_str.c_str(), &stbuf, 0); // TODO: Add stat information.  Consider FUSE_FILL_DIR_PLUS.
+
+                        /// List attribute cache: cache the found file locally without the contents
+                        if(list_attribute_cache == true)
+                        {
+                            off_t length = listResults[i].content_length;
+                            int fd = open((mntPathString+prev_token_str).c_str(), O_RDWR | O_CREAT, S_IRWXU | S_IRWXG);
```
AlexanderYukhanov: do you need to get a lock for this file first?

blobfuse/fileapis.cpp
seguler: Ftruncate fills the with zeros: https://linux.die.net/man/2/ftruncate

Zeroes (file holes) are not supposed to take up space in UNIX systems. Though haven't seen any guarantees in the POSIX spec. Any ideas how we can validate this ? Or is there a better approach ?

blobfuse/directoryapis.cpp
seguler: Hmm good catch. Alternatively I can use O_EXCL flag here, and this would ensure that if some other thread created the same file before this line does, this open will fail with EEXIST. (Explained here: http://man7.org/linux/man-pages/man2/open.2.html) - thoughts ? @AlexanderYukhanov @asorrin-msft

AlexanderYukhanov: may be, but what if some other will to create/delete it just after this line? i think if the introduced a synchronization mechanism - mutex + flock in one place, we need to follow it in other places. otherwise, it will be too difficult to reason about synchronization.

blobfuse/fileapis.cpp
AlexanderYukhanov: i believe it all depends on if file system support sparse files or not. At least some time about tmpfs didn't support sparse files, some ntfs drivers didn't, fat didn't, ext3,4 do. i remember there were filesystems which supported sparse files but still allocated one or two blocks.
if you would like to use this approach, you'll better give requirement for cache file system.

blobfuse/directoryapis.cpp
asorrin-msft: Yes, I was just looking through this, we definitely need to grab the mutex lock.  This may also massively increase the size of the map, we may have to prioritize fixing that

blobfuse/blobfuse.h
```diff
@@ -27,7 +27,7 @@
 
 // Set this to 1 to enable debug output.
 // Prints directly to the console, so this is only useful is you mount in "-f" mode.
-#define AZS_PRINT 0
+#define AZS_PRINT 1
```
AlexanderYukhanov: don't forget to reset to 0 before merge

blobfuse/blobfuse.h
```diff
@@ -37,8 +37,34 @@
 
 using namespace microsoft_azure::storage;
 
-// Internal class used for locking, see fileapis.cpp.
-class file_lock_map;
+// We use two different locking schemes to protect files / blobs against data corruption and data loss scenarios.
+// The first is an in-memory std::mutex, the second is flock (Linux).  Each file path gets its own mutex and flock lock.
+// The in-memory mutex should only be held while control is in a method that is directly communicating with Azure Storage.
+// The flock lock should be held continuously, from the time that the file is opened until the time that the file is closed.  It should also be held during blob download and upload.
```
AlexanderYukhanov: please split the line

blobfuse/blobfuse.h
```diff
@@ -37,8 +37,34 @@
...
+// Blob download should hold the flock lock in exclusive mode.  Read/write operations should hold it in shared mode.
+// Explanations for why we lock in various places are in-line.
+
+// This class contains mutexes that we use to lock file paths during blob upload / download / delete.
+// Each blob / file path gets its own mutex.
+// This mutex should never be held when control is not in an open(), flush(), or unlink() method.
+class file_lock_map
```
AlexanderYukhanov: i put several comments on this class in other CR, do you want me to dup them here as all?

blobfuse/fileapis.cpp
seguler: Now have a sparse file test at the startup. If sparse files are not supported, then the list attribute cache gets disabled.

blobfuse/blobfuse.h
seguler: done

blobfuse/blobfuse.h
seguler: let's address this outside of this PR.

blobfuse/directoryapis.cpp
```diff
@@ -142,6 +159,46 @@ int azs_readdir(const char *path, void *buf, fuse_fill_dir_t filler, off_t, stru
...
+                                }
+                                close(fd);
+
+                                flock(fd, LOCK_UN);
+                            }
+                            else
+                            {
+                                if (AZS_PRINT)
+                                {
+                                    fprintf(stdout, "Lock acquisition failed, errno = %d\n", errno);
```
AlexanderYukhanov: do we need to unlink directorySignifier here?

seguler: Yes!

blobfuse/directoryapis.cpp
```diff
@@ -142,6 +159,46 @@ int azs_readdir(const char *path, void *buf, fuse_fill_dir_t filler, off_t, stru
                         stbuf.st_nlink = 1;
                         stbuf.st_size = listResults[i].content_length;
                         fillerResult = filler(buf, prev_token_str.c_str(), &stbuf, 0); // TODO: Add stat information.  Consider FUSE_FILL_DIR_PLUS.
+
+                        /// List attribute cache: cache the found file locally without the contents
```
AlexanderYukhanov: you are not updating attributes of already cached files, is it expected behavior?

blobfuse/directoryapis.cpp
```diff
@@ -142,6 +159,46 @@ int azs_readdir(const char *path, void *buf, fuse_fill_dir_t filler, off_t, stru
...
+                        if(list_attribute_cache == true)
+                        {
+                            off_t length = listResults[i].content_length;
+
+                            // mutex lock the file path
+                            auto fmutex = file_lock_map::get_instance()->get_mutex(("/" + listResults[i].name).c_str());
```
AlexanderYukhanov: are you locking correct key? is it "a path as seen by user app"?

seguler: Yes - I just verified. Both in azs_open, and in here, the path is exactly same. It is the relative path under the mountpoint

blobfuse/directoryapis.cpp
seguler: You are right. This solution is missing changes on file size that may happen from other nodes... No data corruption, but until you open the file, you keep seeing the outdated file size. I will address this.

seguler: As discussed offline - we decided that upcoming garbage collection feature will address this. There will be no cached files on disk that go beyond the file_cache_timeout_in_seconds

PR #36, opened 2017-12-14
Multithreaded downloading

AlexanderYukhanov:
1. Now the code will download a blob in multiple threads
2. Added errors handling in blob_client_wrapper::download_blob_to_file
3. Added handling of race condition between reader and blob modification on the server
4. Implemented fisrt chunk downloading for getting blob attributes instead of making an additional get_blob_property code


AlexanderYukhanov:
Please note, the perf testing is still in progress. Earlier feedback about the approach is welcome -)

blobfuse/fileapis.cpp
```diff
@@ -96,9 +96,8 @@ int azs_open(const char *path, struct fuse_file_info *fi)
             fprintf(stderr, "Failed to create file or directory on cache directory: %s, errno = %d.\n", mntPathString.c_str(),  errno);
             return -1;
         }
-        std::ofstream filestream(mntPathString, std::ofstream::binary | std::ofstream::out);
         errno = 0;
-        int fd = open(mntPath, O_WRONLY);
+        int fd = open(mntPath, O_CREAT|O_WRONLY, 0770);
```
seguler: This would fail if the file was created already - for the attribute cache, or file cache previously. Right ? Why did you add the O_CREAT option ?

AlexanderYukhanov:
No, it will not,  it will open existing file if it exists.

AlexanderYukhanov:
Reading a large file small chunk by chunk is super inefficient based on the testings. It seems much more efficient to split the file on number of blocks equal to the number of downloaders and download those large blocks in parallel.
Here are the numbers for downloading 100 GB file (average value of 5 samples)
1. curl 34m5s
2. nfs 2 premium drives: 4m40s
3. prefetcher (256 threads, 16Mb chunks): 2m6s
4. blobfuse singlethreaded: 35m50s
5. blobfuse muthithreaded with 9 threads: 7m50s

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
```diff
@@ -10,8 +13,10 @@
 
 namespace microsoft_azure {
     namespace storage {
+
+const unsigned long long DOWNLOAD_CHUNK_SIZE = 16 * 1024 * 1024;
+
         class mempool
-        {
         public:
```
seguler: Missing a bracket before this line

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
```diff
@@ -10,8 +13,10 @@
 
 namespace microsoft_azure {
     namespace storage {
+
+const unsigned long long DOWNLOAD_CHUNK_SIZE = 16 * 1024 * 1024;
```
seguler: spacing

AlexanderYukhanov: not sure what spacing to use -) it's not consistent though this file. would you object auto reformating the file?

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
AlexanderYukhanov: done

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
```diff
@@ -623,108 +630,101 @@ namespace microsoft_azure {
...
                 }
 
+                // Wait for workers to complete downloading.
                 for(size_t i = 0; i < task_list.size(); ++i)
                 {
                     task_list[i].wait();
+                    auto result = task_list[i].get();
+                    // let's report the first encountered error for consistency.
+                    if (0 != result && errcode == 0) {
+                        errcode = result;
```
JasonYang-MSFT: If error happens in downloading, no need to wait for all tasks to finish.

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
```diff
@@ -623,108 +630,101 @@ namespace microsoft_azure {
...
+                }
+                if (-1 == r) {
+                    close(fd);
+                    return;
+                }
+                close(fd);
 
+                // Download the rest.
+                const auto left = length - firstChunk.response().size;
+                const auto chunk_size = std::max(DOWNLOAD_CHUNK_SIZE, (left + downloaders - 1)/ downloaders);
```
JasonYang-MSFT: This may make the chunk size too large. If the chunk size gets too large, we may also change the retry policy in get_chunk_to_stream_sync to resume from offset of left stream instead of start.

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
seguler: Space instead of tabs is what we're aiming to use. Yes to auto reformat! :)

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
AlexanderYukhanov: not sure. we cannot just return an error leaving threads working and there is no cancellation mechanism available. So, the best thing i can do is wait.

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
AlexanderYukhanov: why "too large"? this behaviour will not give bigger chunks than current single threaded download -) since we are not allocating a memory for the chunks, "too large" is not an issue, imho.

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
AlexanderYukhanov: sure, will reformat the file before submitting a merge

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
asorrin-msft: I think the idea is to improve the case of a failure partway through downloading a very large file.  If we get a network issue when we've downloaded 90 MB of a 100 MB chunk, and the issue is temporary or retryable, better to only download the last 10 MB in the retry than the whole 100 MB.
Not a huge concern for me atm

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
```diff
@@ -623,108 +630,101 @@ namespace microsoft_azure {
...
+                        output.close();
+                        if(!chunk.success())
                         {
-                            std::lock_guard<std::mutex> lock(mutex);
-                            ++parallel;
-                            //std::cout << "parallel done: " << parallel << std::endl;
-                            cv.notify_one();
+                           // Looks like the blob has been replaced by smaller one - ask user to retry.
+                           if (constants::code_request_range_not_satisfiable == chunk.error().code) {
+                              return EAGAIN;
```
asorrin-msft: at the moment I think this will propagate up to FUSE.  Might be a good idea to check and retry in this case, either here or in fileapis.cpp

AlexanderYukhanov: no, it will be propagated up to the user giving something like "temporary ..."

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
AlexanderYukhanov: unfortunately retry logic is not implemented anyway - lite retries only on getting error status code and doesn't retry in case of network issues.

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
AlexanderYukhanov: as promised, reformatted the file

seguler:
Works on GCC 4.8.2 now. Great job btw - 400MB/s throughput on the download!

Issue #37, opened 2017-12-15
HTTPS not working on CentOS 7.4


seguler:
It seems there is an issue with use-https option. listing and all operations fail silently when using https.

Workaround is to mount blobfuse with -f option:

blobfuse /path/to/mount -f --use-https=true --tmp-path=/mnt/blobfusetmp --config-file=../connection.cfg

seguler:
This is now fixed in blobfuse 0.2. Also note that https is now the default.

PR #39, opened 2017-12-22
Fixing a bug where you can't use HTTPS in daemon mode in some distros.

asorrin-msft:


seguler:
Validated this on RHEL 7.4. Resolves issue #37

blobfuse/blobfuse.cpp
```diff
@@ -205,10 +205,21 @@ int main(int argc, char *argv[])
...
+            // This can cause an issue, because "It is an error to try to use a 
+            // PKCS#11 crypto module in a process before it has been initialized in that 
+            // process, even if the module was initialized in the parent process."
+            // (https://developer.mozilla.org/en-US/docs/Mozilla/Projects/NSS/Reference/NSS_environment_variables)
+            // To implement this, NSS does an assertion to check for this case.
+            // There are workarounds in NSS and libcurl to re-initialize in a fork'd process, but certain 
+            // builds of libcurl do not trigger this workaround,
+            // and so HTTPS calls in the child process will fail.
+            // Setting this environment variable fixes the issue by disabling  the assertion.
+            char *snf = (char *)"NSS_STRICT_NOFORK=DISABLED";
```
AlexanderYukhanov: Do you know about any side effects? NSS_STRICT_NOFORK=DISABLED just disables the assert but it still true "It is an error to try to use a PKCS#11 crypto module in a process before it has been initialized"

MichaelRiss:
Validated on Fedora 26. Works as root and as normal user.
However, for building I had to additionally cherry-pick commit db6585646b7084f0047883638579d2ca596a3c3c from the master branch, else the fall-through cases throw an error and stop the build process.

asorrin-msft:
Abandoning this change in favor of https://github.com/Azure/azure-storage-fuse/pull/41, just in case there are any unintended side effects of violating the PKCS#11 spec

PR #40, opened 2017-12-28
Purge the cache periodically as files get closed and timed out

seguler:
All tests in test.py pass. Also verified that opened files do not get purged. Setting file-cache-timeout-in-seconds=0 works well as well with simultaneous threads accessing files.


blobfuse/utilities.cpp
```diff
@@ -19,6 +19,87 @@ std::string prepend_mnt_path_string(const std::string path)
...
+                else
+                {
+                    unlink(mntPath);
+                }
+
+                flock(fd, LOCK_UN);
+                close(fd);
+            }
+
+            cleanup.pop_front();
```
AlexanderYukhanov: is it safe to delete elements from cleanup even if we failed to delete a file (e.g. failed to get a lock)?

blobfuse/utilities.cpp
```diff
@@ -19,6 +19,87 @@ std::string prepend_mnt_path_string(const std::string path)
...
+        //if deque is empty, skip
+        if(cleanup.empty())
+        {
+       	    //run it every 1 second
+            usleep(1000);
+            continue;
+        }
+
+        //check if the closed time is old enough to delete
+        if((time(NULL) - cleanup.front().closed_time) > file_cache_timeout_in_seconds)
```
AlexanderYukhanov: i don't think front() is a thread-safe operation

blobfuse/utilities.cpp
AlexanderYukhanov: is not threadsafe

blobfuse/fileapis.cpp
```diff
@@ -439,6 +415,13 @@ int azs_release(const char *path, struct fuse_file_info * fi)
         // It will not release any locks acquired from other calls to open(), in this process or in others.
         flock(((struct fhwrapper *)fi->fh)->fh, LOCK_UN);
         close(((struct fhwrapper *)fi->fh)->fh);
+
+        // store the file in the cleanup list
+        file_to_delete file;
+        file.path = strdup(path);
+        file.closed_time = time(NULL); 
+        cleanup.push_back(file);
```
AlexanderYukhanov: not thread-safe

blobfuse/blobfuse.h
```diff
@@ -37,8 +38,44 @@
...
+    static std::shared_ptr<file_lock_map> _instance;
+    static std::mutex s_mutex;
+    std::mutex m_mutex;
+    std::map<std::string, std::shared_ptr<std::mutex>> m_lock_map;
+};
+
+// deque to age cached files based on timeout
+struct file_to_delete
+{
+    const char *path;
```
AlexanderYukhanov: looks like you are going to leak it, why not to have a destructor or store it in a smart pointer

blobfuse/utilities.cpp
```diff
@@ -19,6 +19,87 @@ std::string prepend_mnt_path_string(const std::string path)
     return str_options.tmpPath + "/root" + path;
 }
 
+// cleanup function to clean cached files that are too old
+void gc_cache()
+{
+
+    while(true){
+
```
AlexanderYukhanov: i do not see a call to sleep for a case when cleanup deque is not empty but nothing to cleanup. without sleep this thread will consume a whole core.

blobfuse/utilities.cpp
```diff
@@ -19,6 +19,87 @@ std::string prepend_mnt_path_string(const std::string path)
...
+                        {
+                            fprintf(stdout, "Failed to lock file %s for clean up\n", cleanup.front().path);
+                        }
+                    }
+                    else
+                    {
+                        // Failed to acquire the lock for some other reason.  We close the open fd, and continue.
+                        if (AZS_PRINT)
+                        {
+                            fprintf(stdout, "Failed to lock file %s for clean up\n", cleanup.front().path);
```
AlexanderYukhanov: don't you want to print errno?

blobfuse/utilities.cpp
```diff
@@ -19,6 +19,87 @@ std::string prepend_mnt_path_string(const std::string path)
...
+                        }
+                    }
+                }
+                else
+                {
+                    unlink(mntPath);
+                }
+
+                flock(fd, LOCK_UN);
```
AlexanderYukhanov: should not flock(fd, LOCK_UN) be under else section otherwise you may unlock a lock which you do not own?

blobfuse/utilities.cpp
seguler: I think so. If we failed to get a lock, means there is another lock on the file, and we should let it go. That another lock will eventually close the file and re-add it to this deque. Tested this and verified.

blobfuse/utilities.cpp
seguler: There is only a single thread working on the front of the deque. What is there to be concerned of ?

blobfuse/utilities.cpp
seguler: Single thread.

blobfuse/fileapis.cpp
seguler: Oops :)

blobfuse/utilities.cpp
seguler: will fix this

blobfuse/utilities.cpp
seguler: added a sleep in the else statement

blobfuse/blobfuse.h
```diff
@@ -37,8 +38,45 @@
...
+};
+
+// deque to age cached files based on timeout
+struct file_to_delete
+{
+    std::string path;
+    time_t closed_time;    
+};
+extern std::deque<file_to_delete> cleanup;
+extern std::mutex deque_lock;
```
AlexanderYukhanov: is not deque_lock is too generic? it sound like we are not going to have any more deques in our code and will protect all of them with the same lock

blobfuse/blobfuse.h
```diff
@@ -37,8 +38,45 @@
...
+void gc_cache();
```
AlexanderYukhanov: look, GC = data, locks, thread, and a method to add an entry for collection.
Doesn't it sound like a class GarbageCollector {... start(); add_item(path)} ?

blobfuse/utilities.cpp
```diff
@@ -19,6 +19,108 @@ std::string prepend_mnt_path_string(const std::string path)
...
+        if(is_empty)
+        {
+            //run it every 1 second
+            usleep(1000);
+            continue;
+        }
+
+
+        //check if the closed time is old enough to delete
+        if((time(NULL) - file.closed_time) > file_cache_timeout_in_seconds)
```
AlexanderYukhanov: can we invert and reduce nesting?

if((time(NULL) - file.closed_time) < file_cache_timeout_in_seconds)
{
   usleep(1000);
   continue;
}

....

PR #41, opened 2017-12-30
Httpsfix2

asorrin-msft:
Changing the creation of the azure_blob_client_wrapper to be in azs_init(), instead of main().  This causes curl_global_init() to be called in the correct process, fixing errors we're seeing when using HTTPS in daemon mode.  This replaces a prior PR, https://github.com/Azure/azure-storage-fuse/pull/39 - due to the concern about unintended consequences of doing the NSS initialization in a separate process as making the HTTPS requests.

### DIFF
--- a/blobfuse/blobfuse.cpp
+++ b/blobfuse/blobfuse.cpp
@@ -120,8 +120,16 @@ int read_config(std::string configFile)
 
 void *azs_init(struct fuse_conn_info * conn)
 {
-    /*cfg->kernel_cache = 1;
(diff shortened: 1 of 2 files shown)


seguler:
Validated this on RHEL 7.4.  @MichaelRiss let us know if this fix doesn't work for you

MichaelRiss:
Yes, this fix also works on Fedora 26, both as root as well as user.
But again, I had to cherry-pick commit db6585646b7084f0047883638579d2ca596a3c3c to make the compilation work.

seguler:
Thanks for testing! We're going to merge all those in a few days, and update the binaries in Linux repos so you can easily install without cloning the repo.

PR #46, opened 2018-01-05
Fix performance for small files

AlexanderYukhanov:


AlexanderYukhanov:
have no clue so far why these changes make so huge difference-) will take a look

AlexanderYukhanov:
Ok, according to a profile the problem was caused by creation of a large temporary string via ostringstream::str() call.

PR #58, opened 2018-01-27
Read account name and key from environment variables

seguler:


seguler:
Closing as it is targeting master

PR #59, opened 2018-01-27
Increasing retry count and adding jitter for resiliency

seguler:


seguler:
Closing as it is targeting master

Issue #57, opened 2018-01-27
data lost during write


AlexanderYukhanov:
Simple repro:
f1 = open('/mnt/hi.txt', 'w')
f2 = open('/mnt/hi.txt', 'w')
f2.close()
... sleep long enough for hi.txt to be removed from cache
f1.write("I am lost")
f1.close()

Expected behavior is hi.txt contains "I am lost", actual - empty file.

AlexanderYukhanov:
If the previous example seems stupid, here is a much more realistic and confusing scenario (caused by the same bug):
I have a long running job which writes a log file. Some time later after job started I want to check its progress by calling 'stat filename' and 'tail filename'. Should work ? Nope, tail will delete the file.

seguler:
Alright this is an issue with multiple file handles. I believe the issue was a regression on blobfuse 0.2.

Let's fix it immediately

seguler:
This is now fixed in blobfuse 0.2.2. Thanks for reporting and identifying the issue Alex.
https://github.com/Azure/azure-storage-fuse/releases/tag/v0.2.3

PR #62, opened 2018-01-28
Fix issue 57 - take shared flock on the file

seguler:
#57

### DIFF
--- a/blobfuse/tests.py
+++ b/blobfuse/tests.py
@@ -1126,24 +1126,22 @@ class TestFuse(unittest.TestCase):
         testFileName = "testFile_new" + str(random.randint(0, 10000))
         testFilePath = os.path.join(self.blobstage, testFileName)
 
-        # 1. test with a new file 2. test with an existing file 
-        for i in range(0,2):
-            repeat = random.randint(1, 10)
-            fd1 = os.open(testFilePath, os.O_WRONLY | os.O_CREAT)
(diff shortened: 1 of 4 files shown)


blobfuse/tests.py
```diff
@@ -1118,6 +1118,32 @@ def test_file_simultaneous_write(self):
...
+        time.sleep(130)
+
+        testData = "random data"
+           
+        for i in range(0, repeat): 
+            os.write(fd1, testData.encode())
+
+        os.close(fd1)
+
+        self.assertEqual(os.stat(testFilePath).st_size, len(testData) * repeat)
```
asorrin-msft: Minor - should be len(testData.encode())

blobfuse/tests.py
```diff
@@ -1118,6 +1118,32 @@ def test_file_simultaneous_write(self):
...
+        testFileName = "testFile_new" + str(random.randint(0, 10000))
+        testFilePath = os.path.join(self.blobstage, testFileName)
+
+        repeat = random.randint(1, 10)
+        fd1 = os.open(testFilePath, os.O_WRONLY | os.O_CREAT)
+        fd2 = os.open(testFilePath, os.O_WRONLY)
+        os.close(fd2)
+       
+        # sleep until cache times out 
+        time.sleep(130)
```
asorrin-msft: This will more than double the length of the test pass.  Please add a # TODO to fix this :)

PR #67, opened 2018-01-30
Fix errors handling during upload

AlexanderYukhanov:
1. Fixed handling of time outed requests
2. In case of a failure, upload must wait for all tasks to complete
3. errno cannot be used to report problems between threads


seguler:
This was already merged, right ? can we close it ?

AlexanderYukhanov:
yes, it's the one you merged already

PR #70, opened 2018-02-01
Reset input stream before retries

AlexanderYukhanov:


AlexanderYukhanov:
Tested by injecting 503 in put response.

PR #73, opened 2018-02-02
 Encode url to fix #51 Files named with spaces not supported + test

### LINKED ISSUE #51: Files named with spaces not supported
cp Sorting\ icons.psd /mnt/mount
cp: cannot stat '/mnt/mount/Sorting icons.psd': Unknown error 400

seguler:


seguler:
All tests pass

azure-storage-cpp-lite/src/storage_url.cpp
```diff
@@ -2,10 +2,95 @@
 
 namespace microsoft_azure {
     namespace storage {
+        bool is_alnum(char ch)
```
AlexanderYukhanov: why not to use curl_easy_escape if we are depending on curllib anyway?

azure-storage-cpp-lite/src/storage_url.cpp
```diff
@@ -2,10 +2,95 @@
...
+            case ',':
+            case ';':
+            case '=':
+                return true;
+            default:
+                return false;
+            }
+        }
+
+        bool is_path_character(char ch)
```
AlexanderYukhanov: what does it mean?

azure-storage-cpp-lite/src/storage_url.cpp
```diff
@@ -2,10 +2,95 @@
...
+        }
+
+        std::string encode_url_path(const std::string& path)
+        {
+            const char* const hex = "0123456789ABCDEF";
+            std::string encoded;
+            for(size_t index = 0; index < path.size(); ++index)
+            {
+                char ch = path[index];
+                if(!is_path_character(ch)
```
AlexanderYukhanov: what about & ?

azure-storage-cpp-lite/src/storage_url.cpp
```diff
@@ -2,10 +2,95 @@
...
+        bool is_query_character(char ch)
+        {
+            return is_path_character(ch) || ch == '?';
+        }
+
+        std::string encode_url_path(const std::string& path)
+        {
+            const char* const hex = "0123456789ABCDEF";
+            std::string encoded;
+            for(size_t index = 0; index < path.size(); ++index)
```
AlexanderYukhanov: foreach?

azure-storage-cpp-lite/src/storage_url.cpp
```diff
@@ -2,10 +2,95 @@
 
 namespace microsoft_azure {
     namespace storage {
+        bool is_alnum(char ch)        
+        {
+            return (ch >= 'A' && ch <= 'Z') || (ch >= 'a' && ch <= 'z') || (ch >= '0' && ch <= '9');
+        }
+
+        bool is_unreserved(char ch)
```
AlexanderYukhanov: unreserved for what? it's not very clear from the name what it's supposed to return

azure-storage-cpp-lite/src/storage_url.cpp
```diff
@@ -2,10 +2,95 @@
...
+        }
+
+        std::string encode_url_query(const std::string& path)
+        {
+            const char* const hex = "0123456789ABCDEF";
+            std::string encoded;
+            for(size_t index = 0; index < path.size(); ++index)
+            {
+                char ch = path[index];
+                if(!is_query_character(ch)
```
AlexanderYukhanov: you can pass it as a predicate to not duplicate so much code

blobfuse/tests.py
```diff
@@ -53,6 +53,18 @@ def test_WriteReadSingleFile(self):
         os.remove(filepath)
         self.assertEqual(False, os.path.exists(filepath))
 
+    def test_WriteReadSingleFileUnicode(self):
+        file1txt = "你好，世界！"
```
AlexanderYukhanov: can you please check special characters as well? someone may like filenames like hello\world&we^are%all~together=Gagarin?perestroika and so on -)?

azure-storage-cpp-lite/src/storage_url.cpp
seguler: I just tried out curl_easy_escape, but it it not better than this. You have to initialize curl in every place, and then free the memory with curl_free...plus conversions from std to c_str etc...

AlexanderYukhanov: up to you -) we can always put it into a helper function. but it's really up to you.

azure-storage-cpp-lite/src/storage_url.cpp
AlexanderYukhanov: btw, we do not escape & in file path and it brakes requests

Issue #22, opened 2017-12-02
Question for connection.cfg for China Azure blob


nonokangwei:
Does this support country sepcified azure, like China Azure?

seguler:
Not at this moment. @JasonYang-MSFT @vinjiang can we implement the support for China and Gov clouds ?

JasonYang-MSFT:
Only need to change suffix for different clouds at https://github.com/Azure/azure-storage-fuse/blob/master/azure-storage-cpp-lite/include/constants.dat#L7

And change the api-version number to match different clouds:https://github.com/Azure/azure-storage-fuse/blob/master/azure-storage-cpp-lite/include/constants.dat#L93

arsenvlad:
In addition to China and Gov, would be great to make it extensible to support additional sovereign clouds like Azure Germany. Each separate cloud has it's own set of endpoints suffixes: 

- Azure US Government Developer Guide describing the AzureUSGovernment endpoints - https://docs.microsoft.com/en-us/azure/azure-government/documentation-government-developer-guide 
- Azure Germany Developer Guide describing the AzureGermanCloud endpoints - https://docs.microsoft.com/en-us/azure/germany/germany-developer-guide
- Azure China 21Vianet Developer Guide describing the AzureChinaCloud endpoints - https://docs.microsoft.com/en-us/azure/china/china-get-started-developer-guide

Azure CLI 2.x uses constants for each "cloud" (e.g. az cloud list -o json) - https://github.com/Azure/azure-cli/blob/06cd9c2160c2f792a3af76f71fb55368c92cf0d6/src/azure-cli-core/azure/cli/core/cloud.py#L201

Azure PowerShell uses constants for each "environment" (e.g. Get-AzureRmEnvironment) - https://github.com/Azure/azure-powershell/blob/0e40f91f5eb32e6f5670d3bd7e08d8db37a6a5be/src/Common/Commands.Common.Authentication.Abstractions/AzureEnvironmentConstants.cs#L36 

Just for reference, Azure Python SDK provides multi-cloud support in https://github.com/Azure/azure-sdk-for-python/blob/master/doc/multicloud.rst defiled in here https://github.com/Azure/msrestazure-for-python/blob/b55563d1f88e0d34ed70288afc1c5f6729160703/msrestazure/azure_cloud.py#L181

seguler:
[...]

PR #85, opened 2018-02-10
Support for non-public clouds

seguler:
Issue #22 - Adds support for non-public clouds via the use of account URI from connection.cfg

Here is connection.cfg template:
```
accountName myaccount
accountKey mykey==
containerName mycontainer
blobEndpoint myaccount.blob.core.windows.net
```

### DIFF
--- a/azure-storage-cpp-lite/src/storage_account.cpp
+++ b/azure-storage-cpp-lite/src/storage_account.cpp
@@ -5,7 +5,7 @@
 namespace microsoft_azure {
     namespace storage {
 
-        storage_account::storage_account(const std::string &account_name, std::shared_ptr<storage_credential> credential, bool use_https, const std::string &domain_suffix)
+        storage_account::storage_account(const std::string &account_name, std::shared_ptr<storage_credential> credential, bool use_https, const std::string &account_uri)
             : m_credential(credential) {
             if (use_https) {
                 append_all("https://");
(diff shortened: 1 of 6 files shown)


seguler:
changed accountUri to blobEndpoint @asorrin

azure-storage-cpp-lite/src/storage_account.cpp
```diff
@@ -5,7 +5,8 @@
 namespace microsoft_azure {
     namespace storage {
 
-        storage_account::storage_account(const std::string &account_name, std::shared_ptr<storage_credential> credential, bool use_https, const std::string &domain_suffix)
+        // TODO: Clean up table queue and file services
+        storage_account::storage_account(const std::string &account_name, std::shared_ptr<storage_credential> credential, bool use_https, const std::string &blob_endpoint)
```
AlexanderYukhanov: oh, what a strange way to populate and handle those variables

seguler: Are you saying 'passing by reference' is odd ? We can address this later... I am only adding blob_endpoint here..

AlexanderYukhanov: no -) i am saying that this class is doing a very simple thing in an unexpectedly complex way -) I ok with your changes.

Issue #99, opened 2018-02-16
cannot put executable for regular user on blobfuse


AlexanderYukhanov:
$ echo 'echo hi' > /mnt/batch/tasks/shared/LS_root/mounts/bfs/script
$ chmod +x /mnt/batch/tasks/shared/LS_root/mounts/bfs/script
$ /mnt/batch/tasks/shared/LS_root/mounts/bfs/script
-bash: /mnt/batch/tasks/shared/LS_root/mounts/bfs/script: Permission denied
$ ls -l /mnt/batch/tasks/shared/LS_root/mounts/bfs/script
-rwxrwx--- 1 root root 8 Feb 16 00:14 /mnt/batch/tasks/shared/LS_root/mounts/bfs/script

At the same time,
$ echo 'echo hi' > /mnt/batch/tasks/shared/LS_root/mounts/bfs/notrunnable
$  /mnt/batch/tasks/shared/LS_root/mounts/bfs/notrunnable
-bash: /mnt/batch/tasks/shared/LS_root/mounts/bfs/notrunnable: Permission denied
$ sudo /mnt/batch/tasks/shared/LS_root/mounts/bfs/notrunnable
hi

seguler:
PR ? :)

seguler:
Can you test the PR I sent ?

seguler:
Fixed in blobfuse 0.3

Issue #98, opened 2018-02-16
blobfuse reports wrong permissions


AlexanderYukhanov:
$ ls -l /mnt/batch/tasks/shared/LS_root/mounts/bfs
total 4
drwxrwx--- 2 root root 4096 Feb 16 00:03 folder

$ echo hi > /mnt/batch/tasks/shared/LS_root/mounts/bfs/hello.txt
alex@8dffb6b3017a4cfe9cb5e88d67803a73000000:~$ ls -l /mnt/batch/tasks/shared/LS_root/mounts/bfs
total 8
drwxrwx--- 2 root root 4096 Feb 16 00:03 folder
-rw-rw-r-- 1 root root    3 Feb 16 00:07 hello.txt

So, file attributes tell that only root has access to folders and files, but in practice any user has full access

seguler:
PR ? :)

seguler:
Test the PR I've sent. It now assigns the correct user as the owner. However if you mount as allow_other, then this is still bypassed because we haven't implemented azs_access yet.

AlexanderYukhanov:
are there any challenges with azs_access?

seguler:
Yes rw-rw is now fixed. Azs_create was creating it with 660.

No challenge for azs_access. It can be implemented by saving ACLs in blob metadata... However I'd prefer to wait for the service side solution for this

seguler:
Fixed in blobfuse 0.3

Issue #95, opened 2018-02-13
Implement File modified time


martinsmmm:
Thanks for making the fuse driver.

Any chance to provide some update about 'if' and 'when' the attributes will be available ? What is the plan ?

We are particularly interested in file modification time.

> Some file system APIs have not been implemented: readlink, symlink, link, chmod, chown, fsync, lock and extended attribute calls.

Thank you!

cvalla:
Hello and thank you for your great job.
Files date attribute are all set to 1 gen 1970 when you mount the azure storage.
On Ubuntu 16.04 Installed blobfuse from the microsoft deb repository yesterday.
mounted on /Azure

ls -latr command output is like:

-rwxrwx---  1 root root  858680 gen  1  1970 FILE1.zip
-rwxrwx---  1 root root 1001214 gen  1  1970 FILE2.zip

seguler:
Oh you are right. I was confused. We implemented timestamp support in this PR #29 but never merged... We should implement this.

seguler:
I have implemented this in PR #102 - I am not sure if this PR will make to a release, but test and let us know here.

One caveat is that files timestamps are retrieved from Blob Last-Modified time, however for directories there is no last-modified time on Blob storage. So I am showing the time of list operation for directories.

martinsmmm:
Hi @seguler / all,
Sorry for slow response - get sick last week. Great news!

I will test it during tomorrow and share results. As per your description this is what we are looking for.

Our use case is: We are using Blobstorage as backend/storage for FTP solution. In some cases we need attributes to know which files (luckily not directories) are new (since 'last time') for processing pipeline (Hello Biztalk).

Thank you,
Martin

martinsmmm:
Hi all,
Tested the PR #102 .

The attributes are showing correctly. Thanks @seguler  

```
root@xxx:/xxx/xxx/xxx/xxxtest# ls -la
total 0
drwxrwx--- 3 root root       4096 Feb 20 03:24 .
drwxrwx--- 3 root root       4096 Feb 20 03:24 ..
-rwxrwx--- 1 root root   10485760 Feb  5 03:45 10MB.zip
drwxrwx--- 3 root root       4096 Feb 20 03:24 380855
-rwxrwx--- 1 root root 3006922752 Feb  9 20:02 box-disk1.vmdk
-rwxrwx--- 1 root root          0 Feb  5 03:45 empty-file.txt
[...]

PR #102, opened 2018-02-18
Correct attributes:uid, gid and modified time

seguler:
Addresses #95, #98, #99 

One caveat is that directory timestamps show the listing time (`now`) since there is no support for dirs in blob storage.

### DIFF
--- a/azure-storage-cpp-lite/include/blob/blob_client.h
+++ b/azure-storage-cpp-lite/include/blob/blob_client.h
@@ -415,7 +415,7 @@ namespace microsoft_azure { namespace storage {
         /// <param name="destPath">The target file path.</param>
         /// <param name="parallel">A size_t value indicates the maximum parallelism can be used in this request.</param>
         /// <returns>A <see cref="storage_outcome" /> object that represents the properties (etag, last modified time and size) from the first chunk retrieved.</returns>
-        storage_outcome<chunk_property> download_blob_to_file(const std::string &container, const std::string &blob, const std::string &destPath, size_t parallel = 9);
+        void download_blob_to_file(const std::string &container, const std::string &blob, const std::string &destPath, chunk_property &returned_properties, size_t parallel = 9);
 
         /// <summary>
         /// Gets the property of a blob.
--- a/azure-storage-cpp-lite/include/blob/blob_client.h
+++ b/azure-storage-cpp-lite/include/blob/blob_client.h
@@ -414,7 +414,8 @@ namespace microsoft_azure { namespace storage {
         /// <param name="size">The size of the data to download from the blob, in bytes.</param>
         /// <param name="destPath">The target file path.</param>
         /// <param name="parallel">A size_t value indicates the maximum parallelism can be used in this request.</param>
-        void download_blob_to_file(const std::string &container, const std::string &blob, const std::string &destPath, size_t parallel = 9);
+        /// <returns>A <see cref="storage_outcome" /> object that represents the properties (etag, last modified time and size) from the first chunk retrieved.</returns>
+        storage_outcome<chunk_property> download_blob_to_file(const std::string &container, const std::string &blob, const std::string &destPath, size_t parallel = 9);
 
         /// <summary>
         /// Gets the property of a blob.
--- a/azure-storage-cpp-lite/include/blob/blob_client.h
+++ b/azure-storage-cpp-lite/include/blob/blob_client.h
@@ -415,7 +415,7 @@ namespace microsoft_azure { namespace storage {
         /// <param name="destPath">The target file path.</param>
         /// <param name="parallel">A size_t value indicates the maximum parallelism can be used in this request.</param>
         /// <returns>A <see cref="storage_outcome" /> object that represents the properties (etag, last modified time and size) from the first chunk retrieved.</returns>
-        void download_blob_to_file(const std::string &container, const std::string &blob, const std::string &destPath, blob_property &returned_properties, size_t parallel = 9);
+        void download_blob_to_file(const std::string &container, const std::string &blob, const std::string &destPath, time_t &returned_last_modified, size_t parallel = 9);
 
         /// <summary>
         /// Gets the property of a blob.
--- a/azure-storage-cpp-lite/include/blob/blob_client.h
+++ b/azure-storage-cpp-lite/include/blob/blob_client.h
@@ -415,7 +415,7 @@ namespace microsoft_azure { namespace storage {
         /// <param name="destPath">The target file path.</param>
         /// <param name="parallel">A size_t value indicates the maximum parallelism can be used in this request.</param>
         /// <returns>A <see cref="storage_outcome" /> object that represents the properties (etag, last modified time and size) from the first chunk retrieved.</returns>
-        void download_blob_to_file(const std::string &container, const std::string &blob, const std::string &destPath, chunk_property &returned_properties, size_t parallel = 9);
+        void download_blob_to_file(const std::string &container, const std::string &blob, const std::string &destPath, blob_property &returned_properties, size_t parallel = 9);
 
         /// <summary>
         /// Gets the property of a blob.
--- a/azure-storage-cpp-lite/include/constants.dat
+++ b/azure-storage-cpp-lite/include/constants.dat
@@ -49,6 +49,7 @@ DAT(header_if_match, "If-Match")
 DAT(header_if_modified_since, "If-Modified-Since")
 DAT(header_if_none_match, "If-None-Match")
 DAT(header_if_unmodified_since, "If-Unmodified-Since")
+DAT(header_last_modified, "Last-Modified")
 DAT(header_origin, "Origin")
 DAT(header_user_agent, "User-Agent")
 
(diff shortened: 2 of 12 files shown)


blobfuse/fileapis.cpp
```diff
@@ -217,7 +214,8 @@ int azs_create(const char *path, mode_t mode, struct fuse_file_info *fi)
     ensure_files_directory_exists_in_cache(mntPathString);
 
     // FUSE will set the O_CREAT and O_WRONLY flags, but not O_EXCL, which is generally assumed for 'create' semantics.
-    res = open(mntPath, fi->flags | O_EXCL, mode);
+    // overwrite mode_t because user should always 770 permission until we have full ACL support
+    res = open(mntPath, fi->flags | O_EXCL, 0770);
```
AlexanderYukhanov: maybe 0777? we always mount with -o allow_other

seguler: I just tested it - these modes does not really matter. Even when it is 770, some other user is able to read, write and execute if you mount it with allow_other option. Only way these permissions get enforced is to mount with -o default_permissions option. Then the kernel starts enforcing these, and you can no longer access the files from other users.

I'll leave these as is until we implement ACL support.

azure-storage-cpp-lite/include/constants.dat
```diff
@@ -49,6 +49,7 @@ DAT(header_if_match, "If-Match")
 DAT(header_if_modified_since, "If-Modified-Since")
 DAT(header_if_none_match, "If-None-Match")
 DAT(header_if_unmodified_since, "If-Unmodified-Since")
+DAT(header_last_modified, "Last-Modified")
```
asorrin-msft: Time will still be wrong with this approach, because when you download a blob into the file cache, the modified time will be the timestamp of the download, not the timestamp on the blob.  Fixing this is dependent upon returning the blob properties from a download, which is on my backlog.

blobfuse/directoryapis.cpp
```diff
@@ -64,6 +64,8 @@ int azs_readdir(const char *path, void *buf, fuse_fill_dir_t filler, off_t, stru
                 {
                     struct stat stbuf;
                     stbuf.st_mode = S_IFDIR | 0770;
+                    stbuf.st_uid = fuse_get_context()->uid;
```
asorrin-msft: Tests please

blobfuse/utilities.cpp
```diff
@@ -303,8 +303,11 @@ int azs_getattr(const char *path, struct stat *stbuf)
     if (strlen(path) == 1)
     {
         stbuf->st_mode = S_IFDIR | 0770; // TODO: proper access control.
+        stbuf->st_uid = fuse_get_context()->uid;
+        stbuf->st_gid = fuse_get_context()->gid;
         stbuf->st_nlink = 2; // Directories should have a hard-link count of 2 + (# child directories).  We don't have that count, though, so we jsut use 2 for now.  TODO: Evaluate if we could keep this accurate or not.
         stbuf->st_size = 4096;
+        stbuf->st_mtime = time(NULL);
```
asorrin-msft: If we're going to address the timing issue, we should decide what to do about directories.  I'm not convinced that "right now" is a good option for a directory timestamp; that could be confused for a correct value, where as '1 Jan 1970' is more clearly wrong.

azure-storage-cpp-lite/include/constants.dat
seguler: Done. download_blob_to_file now returns storage_outcome<chunk_property> which contains last modified time

blobfuse/directoryapis.cpp
seguler: Added tests

blobfuse/utilities.cpp
seguler: Reset the directory timestamps back to 1970

blobfuse/fileapis.cpp
AlexanderYukhanov: The issue is that user will see a permission which doesn't reflect reality.  Exactly as you told in your comment.

azure-storage-cpp-lite/src/blob/blob_client.cpp
```diff
@@ -65,6 +65,7 @@ storage_outcome<chunk_property> blob_client::get_chunk_to_stream_sync(const std:
         property.etag = http->get_header(constants::header_etag);
         property.totalSize = get_length_from_content_range(http->get_header(constants::header_content_range));
         std::istringstream(http->get_header(constants::header_content_length)) >> property.size;
+        property.last_modified = curl_getdate(http->get_header(constants::header_last_modified).c_str(), NULL);
```
AlexanderYukhanov: BTW, what does curl_getdata returns in case of NULL? Should we provide 1970 as a default value for last modified header?

azure-storage-cpp-lite/include/get_blob_request_base.h
```diff
@@ -32,12 +32,13 @@ namespace microsoft_azure {
         public:
             chunk_property()
                :totalSize{0},
-               size{0}
+               size{0}
```
AlexanderYukhanov: if we decided to provide initialization list, let's add last_modified

azure-storage-cpp-lite/include/blob/blob_client.h
```diff
@@ -414,7 +414,8 @@ namespace microsoft_azure { namespace storage {
         /// <param name="size">The size of the data to download from the blob, in bytes.</param>
         /// <param name="destPath">The target file path.</param>
         /// <param name="parallel">A size_t value indicates the maximum parallelism can be used in this request.</param>
-        void download_blob_to_file(const std::string &container, const std::string &blob, const std::string &destPath, size_t parallel = 9);
+        /// <returns>A <see cref="storage_outcome" /> object that represents the properties (etag, last modified time and size) from the first chunk retrieved.</returns>
+        storage_outcome<chunk_property> download_blob_to_file(const std::string &container, const std::string &blob, const std::string &destPath, size_t parallel = 9);
```
AlexanderYukhanov: common, why download_blob_to_file returns chunk_property? let's introduce new type?

blobfuse/fileapis.cpp
```diff
@@ -109,12 +109,18 @@ int azs_open(const char *path, struct fuse_file_info *fi)
...
-            azure_blob_client_wrapper->download_blob_to_file(str_options.containerName, pathString.substr(1), mntPathString);
+            auto outcome = azure_blob_client_wrapper->download_blob_to_file(str_options.containerName, pathString.substr(1), mntPathString);
             if (errno != 0)
             {
                 remove(mntPath);
                 return 0 - map_errno(errno);
             }
+            
+            // preserve the last modified time
+            struct utimbuf new_time;
```
AlexanderYukhanov: formally you need either to memset this variable to 0 or do something like
struct utimbuf new_time = {};
or 
new_time = utimbuf();

azure-storage-cpp-lite/include/blob/blob_client.h
seguler: Because it is the chunk_property. We retrieve these properties from the first chunk. Should I use blob_property instead ?

AlexanderYukhanov: or create a new type containing size, etag and last modification time.

AlexanderYukhanov: There is 2 additional issues:
1. previously storage_output<> was not exposed to blobfuse
2. interface becoming inconsistent - it should either setup errno as it does now or should return strorage_output from all methods or should use exceptions.
Let's decide which approach to use and stay with it? In our current approach the download_file_... should return nothing, etag and last modification time should be output parameters, error should be returned with errno.
I will be happy if you change the interface to throw exceptions or return storage_outcome from each method

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
```diff
@@ -622,12 +622,12 @@ namespace microsoft_azure {
             }
         }
 
-        storage_outcome<chunk_property> blob_client_wrapper::download_blob_to_file(const std::string &container, const std::string &blob, const std::string &destPath, size_t parallel)
+        void blob_client_wrapper::download_blob_to_file(const std::string &container, const std::string &blob, const std::string &destPath, chunk_property &properties, size_t parallel)
```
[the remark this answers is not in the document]
seguler: Changing to blob_propery was much more involved bcs download chunk async returns chunk property and we retrieve these properties from the first chunk. So I left it that way. If you have a easy way to convert this to blob properly send a PR to my fork :)

azure-storage-cpp-lite/include/blob/blob_client.h
```diff
@@ -414,7 +414,8 @@ namespace microsoft_azure { namespace storage {
         /// <param name="size">The size of the data to download from the blob, in bytes.</param>
         /// <param name="destPath">The target file path.</param>
         /// <param name="parallel">A size_t value indicates the maximum parallelism can be used in this request.</param>
-        void download_blob_to_file(const std::string &container, const std::string &blob, const std::string &destPath, size_t parallel = 9);
+        /// <returns>A <see cref="storage_outcome" /> object that represents the properties (etag, last modified time and size) from the first chunk retrieved.</returns>
```
seguler: Needs to be fixed

PR #108, opened 2018-02-24
increase block size to 16MB for uploads

seguler:
Increases supported files to 800GB


AlexanderYukhanov:
Looks like you are on an outdated branch. I refactored this code in master and now you need to change only "const unsigned long long UPLOAD_CHUNK_SIZE = 4 * 1024 * 1024;" at the top of the file

AlexanderYukhanov:
And I don't like this limitation at all. We know the size of the file which we are going to upload, so we can calculate required chunk size to upload the whole file

seguler:
Ok will address this in a different PR. Closing this one

Issue #107, opened 2018-02-22
issue while copying large files


vinulohar007:
Hi Team,

I am facing an issue while creating large files on mounted blob storage. 

In my scenario I have a blob storage mounted on the SUSE Linux machine and we are having SAP DB. So we are dumping the DB backup directly to the mounted blob storage. Backup size is about 250GB.
We also have a large space in temp file system. When we have started the backup on blob storage files are started to create in temp space but it has not created on blob. we only see 0 byes of files.

Backup is completed in temp space but files not uploaded to blob and mounted blob is seems unmounted.

@seguler 
Can you please help here?

 **# more mount.sh**
#!/bin/bash
BLOBFS_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
cd $BLOBFS_DIR/build
./blobfuse $1 --tmp-path=/databck/blobfusetmp -o uid=1004 -o allow_other  -o gid=1001 -o attr_timeout=240 -o entry_timeout=240 -o negative_timeout=120  --
file-cache-timeout-in-seconds=120 --config-file=../connection.cfg

Do I need to add any parameter here or need to modify any parameter with respect to timeout?

Thanks,
Vinayak lohar

seguler:
@asorrin-msft what is the max file size we support ? Is it 200GB at the moment ?

seguler:
I just verified. The largest file we support is 200GB. I don't see any reason to push this up to 800GB though... Perhaps we can do this in the next release.

vinulohar007:
Hi @seguler,

When can we expect the next release?

Can you please provide the link to the rpm file of blobfuse. 

I have tried to add the Microsoft repo but I have not found the blobfuse rpm there. 


Thanks
Vinayak

seguler:
Here is the rpm for rhel 7.3: https://packages.microsoft.com/rhel/7.3/prod/blobfuse-0.2.4-el7.rpm

What is your Linux distro/version ? Let me validate if rpm works for that release.

vinulohar007:
We are using suse Linux 12 sp2

vinulohar007:
Hi @seguler,

I have just checked blobfuse rpm is not available there in the sles repo

https://packages.microsoft.com/sles/12/prod/


Please help us with the same if available.  

Thanks
Vinayak

vinulohar007:
@seguler ,

Any update on the repository ?


Thanks,
Vinayak  Lohar

[...]

PR #110, opened 2018-02-26
Increase min block_size to 16mb, and handle blobs up to 4.77TB

seguler:
Addresses #107

### DIFF
--- a/azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
+++ b/azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
@@ -14,7 +14,7 @@
 namespace microsoft_azure {
     namespace storage {
         const unsigned long long DOWNLOAD_CHUNK_SIZE = 16 * 1024 * 1024;
-        const unsigned long long UPLOAD_CHUNK_SIZE = 4 * 1024 * 1024;
+        const unsigned long long MIN_UPLOAD_CHUNK_SIZE = 16 * 1024 * 1024;
         class mempool
         {
         public:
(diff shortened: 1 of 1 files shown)


azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
```diff
@@ -456,7 +456,14 @@ namespace microsoft_azure {
...
 
             int result = 0;
-            int block_size = 4*1024*1024;
+
+            //support blobs up to 4.77TB
+            //need to round to the nearest multiple of 4MB for efficiency
+            unsigned long long min_block = fileSize / 50000; 
+            int remainder = min_block % 4*1024*1024;
+            min_block += 4*1024*1024 - remainder;
+            unsigned long long block_size = min_block < MIN_UPLOAD_CHUNK_SIZE ? MIN_UPLOAD_CHUNK_SIZE : min_block;
```
asorrin-msft: Couple things:
- We should check that the file is no greater than 100 * 1024 * 1024 * 50 000 bytes
- We should make sure that the rounding up doesn't go over 100 MB - I don't think it can with the above check, but it won't hurt, just     block_size = block_size > 100MB ? 100MB : block_size
- Very minor - if we first check if (fileSize > (50000 * MIN_UPLOAD_CHUNK_SIZE)), we can skip the math in the majority of cases (any blobs smaller than 800 GB)

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
```diff
@@ -471,7 +478,7 @@ namespace microsoft_azure {
             std::condition_variable cv;
             std::mutex cv_mutex;
 
-            for(long long offset = 0, idx = 0; offset < fileSize; offset += UPLOAD_CHUNK_SIZE, ++idx)
+            for(long long offset = 0, idx = 0; offset < fileSize; offset += block_size, ++idx)
```
asorrin-msft: Just checking - you tested locally?  We need to make some 'scale' tests - ones that we only run at ship time, but that test stuff like this.  For now, though, as long as we test locally with a >1TB file should be ok.

seguler: Yes - 250GB and 1TB file tests succeeded. 1TB takes a loooong time due to us writing to the disk first, and then uploading to Blob.

PR #112, opened 2018-02-28
Fix retry logic for download

AlexanderYukhanov:
1. Restored state of CurlEasyRequest after initial 503
2. Added handling of curl errors

### DIFF
--- a/azure-storage-cpp-lite/src/http/libcurl_http_client.cpp
+++ b/azure-storage-cpp-lite/src/http/libcurl_http_client.cpp
@@ -55,7 +55,9 @@ namespace microsoft_azure {
             m_slist = curl_slist_append(m_slist, "Expect:");
             check_code(curl_easy_setopt(m_curl, CURLOPT_HTTPHEADER, m_slist));
 
-            return curl_easy_perform(m_curl);
+            const auto result = curl_easy_perform(m_curl);
+            check_code(result); // has nothing to do with checks, just resets errno for succeeded ops.
+            return result;
         }
 
         size_t CurlEasyRequest::header_callback(char *buffer, size_t size, size_t nitems, void *userdata) {
--- a/azure-storage-cpp-lite/src/http/libcurl_http_client.cpp
+++ b/azure-storage-cpp-lite/src/http/libcurl_http_client.cpp
@@ -24,7 +24,9 @@ namespace microsoft_azure {
             }
         }
 
-        http_base::http_code CurlEasyRequest::perform() {
+        CURLcode CurlEasyRequest::perform() {
+            check_code(curl_easy_setopt(m_curl, CURLOPT_WRITEFUNCTION, write));
+            check_code(curl_easy_setopt(m_curl, CURLOPT_WRITEDATA, this));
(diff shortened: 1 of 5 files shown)


AlexanderYukhanov:
Was able to run alexnet against imagenet on 16 NC24r.

seguler:
You're AWESOME!

seguler:
One thing we discussed today earlier:

This retry also applies to the curl errors like 'Host not found'.. In this case, fuse client will still try 100 times (every 3 seconds) which will be very painful for simple scenarios like you configure a wrong blobEndpoint... So should we decrease the amount of retries to 10 or something ? Or maybe make it configurable and default to 10 ? @asorrin-msft

azure-storage-cpp-lite/src/http/libcurl_http_client.cpp
```diff
@@ -51,9 +55,7 @@ namespace microsoft_azure {
             m_slist = curl_slist_append(m_slist, "Expect:");
             check_code(curl_easy_setopt(m_curl, CURLOPT_HTTPHEADER, m_slist));
 
-            check_code(curl_easy_perform(m_curl));
-
-            return m_code;
+            return curl_easy_perform(m_curl);
```
asorrin-msft: We've had problems in the past where CURL sometimes sets errno when it shouldn't (perhaps even when it returns CURL_OK?  I don't remember for sure.)  That's why check_code resets errno to 0.  I worry if we don't do this here, those bugs might come back

AlexanderYukhanov: but nobody checks errno in this thread as far as I can tell

asorrin-msft: there's no chance that this thread ever gets back to blobfuse?

asorrin-msft:
@seguler - I'm not super concerned about this at the moment.  Before we call fuse_main to actually mount, we do a container_exists() call to validate that we can actually communicate with Storage.  If there's a permanent issue with connecting to Storage, yes this exists call will retry many times before it fails, but at least we won't have actually mounted.  I would assume that most errors coming from libcurl after that are transient.  Of course, we should definitely improve this, but I'm ok with it for now.

azure-storage-cpp-lite/src/http/libcurl_http_client.cpp
AlexanderYukhanov: as far as I see, blob_client always uses anync_executor ever for synchronous calls. But if it will make you more confident I will call check_code here as well.

seguler: Alex we need this merged today :)

seguler: Also this needs to target dev branch right ?

AlexanderYukhanov:
done

azure-storage-cpp-lite/src/http/libcurl_http_client.cpp
AlexanderYukhanov: done

Issue #106, opened 2018-02-21
Blobfuse mount fails with "terminate called after throwing an instance of std::logic_error"


fedexist:
Here is the error:

`terminate called after throwing an instance of 'std::logic_error'`
`  what():  basic_string::_S_construct null not valid`
`Aborted (core dumped)`

after trying to run both

`blobfuse /mnt/azure --config-file=connection.cfg`
`blobfuse /mnt/azure --container-name=container`

connection.cfg
`accountName xxxxxxxxx  `
`accountKey xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx==  `
`containerName container
`

I've tried on both Ubuntu 16.04 and CentOS 7, installing blobfuse via yum and apt and building it from sources.

Storage account is a Standard_LRS StorageV2 account (tried both hot and cool storage), accessible via other tools such as Minio Azure Gateway or CloudBerry Drive.

Any idea what's the problem here?

seguler:
It seems --tmp-path is missing. This happens to many... we'll put a default for this next release

fedexist:
I see, thank you. 
A default option or at least throwing an informative error if that option is not present would be great!

andyzhangx:
+1 for this fix

seguler:
Fixed in blobfuse 0.3

PR #113, opened 2018-03-01
Fixing issue #106, better error on no --tmp-path

asorrin-msft:


seguler:
Great - this now fixes #106

PR #123, opened 2018-03-07
fix non-zero offset truncate

anhvoms:


anhvoms:
I wonder if we can optimize this by having some threshold. I.e., if the file size after truncation small enough, we can stream the exact number of bytes from blob -> local file, instead of downloading the whole blob and truncate it. This probably doesn't work well for very large blob because streaming is not as efficient as downloading blocks (please correct me if I'm wrong).

blobfuse/fileapis.cpp
```diff
@@ -468,70 +468,50 @@ int azs_truncate(const char * path, off_t off)
...
     int statret = stat(mntPath, &buf);
-    if (statret == 0)
+    // If the file exists in cache and the cache has not expired
+    if (statret == 0 && ((time(NULL) - buf.st_mtime) <= file_cache_timeout_in_seconds))
     {
-        // The file exists in the local cache.  So, we call truncate() on the file in the cache, then upload a zero-length blob to the service, overriding any data.
-        int truncret = truncate(mntPath, 0);
-        if (truncret == 0)
+        // The file exists in the local cache.  So, we call truncate() on the file in the cache
+        if( truncate(mntPath, off) != 0)
```
seguler: @asorrin-msft do we need a mutex lock here ?

seguler: It seems we didn't have before...

blobfuse/fileapis.cpp
```diff
@@ -468,70 +468,50 @@ int azs_truncate(const char * path, off_t off)
...
-            close(fd);
-
-            // We want to upload a zero-length blob.
-            std::istringstream emptyDataStream("");
-
-            std::vector<std::pair<std::string, std::string>> metadata;
             errno = 0;
-            azure_blob_client_wrapper->upload_block_blob_from_stream(str_options.containerName, pathString.substr(1), emptyDataStream, metadata);
+            time_t last_modified = {};
+            azure_blob_client_wrapper->download_blob_to_file(str_options.containerName, pathString.substr(1), mntPathString, last_modified);
```
seguler: We need a mutex here because another thread may be downloading to this path in parallel... just like we have mutex protection in azs_open

blobfuse/fileapis.cpp
seguler: ok all good

blobfuse/fileapis.cpp
```diff
@@ -468,70 +468,50 @@ int azs_truncate(const char * path, off_t off)
...
             if (errno != 0)
             {
-                return 0 - map_errno(errno); // TODO: Investigate what might happen in this case - the blob has been truncated locally, but not on the service.
+                remove(mntPath);
+                return 0 - map_errno(errno);
+            }
+            if( truncate(mntPath, off) != 0 )
```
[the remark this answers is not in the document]
seguler: Two options here: truncate and ftruncate... So it can be called on a not opened file too..
http://pubs.opengroup.org/onlinepubs/9699919799/functions/truncate.html
http://pubs.opengroup.org/onlinepubs/9699919799/functions/ftruncate.html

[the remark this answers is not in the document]
seguler: It seems there is no fd being passed into azs_truncate. Therefore I assume we'll not receive a close for this... So we should upload the file here in azs_truncate

seguler:
Thanks Anh - this is now fixed in the following PR: https://github.com/Azure/azure-storage-fuse/pull/154

PR #131, opened 2018-03-13
Fixing statfs per POSIX

seguler:


seguler:
Tests to follow

blobfuse/directoryapis.cpp
```diff
@@ -228,13 +228,26 @@ int azs_rmdir(const char *path)
...
-    const char * mntPath;
-    std::string mntPathString = prepend_mnt_path_string(pathString);
-    mntPath = mntPathString.c_str();
 
-    int res = statvfs(mntPath, stbuf);
-    if (res == -1)
-        return -errno;
+    int dir_result = is_directory_empty(str_options.containerName, "/", pathString.substr(1));
+    bool path_exists = false;
+    if(strlen(path) == 1 || dir_result == D_NOTEMPTY)
```
AlexanderYukhanov: what does it mean if path length is 1? why we consider it to exist?

seguler: It is root of fuse "/"

blobfuse/directoryapis.cpp
```diff
@@ -228,13 +228,26 @@ int azs_rmdir(const char *path)
...
+    else if(azure_blob_client_wrapper->get_blob_property(str_options.containerName, std::string(&(path[1]))).valid()) 
+    {
+        path_exists = true;
+    }
 
-    return 0;
+    if(path_exists == true)
+    {
+        // return tmp path stats
+        int res = statvfs(str_options.tmpPath.c_str(), stbuf);
```
AlexanderYukhanov: it's still not valid for -o allow_other ?

blobfuse/directoryapis.cpp
AlexanderYukhanov: why not to check if it's "/" explicitly?

blobfuse/directoryapis.cpp
AlexanderYukhanov: struct statvfs {
    unsigned long  f_bsize;    /* file system block size */
    unsigned long  f_frsize;   /* fragment size */
    fsblkcnt_t     f_blocks;   /* size of fs in f_frsize units */
    fsblkcnt_t     f_bfree;    /* # free blocks */
    fsblkcnt_t     f_bavail;   /* # free blocks for unprivileged users */
    fsfilcnt_t     f_files;    /* # inodes */
    fsfilcnt_t     f_ffree;    /* # free inodes */
    fsfilcnt_t     f_favail;   /* # free inodes for unprivileged users */
    unsigned long  f_fsid;     /* file system ID */
    unsigned long  f_flag;     /* mount flags */
    unsigned long  f_namemax;  /* maximum filename length */
};
is it fare to return free blocks and inodes for temp directory instead of storage?

seguler: Well - I don't know. We are bound to the local filesystems limits (f_flag, f_namemax) for files that are open... We can also not efficiently report the available space in a blob storage account... Right ? 

Let's not complicate this by trying to report things from Azure. Simply pass the local filesystem limits here...

blobfuse/directoryapis.cpp
seguler: I was following the precedence in the code to be consistent. It seems Adam used strlen here as well: https://github.com/Azure/azure-storage-fuse/blob/ed79ab5108cbcb70dadc540e435d5622938c3d58/blobfuse/utilities.cpp#L340

blobfuse/directoryapis.cpp
asorrin-msft: Would it be better to zero out the size-related fields, as an indication that we don't know?  Or would this cause more problems than it solves, by breaking applications that might check this value?  We could also set it to the equivalent of 500TB, or something.
Also, we may want to subtract the length of the tmp_path string from f_namemax

blobfuse/directoryapis.cpp
```diff
@@ -228,13 +228,26 @@ int azs_rmdir(const char *path)
...
     std::string pathString(path);
-    const char * mntPath;
-    std::string mntPathString = prepend_mnt_path_string(pathString);
-    mntPath = mntPathString.c_str();
 
-    int res = statvfs(mntPath, stbuf);
-    if (res == -1)
-        return -errno;
+    int dir_result = is_directory_empty(str_options.containerName, "/", pathString.substr(1));
+    bool path_exists = false;
```
asorrin-msft: Can we simplify and just call azs_get_attr to determine if it's a valid file or directory?

blobfuse/directoryapis.cpp
seguler: f_namemax is max filename length.. Not the path name...

Many applications check disk available size and if it is 0, throw an error. This is very common - that is why I had to implement this stat function in the first place

blobfuse/directoryapis.cpp
seguler: @asorrin-msft keep this or change to path=="/" ?

blobfuse/directoryapis.cpp
seguler: Ok I can do that

PR #132, opened 2018-03-13
Sas support

rickle-msft:


seguler:
AWESOME!!!

connection.cfg
```diff
@@ -1,3 +1,5 @@
 accountName <account-name-here>
-accountKey <account-key-here>
+accountKey <account-name-here-delete-next-line>
```
seguler: account-key-here... typo

seguler: Deleting one of these is not required, right ?

rickle-msft: Yes it is.

seguler: Why can't we change the flow so that we take accountKey if available, then switch to SAS... And still accept both.. Same goes to environment variables too. Accept both methods, but prefer AccountKey

I am sure there will be users who forget to remove one of these lines and end up failing.

rickle-msft: When I discussed with Adam, he wanted to fail if both were specified. The error message you get if both (or neither) is specified is very clear that exactly one should be present.

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
```diff
@@ -143,7 +145,14 @@ namespace microsoft_azure {
 
             try
             {
-                std::shared_ptr<storage_credential>  cred = std::make_shared<shared_key_credential>(accountName, accountKey);
+                std::shared_ptr<storage_credential>  cred;
+		if (account_key.length() > 0) {
```
AlexanderYukhanov: please fix formatting

blobfuse/blobfuse.cpp
```diff
@@ -79,10 +80,17 @@ int read_config_env()
     {
         str_options.accountKey = env_account_key;
     }
-    else
+    
+    if(env_sas_token!=NULL)
     {
-        fprintf(stderr, "AZURE_STORAGE_ACCESS_KEY environment variable is empty.\n");
-        return -1;
+	str_options.sasToken = env_sas_token;
```
AlexanderYukhanov: formatting

blobfuse/blobfuse.cpp
```diff
@@ -152,9 +165,10 @@ int read_config(std::string configFile)
         fprintf(stderr, "Account name is missing in the configure file.\n");
         return -1;
     }
-    else if(str_options.accountKey.size() == 0)
+    else if((str_options.accountKey.size() == 0 && str_options.sasToken.size() == 0) ||
```
AlexanderYukhanov: .empty()

blobfuse/blobfuse.cpp
```diff
@@ -152,9 +165,10 @@ int read_config(std::string configFile)
         fprintf(stderr, "Account name is missing in the configure file.\n");
         return -1;
     }
-    else if(str_options.accountKey.size() == 0)
+    else if((str_options.accountKey.size() == 0 && str_options.sasToken.size() == 0) || 
+	    (str_options.accountKey.size() != 0 && str_options.sasToken.size() != 0))
     {
-        fprintf(stderr, "Account key is missing in the configure file.\n");
+        fprintf(stderr, "Exactly one of Account Key and SAS token must be specified in the configure file. The other line should be deleted.\n");
```
AlexanderYukhanov: configure is verb. Please replace with "configuration" or "config"

blobfuse/blobfuse.cpp
```diff
@@ -125,6 +133,11 @@ int read_config(std::string configFile)
             std::string accountKeyStr(value);
             str_options.accountKey = accountKeyStr;
         }
+        else if(line.find("sasToken") != std::string::npos)
```
AlexanderYukhanov: can you please replace commented out code before and after this line:
 /*            if(!is_lowercase_string(accountNameStr))
...*/

blobfuse/blobfuse.cpp
```diff
@@ -171,7 +185,8 @@ int read_config(std::string configFile)
 
 void *azs_init(struct fuse_conn_info * conn)
 {
-    azure_blob_client_wrapper = std::make_shared<blob_client_wrapper>(blob_client_wrapper::blob_client_wrapper_init(str_options.accountName, str_options.accountKey, 20, str_options.use_https, str_options.blobEndpoint));
+    azure_blob_client_wrapper = std::make_shared<blob_client_wrapper>(blob_client_wrapper::blob_client_wrapper_init(str_options.accountName, str_options.accountKey, str_options.sasToken, 20, str_options.use_https,
```
AlexanderYukhanov: what is 20 here? can you add a comment or named const?

blobfuse/blobfuse.cpp
rickle-msft: Do you mean delete?

asorrin-msft: Yeah, that code didn't work... I left it in b/c I thought it might be a good idea to keep it, but client-side checks like these can come back to bite us, if something changes on the service.  Let's just remove it.

blobfuse/blobfuse.cpp
```diff
@@ -79,10 +80,17 @@ int read_config_env()
...
     {
-        fprintf(stderr, "AZURE_STORAGE_ACCESS_KEY environment variable is empty.\n");
-        return -1;
+	str_options.sasToken = env_sas_token;
+    }
+
+    if((env_account_key==NULL && env_sas_token==NULL) ||
+       (env_account_key!=NULL && env_sas_token!=NULL))
+    {
+	fprintf(stderr, "Exactly one of AZURE_STORAGE_ACCESS_KEY and AZURE_STORAGE_SAS_TOKEN environment variables must be specified.\n");
```
asorrin-msft: Can you prefix this with something like "If not using the config file, ..."

blobfuse/blobfuse.cpp
```diff
@@ -147,17 +144,18 @@ int read_config(std::string configFile)
         data.clear();
     }
 
-    if(str_options.accountName.size() == 0)
+    if(str_options.accountName.empty())
     {
         fprintf(stderr, "Account name is missing in the configure file.\n");
```
asorrin-msft: Could you fix the text here?

blobfuse/blobfuse.cpp
```diff
@@ -171,7 +169,8 @@ int read_config(std::string configFile)
 
 void *azs_init(struct fuse_conn_info * conn)
 {
-    azure_blob_client_wrapper = std::make_shared<blob_client_wrapper>(blob_client_wrapper::blob_client_wrapper_init(str_options.accountName, str_options.accountKey, 20, str_options.use_https, str_options.blobEndpoint));
+    azure_blob_client_wrapper = std::make_shared<blob_client_wrapper>(blob_client_wrapper::blob_client_wrapper_init(str_options.accountName, str_options.accountKey, str_options.sasToken, 20/*concurrency*/, str_options.use_https, 
+														    str_options.blobEndpoint));
```
asorrin-msft: Formatting looks weird here, but maybe that's just Github?

connection.cfg
asorrin-msft: I definitely prefer to fail rather than to have to guess.  Could we just add a comment in the sample connection.cfg, explaining?  It looks to me like the parser will ignore any lines it doesn't recognize, so we could just add a line, something like:
\# Please provide either an account key or a SAS token, and delete the other line:

blobfuse/blobfuse.cpp
rickle-msft: I think this and some of the other formatting comments are GitHub. In my editor, str_options.accountName and str_options.blobEndpoint align

AlexanderYukhanov: Please configure your editor to use spaces instead of tabs

azure-storage-cpp-lite/include/storage_credential.h
```diff
@@ -43,7 +43,12 @@ namespace microsoft_azure {
         class shared_access_signature_credential : public storage_credential {
         public:
             shared_access_signature_credential(const std::string &sas_token)
-                : m_sas_token(sas_token) {}
+                : m_sas_token(sas_token) {
+		// If there is a question mark at the beginning of the sas token, erase it for easier processing in sign_request.
+		if (m_sas_token[0] == '?') {
```
AlexanderYukhanov: !m_sas_tocken.empty() &&

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
```diff
@@ -118,15 +118,17 @@ namespace microsoft_azure {
...
+            return blob_client_wrapper_init(account_name, account_key, sas_token, concurrency, false, NULL);
         }
 
 
-        blob_client_wrapper blob_client_wrapper::blob_client_wrapper_init(const std::string &account_name, const std::string &account_key, const unsigned int concurrency, const bool use_https, const std::string &blob_endpoint)
+        blob_client_wrapper blob_client_wrapper::blob_client_wrapper_init(const std::string &account_name, const std::string &account_key, const std::string &sas_token,  const unsigned int concurrency, const bool use_https, 
+									  const std::string &blob_endpoint)
         {
-            if(account_name.length() == 0 || account_key.length() == 0)
+            if(account_name.length() == 0 ||
```
AlexanderYukhanov: .empty()

blobfuse/blobfuse.cpp
```diff
@@ -79,10 +80,17 @@ int read_config_env()
     {
         str_options.accountKey = env_account_key;
     }
-    else
+    
+    if(env_sas_token!=NULL)
```
AlexanderYukhanov: if(env_sas_tocken)
if(nullptr !=..)...

PR #133, opened 2018-03-14
Tests

asorrin-msft:


CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 2.8.2)
```
seguler: This essentially drops the support for RHEL 6, because only Cmake 2.8.12 is available on that distro.. Can you go back to 2.8 ?

blobfuse/blobfuse.cpp
```diff
@@ -205,9 +205,10 @@ void print_usage()
     fprintf(stdout, "See https://github.com/Azure/azure-storage-fuse for detailed installation and configuration instructions.\n");
 }
 
-int main(int argc, char *argv[])
+
+void set_up_callbacks()
 {
-    // Here, we set up all the callbacks that FUSE requires.
+        // Here, we set up all the callbacks that FUSE requires.
```
seguler: Extra spaces ?

test/cpplitetests.cpp
```diff
@@ -0,0 +1,510 @@
...
+TEST_F(BlobClientWrapperTest, BlobPutDownload)
+{
+    errno = 0;
+    std::vector<list_blobs_hierarchical_item> blobs = list_all_blobs(container_name, "/", "");
+    ASSERT_EQ(0, errno);
+    ASSERT_EQ(0, blobs.size());
+
+    // Create a file
+    std::string file_path = tmp_dir + "/tmpfile";
+    std::string file_text = "some file text here.";
```
seguler: We are not testing parallel upload/download because the file is smaller than 16MB. Any chance we create a random 100MB file or something and validate that instead of this ? Validation can use md5 sum

test/cpplitetests.cpp
```diff
@@ -0,0 +1,510 @@
...
+    blobs = list_all_blobs(container_name, "/", "");
+    ASSERT_EQ(0, errno);
+    ASSERT_EQ(1, blobs.size());
+    ASSERT_EQ(file_size, blobs[0].content_length) << "Blob found, but size incorrect.";
+
+    std::string dest_path = tmp_dir + "/destfile";
+
+    time_t lmt;
+    test_blob_client_wrapper->download_blob_to_file(container_name, blob_1_name, dest_path, lmt);
+    // TODO: test that download returns correct blob properties once implemented.
```
seguler: Test the lmt

test/cpplitetests.cpp
seguler: Ok just saw you are testing large files as well.. Can we do a checksum on that to compare the actual data rather than the file size ?

asorrin-msft: The more-extensive tests below (run_upload_download) do validate the content via RNG

CMakeLists.txt
asorrin-msft: Fixed, turns out 2.8.2 was just for a sample, actual minimum is 2.6.4

blobfuse/blobfuse.cpp
asorrin-msft: Fixed

test/cpplitetests.cpp
asorrin-msft: Good point, these tests were written before the LMT stuff was in, then rebased.  I'll add a new test just for LMT testing.

asorrin-msft: Changed my mind, added checks in with the blob properties test

PR #134, opened 2018-03-15
Changing blobfuse to log to syslog

asorrin-msft:
This adds log statements throughout the blobfuse part of the codebase.  Cpplite will be done separately.

### DIFF
--- a/blobfuse/blobfuse.cpp
+++ b/blobfuse/blobfuse.cpp
@@ -74,7 +74,7 @@ int read_config_env()
     }
     else
     {
-        syslog (LOG_CRIT, "Unable to start blobfuse.  No config file was specified and AZURE_STORAGE_ACCESS_KEY environment variable is empty.");
+        syslog(LOG_CRIT, "Unable to start blobfuse.  No config file was specified and AZURE_STORAGE_ACCESS_KEY environment variable is empty.");
         fprintf(stderr, "No config file was specified and AZURE_STORAGE_ACCOUNT environment variable is empty.\n");
         return -1;
     }
@@ -85,7 +85,7 @@ int read_config_env()
     }
     else
     {
-        syslog (LOG_CRIT, "Unable to start blobfuse.  No config file was specified and AZURE_STORAGE_ACCESS_KEY environment variable is empty.");
+        syslog(LOG_CRIT, "Unable to start blobfuse.  No config file was specified and AZURE_STORAGE_ACCESS_KEY environment variable is empty.");
         fprintf(stderr, "No config file was specified and AZURE_STORAGE_ACCESS_KEY environment variable is empty.\n");
         return -1;
     }
@@ -99,7 +99,7 @@ int read_config(std::string configFile)
     std::ifstream file(configFile);
     if(!file)
     {
-        syslog (LOG_CRIT, "Unable to start blobfuse.  No config file found at %s.", configFile.c_str());
+        syslog(LOG_CRIT, "Unable to start blobfuse.  No config file found at %s.", configFile.c_str());
         fprintf(stderr, "No config file found at %s.\n", configFile.c_str());
         return -1;
     }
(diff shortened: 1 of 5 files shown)


blobfuse/utilities.cpp
```diff
@@ -259,10 +234,7 @@ std::vector<list_blobs_hierarchical_item> list_all_blobs_hierarchical(std::strin
         {
             failcount++;
             success = false;
-            if (AZS_PRINT)
-            {
-                fprintf(stdout, "list_blobs_hierarchical failed %d time with errno = %d\n", failcount, errno);
-            }
+            syslog(LOG_WARNING, "list_blobs_hierarchical failed for the %d time with errno = %d.\n", failcount, errno);
```
seguler: warning only when it fails ?

blobfuse/utilities.cpp
```diff
@@ -358,21 +327,26 @@ int azs_getattr(const char *path, struct stat *stbuf)
...
+        if (res == -1)
         {
-            printf("LSTAT res = %d, errno = %d, ENOENT = %d\n", res, errno, ENOENT);
+            int lstaterrno = errno;
+            syslog(LOG_ERR, "lstat on file %s in local cache during get_attr failed with errno = %d.\n", mntPathString.c_str(), lstaterrno);
+            return -lstaterrno;
+        }
+        else
+        {
+            syslog(LOG_DEBUG, "lstat on file %s in local cache succeeded.\n", mntPathString.c_str());
```
seguler: I wonder if we also had the function name prepended to these text. If I see this in my syslog, how do I know that this happened during a getattr call ?

blobfuse/utilities.cpp
asorrin-msft: Yes, because it might succeed on retry... this stuff should be ripped out anyway, once cpplite has better retry support.

blobfuse/utilities.cpp
```diff
@@ -427,15 +393,19 @@ int azs_getattr(const char *path, struct stat *stbuf)
         }
         else
         {
-            return -(ENOENT); // -2 = Entity does not exist.
+            syslog(LOG_DEBUG, "Entity %s does not exist.  Returning ENOENT (%d) from get_attr.\n", path, ENOENT);
```
seguler: For instance; rewriting all of these as:

"getattr call - Entity %s does not exist. Returning ENOENT (%d)"

asorrin-msft: I tried to make the info / warning / error logs readable on their own, but for debug, my thinking was that someone reading the debug log will have the other debug log entries right there, and every method logs when it's first called into.  I tried to add the input path argument (or something equivalent, like mntPath or blob name) to every log statement, to help detangle them if stuff is happening in parallel.

seguler: I don't think this justifies it :) I can't understand looking at the following log and determine which call it is coming from... Think about the support team who may end up looking at these logs one day

blobfuse/utilities.cpp
seguler: Ok - I think we have a syslog statement for failed readdirs.. So it is fine

blobfuse/utilities.cpp
asorrin-msft: Ok, I'm not super convinced, but no harm in adding it :)  I'll get to this tomorrow.

blobfuse/blobfuse.cpp
```diff
@@ -13,14 +13,14 @@ namespace {
 }
```
seguler: Usage message is not updated with the new command line option: https://github.com/Azure/azure-storage-fuse/blob/ed79ab5108cbcb70dadc540e435d5622938c3d58/blobfuse/blobfuse.cpp#L201

seguler:
Tested and validated on Ubuntu 16.04 with debug log levels.

blobfuse/blobfuse.cpp
asorrin-msft: See next line - it was getting too long, so I split it up over two lines :)

asorrin-msft:
Cool, thanks for testing!  On my centos 7.4 machine, I had to configure rsyslog to print debug messages, by default it ignores them.  Do you think we need to give instructions on this?  Or is this obvious enough that we shouldn't worry?

PR #136, opened 2018-03-15
helper copy tool for concurrent data transfers

seguler:
We're crazy for doing this; but here is a concurrent copy tool that can be used with blobfuse.

In my tests, it improves the upload/download performance by 5X on average through blobfuse.

Here are a few samples:
1. Copy a directory into another directory
./blobcp -s /home/seguler/mount/images45K/images/ -d /mnt/

2. Copy a directory of files that matches a pattern (regular file spec)
./blobcp -s /home/seguler/mount/images45K/images/ -d /mnt/ -pattern "mysamplefile.9*"


seguler:
If all agrees, I can push another commit that installs blobcp along with blobfuse during the install so that users call blobcp from any where.

tools/blobcp.go
```diff
@@ -0,0 +1,116 @@
...
+var workers = flag.Int("n", 32, "Number of workers")
+var source = flag.String("s", "", "Source path to walk through")
+var dest = flag.String("d", "", "Destination to copy to")
+var include = flag.String("pattern", "", "Path pattern to match in source path")
+var processed int64 = 0
+
+func main() {
+    flag.Parse()
+    if *source == "" {
+        log.Printf("-s source path is missing")
```
AlexanderYukhanov: Print of Println since there is no format message

tools/blobcp.go
```diff
@@ -0,0 +1,116 @@
...
+        flag.PrintDefaults()
+    } else if *dest == "" {
+        log.Printf("-d destination path is missing")
```
AlexanderYukhanov: same as above

tools/blobcp.go
```diff
@@ -0,0 +1,116 @@
...
+        flag.PrintDefaults()
+    }
+
+    pending := sync.WaitGroup{}
+    files := make(chan string)
+
+    go func() {
```
AlexanderYukhanov: do we need it?

tools/blobcp.go
```diff
@@ -0,0 +1,116 @@
...
+        }
+    }()
+
+    start := time.Now()
+    pending.Add(1)
+    startWorkers(files, &pending)
+    submitFiles(*source, *include, files, &pending)
+    pending.Wait()
+
+    fmt.Printf("Processed %v files in %v sec", processed, time.Now().Sub(start))
```
AlexanderYukhanov: if we  want some stats, let's add mb per sec, total size ?

tools/blobcp.go
```diff
@@ -0,0 +1,116 @@
...
+        last := processed
+        for {
+            <-time.After(5 * time.Second)
+            fmt.Printf("Processed %v, %v per second       \r", processed, (processed-last)/5)
+            last = processed
+        }
+    }()
+
+    start := time.Now()
+    pending.Add(1)
```
AlexanderYukhanov: let's move to start workers to make same guy to increase and decrease it

tools/blobcp.go
```diff
@@ -0,0 +1,116 @@
...
+    for i := 0; i < *workers; i++ {
+        go func() {
+            for p := range files {
+                processFile(p)
+                atomic.AddInt64(&processed, 1)
+                pending.Done()
+            }
+        }()
+    }
+    pending.Done()
```
AlexanderYukhanov: why startWorkers doing this? this should be done at the end of submitFiles instead

AlexanderYukhanov: why do we need it and why do we need pending.add in main?

AlexanderYukhanov:
I think you committed the binary by mistake?

AlexanderYukhanov:
Do you want to add documentation update for this tool in the same PR?
1. reasoning 
2. how to build
3. how to run

tools/blobcp.go
seguler: pending.add in main is needed to avoid race condition, and exit the main before anything is executed.

seguler: this pending.Done is to clear the pending.Add(1) from the main function

seguler:
Why not commit the binary as well :) instead of requiring everyone to download Go etc....

tools/blobcp.go
seguler: How do I get total size for this ?

tools/blobcp.go
seguler: What do you mean ? Doesn't this have to be in a go routine ?

tools/blobcp.go
```diff
@@ -0,0 +1,119 @@
...
+
+    pending := sync.WaitGroup{}
+    files := make(chan string)
+
+    go func() {
+        last := processed
+        for {
+            <-time.After(5 * time.Second)
+            _processed := atomic.LoadInt64(&processed)
+            fmt.Printf("Processed %v, %v per second \r", _processed, (_processed-last)/5)
```
AlexanderYukhanov: add more more spaces

AlexanderYukhanov: -)

tools/blobcp.go
```diff
@@ -0,0 +1,119 @@
...
+        flag.PrintDefaults()
+        os.Exit(1)
+    } else if *dest == "" {
+        log.Println("-d destination path is missing")
+        flag.PrintDefaults()
+        os.Exit(1)
+    }
+
+    pending := sync.WaitGroup{}
+    files := make(chan string)
```
AlexanderYukhanov: sum buffer?
make(chan string, 1024)

asorrin-msft:
Wait, what?  Why are we committing the binary?  I thought we were going to package the binary in with the blobfuse binaries; you wouldn't have to install go unless you want to build from source

seguler:
TODO:
1. Reformat spaces
2. Add README file, modify CMakeLists to install blobcp

AlexanderYukhanov:
lgtm

seguler:
@asorrin-msft I think this is ready to merge now.

PR #139, opened 2018-03-17
fix gc issue with renames; add --version --help support

seguler:


asorrin-msft:
This won't completely fix the issue, because if a file is already open and then gets renamed, it won't get cleaned up.  This is much better than what we had before, though, so I think we should merge this in and fix the corner case later.

PR #145, opened 2018-03-21
Code cleanup

rickle-msft:


seguler:
Can you test if you can write some data with AzCopy and then overwrite with blobfuse ? This is to make sure block ids are in the same length.

PR #146, opened 2018-03-21
Timeouts

rickle-msft:
Absolute operation timeouts for requests not expecting a response body. Timeouts based on whether a minimum throughput is maintained throughout the request for requests expecting a response body.

### DIFF
--- a/azure-storage-cpp-lite/include/http/libcurl_http_client.h
+++ b/azure-storage-cpp-lite/include/http/libcurl_http_client.h
@@ -148,6 +148,16 @@ namespace microsoft_azure {
                 return m_input_stream;
             }
 
+            void set_absolute_timeout() override {
+                check_code(curl_easy_setopt(m_curl, CURLOPT_TIMEOUT, 30L)); // Timeout of 30 seconds
+            }
+
+            void set_data_rate_timeout() override {
+                // If the download speed is less than 17KB/sec for more than a minute, timout. This time was selected because it should ensure that downloading each megabyte take no more than a minute.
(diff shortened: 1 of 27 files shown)


seguler:
Have you tested this ?

rickle-msft:
My how the tables have turned!! Using the proxy, I tested the absolute timeouts. There's not a good way for me to let data pass through at a specific rate to test the data-rate timeout, but since it's just a curl option that I implemented following their sample and it's pretty straightforward, I trusted that would work very similarly.

azure-storage-cpp-lite/include/http/libcurl_http_client.h
```diff
@@ -148,6 +148,16 @@ namespace microsoft_azure {
                 return m_input_stream;
             }
 
+            void set_absolute_timeout() override {
+                check_code(curl_easy_setopt(m_curl, CURLOPT_TIMEOUT, 30L)); // Timeout of 30 seconds
```
seguler: I think for operations like get_blob_properties we should set this to 5 second. 30 second for those atomic operations is too long

asorrin-msft: I don't know, 30 seconds is fairly standard for us.  It shouldn't be common, but if we get unlucky and hit a partition move, more than 5 seconds can happen

seguler: ok - if it is standard across other things, let's keep it

seguler: Well okay - isn't it actually faster to time out at 5-10 seconds, and then let it retry ? :)

PR #147, opened 2018-03-21
Exponential retries

rickle-msft:
With a max delay of 60s. All retries and delays included, a single operation should max out at about an hour.

### DIFF
--- a/azure-storage-cpp-lite/include/retry.h
+++ b/azure-storage-cpp-lite/include/retry.h
@@ -67,8 +67,10 @@ namespace microsoft_azure {
             retry_info evaluate(const retry_context &context) const override {
                 if (context.numbers() == 0) {
                     return retry_info(true, std::chrono::seconds(0));
-                } else if (context.numbers() < 100 && can_retry(context.result())) {
-                    return retry_info(true, std::chrono::seconds(5+rand()%5));
+                } else if (context.numbers() < 300 && can_retry(context.result())) {
+                    double delay = (pow(2, context.numbers()-1)-1) * 4;
+                    delay *= (((double)rand())/RAND_MAX)/2 + 0.8;
+                    return retry_info(true, std::chrono::seconds((int)delay));
                 }
                 return retry_info(false, std::chrono::seconds(0));
             }
--- a/azure-storage-cpp-lite/include/retry.h
+++ b/azure-storage-cpp-lite/include/retry.h
@@ -1,6 +1,7 @@
 #pragma once
 
 #include <chrono>
+#include <algorithm>
 
 #include "storage_EXPORTS.h"
 
@@ -67,9 +68,10 @@ namespace microsoft_azure {
             retry_info evaluate(const retry_context &context) const override {
                 if (context.numbers() == 0) {
                     return retry_info(true, std::chrono::seconds(0));
-                } else if (context.numbers() < 300 && can_retry(context.result())) {
-                    double delay = (pow(2, context.numbers()-1)-1) * 4;
+                } else if (context.numbers() < 60 && can_retry(context.result())) {
+                    double delay = (pow(2, context.numbers()-1)-1);
                     delay *= (((double)rand())/RAND_MAX)/2 + 0.8;
+                    delay = std::min(delay, 60.0); // Maximum backoff delay of 1 minute
                     return retry_info(true, std::chrono::seconds((int)delay));
                 }
                 return retry_info(false, std::chrono::seconds(0));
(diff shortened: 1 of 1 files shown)


azure-storage-cpp-lite/include/retry.h
```diff
@@ -67,8 +68,11 @@ namespace microsoft_azure {
             retry_info evaluate(const retry_context &context) const override {
                 if (context.numbers() == 0) {
                     return retry_info(true, std::chrono::seconds(0));
-                } else if (context.numbers() < 100 && can_retry(context.result())) {
-                    return retry_info(true, std::chrono::seconds(5+rand()%5));
+                } else if (context.numbers() < 60 && can_retry(context.result())) {
```
seguler: I propose we reduce this to 10-20 retries, because with backoff implemented, retries will now take much more longer... Nothing should retry for more than a few minutes... What do you all think ?

rickle-msft: When Adam and I were discussing, we were thinking we reaaaally don't want to fail if at all possible since it can be tough for the user to figure out what to do. These retries will max out at a little less than an hour, which we thought would be more important for large data transfers

rickle-msft: As it stands now, yes.

azure-storage-cpp-lite/include/retry.h
```diff
@@ -67,8 +68,11 @@ namespace microsoft_azure {
...
+                    double delay = (pow(2, context.numbers()-1)-1);
```
asorrin-msft: I think we should reduce the base here, maybe 1.2?  2^x can grow quite quick, I think it would be good to have a few quick retries before starting to get into longer backoffs.

azure-storage-cpp-lite/include/retry.h
```diff
@@ -67,8 +68,11 @@ namespace microsoft_azure {
...
+                    delay *= (((double)rand())/RAND_MAX)/2 + 0.8;
```
asorrin-msft: Minor - this should probably be +0.75 :)  +0.8 puts the jitter in [0.8, 1.3], which biases above 1 (again though, not that it really matters here :) )

azure-storage-cpp-lite/include/retry.h
```diff
@@ -67,8 +68,11 @@ namespace microsoft_azure {
...
+                    delay = std::min(delay, 60.0); // Maximum backoff delay of 1 minute
```
asorrin-msft: Can you move the jitter after the min(delay, 60) check?  This also doesn't matter too much, but I'd feel a bit more comfortable if the max-retry case has the jitter as well

azure-storage-cpp-lite/include/retry.h
asorrin-msft: Depends on whether failure is worse than delay... the issue for me is that consumers have very little they can do in the form of error-recovery (especially for lift-and-shift apps, written assuming a local HD).  If there's any chance at all of recovery, it might be worth it.  OTOH, if there's a nonrecoverable network issue, failing faster is always better...

How about we change the base to 1.2 (see below), and set the max retry count to 26?  If my math is right, this would give a max of 10 minutes before retries are done, give or take from jitter

seguler: I think it is a good idea. If it takes more than 10 minutes, user should consider a different solution (sharding data across storage accounts)

asorrin-msft: So, I just realized - in this case, 10 minutes is all the time spent waiting.  This doesn't include time spent inside the actual request.  Worst case - each try in a non-data-transfer operation takes the full 30 seconds to fail before retry, that adds 30 seconds to each, which would total 10 min + 26/2 = 23 minutes total.  (Data-transfer operations are practically unbounded, pathological worst case at 55 days per try, but I'm not as concerned about this.)

AlexanderYukhanov: unfortunately, the user has no way to terminate the operation if this happened. he cannot even kill his process. may be it's better to make retries configurable and fail fast by default?

azure-storage-cpp-lite/include/retry.h
asorrin-msft: Yes, it caps out at 60 seconds

azure-storage-cpp-lite/include/retry.h
asorrin-msft: Making retries configurable is a good idea, but I think we should do that in a separate commit/PR, if that's ok?

azure-storage-cpp-lite/include/retry.h
AlexanderYukhanov: I think 0.75 will be easier to reason about than 0.8. And it would be nice to give them names

seguler:
lgtm

PR #155, opened 2018-03-30
Timeout

asorrin-msft:
This is a minor change on top of https://github.com/Azure/azure-storage-fuse/pull/146/files.  It sets different timeouts on different operations (5 seconds for most operations, 30 seconds for listing & larger-payload operations.)  Logic for data operations (put blob, put block, get blob) is unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,78 +12,75 @@ Please note that this tool is currently in Preview, and need your feedback for i
 - Parallel download and upload features for fast access to large blobs
 - Allows multiple nodes to mount the same container for read-only scenarios.
 
-## Considerations
-Please take careful note of the following points, before using blobfuse:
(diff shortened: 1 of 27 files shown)


AlexanderYukhanov:
are timeout setting autoreset after a request has been completed?

seguler:
I am not sure what you mean @AlexanderYukhanov - why would we reset any timeout setting ? Isn't it an absolute number every time ? E.g. 5 second for each get blob properties request.

seguler:
@asorrin-msft @rickle-msft ; Alex has a good point. This curl handle is being reused all the time... Does the timeout get reset ? For instance, if you have listed blobs in 10 seconds, and then used the same handle for a large file upload. Would it timeout ? Has this been tested ?

asorrin-msft:
Latest commit should fix any potential issues by resetting the values to zero (meaning "no timeout", according to libcurl docs)

AlexanderYukhanov:
LGTM

seguler:
lgtm

Issue #158, opened 2018-04-03
Compile on Debian 9


difrancosimone:
Hello!

i got an error compiling on Debian 9 64bit, on retry.h file.
The error is: 
retry.h:72:65: error: ‘pow’ was not declared in this scope
                     double delay = (pow(1.2, context.numbers()-1)-1);

i solved adding the

    #include <math.h>

line to the retry.h file at line 11.

I hope this might help.

seguler:
I have also just encountered this issue on Ubuntu 17.10 as well

seguler:
Thanks resolved!

PR #166, opened 2018-04-16
File handle leak fix, debugging improvements

asorrin-msft:
- Fixing some slow file handle leaks
- Adding some cpp-lite logging statements
- Changing the build to include debug symbols

### DIFF
--- a/azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
+++ b/azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
@@ -7,6 +7,7 @@
 #include <fcntl.h>
 #include <iostream>
 #include <fstream>
+#include <uuid/uuid.h>
 
 #include "blob/blob_client.h"
 #include "storage_errno.h"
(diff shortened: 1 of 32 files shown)


seguler:
issue #158 is not fixed - can we include that too ?

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
```diff
@@ -149,7 +149,7 @@ namespace microsoft_azure {
             }
             catch(const std::exception &ex)
             {
-                std::cerr << ex.what() << std::endl;
+                syslog(LOG_ERR, "Unknown failure in blob_client_wrapper_init.  ex.what() = %s.", ex.what());
                 errno = unknown_error;
```
Mic92: Why is the error set to unknown here if the exception tells exactly what happens here?

azure-storage-cpp-lite/src/blob/blob_client_wrapper.cpp
```diff
@@ -149,7 +149,7 @@ namespace microsoft_azure {
             }
             catch(const std::exception &ex)
             {
-                std::cerr << ex.what() << std::endl;
+                syslog(LOG_ERR, "Unknown failure in blob_client_wrapper_init.  ex.what() = %s.", ex.what());
```
Mic92: It is not unknown. The exact error was catched. Just say `"Failed to create blob client: %s", ex.what()`

asorrin-msft: Thanks!  Latest commit changes it to: 
syslog(LOG_ERR, "Failed to create blob client.  ex.what() = %s.", ex.what());

asorrin-msft:
Sercan - should be done with latest commit, but I haven't tested on Debian.  Worked on Ubuntu 17, although not this exact commit hash, so we should try that again too.

PR #167, opened 2018-04-17
Use env to locate bash

jbgi:


asorrin-msft:
Thanks!  Merging now.  PR's in general to blobfuse should generally be submitted against the dev branch, but this is fine :)

PR #179, opened 2018-05-08
Fixing blob deletion failure reporting

AlexanderYukhanov:
Fixed:
1. wrong logging of local file deletion status
2. fixed blob deletion status check
3. now azs_unlink will report an error if blob deletion actually failed


seguler:
@rickle-msft can you review and merge ?

PR #183, opened 2018-05-14
Building azure-storage-cpp-lite standalone

damienpontifex:
I was interested in using just the azure-storage-cpp-lite code but ran into a few build issues on my mac. Hopefully this is a good change for the project as well, but focuses around error fixes on build and reduced dependency requirement for that component

- azure-storage-cpp-lite links against OpenSSL so use that to do HMAC hash and able to remove additional dependency on GnuTLS
- Build also failed with a few issues just from overrides and unused lambda captures
- Also building sample with cmake

Steps to reproduce build fail:
```bash
cd azure-storage-cpp-lite
mkdir build && cd build
cmake ..
make azure-storage
```


seguler:
Awesome! You're great! 

Couple of issues:
- We intentionally moved to GnuTLS because its license plays better with the license we have. We cannot accept the change to OpenSSL
- Why was there a need to upgrade to Cmake 3.5 ? Most users on older distros only have Cmake 2.8.

Let's wait for others to review as well. This is very helpful since we have a plan to move this library out of this repository in the future. It is a great start.

damienpontifex:
@seguler Added an option to use OpenSSL which defaults to the existing configuration of GnuTLS. Given the default build would be with GnuTLS is this acceptable within the licensing issues you mentioned? Not sure if it covers any inclusion or just the use of it in the final build product.

damienpontifex:
@seguler any update of feedback on this?

seguler:
I think this looks good. OpenSSL license does not allow us to ship binaries that link with OpenSSL. Your PR is good, I think. 

@asorrin-msft and @JasonYang-MSFT for approval

damienpontifex:
@seguler final point: did you want me to standardise the CMake version to 2.8 for both azure-storage-fuse/CMakeLists.txt and azure-storage-cpp-lite/sample/CMakeLists.txt ?

seguler:
@damienpontifex - just encountered your fork of tensorflow. Are you going to use this library there ? That'd be good, because this is a library we will support going forward.

damienpontifex:
@seguler yes, that is the plan. Creating a [Custom Filesystem Plugin](https://www.tensorflow.org/extend/add_filesys) for azure blob storage. See WIP issue I created here https://github.com/tensorflow/tensorflow/issues/18852 

I had started that work using the curl library. I got it implemented and ironing out use cases until I found this library. Started some work now to remove my custom code and use this library. Much nicer 😄

damienpontifex:
@seguler here's the work to use this library for the file system plugin https://github.com/damienpontifex/tensorflow/commits/az-storage-fuse

Currently have the bazel dependency pointed to my fork + commit so I can build using TensorFlow's existing dependency on boringssl (i.e. their openssl fork) but wouldn't make the PR back to TensorFlow without changing that to point to this as the main repo, so looking forward to seeing if these changes are accepted.

damienpontifex:
@asorrin-msft and @JasonYang-MSFT just checking in to see if there's any feedback on this?

seguler:
All good - thanks for the contribution @damienpontifex

PR #187, opened 2018-06-07
Fix catch value warning

mhsjlw:
Fixes compilation on my machine (as warnings are treated as errors)


seguler:
All good - thanks for the contribution!

PR #198, opened 2018-07-11
Adding a sample AzureUSGovernment config file with blobEndpoint config example

daweins:
I'll add some text in the wiki documentation to reference & explain this as well. Thank you for adding the Sovereign Cloud capability - I've had customers complain that the behavior for the blobEndpoint config is not well documented, so I'm attempting to flesh it out.


seguler:
Thanks @daweins

PR #205, opened 2018-08-15
linking uuid library at client target level for standalone build

damienpontifex:
Had already looked at standalone build of azure-storage-cpp-lite in #183 but found uuid dependency wasn't included at this level so would fail to build this target on ubuntu

### DIFF
--- a/azure-storage-cpp-lite/CMakeLists.txt
+++ b/azure-storage-cpp-lite/CMakeLists.txt
@@ -116,6 +116,7 @@ else(MSVC)
   find_package(Threads REQUIRED)
   find_package(CURL REQUIRED)
   find_package(OpenSSL REQUIRED)
+  pkg_search_module(UUID REQUIRED uuid)
 
   if (NOT ${USE_OPENSSL})
     find_package(GnuTLS REQUIRED)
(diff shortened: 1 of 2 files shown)


azure-storage-cpp-lite/CMakeLists.txt
```diff
@@ -137,7 +138,7 @@ add_library(azure-storage SHARED ${AZURE_STORAGE_HEADER} ${AZURE_STORAGE_SOURCE}
 if(MSVC)
   target_link_libraries(azure-storage bcrypt libcurl)
 else(MSVC)
-  target_link_libraries(azure-storage ${CURL_LIBRARIES} ${OPENSSL_CRYPTO_LIBRARY} ${OPENSSL_SSL_LIBRARY} ${EXTRA_LIBRARIES})
+  target_link_libraries(azure-storage ${CURL_LIBRARIES} ${UUID_LIBRARIES} ${OPENSSL_CRYPTO_LIBRARY} ${OPENSSL_SSL_LIBRARY} ${EXTRA_LIBRARIES})
```
katmsft: I think in order to make UUID library available for the project, `$(UUID_INCLUDE_DIR)` should also be added to L136`add_library(azure-storage SHARED ${AZURE_STORAGE_HEADER} ${AZURE_STORAGE_SOURCE})`

damienpontifex: I originally thought the same, but looking at the CMakeLists.txt in the root directory, it only had the `${UUID_LIBRARIES}` added and not the `${UUID_INCLUDE_DIR}`, so did the same here. Thoughts?

seguler: I think our CMakelist is incorrect in the root. We use uuid.h, so we should add include_dir.. @asorrin-msft  could you confirm ?

damienpontifex: I've updated the storage library and root with UUID_INCLUDE_DIR.

I think there's also missing search for gcrypt and fuse header files etc as I also faced not found issues of those on a fresh ubuntu docker container I was using to build. Thought I'd add those to a separate PR if agreed

PR #212, opened 2018-08-26
Adding Symlink support

seguler:
- Implemented symlink and readlink functions
- Added support for symlinks in readdir. Readdir returns d_type=S_IFLNK for symlinks
- Symlink blobs are identified with metadata 'is_symlink=true'
- Symlink blobs contain the destination path as a plain text in the blob
- Implemented tests for symlink as a directory and as a file

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,0 +1,17 @@
+### Which version of the blobfuse was used?
+
+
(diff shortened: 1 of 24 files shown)


james-callahan:
Is this likely to be merged? What is the current status?

vibhansa-msft:
Taken care of as part of ADLS and cpplite upgrade part. symlink are supported with that change-set.

PR #214, opened 2018-08-31
Enable blobfuse to work with NFS

### LINKED ISSUE #213: NFS unsupported by azure-storage-fuse
### The Problem

Exporting a blobfuse folder as an NFS export does not fully work.  Files written from an NFS client can be seen for a short while by all NFS clients and also show up in the blobfuse temporary cache.  However, they are never uploaded to the storage account.  The blobfuse GC process eventually removes them.

A dump of the blobfuse debug logs shows that the ```azs_flush``` callback is never called from NFS, and since blobfuse only uploads the blob in the ```azs_flush```, this explains why the blob is never uploaded.  This is confirmed on the SSHFS Fuse driver mailing list (https://sourceforge.net/p/fuse/mailman/message/8901469/).  SSHFS calls ```flush``` from ```release```, and then uploads the file at the end of the ```release``` call: https://github.com/libfuse/sshfs/blob/master/sshfs.c#L2664-L2676.

### The Repro

1. Setup Blob Fuse on Ubuntu 18.04
2. Export the share using the "/etc/exports" and don't forget to add ```fsid=NNN``` where ```NNN``` can be any number (example: ```echo -e "\n${FUSE_BASE}   *(rw,fsid=1,sync,no_root_squash)" >> /etc/exports```) and restart nfs server.
3. mount using some NFS clients, and write some files.  The files will be seen by all NFS clients and the blobfuse cache, but never uploaded.  The blobfuse GC will kick in and eventually remove the files.

### Solution

A potential solution is following what was done by SSHFS and either call ```azs_flush``` from```azs_release```, or upload the blob in the release.

Ubuntu 18.04 (https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb)

anhowe:
Fixes #213.

Similar to SSHFS, a flush is added to the release.  A guard based on mtime ensures the same file is not uploaded more than once.  This has been tested locally and through NFS.

### DIFF
--- a/blobfuse/blobfuse.h
+++ b/blobfuse/blobfuse.h
@@ -100,9 +100,10 @@ struct fhwrapper
 {
     int fh; // The handle to the file in the file cache to use for read/write operations.
     bool upload; // True if the blob should be uploaded when the file is closed.  (False when the file was opened in read-only mode.)
+    uint32_t last_upload_mtime; // The modified time of the file since last upload.  This avoids uploading a file we have already uploaded.
     fhwrapper(int fh, bool upload) : fh(fh), upload(upload)
     {
-
+        last_upload_mtime = 0;
     }
 };
 
--- a/blobfuse/fileapis.cpp
+++ b/blobfuse/fileapis.cpp
@@ -327,6 +327,18 @@ int azs_flush(const char *path, struct fuse_file_info *fi)
                 }
             }
 
+            // skip upload of unchanged file
+            if (buf.st_mtime > ((struct fhwrapper *)fi->fh)->last_upload_mtime)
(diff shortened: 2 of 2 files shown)


marvinpuethe:
Is this going to be merged? Works perfectly for me.

vibhansa-msft:
We have recently previewed our new generation of blobfuse (https://github.com/Azure/azure-storage-fuse/releases/tag/blobfuse2-2.0.0-preview.1)
It handles this condition where 'flush' is not called by the system and instead 'close' is hit directly. We will not be porting this feature to blobfuse hence closing this request,

PR #225, opened 2018-10-05
Adding ability to use OAuth access token as credential to storage library

damienpontifex:
Adding [Azure Active Directory auth](https://docs.microsoft.com/en-us/azure/storage/common/storage-auth-aad) as a storage credential option. Similar naming convention to node and C# libraries that have a current implementation

### DIFF
--- a/azure-storage-cpp-lite/include/storage_credential.h
+++ b/azure-storage-cpp-lite/include/storage_credential.h
@@ -75,6 +75,8 @@ namespace microsoft_azure {
                 return m_token;
             }
 
+            void set_token(const std::string& token);
+
         private:
             std::string m_token;
         };
--- a/azure-storage-cpp-lite/include/storage_credential.h
+++ b/azure-storage-cpp-lite/include/storage_credential.h
@@ -65,5 +65,19 @@ namespace microsoft_azure {
             void sign_request(const storage_request_base &, http_base &, const storage_url &, const storage_headers &) const override {}
         };
 
+        class token_credential : public storage_credential {
+        public:
+            AZURE_STORAGE_API token_credential(const std::string &token);
+
+            void sign_request(const storage_request_base &, http_base &, const storage_url &, const storage_headers &) const override;
+
+            const std::string &token() const {
+                return m_token;
+            }
+
+        private:
+            std::string m_token;
+        };
+
     }
 }
--- a/azure-storage-cpp-lite/src/storage_credential.cpp
+++ b/azure-storage-cpp-lite/src/storage_credential.cpp
@@ -80,5 +80,12 @@ namespace microsoft_azure {
             std::string transformed_url = transform_url(h.get_url());
             h.set_url(transformed_url);
         }
+
+        AZURE_STORAGE_API token_credential::token_credential(const std::string &token)
+            : m_token(token) {}
+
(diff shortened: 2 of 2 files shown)


seguler:
Wohooo! Thank you Damien!

damienpontifex:
@seguler thought it could also be convenient to integrate a convenience function for calling the GET request to 'http://169.254.169.254/metadata/identity/oauth2/token?api-version=2018-02-01&resource=https%3A%2F%2Fstorage.azure.com%2F' on an Azure VM to then fill the token but 
1. Didn't want to overload this PR
1. Not sure if it's something that would want to be integrated into this library or just leave it purely independent for the developer

seguler:
I am not sure, I've pinged other experts to respond on this. Meanwhile what's missing seems to be a way to refresh the token after it expires. We usually have a update method on the token credential in our other SDKs (e.g. Java: https://github.com/Azure/azure-storage-java/blob/ebdc21465a57ed5cb6fdd119848884dc1bb87d34/src/main/java/com/microsoft/azure/storage/blob/TokenCredentials.java#L64). Could you also add an update_token method ?

I am thinking the user will have to make the call to MSI, get the token... and also create a thread to go off and refresh the token before it times out, which should call the update_token on the token_credential...

katmsft:
The PR overall seems to be good to me, however, regarding the refreshing token mechanism I have some comments.
Current implementation would require the user to monitor the asynchronous task that is using the OAuth token, and ensure the task always has valid token if the task sends multiple request during its lifetime, otherwise there would be an authentication failure.
A more convenient way is for us to provide a callback as parameter that user can set which gets the token when called. During every sign step, the SDK evaluates if the token is expired, if it is, the callback is called updating the token.
I believe several major Azure client library has implement the Oauth token support in this way, and it would be better that the azure-storage-cpplite follows the design.

seguler:
@JeffreyRichter - could you please advise on the OAuth refresh token design in the C++ SDK ? I thought we left the update of token to user to handle in all the SDKs. Am I mistaken ? What is the approach to take and be consistent across all Azure SDKs ?

JeffreyRichter:
I've not had any involvement with the C/C++ SDK. However, for all the SDKs I have been involved with the design had been that the TokenCredential type is created with a token and a callback method. The callback is involved periodically so that there token can be refreshed and updated inside the TokenCredential object. By design, our Storage SDKs do not take any hard dependency on any library or mechanism for acquiring or refreshing a token. 

Some other SDKs do check if the token is close to expiring on each I/O operation and, if so, involved a callback to refresh the token. I think this is a bad design because this does not prevent or guarantee that the token is always valid and so all user code that makes an I/O request should be written in such a way as to recover from this failure. In addition, if multiple threads see an expiring token, then the callback method must be implemented in a thread safe way which greatly complicates the user's callback and may end up blocking threads which hurts resource consumption and scalability. For these reasons, we choose a different, simpler, more efficient and scalable design.

damienpontifex:
Maybe this PR covers the most basic case and could be built up with another to add better handling of this mechanism? 

If wanting a callback mechanism, are there links to other implementations on the reference to the appropriate behaviour to ensure consistency and provide guidance?

damienpontifex:
@seguler where did we land around this?

damienpontifex:
Closing as https://github.com/Azure/azure-storage-cpplite seems to be cpp storage repo now so not sure if this will change usage of cpplite sources in this repo

PR #226, opened 2018-10-05
Azure Pipelines build

damienpontifex:
Azure pipelines build that can be used to add CI and PR build validation 
- Build libazure-storage.so standalone
- Build blobfuse


damienpontifex:
Closing as https://github.com/Azure/azure-storage-cpplite seems to be cpp storage repo now so not sure if this will change usage of cpplite sources in this repo

Issue #200, opened 2018-07-16
connection.cfg file encryption


ghost:
In connection.cfg file we mentioned account key and that is in plain text. Is there any way to encrypt this account key to keep secure.

![image](https://user-images.githubusercontent.com/22212578/42763670-65d84636-8931-11e8-9b41-a012632a5715.png)

seguler:
Nope - doesn't support encryption. If you are looking to secure things, the best practice is to put these into environment variables and mount blobfuse using env variables. Only your user will be able to access these variables.

vijayraje23:
@seguler Thanks for quick response. I found one workaround and it worked as per need. Please review it and let me know if this is the correct way to do it.

#**Create mount.sh file to encrypt and decrypt account key.**
cd /etc/vsftpd

#**generate a 2048-bit RSA key and store it in rsakey.txt.**
openssl genrsa -out .rsakey.txt 2048

**#generate encrypted account key using the RSA key and store it in accountkey.txt.**
echo "insert the actual key taken from azure portal" | openssl rsautl -inkey .rsakey.txt -encrypt >.accountkey.txt

**vi /etc/vsftpd/mount.sh**
#!/bin/bash
file=$(openssl rsautl -inkey .rsakey.txt -decrypt <.accountkey.txt)
echo "accountName blobftp" >/etc/vsftpd/.connection.cfg
echo "accountKey $file" >>/etc/vsftpd/.connection.cfg
echo "containerName blobfuseftp" >>/etc/vsftpd/.connection.cfg

blobfuse /blobftp/ftp/localuser --tmp-path=/mnt/blobfusetmp --config-file=/etc/vsftpd/.connection.cfg  -o allow_other -o nonempty
sleep 2
echo "#Decrypting AccountKey in .connection.cfg file"
echo "accountName blobftp" >/etc/vsftpd/.connection.cfg
echo "accountKey encrypted" >>/etc/vsftpd/.connection.cfg
echo "containerName blobfuseftp" >>/etc/vsftpd/.connection.cfg

**set this mount.sh script file to run after reboot. so now my .connection.cfg file looks like encrypted.**
![image](https://user-images.githubusercontent.com/22212578/42796793-1caaaa70-895a-11e8-815f-ee36c30d664d.png)



Thanks again !!!!

seguler:
I am not sure what the aim here is. If some intruder has access to your machine/your user, they can decrypt this key just like you are doing. Right ?

[...]

PR #229, opened 2018-10-14
connection.cfg file encryption

vijayraje23:
Can we have auto encryption for connection.cfg file as this file contain important information about storage.

I had discussed this issue and asked me to create PR for the same. Please refer below link for more info.

https://github.com/Azure/azure-storage-fuse/issues/200


seguler:
Thanks - I meant send a PR with the feature implemented. But that is ok.

PR #234, opened 2018-10-26
Added azure-storage-cpplite as submodule

katmsft:


vibhansa-msft:
Taken care as part of ADLS and cpplite upgrade PR.

PR #258, opened 2019-03-22
Fix by value exception catching

johnBuffer:
Fix compilation error with GCC 8.2.1 :

> error: catching polymorphic type 'class std::exception' by value [-werror=catch-value=]

Compiled successfully after modification


johnBuffer:
Close because of wrong source branch

PR #259, opened 2019-03-22
Fix by value exception catching

johnBuffer:
Fix compilation error with GCC 8.2.1 :

> error: catching polymorphic type 'class std::exception' by value [-werror=catch-value=]

Compiled successfully after modification


seguler:
@johnBuffer - could you accept the following contribution license agreement ? Once you do that, we'll merge this.

https://cla.opensource.microsoft.com/Azure/azure-storage-fuse?pullRequest=259

johnBuffer:
@seguler apparently I already did, when I follow your link everything is already filled

PR #274, opened 2019-04-29
Automatic garbage collection on cache disk when disk space is full

seguler:


amnguye:
Pull request merged by PR #300

PR #312, opened 2019-08-30
Update connection.cfg

flybd5:
Correcting reference to what to put in


vibhansa-msft:
Suggested change has already been integrated in master.

PR #336, opened 2019-12-03
Support Managed Identity Authentication in blobfuse

adreed-msft:


blobfuse/OAuthTokenCredentialManager.cpp
```diff
@@ -0,0 +1,216 @@
...
+
+/// <summary>
+/// SetUpMSICallback sets up a refresh callback for MSI auth. This should be used to create a OAuthTokenManager instance.
+/// </summary>
+std::function<OAuthToken(std::shared_ptr<CurlEasyClient>)> SetUpMSICallback(std::string client_id_p, std::string object_id_p, std::string resource_id_p)
+{
+    // Create the URI token request
+    std::shared_ptr<microsoft_azure::storage::storage_url> uri_token_request_url = std::make_shared<microsoft_azure::storage::storage_url>();
+    uri_token_request_url->set_domain(constants::msi_request_uri);
+    uri_token_request_url->append_path(constants::msi_request_path);
```
vivram: Can you combine msi_request_uri and msi_request_path and instead use a single variable? The default constant should be used only if the environment variable MSI_ENDPOINT is set. Consider using the same approach as [here](https://github.com/Azure/azure-sdk-for-net/blob/b0ff97d220ec1e83d43767d693b17a3b6245314e/sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs#L87) to construct this endpoint url.

blobfuse/OAuthTokenCredentialManager.cpp
```diff
@@ -0,0 +1,216 @@
...
+    {
+        uri_token_request_url->add_query(constants::param_mi_res_id, resource_id_p);
+    }
+
+    return [uri_token_request_url](std::shared_ptr<CurlEasyClient> httpClient) {
+        // prepare the CURL handle
+        std::shared_ptr<CurlEasyRequest> request_handle = httpClient->get_handle();
+
+        request_handle->set_url(uri_token_request_url->to_string());
+        request_handle->add_header(constants::header_metadata, "true");
```
vivram: Add MSI_SECRET header if the environment variable is set similar to what is done [here](https://github.com/Azure/azure-sdk-for-net/blob/b0ff97d220ec1e83d43767d693b17a3b6245314e/sdk/identity/Azure.Identity/src/ManagedIdentityClient.cs#L325)

.gitignore
```diff
@@ -50,3 +50,11 @@ modules.order
 Module.symvers
 Mkfile.old
 dkms.conf
+cmake-build-debug/
+cmake-build-debug/*
+cmake-build-release/
+cmake-build-release/*
```
adreed-msft: This can be simplified to `cmake-build-*/*`

azure-storage-cpp-lite/include/storage_credential.h
```diff
@@ -43,27 +46,51 @@ namespace microsoft_azure {
...
         public:
             void sign_request(const storage_request_base &, http_base &, const storage_url &, const storage_headers &) const override {}
         };
 
+        class token_credential : public storage_credential {
+        public:
+            AZURE_STORAGE_API token_credential();
+            // DEPRECATION NOTICE: This is mostly meant for compatibility with sample.cpp
+            // The new token credential uses the OAuthTokenCredentialManager, and explicitly set tokens disable this functionality!
+            AZURE_STORAGE_API token_credential(const std::string &token);
```
adreed-msft: TODO: Perhaps mark changes made within CPPlite for the sake of blobfuse in comments so that when it comes time to merge in a new version of CPPlite we can pretty easily decipher what needs to stay and what gets replaced

azure-storage-cpp-lite/src/storage_credential.cpp
```diff
@@ -76,9 +86,46 @@ namespace microsoft_azure {
...
+                const storage_url &,
+                const storage_headers &) const
+        {
             std::string transformed_url = transform_url(h.get_url());
             h.set_url(transformed_url);
         }
+
+        AZURE_STORAGE_API token_credential::token_credential(){
+            m_credmgr_ptr = GetTokenManagerInstance(EmptyCallback);
+        }
```
adreed-msft: TODO: Add comment expecting a token manager instance to already have been initialized, for usability purposes.

azure-storage-cpp-lite/src/storage_url.cpp
```diff
@@ -110,5 +110,92 @@ namespace microsoft_azure {
...
+                            output->set_domain(std::string(runningString));
+                            // empty the buffer, do not append the new char to the string because storage_url handles it for us, rather than checking itself
+                            runningString.clear();
+                            segment++;
+                        }
+                        else
+                        {
+                            runningString += *charptr;
+                            if (charptr == url.end() - 1)
+                                goto domainfinalchar; // Because these jumps are local, it's arguably still easy to debug.
```
adreed-msft: Actually, this can be wildly simplified, in the same way that segment 0 is handled.

TODO: Simplify this code that way.

blobfuse/OAuthTokenCredentialManager.cpp
```diff
@@ -0,0 +1,226 @@
...
+/// Returns current oauth_token, implicitly refreshing if the current token is invalid.
+/// Note that this can throw an error if the refresh fails, so be prepared to catch.
+/// </summary>
+OAuthToken OAuthTokenCredentialManager::get_token()
+{
+    if (is_token_expired()) {
+        if (token_mutex.try_lock()) {
+            try {
+                refresh_token();
+            } catch (std::exception &ex) {
```
adreed-msft: >std::exception &e [](start = 21, length = 17)

Being passed through multiple exception covers like this seems to make ex.what() turn into "std::exception()". TODO: Make the actual error get dropped, rather than that.

connection.cfg
```diff
@@ -1,5 +1,5 @@
 accountName <account-name-here>
 # Please provide either an account key or a SAS token, and delete the other line.
-accountKey <account-name-here-delete-next-line>
+accountKey <account-key-here-delete-next-line>
 sasToken <shared-access-token-here-delete-previous-line>
-containerName <container-name-here>
```
amnguye: Keep containerName in the connection.cfg file

README.md
```diff
@@ -49,11 +49,49 @@ For more information, see the [wiki](https://github.com/Azure/azure-storage-fuse
...
+    * `AZURE_STORAGE_IDENTITY_RESOURCE_ID`: Only one of these three parameters are needed if multiple identities are present on the system.
+    * `AZURE_STORAGE_MANAGED_IDENTITY_ENDPOINT`: Specifies a custom managed identity endpoint, as IMDS may not be available under some scenarios. Uses the `AZURE_STORAGE_IDENTITY_CLIENT_ID` parameter as the `Secret` header.
+
+### Config file options
+
+- General options
+    * `accountName`: Specifies the storage account blobfuse targets.
+    * `blobEndpoint`: Specifies the blob endpoint to use. Defaults to *.blob.core.windows.net, but is useful for targeting storage emulators.
+    * `authType`: Overrides the currently specified auth type. Options: Key, SAS, MSI
+
```
amnguye: Saw a TODO in the code to make this type insensitive (lower,upper). So maybe we should add something here that says to make sure you get the cases right or else blobfuse won't know what you're talking about. We can remove it later after we get it type insensitive

blobfuse/blobfuse.cpp
```diff
@@ -169,13 +255,18 @@ int read_config(const std::string configFile)
         fprintf(stderr, "Account name is missing in the config file.\n");
         return -1;
     }
-    else if((str_options.accountKey.empty() && str_options.sasToken.empty()) || 
-	    (!str_options.accountKey.empty() && !str_options.sasToken.empty()))
+    /// TODO: make a check for checking if some parameters are used
+    /*
```
amnguye: should we still do this in this PR?

blobfuse/blobfuse.cpp
```diff
@@ -191,15 +282,92 @@ int read_config(const std::string configFile)
 
 void *azs_init(struct fuse_conn_info * conn)
 {
+    const int defaultMaxConcurrency = 20;
```
amnguye: This is probably on me, but we should move this to constants and replace all the magic 20s with this value.

blobfuse/blobfuse.cpp
```diff
@@ -468,8 +636,51 @@ int validate_storage_connection()
...
+            std::function<OAuthToken(std::shared_ptr<CurlEasyClient>)> MSICallback = SetUpMSICallback(
+                    str_options.clientId,
+                    str_options.objectId,
+                    str_options.resourceId,
+                    str_options.msiEndpoint);
+
+            std::shared_ptr<OAuthTokenCredentialManager> tokenManager = GetTokenManagerInstance(MSICallback);
+
+            if (!tokenManager->is_valid_connection()) {
+                // todo: isolate definitions of errno's for this function so we can output something meaningful.
```
amnguye: I agree with this. Let's talk about this later

PR #337, opened 2019-12-03
Support Service Principal Name Authentication in blobfuse

adreed-msft:
This PR is in conjunction with MSI authentication, and commits including and past 63f791: "Support SPN based authentication" should be reviewed in this PR.

### DIFF
--- a/blobfuse/OAuthTokenCredentialManager.cpp
+++ b/blobfuse/OAuthTokenCredentialManager.cpp
@@ -265,7 +265,7 @@ std::function<OAuthToken(std::shared_ptr<CurlEasyClient>)> SetUpMSICallback(std:
     };
 }
 
-std::function<OAuthToken(std::shared_ptr<CurlEasyClient>)> SetUpSPNCallback(std::string tenant_id_p, std::string client_id_p, std::string client_secret_p)
+std::function<OAuthToken(std::shared_ptr<CurlEasyClient>)> SetUpSPNCallback(std::string tenant_id_p, std::string client_id_p, std::string client_secret_p, std::string aad_endpoint_p)
 {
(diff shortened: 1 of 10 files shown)


blobfuse/OAuthTokenCredentialManager.cpp
```diff
@@ -230,6 +264,88 @@ std::function<OAuthToken(std::shared_ptr<CurlEasyClient>)> SetUpMSICallback(std:
...
+
+                try {
+                    std::string json_request_result(std::istreambuf_iterator<char>(ios.istream()),
+                                                    std::istreambuf_iterator<char>());
+                    req_result = json_request_result;
+                } catch(std::exception&) {}
+
+                std::ostringstream errStream;
+                errStream << "Failed to retrieve OAuth Token (CURLCode: " << curl_code << ", HTTP code: " << http_code_result << "): " << req_result;
+                throw std::runtime_error(errStream.str());
```
amnguye: We should also log this error in both syslog and to user

blobfuse/OAuthTokenCredentialManager.cpp
```diff
@@ -230,6 +264,88 @@ std::function<OAuthToken(std::shared_ptr<CurlEasyClient>)> SetUpMSICallback(std:
...
+            } else {
+                std::string json_request_result(std::istreambuf_iterator<char>(ios.istream()),
+                                                std::istreambuf_iterator<char>());
+
+                try {
+                    json j;
+                    j = json::parse(json_request_result);
+                    parsed_token = j.get<OAuthToken>();
+                } catch(std::exception& ex) {
+                    throw std::runtime_error(std::string("Failed to parse OAuth token: ") + std::string(ex.what()));
```
amnguye: same as above

Issue #87, opened 2018-02-12
Limit file cache size


asorrin-msft:
blobfuse assumes there is enough disk space for the file cache.  If disk space is exhausted, operations may fail in mysterious ways.  We should at minimum detect this and error out intelligently, or (better) GC some items in the cache until there's enough space.

Tatsinnit:
👍 for this bug 

💡  Note: please also update `min` or `max` limit (if there is any) in the documentation here: https://docs.microsoft.com/en-us/azure/storage/blobs/storage-how-to-mount-container-linux#optional-use-a-ramdisk-for-the-temporary-path

syagev:
One of the points of having a network (e.g. "blob") storage is for storing big amounts of data that don't fit on local disks... I think this issue deserves more attention.

I'm using Premium Blobs to store datasets for deep-learning. This means my worker-nodes are iterating the entire storage. It's therefore quite critical that the cache is elegantly cleaned up every now and then.

vibhansa-msft:
--cache-size-mb command line parameter can be used to control the amount of cache. Kindly refer to README. This is available version 1.3.1 onwards.

j-baron:
Hi, thank you for the quick reply, I updated to 1.3.6 but its still ignoring the flag and using much more than it is meant to, when I set the flag to:  - --cache-size-mb=6000 I still see usage for prometheus running of more than 18GB on the node on the /mnt. This is three times more than the expected figure. Thank you

Issue #79, opened 2018-02-07
RFE: Support for very large datasets - block-based cache


mkiernan:
I'm attempting to copy a 270GB file from an azure storage blob container onto a Linux 7.1 machine. 

I see the file appear in the temp directory, but it never makes it to the target, and the command itself bombs out with the below error. 

Is there any way to increase the timeouts here to avoid this error ?

cp /mnt/hpcpoc/bcbio-data.tar.gz .
cp: cannot open ‘/mnt/hpcpoc/bcbio-data.tar.gz’ for reading: No such file or directory

The system log shows : 

[ 2112.479803] fuse init (API version 7.22)
[ 2112.484293] SELinux: initialized (dev fuse, type fuse), uses genfs_contexts
[ 2112.495617] SELinux: initialized (dev fusectl, type fusectl), uses genfs_contexts
[ 2252.473409] SELinux: initialized (dev fuse, type fuse), uses genfs_contexts
[ 2410.892831] SELinux: initialized (dev fuse, type fuse), uses genfs_contexts
[ 2522.446944] SELinux: initialized (dev tmpfs, type tmpfs), uses transition SIDs
[ 2761.277854] INFO: task cp:48332 blocked for more than 120 seconds.
[ 2761.282241] "echo 0 > /proc/sys/kernel/hung_task_timeout_secs" disables this message.
[ 2761.287717] cp              D ffff881cc6e33680     0 48332  30695 0x00000084
[ 2761.292772]  ffff881c8493bb60 0000000000000086 ffff881c80e6e660 ffff881c8493bfd8
[ 2761.297958]  ffff881c8493bfd8 ffff881c8493bfd8 ffff881c80e6e660 ffff881c8431c640
[ 2761.304768]  ffff881cb0088000 ffff881c8493bb90 ffff881c8431c720 000000000000000e
[ 2761.311518] Call Trace:
[ 2761.313653]  [<ffffffff816098d9>] schedule+0x29/0x70
[ 2761.316914]  [<ffffffffa038853d>] __fuse_request_send+0x13d/0x2c0 [fuse]
[ 2761.323007]  [<ffffffff81098240>] ? wake_up_bit+0x30/0x30
[ 2761.328180]  [<ffffffffa03886d2>] fuse_request_send+0x12/0x20 [fuse]
[ 2761.332465]  [<ffffffffa0391189>] fuse_do_open+0xd9/0x190 [fuse]
[...]

PR #379, opened 2020-04-07
mention the implication of having a cache + alt. impl.

perbu:
I've run into the implications of a local cache and would like to place some more emphasis on the consequences.

Also, there is an alternative implication with a slightly different focus, which might be useful in certain scenarios.

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,8 +153,13 @@ By default, blobfuse will log to syslog.  The default settings will, in some cas
 ### Current Limitations
 - Some file system APIs have not been implemented: readlink, symlink, link, chmod, chown, fsync, lock and extended attribute calls.
 - Not optimized for updating an existing file. blobfuse downloads the entire file to local cache to be able to modify and update the file
+- If the local cache isn't big enough to fit the file you are reading from or writing to the operation will fail. Make sure the cache is big enough to host the biggest file you plan to read or write.
 - When using enabling the "--use-attr-cache" feature, there may be an issue with overflow and will not clear the attribute cache until blobfuse is unmounted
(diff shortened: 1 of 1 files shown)


NaraVen:
Thanks for the suggestion, AzCopy is a much more performant and faster alternative to BlobFuse, if there are latency issues with POSIX, they will be better served by using AzCopy instead.

perbu:
AzCopy isn't a filesystem so I don't understand why you are comparing it to Blobfuse. For a lot of ML workloads you work on huge read-only datasets consisting of files that are to big to copy into the container so AzCopy isn't helpful. For my project it was either Goofys or migrate away from Azure - we absolutely needed a fuse implementation.

I know Goofys isn't a Microsoft product but I still think you should mention them as they make it possible to use blob storage in cases where Blobfuse fails.

NaraVen:
From the scratching the surface I find Goofy’s different from blobfuse. AZCopy on the other hand copies files just like blobfuse. Appreciate you analysis,

amnguye:
Hi,

Thanks for submitting a documentation PR. I haven't seen any issues posted by you on the blobfuse repo. If you have cases where blobfuse fails or any semantics that you are unhappy with blobfuse, feel free to make an issue on the blobfuse issues tab and we will attempt to address your issues accordingly. Whether that be a bug or asking for a feature.

I understand your intentions to have other customers be open to other tools to access Azure Blobs however with this PR we would also have to add in the rest of our documentation to make suggestions to other tools to access Azure Blobs (SDKs, azcopy, storage explorer, and so on). This wouldn't be the right place to document this. This would be better documented in our other Microsoft Docs when users are deciding what tools to use to access their data.

perbu:
Thank you for the analysis.

I've not open new issues as there are already open issues wrt. copying objects onto local storage (#87, #79) that cover the topic extensively and I have little new to add to the topic. 

I'll close the PR, then. 

Per.

PR #394, opened 2020-05-04
Bug/msi client

NaraVen:
This PR has the following changes
1) fixed blobnotfound error related to auth after token expiry
2) fixed blobnotfound error due to Getproperty call on dirs
3) fixedparam unexplainable parsing issue for authtypes.
4) removed unnecessary repeat call to getblobproperty and Get_Hierarchical_list

### DIFF
--- a/blobfuse/blobfuse.cpp
+++ b/blobfuse/blobfuse.cpp
@@ -217,10 +217,6 @@ int read_config(const std::string configFile)
         fprintf(stderr, "No config file found at %s.\n", configFile.c_str());
         return -1;
     }
-    else
-    {
-       fprintf(stderr, "reading config");
-    }
 
     std::string line;
     std::istringstream data;
@@ -339,7 +335,6 @@ void *azs_init(struct fuse_conn_info * conn)
 {
     // TODO: Make all of this go down roughly the same pipeline, rather than having spaghettified code
     auth_type AuthType = get_auth_type();
-    fprintf(stdout, "The authtype is %d\n", AuthType);
 
     if (str_options.use_attr_cache)
     {
@@ -417,9 +412,7 @@ void *azs_init(struct fuse_conn_info * conn)
                         str_options.spnClientSecret,
                         str_options.aadEndpoint);
             }
-
-                        fprintf(stdout , "MSI params: identityclientId %s\n", str_options.identityClientId.c_str());
-
+            
             GetTokenManagerInstance(OTMCallback);
             //2. try to make blob client wrapper using oauth token
             azure_blob_client_wrapper = blob_client_wrapper_init_oauth(
@@ -490,7 +483,7 @@ void print_usage()
 
 void print_version()
 {
-    fprintf(stdout, "blobfuse 1.2.2\n");
+    fprintf(stdout, "blobfuse 1.2.5\n");
 }
 
 int set_log_mask(const char * min_log_level_char)
@@ -539,7 +532,7 @@ int set_log_mask(const char * min_log_level_char)
     }
 
     syslog(LOG_CRIT, "Unable to start blobfuse. Error: Invalid log level \"%s\"", min_log_level.c_str());
-    fprintf(stdout, "Error: Invalid log level \"%s\".  Permitted values are LOG_OFF, LOG_CRIT, LOG_ERR, LOG_WARNING, LOG_INFO, LOG_DEBUG.\n", min_log_level.c_str());
+    fprintf(stderr, "Error: Invalid log level \"%s\".  Permitted values are LOG_OFF, LOG_CRIT, LOG_ERR, LOG_WARNING, LOG_INFO, LOG_DEBUG.\n", min_log_level.c_str());
     fprintf(stdout, "If not specified, logging will default to LOG_WARNING.\n\n");
     return 1;
 }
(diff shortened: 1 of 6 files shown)


connection.cfg
```diff
@@ -1,8 +1,9 @@
...
\ No newline at end of file
+accountName <account-name-here>
+# Please provide either an account key or a SAS token, and delete the other line.
+accountKey <account-key-here-delete-next-line>
+sasToken <shared-access-token-here-delete-previous-line>
+authType <MSI/SAS/SPN/Key/empty>
+containerName <insert-container-name-here>
+
+# Check README.md for valid configuration options for each auth type.
+identityClientId a7c18d79-3c0b-440a-aec3-b82ada05bf08
```
amnguye, line 9: Remove IdentityClientId

blobfuse/blobfuse.cpp
```diff
@@ -215,6 +217,10 @@ int read_config(const std::string configFile)
         fprintf(stderr, "No config file found at %s.\n", configFile.c_str());
         return -1;
     }
+    else
+    {
+       fprintf(stderr, "reading config");
```
amnguye, line 222: This will print out as an error/to the console, so do we want this to be printing out whenever a configuration file is being read? If we find a configuration file we shouldn't be printing out anything

blobfuse/blobfuse.cpp
```diff
@@ -410,6 +418,8 @@ void *azs_init(struct fuse_conn_info * conn)
                         str_options.aadEndpoint);
             }
 
+                        fprintf(stdout , "MSI params: identityclientId %s\n", str_options.identityClientId.c_str());
```
amnguye, line 421: This will always print out, is this for debugging purposes? if anything it should be put in the syslog and under conditions of log_debug flag

blobfuse/utilities.cpp
```diff
@@ -249,7 +249,7 @@ int ensure_files_directory_exists_in_cache(const std::string& file_path)
             }
 
             // Ignore if some other thread was successful creating the path
-	    if(errno == EEXIST)
+        if(errno == EEXIST)
```
amnguye, line 252: NIT - add indent back

blobfuse/blobfuse.cpp
NaraVen: done

blobfuse/blobfuse.cpp
NaraVen: removed

NaraVen:
> If you leave the connection.cfg file as is (aka if you use the connection.cfg as it is shown in the repo) it will try to get the oauth token. It should be throwing an error right then and there and should not attempt to get an oauth token.
> 
> There's a breaking change where if I don't specify the authType in the configuration file (just using accountName, accountKey and containerName), it will default to calling msi auth and prints out unnecessary statements showing it is trying to get the oauth token.
> 
> We can take this testing offline, but these errors are not hard to reproduce.

I tested with accountkey and secret pure commandline without config it worked.
I modified the config with the right values including Key for authType and it worked, it did not default to msi. 
I don't understand your concern. The connection.cfg is just a placeholder it should never be used as is, that willl be obvious to anyone who opens it.

amnguye:
Try testing it with no secret. I was able to reproduce this with no environment variables and populating the configuration.cfg file with the accountName, accountKey and containerName.

Yes the connection.cfg should never be used as is, however it is a possibility the user will forget to populate their configuration file or save it. When that happens, it will try to retrieve the oauth token endlessly, and the user will be confused. Blobfuse should react by throwing an authentication error right away. This is just a base test case.

NaraVen:
Fixing the config is not part of this PR. I agree it is a well known bug, I will create a work item and fix it in another PR. This PR is for BlobNotFound issues. The two issues are unrelated.

Get Outlook for iOS<https://aka.ms/o0ukef>
________________________________

amnguye:
These issues are related because they are introduced in this PR. When I pull master and build and run that, they are not produced.

NaraVen:
I cannot reproduce this issue. I am testing it exactly how you have mentioned but just in case I have missed anything here are the details.
I have no environment variables for storage account or key. 
I have the below 3  entries in my config files.
accountName wcu1437749890
containerName nara-blobstore-3ae6841
accountKey anZ9808984395877435874538798453+bUaCrJiSDFrIY2DnvdFqqVHAvx4WmjGQ==

NaraVen:
I cannot reproduce this Amanda. I used the same test criteria you used. Let us wait for Vikas Bhansali @vibhansa-msft  to come online and give this a try. 
When I specify the containerName, accountName and accountKey in the config file it mounts and runs correctly.
When I specify the containerName, accountName only it gives me a failed to auth error and does not mount which is correct.
I never run into the scenario where it keeps tryign to retrieve and OAuthToken and confuse the user.
Setup details -------------------------
I do not have any environment variables. 
The config file looks as below.
accountName fake9890
containerName fake-nara-blobstore-3cd50821-00bf-483d-b11f-2ac7d3ae6841
accountKey fakeKey----------

connection.cfg
```diff
@@ -1,8 +1,8 @@
-accountName <account-name-here>
-# Please provide either an account key or a SAS token, and delete the other line.
-accountKey <account-key-here-delete-next-line>
-sasToken <shared-access-token-here-delete-previous-line>
-authType <MSI/SAS/SPN/Key/empty>
-containerName <insert-container-name-here>
-
```
vibhansa-msft, line 7: Do not see any change in this file.

NaraVen: No change. This is a display issue.

NaraVen:
It worked in MSI, non-MSI with storage account and key on both 18.04 and 16.04 so I am finishing the PR

PR #398, opened 2020-05-12
Bug/file not found bad header

NaraVen:
This PR has the following  bug fixes
1) Header encoding to match the latest Storage API ( to prevent error 400 on list requests)
2) Upgrade to storage API "2018-11-09"
3) Fix for directory creation in root directory ( where it created a no name directory underneath)
4) Fix for repeated Oauth token retrieval : Defaulting to Authtype MSI when the authtype was jumbled up or left at default msi/sas/spn
5) Test integration with CI: Adding tests to build trigger Yaml , I had to comment out some googletests that fail repeatedly.


azure-pipelines.yml
```diff
@@ -19,11 +20,33 @@ jobs:
...
+    pool:
+      vmImage:  'ubuntu-18.04'
+    steps:
+      - script: |
+            sudo apt-get install pkg-config libfuse-dev cmake libcurl4-gnutls-dev libgnutls28-dev uuid-dev libgcrypt20-dev libboost-all-dev gcc g++ -y
+            . ./build.sh
+            cp ./blobfuse $(Build.ArtifactStagingDirectory)
+        displayName: Generate blobfuse build for Ubuntu 18.04
       - task: PublishBuildArtifacts@1
-        displayName: 'Publish Artifacts'
```
vibhansa-msft, line 27: Do we have different set of steps of Ubuntu 16 and 18. If not there is no need to write two different jobs we can write a single one with strategy matrix and it will build both in parallel.

azure-storage-cpp-lite/src/storage_url.cpp
```diff
@@ -1,7 +1,27 @@
+#include <ctime>
+
 #include "storage_url.h"
 
-namespace microsoft_azure {
-    namespace storage {
+#include "constants.h"
+
+#ifdef _WIN32
+#define WIN32_LEAN_AND_MEAN
```
vibhansa-msft, line 8: There are lot of changes in this file, is it just taken from the storage-cpp-lite code directly ?

azure-pipelines.yml
NaraVen: It is all in this piece of code you are reading unless, I don't understand the question. Yes, a couple of steps are different.

azure-storage-cpp-lite/src/storage_url.cpp
NaraVen: code structure is a different so I did not and cannot pull it directly. However, I matched all the encodings carefully with their latest and it works. Even the client tested it. It passed the private preview.

PR #406, opened 2020-05-19
fix #405

### LINKED ISSUE #405: Error building blobfuse on RHEL8+
1.1.1

### Which OS (please include version) are you using?
RHEL8+

### What problem was encountered?
manually build will be failed as per as wiki steps.

We should install 'make' package and disable Boost_USE_STATIC_LIBS on CMakeLists.txt. The complete procedure is below.
```
sudo yum install git make cmake fuse-devel libcurl-devel gcc gcc-c++ gnutls-devel libgcrypt-devel libuuid-devel boost-devel fuse -y
git clone https://github.com/Azure/azure-storage-fuse/
cd ./azure-storage-fuse/
sudo sed -i 's/set(Boost_USE_STATIC_LIBS ON)/set(Boost_USE_STATIC_LIBS OFF)/g' ./CMakeLists.txt
./build.sh
```

It seems that there are similar and related isssues.
https://github.com/Azure/azure-storage-fuse/issues/319
https://github.com/Azure/azure-storage-fuse/issues/196

juyama:
- Update CMakeLists.txt to fix #405


CMakeLists.txt
```diff
@@ -131,7 +131,7 @@ if(UNIX)
   find_package(GnuTLS REQUIRED)
   find_package(PkgConfig)
   find_package(Boost REQUIRED)
-  set(Boost_USE_STATIC_LIBS ON)
```
NaraVen, line 134: We need this on for certain features to work.

NaraVen: you can conditionally turn this on for RHEL if you want to continue with this PR. We do not want to turn it on for CentOS

juyama: Thanks for your review, Nara. I understand this PR will be breaking changes for Ubuntu.
Could you please consider to update the installation steps on wiki for RHEL8?

PR #439, opened 2020-07-15
Naraven/rhel compliant

NaraVen:
1) Made command line args RHEL compliant"
2) user agent blobfuse
3) removed validation to allow any order in config file
4) made version number -beta

### DIFF
--- a/blobfuse/src/BlobfuseGlobals.cpp
+++ b/blobfuse/src/BlobfuseGlobals.cpp
@@ -25,10 +25,20 @@ inline bool is_lowercase_string(const std::string &s)
     })));
 }
 
+std::string trim(const std::string& str) 
+{
+    const size_t start = str.find_first_not_of(' ');
+    if (std::string::npos == start) {
+        return std::string();
+    }
+    const size_t end = str.find_last_not_of(' ');
+    return str.substr(start, end - start + 1);
+}
+
 AUTH_TYPE get_auth_type(std::string authStr) 
 {
     if(!authStr.empty()) {
-        std::string lcAuthType = to_lower(authStr);
+        std::string lcAuthType = trim(to_lower(authStr));
         syslog(LOG_DEBUG, "Auth type set as %s", lcAuthType.c_str());
         int lcAuthTypeSize = (int)lcAuthType.size();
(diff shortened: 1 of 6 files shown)


cpplite/include/constants.dat
```diff
@@ -116,7 +116,7 @@ DAT(header_value_payload_format_nometadata, "application/json;odata=nometadata")
 DAT(header_value_payload_format_fullmetadata, "application/json;odata=fullmetadata")
 DAT(header_value_storage_blob_version, "2018-11-09")
 
-DAT(header_value_user_agent, "azure-storage-cpplite/0.3.0")
```
vibhansa-msft, line 119: There is no need to change this. We have taken this from the cpplite lib itself.

blobfuse/src/BlobfuseGlobals.cpp
```diff
@@ -29,22 +29,27 @@ AUTH_TYPE get_auth_type(std::string authStr)
 {
     if(!authStr.empty()) {
         std::string lcAuthType = to_lower(authStr);
-        if (lcAuthType == "msi") {
+        syslog(LOG_DEBUG, "Auth type set as %s", lcAuthType.c_str());
+        int lcAuthTypeSize = (int)lcAuthType.size();
+         // sometimes an extra space or tab sticks to authtype thats why this size comparison, it is not always 3 lettered
+        if(lcAuthTypeSize > 0 && lcAuthTypeSize < 5)
```
vibhansa-msft, line 35: Even if do the size check to safeguard extra space and tab, below we are comparing the data without handling the same. We should rather trim the value first and then do the comparisons.

cpplite/include/constants.dat
NaraVen: No it should show Azure-storage-fuse and not cpplite. The logs are triaged based on Azure-storage-fuse. I understand it is not correct to change the Cpplite code base but the endproduct wins here. So the logs have to show Azure-storage-fuse. Leaving this as is

blobfuse/src/BlobfuseGlobals.cpp
NaraVen: done trimmed.

PR #446, opened 2020-08-03
Narven/perf fix

NaraVen:
Added Exists method to improve ADLS perf
removed caching for ADLS getProperties

### DIFF
--- a/blobfuse/src/DfsProperties.cpp
+++ b/blobfuse/src/DfsProperties.cpp
@@ -180,10 +180,10 @@ int adls_client_ext::adls_exists(const std::string &filesystem, const std::strin
     {
         exists=1;
     }
-    if (!success() && errno != 404)
+    if (errno == 404)
     {
         exists=0;
-    }
+    } 
     return exists;
 }
 
--- a/blobfuse/src/DfsProperties.cpp
+++ b/blobfuse/src/DfsProperties.cpp
@@ -1,5 +1,5 @@
 //
-// Created by adreed on 2/7/2020.
+// ADLS specific properties
 //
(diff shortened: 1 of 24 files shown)


blobfuse/utilities.cpp
```diff
@@ -276,9 +274,6 @@ int azs_getattr(const char *path, struct stat *stbuf)
 
                 stbuf->st_uid = fuse_get_context()->uid;
                 stbuf->st_gid = fuse_get_context()->gid;
-                //stbuf->st_mtime = blob_property.last_modified;
```
vibhansa-msft, line 279: We need to preserve the correct last_modified time.

blobfuse/src/DfsProperties.cpp
```diff
@@ -99,7 +99,8 @@ dfs_properties adls_client_ext::get_dfs_path_properties(const std::string &files
         props.owner = http->get_response_header(constants::header_ms_owner);
         props.group = http->get_response_header(constants::header_ms_group);
         props.permissions = http->get_response_header(constants::header_ms_permissions);
-        props.acl = http->get_response_header(constants::header_ms_acl);
```
vibhansa-msft, line 102: This gives file/dir permission related info in case of dfs endpoint.

NaraVen: acl was not returned. Only permissions. So I removed.

blobfuse/utilities.cpp
NaraVen: yes below the comment we are preserving it. This is just the comment.

PR #473, opened 2020-09-18
Update README.md with authType SPN

sihegde:
I am proposing this change as the content didn't had any information on authType 'SPN' that we can use for Service Principal Authentication. The connection.cfg has details where it is mentioned that SPN should be used as authType. However, since README.md had no information provided on it, I have added those changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,6 +78,10 @@ For more information, see the [wiki](https://github.com/Azure/azure-storage-fuse
             - Client ID (Use this if you are using a custom MSI endpoint)
             - Object ID
             - Resource ID
+- Service Principal Name (`authType SPN`)
+    - Client ID
+    - Tenant ID
+    - Client Secret
 
 ### Environment variables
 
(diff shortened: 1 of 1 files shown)


vibhansa-msft (commented):
Thank you for your contribution. 
Under "Config file options" and "Environment Variables" sections these details are already present. We can detail it out in "Valid authentication" section as well, if that makes more sense for the user.

sihegde:
Hi @vibhansa-msft ,

Thanks for the update! yes, both Config File options and Environment Variables sections have details about the Service principal authentication. However, it is bit confusing as there is no mention of 'authType' that has to be used for this. So I have added only 'authType' part.

vibhansa-msft:
> Hi @vibhansa-msft ,
> 
> Thanks for the update! yes, both Config File options and Environment Variables sections have details about the Service principal authentication. However, it is bit confusing as there is no mention of 'authType' that has to be used for this. So I have added only 'authType' part.

Yes for the second part of the change I do agree with you.

vibhansa-msft:
I will approve this change-set and merge. We will re-phrase the first part of your change-set if required.

vibhansa-msft (approved):

sihegde:
@vibhansa-msft ,
Thank you!  sure..

PR #479, opened 2020-09-22
Narven/delete bug fix

NaraVen:
Version upgrade
Removed data comparison for token check for performance
Reenabled MSI to run under a separate thread.

### DIFF
--- a/blobfuse/src/OAuthTokenCredentialManager.cpp
+++ b/blobfuse/src/OAuthTokenCredentialManager.cpp
@@ -86,7 +86,7 @@ void OAuthTokenCredentialManager::StartTokenMonitor()
     #ifdef TOKEN_REFRESH_THREAD
     std::thread t1(std::bind(&OAuthTokenCredentialManager::TokenMonitor, this));
     t1.detach();
-    syslog(LOG_ERR,"OAUTH Token : Token empiry monitor started");
+    syslog(LOG_WARNING,"OAUTH Token : Token expiry monitor started");
     #endif
 }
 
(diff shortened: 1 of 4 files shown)


blobfuse/src/OAuthTokenCredentialManager.cpp
```diff
@@ -194,36 +185,14 @@ OAuthToken OAuthTokenCredentialManager::refresh_token()
...
-                // If we were working in Go, I (Adele) would suggest using a sync.WaitGroup for this functionality.
-                syslog(LOG_ERR, "OAUTH Token : Wait for token to get refreshed as buffer time has expired");
-                token_mutex.lock();
-            }
-        }
-
-        // Note that we don't always lock in this function and sometimes return early.
-        // Be sure to ensure you're safely manipulating the lock when modifying this function.
-        token_mutex.unlock();
+    #ifndef TOKEN_REFRESH_THREAD
```
vibhansa-msft, line 188: This check shall be reverse, we want to use previously acquired token if the monitor thread is running.

blobfuse/src/OAuthTokenCredentialManager.cpp
```diff
@@ -194,36 +185,14 @@ OAuthToken OAuthTokenCredentialManager::refresh_token()
...
+    // since the thread would have updated the expired token just check if valid_authentication
+   // and return the current token, control will fall to the if is_token_expired() below if this directive is undefined.
+    if (valid_authentication)
+    {
+        return current_oauth_token;
     }
-    #else
+    #endif
```
vibhansa-msft, line 195: No need to call refresh_token() if the thread is running, we just need to wait for the thread to refresh the token.

NaraVen: if valid_authentication has failed the thread has gone wrong so refreshing the token here even if the #def TOKEN_REFRESH_THREAD isdefined

blobfuse/src/OAuthTokenCredentialManager.cpp
NaraVen: done

blobfuse/src/OAuthTokenCredentialManager.cpp
vibhansa-msft: understood the logic.

vibhansa-msft (approved):

PR #500, opened 2020-10-08
Rename in adls failing even after operation being sucessful

vibhansa-msft:
- if file/dir does not exists in local cache then even when rename is successful, on cli we return failure "Input/Output error"

### DIFF
--- a/blobfuse/src/DataLakeBfsClient.cpp
+++ b/blobfuse/src/DataLakeBfsClient.cpp
@@ -340,11 +340,14 @@ std::vector<std::string> DataLakeBfsClient::Rename(std::string sourcePath, std::
         std::string srcMntPathString = prepend_mnt_path_string(sourcePath);
         std::string dstMntPathString = prepend_mnt_path_string(destinationPath);
 
-        long int rename_ret = rename_cached_file(srcMntPathString.c_str(), dstMntPathString.c_str());
+        bool cache_cleared = false;
+        long int rename_ret = rename_cached_file(srcMntPathString.c_str(), dstMntPathString.c_str(), cache_cleared);
         if(rename_ret != 0) {
             syslog(LOG_ERR, "Failure to rename source file %s in the local cache.  Errno = %d.\n", sourcePath.c_str(), errno);
         } 
-        file_paths_to_remove.push_back(sourcePath);
+        
+        if (cache_cleared)
+            file_paths_to_remove.push_back(sourcePath);
     }
     return file_paths_to_remove;
 }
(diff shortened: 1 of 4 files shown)


blobfuse/src/DataLakeBfsClient.cpp
```diff
@@ -325,24 +325,30 @@ std::vector<std::string> DataLakeBfsClient::Rename(const std::string /*sourcePat
...
 
-    if(rename_ret != 0)
-    {
-        syslog(LOG_ERR, "Failure to rename source file %s in the local cache.  Errno = %d.\n", sourcePath.c_str(), errno);
+        if(rename_ret != 0) {
+            syslog(LOG_ERR, "Failure to rename source file %s in the local cache.  Errno = %d.\n", sourcePath.c_str(), errno);
+        }
+        
+        file_paths_to_remove.push_back(sourcePath);
+        errno = 0;
```
NaraVen, line 350: 1)
rename_cached_file should not return a non zero errno if rename_ret =0. Please fix line add errno=0 after 473 of rename_cached_file 
2) this line doesn't make sense, to simply reset the errno, if the file is in local cache and rename_cached_file could not rename there is something wrong the error has to bubble up.
3) same with file_paths_to_remove, if the local rename has failed when it is in cache what will we achieve by adding the oldpath to the list to be removed from cache .

NaraVen (changes requested):

blobfuse/src/DataLakeBfsClient.cpp
vibhansa-msft: - Code modified to take care of point 1,2.
- For point 3 : the files are renamed in container but local rename has failed. If we keep the files in local cache directory listing will show both of them.

blobfuse/src/DataLakeBfsClient.cpp
```diff
@@ -325,24 +325,27 @@ std::vector<std::string> DataLakeBfsClient::Rename(const std::string /*sourcePat
...
+        syslog(LOG_ERR, "Failure to rename source file %s in container.  Errno = %d.\n", sourcePath.c_str(), errno);
+    } else {
+        std::string srcMntPathString = prepend_mnt_path_string(sourcePath);
+        std::string dstMntPathString = prepend_mnt_path_string(destinationPath);
+
+        long int rename_ret = rename_cached_file(srcMntPathString.c_str(), dstMntPathString.c_str());
+        if(rename_ret != 0) {
+            syslog(LOG_ERR, "Failure to rename source file %s in the local cache.  Errno = %d.\n", sourcePath.c_str(), errno);
+        } 
+        file_paths_to_remove.push_back(sourcePath);
```
NaraVen, line 347: sourcePath should not be added to the file_paths_to_remove if the file does not exist in cache. So the return from rename_cached_file should return a tuple, the second value of the pair should indicate if the file was in cache or not.

NaraVen (changes requested):

blobfuse/src/DataLakeBfsClient.cpp
vibhansa-msft: Added a check to push to vector only when local cache was cleared.

NaraVen (approved):

Issue #523, opened 2020-11-09
blobfuse uploaded the same file twice to Storage Account when using SCP.


takmnagayams:
Can I suggest a feature that allows files to be uploaded once to Storage Account when using scp?

### Which version of the blobfuse was used?
1.0.2

### Which OS (please include version) are you using?
istributor ID: Ubuntu
Description:    Ubuntu 18.04.5 LTS
Release:        18.04
Codename:       bionic

### What problem was encountered?
When I uploaded a file with scp command, the file was uploaded to the Storage Account twice as following logs.
I didn’t update the file which was uploaded, but it was uploaded to the Storage Account twice.

### Have you found a mitigation/solution?
No

### By default, blobfuse logs errors to syslog. If this is relevant, is there anything in the syslog that might be helpful?

```
Oct 26 13:26:49 blobfuse systemd[1]: Started Session 522 of user tanagaya.
Oct 26 13:26:50 blobfuse blobfuse[2796]: Function azs_getattr, in file /home/vsts/work/1/s/blobfuse/utilities.cpp, line 120: azs_getattr called with path = /tt7.png
Oct 26 13:26:50 blobfuse blobfuse[2796]: Function azs_getattr, in file /home/vsts/work/1/s/blobfuse/utilities.cpp, line 169: Object /mnt/resource/blobfusetmp/root/tt7.png is not in the local cache during getattr.
Oct 26 13:26:50 blobfuse blobfuse[2796]:  ==> REQUEST/RESPONSE :: GET https://tanagayablobfuse.blob.core.windows.net/mycontainer?comp=list&delimiter=/&include=metadata&maxresults=5000&prefix=tt7.png&restype=container?&User-Agent=azure-storage-fuse/1.3.4&x-ms-date=Mon, 26 Oct 2020 13:26:50 GMT&x-ms-version=2018-11-09&Authorization=****&Transfer-Encoding=--------------------------------------------------------------------------------RESPONSE Status :: 200 :: REQ ID : 3afdcc3b-d01e-009a-179b-ab0134000000
Oct 26 13:26:50 blobfuse blobfuse[2796]: Function azs_getattr, in file /home/vsts/work/1/s/blobfuse/utilities.cpp, line 233: In azs_getattr list_segmented_item do loop blob prefix: tt7.png continuation:
[...]

Issue #505, opened 2020-10-10
Listing mounted folder ADLS Gen 2 not working on Centos 7.8.2003


kthangar:
### Which version of the blobfuse was used?
https://github.com/Azure/azure-storage-fuse/releases/download/blobfuse-1.3.4-Linux/blobfuse-1.3.4-RHEL7.8-x86_64.rpm

### Which OS (please include version) are you using?
CentOS Linux release 7.8.2003 (Core)
3.10.0-1127.13.1.el7.x86_64 #1 SMP Tue Jun 23 15:46:38 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
### What problem was encountered?
I am trying to rsync files from centos 7 host to azure storage account.
ls command in mounted folder on centos 7 hangs forever.
Every few secs  "REQUEST/RESPONSE :: GET" messages appears in /var/log/messages.

### Have you found a mitigation/solution?
if i add "-o debug" to blobfuse i was able to mount and rsync files, but getting these errors 
rsync: failed to set permissions on "/home/svc_user/app-uswest2/2020/xxx": Function not implemented (38)
rsync: chgrp "/home/svc_user/app-uswest2/2020/xxx" failed: Function not implemented (38)

---------------------
FUSE library version: 2.9.2
nullpath_ok: 0
nopath: 0
utime_omit_ok: 0
unique: 1, opcode: INIT (26), nodeid: 0, insize: 56, pid: 0
INIT: 7.23
flags=0x0003fffb
max_readahead=0x00020000
   INIT: 7.19
   flags=0x00000031
   max_readahead=0x00400000
   max_write=0x00400000
   max_background=128
   congestion_threshold=96
   unique: 1, success, outsize: 40
unique: 2, opcode: GETATTR (3), nodeid: 1, insize: 56, pid: 30790
getattr /
   unique: 2, success, outsize: 120

### By default, blobfuse logs errors to syslog. If this is relevant, is there anything in the syslog that might be helpful?

Oct  9 09:42:15 san-xxx blobfuse[11536]: Setting logging level to : LOG_DEBUG
Oct  9 09:42:15 san-xxx blobfuse[11536]: release is 3.100000, Kernel version is 3.10.0-1127.13.1.el7.x86_64
[...]

PR #525, opened 2020-11-19
Vibhansa/filenameencode

vibhansa-msft:
1. #523 : Fix for scp uploading the file twice. In Truncate call do not upload the file if current file size and truncate size are same. In case of SCP we receive a truncate call before the flush and here file upload can be ignored.
2. #505 : Fix for issue mentioned in comments regarding "Unexpected amount of permissions"
3. Enable exponential retry for adls account
4. Adding support for '%' in file name for adls accounts.
5. Modify file permissions and last modify/change time only if 'open' call resulted into file download.


blobfuse/fileapis.cpp
```diff
@@ -491,6 +495,14 @@ int azs_truncate(const char * path, off_t off)
             return res;
         }
 
+        // If the file in cache has exactly the same size then there is no need to do trucate and upload
+        struct stat buf;
```
NaraVen, line 503: How do we know that something has not changed just by the size of the file

blobfuse/src/DataLakeBfsClient.cpp
```diff
@@ -40,6 +40,10 @@ bool DataLakeBfsClient::AuthenticateStorage()
         
     if(m_adls_client != NULL)
     {
+        if (m_adls_client->get_blob_client() && m_adls_client->get_blob_client()->context())
```
NaraVen, line 43: this is redundant. The line above is doing it. So remove this one

blobfuse/src/DataLakeBfsClient.cpp
```diff
@@ -40,6 +40,10 @@ bool DataLakeBfsClient::AuthenticateStorage()
...
+            m_adls_client->get_blob_client()->context()->set_retry_policy(
+                        std::make_shared<azure::storage_lite::expo_retry_policy>());
```
NaraVen, line 45: remove

cpplite/adls/include/adls_client.h
```diff
@@ -315,6 +315,11 @@ namespace azure { namespace storage_adls {
             return m_exception_enabled;
         }
 
+        AZURE_STORAGE_ADLS_API std::shared_ptr<azure::storage_lite::blob_client> get_blob_client()
+        {
```
NaraVen, line 319: remove

blobfuse/src/Permissions.cpp
```diff
@@ -36,8 +36,7 @@ mode_t aclToMode(access_control acl)
     }
     else if(permissions.size() != blobfuse_constants::acl_size)
     {
-        syslog(LOG_ERR, "Failure: Unexpected amount of permissions from service : %s", permissions.c_str());
-        return mode;
+        syslog(LOG_DEBUG, "Unexpected amount of permissions from service : %s", permissions.c_str());
```
NaraVen, line 39: Check if it is sticky. Ignore if it is sticky. Issue LOG_WARNING otherwise.

blobfuse/fileapis.cpp
```diff
@@ -335,21 +335,24 @@ int azs_flush(const char *path, struct fuse_file_info *fi)
...
-            {
-                int storage_errno = errno;
-                syslog(LOG_ERR, "Failing blob upload in azs_flush with input path %s because of an error from upload_file_to_blob().  Errno = %d.\n", path, storage_errno);
-                free(path_buffer);
-                return 0 - map_errno(storage_errno);
-            }
-            else
-            {
-                syslog(LOG_INFO, "Successfully uploaded file %s to blob %s.\n", path, blob_name.c_str());
+            if (buf.st_mtime > ((struct fhwrapper *)fi->fh)->mtime_at_upload) {
```
NaraVen, line 338: I haven't tested this yet, so expect comments a little later.

blobfuse/fileapis.cpp
vibhansa-msft: With truncate we can only set the size of a file to a certain number you can not change the data. So if the current size of file is same as what you are setting with truncate then there is no need to take any action. For e.g. if user create a new file of 100 bytes and then truncates it to 100, nothing changes. Just ignoring such calls because truncate for blobfuse results into uploading a new file with given size.

vibhansa-msft: If question is about write something to file and then truncate then anyway we will receive a flush call later which will upload, so that shall also be ignored here. This is what happens in case of SCP where after writing the data we receive a truncate call followed by a flush call.

blobfuse/src/DataLakeBfsClient.cpp
vibhansa-msft: This is not redundant. In case of ADLS we have a blob client object created explicitly and one blob clien tobject adls object creates internally. For the external one we were always setting the exponential retry policy however the internal blob client of adls object was missed. This was the reason for not having exponential retry only in certain flows. With this line we are setting the retry policy for implicit blob client object created inside adls object. Ref : m_adls_client->get_blob_client()

blobfuse/src/DataLakeBfsClient.cpp
vibhansa-msft: Replied in above comments.

cpplite/adls/include/adls_client.h
vibhansa-msft: This method we need to add because there is no way to get the implicit blob client object from within the adls object as of now.

blobfuse/src/Permissions.cpp
vibhansa-msft: From blobfuse we can not set the sticky bit or anything other then 'rwx' triplet so we are not sure for what reason this extra permissions are coming or how the user was able to set them. Anyway while setting or reading the permissions blobfuse only parses 'rwx' triplet (first 9 bytes) so even if these extra bits come, we do not handle those. Have put a debug log just to get info on what these extra permissions are, just in case we face any issue with parsing or setting.

blobfuse/fileapis.cpp
vibhansa-msft: This check is introduced to solve the double upload issue. When we upload we will store the last modified time of file and when the next upload request comes we just check anything in file has changed or not, if no then skip the upload part. As soon as file handle is closed this data will be wiped out so this will hit only when same file is flushed twice before close.

NaraVen (approved):
approving now but I will be pulling out the double upload changes as there is a lot of erdundancy between upload and write in the fhwrapper and that could cause new regressions. So fix for double upload will not be part of release 1.3.6

blobfuse-nightly.yml
```diff
@@ -874,6 +874,8 @@ jobs:
 
         # Prebuild cleanup
         - script: |
+            sudo fusermount -u $(MOUNT_DIR)
+            sudo kill -9 `pidof blobfuse`
```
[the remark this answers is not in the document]
vibhansa-msft: It will give an error but step will not fail. This is just a safe check, in case we cancel a running job then blobfuse remain mounted on self hosted agents and next time we try to run the pipeline it fails. So first we cleanup the last run and then proceed.

blobfuse/include/BlobfuseGlobals.h
```diff
@@ -94,9 +94,8 @@ struct fhwrapper
 {
     int fh; // The handle to the file in the file cache to use for read/write operations.
     bool upload; // True if the blob should be uploaded when the file is closed.  (False when the file was opened in read-only mode.)
-    uint32_t mtime_at_upload = 0;
-
-    fhwrapper(int fh, bool upload) : fh(fh), upload(upload), mtime_at_upload(0)
+    bool write_mode;
+    fhwrapper(int fh, bool write) : fh(fh), upload(false), write_mode(write)
```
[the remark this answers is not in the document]
vibhansa-msft: This is the constructor, so by default we assume it to be false and only when a write is executed we set it to true which indicates that the file has been modified and needs to be uploaded back to container on flush.

blobfuse/fileapis.cpp
```diff
@@ -512,6 +513,7 @@ int azs_truncate(const char * path, off_t off)
         if (truncret == 0)
         {
             AZS_DEBUGLOGV("Successfully truncated file %s in the local file cache.", mntPath);
+            ((struct fhwrapper *)fi.fh)->upload = true;
```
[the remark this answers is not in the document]
vibhansa-msft: Here we have modified the file by resetting its size so we are setting the flag to true so that when flush() is called it uploads the file.

PR #527, opened 2020-11-23
Vibhansa/el7fix

vibhansa-msft:
Postpone of curl and ssl init based on curl version instead of kernel version

### DIFF
--- a/blobfuse/src/BlobfuseGlobals.cpp
+++ b/blobfuse/src/BlobfuseGlobals.cpp
@@ -1,6 +1,7 @@
 #include "BlobfuseGlobals.h"
 #include <ctype.h>
 #include <sys/utsname.h>
+#include <curl/curl.h>
 
 bool gEnableLogsHttp;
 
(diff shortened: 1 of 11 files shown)


blobfuse/src/BlobfuseGlobals.cpp
```diff
@@ -80,6 +81,16 @@ AUTH_TYPE get_auth_type(std::string authStr)
     return INVALID_AUTH;
 }
 
+float libcurl_version = 0.0;
+void  populate_libcurl_version()
+{
+	curl_version_info_data *d = curl_version_info(CURLVERSION_NOW);	
+	if (d) {
+		syslog(LOG_INFO, "CURL Version is %s", d->version);
+		libcurl_version = std::stof(d->version);
```
NaraVen, line 90: add try catch round it. Also add a test case for populate_libcurl_version

NaraVen (changes requested):
since populate_curl_version is a separate method, it can have a test case.

NaraVen (approved):

PR #531, opened 2020-11-30
Vibhansa/filenameencode

vibhansa-msft:
- version update


blobfuse/fileapis.cpp
```diff
@@ -301,8 +301,9 @@ int azs_flush(const char *path, struct fuse_file_info *fi)
         // For some file systems, however, close() flushes data, so we do want to do that before uploading data to a blob.
         // The solution (taken from the FUSE documentation) is to close a duplicate of the file descriptor.
         close(dup(((struct fhwrapper *)fi->fh)->fh));
-        if (((struct fhwrapper *)fi->fh)->write_mode &&
```
NaraVen, line 304: what is the difference here?

NaraVen (approved):

blobfuse/fileapis.cpp
NaraVen: its ok I will approve

PR #536, opened 2020-12-02
consolidated encodings and removed command line arg option

NaraVen:
no other major changes

### DIFF
--- a/blobfuse/blobfuse.cpp
+++ b/blobfuse/blobfuse.cpp
@@ -12,9 +12,6 @@
 #include <include/DataLakeBfsClient.h>
 #include <include/AttrCacheBfsClient.h>
 
-
-bool gEncodeFullFileName = true;
-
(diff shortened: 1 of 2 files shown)


blobfuse/blobfuse.cpp
```diff
@@ -12,9 +12,6 @@
 #include <include/DataLakeBfsClient.h>
 #include <include/AttrCacheBfsClient.h>
 
-
-bool gEncodeFullFileName = true;
```
vibhansa-msft, line 16: This variable is declared as extern in cpplite. we need to remove that as well otherwise compilation will fail at some point.

vibhansa-msft (approved):

blobfuse/blobfuse.cpp
NaraVen: removed, done

PR #549, opened 2021-01-04
Vibhansa/msi pipeline adls

vibhansa-msft:
Adding Block blob and ADLS MSI auth test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,7 +113,7 @@ For more information, see the [wiki](https://github.com/Azure/azure-storage-fuse
 
 - General options
     * `accountName`: Specifies the storage account blobfuse targets.
-    * `blobEndpoint`: Specifies the blob endpoint to use. Defaults to *.blob.core.windows.net, but is useful for targeting storage emulators. (e.g. storageaccountname.blob.region.azurestack.company.com)
+    * `blobEndpoint`: Optional parameter, defaults to `blob.core.windows.net`". This parameter should be specified for zonal dns accounts, custom domain name accounts or storage emulators.(e.g. storageaccountname.blob.region.azurestack.company.com)
     * `authType`: Overrides the currently specified auth type. Options: Key, SAS, MSI, SPN (Using this option is only available for 1.2.0 or above)
     * `logLevel`: Specifies the logging level. Use to change the logging level dynamically. Read `Logging` section for details. For allowed values refer to `--log-level` command line option.
     * `accountType`: Specifies the type of account. Either `block` or `adls` can be specified, `block` is the default value. Same can also be controlled through the command line option `--use-adls=true`. If wrong account type is supplied, certain features may not work as expected. To learn more about ADLS follow the link `(https://docs.microsoft.com/en-us/azure/storage/blobs/data-lake-storage-introduction)`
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -113,7 +113,7 @@ For more information, see the [wiki](https://github.com/Azure/azure-storage-fuse
 
 - General options
     * `accountName`: Specifies the storage account blobfuse targets.
-    * `blobEndpoint`: Specifies the blob endpoint to use. Defaults to *.blob.core.windows.net, but is useful for targeting storage emulators.
+    * `blobEndpoint`: Specifies the blob endpoint to use. Defaults to *.blob.core.windows.net, but is useful for targeting storage emulators. (e.g. storageaccountname.blob.region.azurestack.company.com)
```
NaraVen, line 116: the writing should " Optional parameter, if it is not specified defaults to blob.core.windows.net". This parameter should be specified for zonal dns accounts,  custom domain name accounts or storage emulators.

blobfuse/blobfuse.cpp
```diff
@@ -327,28 +341,45 @@ void *azs_init(struct fuse_conn_info * conn)
...
+            char errStr[] = "Unable to start blobfuse due to a lack of credentials. Please check the readme for valid auth setups.\n";
+            if (stdErrFD != -1) {
+                ssize_t n = write(stdErrFD, errStr, sizeof(errStr));
+                if (n == -1) {
+                    syslog(LOG_ERR, "Failed to report back the status of auth.");
+                }
+            }
+            syslog(LOG_ERR, "%s", errStr);
+            std::thread t1(std::bind(&destroyBlobfuseOnAuthError));
+            t1.detach();
```
NaraVen, line 369: does this detach succeed even if the main thread is executing? In my testing if I am in that dir it did not unmount.
Can we throw an error if unmount fails so that the user knows they have to unmount manually?

blobfuse/blobfuse.cpp
```diff
@@ -304,6 +306,18 @@ int read_config(const std::string configFile)
...
+    if (n == -1) {
+        syslog(LOG_ERR, "Failed to report back failure.");
+    }
+    close(stdErrFD);
+    fuse_unmount(config_options.mntPath.c_str(), NULL);
+}
+
 int configure_tls();
 void *azs_init(struct fuse_conn_info * conn)
 {
```
NaraVen, line 323: kernel_version < 4.16 can be replaced by blobfuse_constants.min_kernel_version or equivalent

NaraVen (changes requested):

blobfuse/blobfuse.cpp
vibhansa-msft: Will change accordingly.

blobfuse/blobfuse.cpp
vibhansa-msft: fuse_unmount is of void return type so we need to explicitly check whether unmount was successful or not. By the call itself we can not judge it.

vibhansa-msft: Have added mount directory check post unmount for the error reporting.

NaraVen (approved):

PR #565, opened 2021-02-08
Vibhansa/listapiblock

vibhansa-msft:
Block listing operation on mount for configured amount of seconds.

### DIFF
--- a/blobfuse/include/BlobfuseGlobals.h
+++ b/blobfuse/include/BlobfuseGlobals.h
@@ -63,6 +63,7 @@ struct configParams
     int defaultPermission;
     int concurrency;
     unsigned long long cacheSize;
+    volatile int  block_lst_sec;
     bool emptyDirCheck;
     std::string mntPath;
 };
@@ -84,6 +85,7 @@ struct cmdlineOptions
     const char *help; // print blobfuse usage
     const char *concurrency; // Max Concurrency factor for blob client wrapper (default 40)
     const char *cache_size_mb; // MAX Size of cache in MBs
+    const char *block_list_on_mount_for_secs; // Block the list api call on mount for n seconds
     const char *empty_dir_check;
     const char *encode_full_file_name; // Encode the '%' symbol in file name
 };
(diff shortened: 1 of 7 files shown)


NaraVen (commented):
conflicts need to be resolved here

blobfuse/src/BlobfuseGcCache.cpp
```diff
@@ -184,7 +188,7 @@ void gc_cache::run_gc_cache()
         else
         {
             // no file was timed out - let's wait a second
-            usleep(1000);
```
NaraVen, line 187: is this related to not listing on mount?

blobfuse/include/BlobfuseGlobals.h
```diff
@@ -84,6 +85,7 @@ struct cmdlineOptions
     const char *help; // print blobfuse usage
     const char *concurrency; // Max Concurrency factor for blob client wrapper (default 40)
     const char *cache_size_mb; // MAX Size of cache in MBs
+    const char *block_list_on_mount_for_secs; // Block the list api call on mount for n seconds
```
NaraVen, line 88: the purpose is to not do the default list so why use time here?

NaraVen (changes requested):

blobfuse/include/BlobfuseGlobals.h
vibhansa-msft: we need to block the list call for X seconds post mount. We do not have any way to determine its an implicit list by fuse driver or user driven operation, so blocking all list for few seconds after mount.

blobfuse/src/BlobfuseGcCache.cpp
vibhansa-msft: this was one ms sleep was was converted to 1 sec sleep that time. Now we need to make this configurable.

blobfuse/include/BlobfuseGlobals.h
NaraVen: blocking alone will not be enough, after mount it should not list at all.

vibhansa-msft: Updated.

NaraVen (approved):

PR #568, opened 2021-02-10
Added content-type as a header on put request.

eliaspuurunen:
Added a proof-of-concept extension to create_block_blob_request. We needed the ability to add content-type at time of upload. We didn't see a way to do this, so added it into the base class.


vibhansa-msft:
#580 has the required changes for content-type.

PR #575, opened 2021-03-05
Correcting parallelism factor for MSI

vibhansa-msft:
MSI download slowness due to improper parallelism fixed.

### DIFF
--- a/blobfuse/src/BlobfuseConstants.cpp
+++ b/blobfuse/src/BlobfuseConstants.cpp
@@ -9,7 +9,7 @@ namespace blobfuse_constants {
     const int max_concurrency_oauth = 2;
     const int max_retry_oauth = 5;
     const int max_concurrency_blob_wrapper = 40;
-    const int def_concurrency_blob_wrapper = 20;
+    const int def_concurrency_blob_wrapper = 12;
 
     const char* oauth_request_uri = "https://login.microsoftonline.com";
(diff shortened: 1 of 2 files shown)


blobfuse/src/BlobfuseConstants.cpp
```diff
@@ -9,7 +9,7 @@ namespace blobfuse_constants {
     const int max_concurrency_oauth = 2;
     const int max_retry_oauth = 5;
     const int max_concurrency_blob_wrapper = 40;
-    const int def_concurrency_blob_wrapper = 20;
```
NaraVen, line 12: Libfuse default is 12, I understand the reason behind this but will lowering from 20 to 12 cause performance issues?

NaraVen (approved):

blobfuse/src/BlobfuseConstants.cpp
vibhansa-msft: Till now default factor was 9 and this config was not used because blobfuse was not sending this as argument to upload/download functions in cpplite and library was using 9 as default factor. In Readme its mentioned as 12 being the default value so have corrected this value to 12. With this branch we will start using this factor of parallelism in both upload and download.

vibhansa-msft: Also this is configurable parameter, so if any customer observes performance issue we can adjust it using command line parameters

PR #580, opened 2021-03-16
Content-type can be updated based on file type using a cli option

vibhansa-msft:
- Populate content-type while uploading blob, based on file extension.

### DIFF
--- a/cpplite/include/blob/create_block_blob_request.h
+++ b/cpplite/include/blob/create_block_blob_request.h
@@ -27,7 +27,6 @@ namespace azure { namespace storage_lite {
             return blob_type::block_blob;
         }
 
-        //  Changes submitted by @eliaspuurunen :  PR #568
         std::string content_type() const override
         {
             return m_content_type;
--- a/cpplite/include/blob/create_block_blob_request.h
+++ b/cpplite/include/blob/create_block_blob_request.h
@@ -27,6 +27,18 @@ namespace azure { namespace storage_lite {
             return blob_type::block_blob;
         }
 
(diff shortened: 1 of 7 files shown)


cpplite/include/blob/create_block_blob_request.h
```diff
@@ -27,6 +27,18 @@ namespace azure { namespace storage_lite {
             return blob_type::block_blob;
         }
 
+        //  Changes submitted by @eliaspuurunen :  PR #568
```
NaraVen, line 30: remove comment

cpplite/src/blob/blob_client.cpp
```diff
@@ -184,6 +186,46 @@ std::future<storage_outcome<void>> blob_client::download_blob_to_buffer(const st
     return context->task_promise.get_future();
 }
 
+std::map<std::string, std::string> contentTypeMap {
+	{".css", "text/css"},
+	{".gif",  "image/gif"},
+	{".htm",  "text/html"},
+	{".html", "text/html"},
+	{".jpeg", "image/jpeg"},
+	{".jpg",  "image/jpeg"},
```
NaraVen, line 195: what about .avi, .wav. .mov, .mp3. , .mp4

vibhansa-msft: This list is taken from azcopy, so whatever they support we also added. Other then that they also go on to check first few bytes of file to determine the type but for now we will rely only on extension.

NaraVen (approved):

PR #581, opened 2021-03-19
Correcting disk usage calculation for non sudo users

vibhansa-msft:


blobfuse/src/BlobfuseGcCache.cpp
```diff
@@ -19,6 +19,18 @@ bool gc_cache::check_disk_space()
...
+    static bool user_detected = false;
+    static bool is_sudo_user = false;
+
+    if (!user_detected) {
+        // If effective uid = 0 then we are running as sudo
+        if (geteuid() == 0)
+            is_sudo_user = true;
+        user_detected = true;
+
+        AZS_DEBUGLOGV("GC_Cache User identified. Privillaged : %d", (int)(is_sudo_user));
```
NaraVen, line 31: typo privileged

blobfuse/src/BlobfuseGcCache.cpp
```diff
@@ -19,6 +19,18 @@ bool gc_cache::check_disk_space()
...
+    }
+
     if (config_options.cacheSize == 0) {
         struct statvfs buf;
         if(statvfs(cache_folder_path.c_str(), &buf) != 0)
```
NaraVen, line 36: Also log  an error if this stavfs lookup fails

blobfuse/src/BlobfuseGcCache.cpp
```diff
@@ -19,6 +19,18 @@ bool gc_cache::check_disk_space()
     double available;
     unsigned long long used;
 
+    static bool user_detected = false;
+    static bool is_sudo_user = false;
+
+    if (!user_detected) {
+        // If effective uid = 0 then we are running as sudo
+        if (geteuid() == 0)
+            is_sudo_user = true;
```
NaraVen, line 28: is it a bad idea to make a global variable for the is_sudo_user? instead of having a flag here to make sure sudo detection is done or not?

blobfuse/src/BlobfuseGcCache.cpp
vibhansa-msft: corrected.

blobfuse/src/BlobfuseGcCache.cpp
vibhansa-msft: Failure log added with errno.

blobfuse/src/BlobfuseGcCache.cpp
vibhansa-msft: These are static variables inside a function so their scope is not global rather limited to this function only.

NaraVen (approved):

PR #582, opened 2021-03-19
Update README.md

Amrinder-Singh29:
The logs may also get logged under var/log/message and not be var/log/syslog everytime. This depends on the Linux OS Family type as well.

Hence, proposing the changes so that Customers can check for both the of them in case they are not able to find syslog one.

### DIFF
--- a/README.md
+++ b/README.md
@@ -183,7 +183,7 @@ Please take careful note of the following points, before using blobfuse:
 	- to go back to your default logging level (provided in command line options) 
 		- remove the `logLevel` entry from config file 
 		- after saving config file send `SIGUSR1` to running instance of blobfuse.
-- By default logs are directed to system-configured syslog file e.g. /var/log/syslog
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -183,7 +183,7 @@ Please take careful note of the following points, before using blobfuse:
 	- to go back to your default logging level (provided in command line options) 
 		- remove the `logLevel` entry from config file 
 		- after saving config file send `SIGUSR1` to running instance of blobfuse.
-- By default logs are directed to system-configured syslog file e.g. /var/log/syslog
+- By default logs are directed to system-configured syslog file e.g. /var/log/syslog or var/log/message (Depending Upon the Linux OS Family)
```
NaraVen, line 186: that's right, but change the phrase system-configured syslog file to system-configured log file

Amrinder-Singh29: Thanks @NaraVen - Thanks for the review. The request has been re-submitted

NaraVen (approved):

PR #584, opened 2021-03-25
unblock gc fix for read only files

NaraVen:


blobfuse/src/BlobfuseGcCache.cpp
```diff
@@ -164,6 +164,19 @@ void gc_cache::run_gc_cache()
             {
                 //clean up the file from cache
                 int fd = open(mntPath, O_WRONLY);
+                if (errno == 13) {
+                    close(fd);
```
vibhansa-msft, line 168: We can add a log here stating this condition has hit and file is write protected. If the chmod fails we will not know whether this code flow was hit or not

blobfuse/src/BlobfuseGcCache.cpp
```diff
@@ -164,6 +164,19 @@ void gc_cache::run_gc_cache()
             {
                 //clean up the file from cache
                 int fd = open(mntPath, O_WRONLY);
+                if (errno == 13) {
```
vibhansa-msft, line 167: errno = 13 : then open was not successful so there is no need to close

blobfuse/src/BlobfuseGcCache.cpp
```diff
@@ -164,6 +164,19 @@ void gc_cache::run_gc_cache()
...
+                    close(fd);
+                    int fdr = open(mntPath, O_RDONLY);
+                    if (fdr > 0) {
+                        errno = 0;
+                        if (fchmod(fdr, config_options.defaultPermission) == 0) {
+                            AZS_DEBUGLOGV("GCClean clould not open for delete, so, Write protection overwritten for %s, resubmitting for cleanup", mntPath);
+                            }
+                        //TODO: If there is lock issue below convert permissions back to what it was.
+                        close(fdr);
+                    }
```
vibhansa-msft, line 177: If readonly also fails then we can dump a log saying this file will not get evicted from cache due to the error.

blobfuse/src/BlobfuseGcCache.cpp
```diff
@@ -164,6 +164,19 @@ void gc_cache::run_gc_cache()
             {
                 //clean up the file from cache
                 int fd = open(mntPath, O_WRONLY);
+                if (errno == 13) {
+                    close(fd);
+                    int fdr = open(mntPath, O_RDONLY);
+                    if (fdr > 0) {
+                        errno = 0;
+                        if (fchmod(fdr, config_options.defaultPermission) == 0) {
```
vibhansa-msft, line 172: chmod failure can also be logged here

NaraVen: done

blobfuse/src/BlobfuseGcCache.cpp
NaraVen: done

blobfuse/src/BlobfuseGcCache.cpp
NaraVen: done

NaraVen: done

blobfuse/src/BlobfuseGcCache.cpp
NaraVen: done

vibhansa-msft (approved):

PR #585, opened 2021-03-25
Narven/typo fix

NaraVen:
typo fix and one manual non-automated test file


test/not-automated/multiprocessing-threading/multithreading.py
```diff
@@ -0,0 +1,38 @@
+from threading import Thread
+import os
+
+def threaded_function():
+    top = "/home/nara/mntblobfuse/128/wiki_pretrain/"
```
vibhansa-msft, line 5: We want to make this script generic and add it to the pipeline ? I see the paths are hard-coded here.

blobfuse/src/BlobfuseGcCache.cpp
```diff
@@ -205,6 +205,10 @@ void gc_cache::run_gc_cache()
                             // Failed to acquire the lock for some other reason.  We close the open fd, and continue.
                             syslog(LOG_ERR, "Did not clean up file %s from file cache because we failed to acquire the flock for an unknown reason, errno = %d.\n", mntPath, errno);
                         }
+                        if (permissionsoverwritten == true)
+                        {
+                            fchmod(fd, originalpermissions);
```
vibhansa-msft, line 210: We shall also reset the permission when open call fails (around line 236)

NaraVen: done

vibhansa-msft (approved):

blobfuse/src/BlobfuseGcCache.cpp
vibhansa-msft, line 210: There is build failure due to this statement :

/home/vsts/work/1/s/blobfuse/src/BlobfuseGcCache.cpp:210:35: error: ‘originalpermissions’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
                             fchmod(fd, originalpermissions);

we need to initialize this variable in the begining.

PR #591, opened 2021-04-09
Update README.md

ShraddhaSSwadi:
We had a discussion with PG where it was mentioned that blobfuse shouldn't be used for concurrent write operations on same file. But, it can be used for concurrent write operations to different files in the same instance.

Hence, updating the description for the --max-concurrency option for the blobfuse. 

Adding this description will help customer to understand for what purpose they can use the --max-concurrency option for.


README.md
```diff
@@ -56,7 +56,7 @@ For more information, see the [wiki](https://github.com/Azure/azure-storage-fuse
...
     * [OPTIONAL] **--no-symlinks=true** : Turns off symlinks. Turning off symlinks will improve performance. Symlinks are on by default. This option is only available after version 1.3.1.
-    * [OPTIONAL] **--max-concurrency=20** : option to override default number of concurrent storage connections, default=12, please note that blobfuse caps this at 40. So if the user specifies a value between 0 to 40, the users value will override the default. However, if the user specifies a value > 40 blobfuse will still only allow 40 concurrent connections.
+    * [OPTIONAL] **--max-concurrency=20** : option to override default number of concurrent storage connections, default=12, please note that blobfuse caps this at 40. So if the user specifies a value between 0 to 40, the users value will override the default. However, if the user specifies a value > 40 blobfuse will still only allow 40 concurrent connections. Blobfuse does not recommend multi-instance write operations. It does allow concurrent write operations to different files in the same instance.
```
NaraVen, line 59: concurrency is for a single instance. A multi-instance comment doesn't belong here? if you see the"Limitations" section in the wiki we have explained that there will be no guaranteed integrity with writes from different nodes that is different instances . https://github.com/Azure/azure-storage-fuse/wiki/4.-Limitations-%7C-Differences-from-POSIX

NaraVen (changes requested):

vibhansa-msft:
As explained in the comments wiki mentions the details about max-concurrency. Config parameter here is about max-concurrency to be used in upload/download of a single file and not a system level limitation. Closing this change request.

PR #595, opened 2021-04-23
Narven/https proxy feature

NaraVen:
new feature to support blobfuse access to storage thru proxy server.


README.md
```diff
@@ -67,6 +67,8 @@ For more information, see the [wiki](https://github.com/Azure/azure-storage-fuse
      * [OPTIONAL] **--cache-poll-timeout-msec=1** : Timeout (milisecond) of checking the blobfuse cache for any possible eviction and free up space. Default is 1 milisecond.
      * [OPTIONAL] **--max-eviction=0** : How many cached files shall be evicted at a time, in case multiple are expired. This is to conserve the CPU usage done for cache eviction. Remaining files will be evicted in the next turn. '0' means there is no upper limit and all expired files will get evicted, may result in 100% CPU usage.
      * [OPTIONAL] **--set-content-type=false** : While uploading blob set the 'content-type' property based on file extension. Disabled by default, to enable set value to 'true'.
+     * [OPTIONAL] **--ca-cert-file=/etc/ssl/certs/proxy.pem** : If external network is only available through a proxy server, this parameter should specify the proxy pem certificate otherwise blobfuse cannot connect to the storage account.
+     * [OPTIONAL] **--httpsproxy=http://10.1.22.4:8080/** : If external network is only available through a proxy server, this parameter should specify the proxy server along with the port which is 8080 unless some alternatives have been made.
```
vibhansa-msft, line 71: We do support connection to container over http as well. Do we need to take care of http-proxy seperatly ?

blobfuse/blobfuse.cpp
```diff
@@ -47,7 +47,7 @@ const struct fuse_opt option_spec[] =
     OPTION("--use-attr-cache=%s", useAttrCache),
     OPTION("--use-adls=%s", use_adls),
     OPTION("--no-symlinks=%s", no_symlinks),    
-    OPTION("--no_symlinks=%s", no_symlinks),
+   // OPTION("--no_symlinks=%s", no_symlinks),
```
vibhansa-msft, line 50: This "_" option was add first and released and in later build we added new option with "-". In our Readme we have mentioned that both works, so I do not see any reason to remove this here.

blobfuse/blobfuse.cpp
```diff
@@ -248,6 +255,18 @@ int read_config(const std::string configFile)
             std::string resourceIdStr(value);
             config_options.resourceId = resourceIdStr;
         }
+        else if(line.find("caCertFile") != std::string::npos)
+        {
+            syslog(LOG_DEBUG, "caCertFile has a non null value");
+            std::string caCertFileStr(value);
+            config_options.caCertFile = caCertFileStr;
```
vibhansa-msft, line 262: We shall have a debug log here as well, stating we have hit this condition and using a particular file for cert

blobfuse/blobfuse.cpp
```diff
@@ -248,6 +255,18 @@ int read_config(const std::string configFile)
...
+        }
+        else if(line.find("httpsProxy") != std::string::npos)
+        {
+            std::string httpsProxyStr(value);
+            config_options.httpsProxy = httpsProxyStr;
```
vibhansa-msft, line 267: Minor : both cert path and https proxy are strings only so no need to read in a temp variable and then assign, we can directly read into config param.

blobfuse/blobfuse.cpp
NaraVen: I think we should remove the extra _option from ReadMe.txt

blobfuse/blobfuse.cpp
NaraVen: that gave me a segmentation fault error so did it this way

vibhansa-msft (changes requested):
There are lot of files which do not have any change and are still part of commit, can we restore them back to their original state.

blobfuse/blobfuse.cpp
vibhansa-msft: We can remove but just in case someone is using it then it will be a breakage for them,

blobfuse/blobfuse.cpp
vibhansa-msft: Ok, then we can keep it this way.

blobfuse/blobfuse.cpp
NaraVen: I see that we don't have it in the readme file

NaraVen: By now, everyone should have fixed their links, I will add it to the revision notes

vibhansa-msft: Earlier it was there in Readme, I remember.  May be we cleaned it up in some commit, Fine then we can remove this.

README.md
NaraVen: yes, we do but when you set http_proxy and https_proxy, normally you should set them at the same URL including the schema part. When, using the proxy for connecting to a HTTPS website, the proxy works in TCP mode, and your TCP connection passes through the proxy carrying encrypted packets from your client to the target website.

blobfuse/blobfuse.cpp
```diff
@@ -70,7 +72,9 @@ const struct fuse_opt option_spec[] =
 int read_config_env()
 {
     char* env_account = getenv("AZURE_STORAGE_ACCOUNT");
+    fprintf(stdout, "done getenv storage account");
```
vibhansa-msft: why do we need to print these messages on the cli output ?

cpplite/include/http/libcurl_http_client.h
```diff
@@ -364,12 +364,12 @@ namespace azure {  namespace storage_lite {
...
 
         //Sets CURL CA BUNDLE location for all the curl handlers.
-        CurlEasyClient(int size, const std::string& ca_path) : m_size(size), m_capath(ca_path)
+        CurlEasyClient(int size, const std::string& ca_path, const std::string& https_proxy) : m_size(size), m_capath(ca_path), m_proxy(https_proxy)
         {
+            syslog(LOG_DEBUG, "In libcurl CurlEasyClient ca path= %s", ca_path.c_str());
             curl_global_init(CURL_GLOBAL_DEFAULT);
             for (int i = 0; i < m_size; i++) {
                 CURL *h = curl_easy_init();
-                curl_easy_setopt(h, CURLOPT_CAPATH, ca_path.c_str());
```
vibhansa-msft, line 372: Where are we using the ca_path variable. In other layers its just coming till here.

README.md
NaraVen: added http_proxy as well so that it will be used for http only connections. Done!

cpplite/include/http/libcurl_http_client.h
NaraVen: Curl uses it.

vibhansa-msft (approved):

PR #629, opened 2021-08-02
Adding background download support

vibhansa-msft:
- Allow user to download the file in background instead of blocking open system call
- sync can force evict a file from cache, behavior is configurable.


README.md
```diff
@@ -76,7 +76,8 @@ For more information, see the [wiki](https://github.com/Azure/azure-storage-fuse
...
      * [OPTIONAL] **--basic-remount-check=false** : Set this to true if you want to check for an already mounted status using /etc/mtab instead of calling the syscall setmntent. Default is true. It is known that for AKS 1.19, blobfuse will throw a segmentation fault error, so set this to false.  This option is only available from version 1.3.8
      * [OPTIONAL] **--pre-mount-validate=false** : Set this to true to skip the cURL version check and just straight validate storage connection before mount. Default is false, so use this only if you know that you have the recent Curl version, otherwise blobfuse will hang. This option is only available from version 1.3.8
-
+    * [OPTIONAL] **--background-download=false** : Instead of downloading the file in 'open' system call, download it in background post 'open'. Follow up 'read'/'write' calls will wait for download to complete. This option is only available from version 1.4.0
```
NaraVen, line 79: Set this true if you want file download to run in the background on open. Setting this to true will put a wait on 'read'/'write' calls till download completes. If the file is already in the local cache this switch is not evaluated. Default value is false. This option is only available from version 1.4.0

README.md
```diff
@@ -76,7 +76,8 @@ For more information, see the [wiki](https://github.com/Azure/azure-storage-fuse
...
+    * [OPTIONAL] **--evict-on-sync=false** : When file cache timeout is set to a higher value and user wants to force evict a file from cache use the sync command to do the same. If set to false, sync command on file is ignored. This option is only available from version 1.4.0
```
NaraVen, line 80: Set this to true if you want the particular file or directory content and attribute cache to be invalidated when the linux "sync" command is issued on a file or on a directory. Default is false.  This option is only available from version 1.4.0

blobfuse/utilities.cpp
```diff
@@ -481,9 +494,13 @@ int azs_access(const char * /*path*/, int /*mask*/)
     return 0; // permit all access
 }
 
-int azs_fsync(const char * /*path*/, int /*isdatasync*/, struct fuse_file_info * /*fi*/)
+int azs_fsync(const char * path, int /*isdatasync*/, struct fuse_file_info *fi)
 {
-    return 0; // Skip for now
```
NaraVen, line 486: we need this for directories also

vibhansa-msft: I tried to test 'sync' command on directory but did not get any call in blobfuse for the same. 'sync <filename>' consistently resulted in fsync call on our end.

vibhansa-msft: Also as you can see this fsync is coming with a handle info, which was returned back in 'open' file call so not sure if libfuse supports fsync of directory or not. We do not support opendir call in blobfuse.

blobfuse/fileapis.cpp
```diff
@@ -12,6 +12,53 @@ std::deque<file_to_delete> cleanup;
...
+        return 0 - map_errno(storage_errno);
+    }
+    else
+    {
+        syslog(LOG_INFO, "Successfully downloaded blob %s into file cache as %s.\n", pathString.c_str()+1, mntPathString.c_str());
+        g_gc_cache->addCacheBytes(mntPathString, size);
+    }
+
+    // preserve the last modified time
+    struct utimbuf new_time;
```
NaraVen, line 39: these 3 lines are not required. It retains the last modified time anyway

blobfuse/fileapis.cpp
```diff
@@ -12,6 +12,53 @@ std::deque<file_to_delete> cleanup;
...
+    new_time.modtime = last_modified;
+    new_time.actime = 0;
+    utime(mntPathString.c_str(), &new_time);
+
+    if (is_delayed) {
+        if (storage_client->isADLS()) {
+            // preserve the last modified time
+            struct utimbuf new_time;
```
NaraVen, line 47: just like above the below 3 lines are not required.

blobfuse/include/BlobfuseGcCache.h
```diff
@@ -20,7 +21,7 @@ class gc_cache
         file_cache_timeout_in_seconds(timeout),
         m_current_usage(0){}
     void run();
-    void uncache_file(std::string path);
+    void uncache_file(std::string path, bool force = false);
```
NaraVen, line 24: will the force work when the file is open?

blobfuse/src/AttrCacheBfsClient.cpp
```diff
@@ -528,4 +528,18 @@ void AttrCacheBfsClient::GetExtraProperties(const std::string pathName, BfsFileP
 int AttrCacheBfsClient::RefreshSASToken(std::string sas)
 {
     return blob_client->RefreshSASToken(sas);
+}
+
+
+void AttrCacheBfsClient::InvalidateFile(const std::string blob)
+{
```
NaraVen, line 535: Either add InvalidateDir or remove this from this PR, since the request is invalidate both a file and a directory

blobfuse/utilities.cpp
NaraVen: The AML CI Jupyter team will use the Jupyter Notebook command to issue a “sync <filename>” whenever they want to ensure file contents are not stale. They will issue a  “sync dirname” command to ensure git status results are not stale, in  other ways refresh dir listing.

These will invoke the blobfuse fsync  or fsyncdir respectively.

Refer to Workitem: https://msazure.visualstudio.com/One/_backlogs/backlog/XStore-DevExp/Epics/?workitem=10196695
The fsync will invalidate file attribute cache on the kernel and the memory and delete the content cache for the file from the disk.

The fsyncdir will invalidate the directory list of attributes cache for all its files and subdirs.\

blobfuse/include/FileLockMap.h
```diff
@@ -20,7 +20,8 @@ class file_lock_map
 public:
     static file_lock_map* get_instance();
     std::shared_ptr<std::mutex> get_mutex(const std::string& path);
-
```
NaraVen, line 23: It is risky to obtain multiple mutexs at the same time. Specify a lock order for the mutexs

blobfuse/src/BlobfuseGcCache.cpp
```diff
@@ -162,8 +171,10 @@ void gc_cache::run_gc_cache()
 
             struct stat buf;
             stat(mntPath, &buf);
-            if (((now - buf.st_mtime) > file_cache_timeout_in_seconds) ||
-                disk_threshold_reached)
+            if (file_cache_timeout_in_seconds == 0 ||
+                disk_threshold_reached ||
+                file.force ||
+                (((now - buf.st_mtime) > file_cache_timeout_in_seconds) && ((now - buf.st_ctime) > file_cache_timeout_in_seconds)))
```
NaraVen, line 177: ctime is not needed, if it is greater than modified time then it is greater than create time.

blobfuse/include/BlobfuseGcCache.h
vibhansa-msft: On close we just submit the file to gc-cache for eviction. If the force flag is truned on its submitted to the front of the list to give it high priority. Rest of the logic for eviction remains same. If file is open or we fail to take exclusive lock it can not be evicted.

blobfuse/src/BlobfuseGcCache.cpp
vibhansa-msft: ctime is change time and we can not rely on mod-time alone because in case of adls mod time reflects the last modified time of the file as per container and it can be long back then current time. Change time is updated everytime we download the file as we set its mod. Hence relying on ctime is more accurtae. In azs_open where we decide to redownload the file same logic is applied by checking both the times.

blobfuse/include/FileLockMap.h
vibhansa-msft: There are two mutexes. One is applied only to block download and deletion of file. As per existing code when file is under download we are not allowed to delete it. In case of delayed download the file locks will be releases as soon as open call returns. However we still need to block read/write calls as the file download may be going in background. Hence we have added another level of locks just to mark that file download is still going on in background and do not allow read/write.

blobfuse/fileapis.cpp
vibhansa-msft: We want last modified time to reflect the one mentioned in container. When we create a local file and dump data into it system will change time to current time, hence post dumping the data we need to reset the last modified time back to one mentioned in container. We were doing the same earlier also just that download of file logic has been moved to another function now.

blobfuse/fileapis.cpp
vibhansa-msft: It can be taken care in above step so will remove this redundant reset of mod time.

blobfuse/utilities.cpp
vibhansa-msft: As of now blobfuse does not register and handel fsyncdir, this will need new registeration then. Also handling fsyncdir also means we need to handle opendir and closedir, which do not have any implementation as of now.

vibhansa-msft: Once fsyncdir is registered, we are getting the dir name as well in the api so relying on that we can ignore opendir and closedir. As per the task item we just need to invalidate the attr-cache on fsyncdir so have done code changes for that. However there is one concern, even if we invalidate the attr-cache, getattr on file may still rely on disk cached file for attributes so we need to do sync on those cached files seperatly.

blobfuse/src/AttrCacheBfsClient.cpp
vibhansa-msft: Added and invalidating attr-cache as per the task item.

blobfuse/utilities.cpp
NaraVen: without fysndir how will the dir attributes be cleared even if you add the code to AttrCacheBfsCllient?
Also when you delete the attribute cache for a dir, you have to delete the attribute for all the files and subdirs in it too. The best thing at this time is to remove remove the whole attribute invalidation from  the PR and keep it to streaming alone.

vibhansa-msft: In fsyncdir we are getting the directory path being synced. Using that in attr-cache layer we are iterating through all elements having that path and invalidating the cache. So its working as per the expectation given in the task. In task it only talks about invalidating the attr-cache and not deleting files on that path to be removed from the disk cache.

NaraVen (approved):

gapra-msft (approved):

PR #639, opened 2021-08-30
Adding MSI_SECRET env variable to init flow of adls_msi auth mode

vibhansa-msft:
ADLS Gen-2 account with MSI auth is not working if MSI_SECRET env variable is set along with msi endpoint.


gapra-msft (approved):
Do we have tests for ADLS MSI via env variable? Should we add some if not?

PR #641, opened 2021-09-06
Vibhansa/read stream parallel download block

vibhansa-msft:
If the block size if bigger then 64MB, instead of downloading in single thread, download in parallel for better performance.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,15 +1,15 @@
 cmake_minimum_required(VERSION 2.8)
 
 # Change this while bumping up the version, needed for RHEL package
-set(VERSION "1.4.0")
+set(VERSION "1.4.1")
 
 # This macro is referred in code to print the current version
 # Change CPACK_DEBIAN_PACKAGE_DESCRIPTION as well in this file while bumping up the version
-add_definitions(-D_BLOBFUSE_VERSION_="1.4.0")
+add_definitions(-D_BLOBFUSE_VERSION_="1.4.1")
 
 set(CPACK_PACKAGE_VERSION_MAJOR 1)
 set(CPACK_PACKAGE_VERSION_MINOR 4)  
-set(CPACK_PACKAGE_VERSION_PATCH 0)
+set(CPACK_PACKAGE_VERSION_PATCH 1)
 set(CPACK_PACKAGE_VERSION_RELEASE 1)
 
 set (BLOBFUSE_HEADER
@@ -127,7 +127,7 @@ if(UNIX)
   set(CPACK_GENERATOR "DEB")
   set(CPACK_DEBIAN_PACKAGE_MAINTAINER "Microsoft - Azure Storage")
   set(CPACK_DEBIAN_PACKAGE_DEPENDS "fuse")
-  set(CPACK_DEBIAN_PACKAGE_DESCRIPTION "blobfuse 1.4.0 - FUSE adapter for Azure Blob Storage")
+  set(CPACK_DEBIAN_PACKAGE_DESCRIPTION "blobfuse 1.4.1 - FUSE adapter for Azure Blob Storage")
   set(CPACK_PACKAGE_NAME "blobfuse")
   set(CPACK_PACKAGE_VENDOR "Microsoft")
   include(CPack)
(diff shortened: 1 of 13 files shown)


blobfuse/src/BlobStreamer.cpp
```diff
@@ -1,6 +1,8 @@
 #include <BlobStreamer.h>
 #include <base64.h>
 
+#define MAX_BLOCK_SIZE_FOR_SINGLE_READ (64 * 1204 * 1024)
```
tasherif-msft, line 4: we're opting to not have it be configurable?

blobfuse/src/BlobStreamer.cpp
```diff
@@ -95,7 +97,30 @@ BlobBlock* BlobStreamer::GetBlock(const char* file_name, uint64_t offset, Stream
...
+                    return NULL;
+                }
+            }
+        }
+
+        if (download_size < MAX_BLOCK_SIZE_FOR_SINGLE_READ) {
+            azclient->DownloadToStream(file_name, block->buff, start_offset, download_size);
+        } else {
+            char *buff = (char*)malloc(download_size);
+            azclient->DownloadToBuffer(file_name, buff, start_offset, download_size, config_options.concurrency);
```
tasherif-msft, line 119: I think we should explain to the user in the README how max-concurrency attribute will be used in the streaming case? since this is ignored if the block size is smaller than 64MB.

blobfuse/src/BlobStreamer.cpp
vibhansa-msft: Block size is still configurable, we are just deciding whether to do download in single thread or in parallel based on 64MB boundary condition.

blobfuse/src/BlobStreamer.cpp
vibhansa-msft: Updated readme for the same.

tasherif-msft (approved):

CMakeLists.txt
```diff
@@ -1,15 +1,15 @@
 cmake_minimum_required(VERSION 2.8)
 
 # Change this while bumping up the version, needed for RHEL package
-set(VERSION "1.4.0")
+set(VERSION "1.4.1")
```
gapra-msft, line 4: why not keep this as 1.4.0? We havent released this yet right?

README.md
```diff
@@ -81,7 +81,7 @@ For more information, see the [wiki](https://github.com/Azure/azure-storage-fuse
     * [OPTIONAL] **--streaming=false** : Enable read streaming of files instead of disk-caching. This option works only with read-only mount. This option is only available from version 1.4.0
     * [OPTIONAL] **--stream-cache-mb=500** : Limit total amount of data being cached in memory to conserve memory footprint of blobfuse.
     * [OPTIONAL] **--max-blocks-per-file=3** : Maximum number of blocks to be cached in memory for a read streaming.
-    * [OPTIONAL] **--block-size-mb=16** : Size (in MB) of a block to be downloaded during streaming.
+    * [OPTIONAL] **--block-size-mb=16** : Size (in MB) of a block to be downloaded during streaming. If configured block-size is <= 64MB then block will be downloaded in a single thread. For higher block size parts of it will be downloaded in parallel. "--max-concurrency" parameter can be used to control parallelism factor. When higher block size is configured, memory usage of blobfuse will be high as these blocks are cached in memory only.
```
gapra-msft, line 84: should we hardcode this 64mb value? why not make this user settable as well?

tasherif-msft:
I'm curious if we should have a separate max-concurrency attr specific to streaming. If the customer sets it then that means regardless of their block size they want to parallelize. This also enables us to avoid the hardcoded 64MB

CMakeLists.txt
vibhansa-msft: Some customers have taken the code to try out streaming and sync feature and there version is already 1.4.0. Later it may create confusion on whether its official build or the local build. Just to avoid that confusion later while debugging issues bumping up this version to 1.4.1

README.md
vibhansa-msft: Just felt whether to use single thread or parallel download shall be left to internal implementation. Some one may configure 1mb block also be to downloaded in parallel thread, which will be overkill.

vibhansa-msft:
> I'm curious if we should have a separate max-concurrency attr specific to streaming. If the customer sets it then that means regardless of their block size they want to parallelize. This also enables us to avoid the hardcoded 64MB

We do not want user to do parallel download even for smaller buffer sizes. if we leave the option open, you can configure even smaller chunks to be downloaded by large number of threads, which will be overkill. So unless chunk size is >64MB we do not want to do things in parallel. This is matching with current blobfuse upload code where files upto 64MB are not broken down into blocks and are downloaded by single thread as single block. Once block size is bigger we try to do stuff in parallel. Reason for reusing the same config param is when streaming is enable that option will be of no use in regular flow, and as it control the same thing (number of parallel threads to run to download/upload file) just reused the same.

gapra-msft (approved):

PR #648, opened 2021-09-22
Mcdhee/phantom

vibhansa-msft:


blobfuse/blobfuse.cpp
```diff
@@ -348,22 +376,22 @@ int read_config(const std::string configFile)
...
-    {
-        syslog (LOG_CRIT, "Unable to start blobfuse. Container name is missing in the config file.");
-        fprintf(stderr, "Unable to start blobfuse. Container name is missing in the config file.\n");
-        return -1;
-    }
-    else
-    {
-        return 0;
+    if (!config_options.enableGen1) {
+        if(config_options.accountName.empty())
```
NaraVen, line 380: accountName cannot be empty for gen1 also. Container name can be empty but if it is not empty it should get appended at the end for the resoruceid. adlsgen1 folder is treated like blob container

blobfuse/blobfuse.cpp
```diff
@@ -762,24 +797,35 @@ bool is_directory_empty(const char *tmpDir) {
...
-read_and_set_arguments(int argc, char *argv[], struct fuse_args *args)
+int read_and_set_arguments(int argc, char *argv[], struct fuse_args *args)
 {
     // FUSE has a standard method of argument parsing, here we just follow the pattern.
     *args = FUSE_ARGS_INIT(argc, argv);
 
     // Check for existence of allow_other flag and change the default permissions based on that
     config_options.defaultPermission = 0770;
     config_options.readOnlyMount = false;
+    config_options.entryTimeout = 240;
```
NaraVen, line 808: any reason this default is added here. datasetfuse has its own defaults optimized for performance as per what they think

blobfuse/blobfuse.cpp
```diff
@@ -1140,95 +1176,93 @@ read_and_set_arguments(int argc, char *argv[], struct fuse_args *args)
         std::string max_retry(cmd_options.max_retry);
         config_options.maxTryCount = stoi(max_retry);
     }
-    
+
+    if (config_options.entryTimeout == -1){
+        config_options.entryTimeout = 240;
```
NaraVen, line 1181: skip it from json, don't default, same reason as above

blobfuse/blobfuse.cpp
```diff
@@ -1331,3 +1378,99 @@ int initialize_blobfuse()
...
+            j["authorityurl"] = "https://login.microsoftonline.com";
+        }
+
+        j["credentialtype"] = "servicePrincipal";
+    } else{
+        syslog(LOG_ERR, "For Gen1 account only SPN auth is supported.");
+        return -1;
+    }
+
+    j["resourceurl"] = "https://datalake.azure.net/";
```
NaraVen, line 1407: here , pls add a check to see if the container name is not empty. If it is not empty then append containername + "/" at the end. ADLS gen1 folder works similar to a container so lets just keep that variable name.

blobfuse/blobfuse.cpp
dheeraj-coding: the json doesn't get these default values so nothing to worry

blobfuse/blobfuse.cpp
dheeraj-coding: append to resourceurl ? or resourceid?

blobfuse/blobfuse.cpp
vibhansa-msft: default set to 0 now

blobfuse/blobfuse.cpp
vibhansa-msft: Does customer really need a feature to mount a specific folder within container. If such feature is not there in existing solution lets skip this part. Also if we decide to add this anyway then shouldn't resourceID also needs similar modification ?

NaraVen: It is there, meaning they can mount root  - empty container, also means the whole account
or
they can mount subfolders , we can use the container-name for subfolder.
I have commented on the resourceid also to append the container name if any

blobfuse/blobfuse.cpp
```diff
@@ -1438,13 +1444,17 @@ int mount_rust_fuse(char* argv[]){
         j["requiredfreespaceinmb"] = config_options.requiredFreeSpace;
     }
 
+    if (config_options.tmpPath != ""){
```
NaraVen, line 1443: perfect

PR #666, opened 2021-11-10
Mask out SYNC and DIRECT flags in open call as write operation fails …

vibhansa-msft:
#665 : Mask out SYNC and DIRECT flags in open call as write operation fails if they are given in open call


blobfuse/fileapis.cpp
```diff
@@ -62,6 +62,11 @@ int DownloadFileToDisk(std::string pathString, std::string mntPathString, bool i
 int azs_open(const char *path, struct fuse_file_info *fi)
 {
     syslog (LOG_DEBUG, "azs_open called with path = %s, fi->flags = %X.\n", path, fi->flags);
+    if (fi->flags & O_SYNC || fi->flags & __O_DIRECT) {
```
gapra-msft, line 65: Can we add a comment here for why we do this?

gapra-msft (approved):

Issue #668, opened 2021-11-11
Rewriting file on one node is not always noticed on another - with file-cache-timeout-in-seconds=0


cris-b:
### Which version of the blobfuse was used?

1.3.8, 1.4.1

### Which OS (please include version) are you using?

Ubuntu 18.04 (AKS)

### What problem was encountered?

When running blobfuse on two nodes, with the cache timeout set to zero, I observed that sometimes one node does not get the correct file content, in this pattern:

node 1:  echo hello > helloworld
node 2:  cat helloworld
hello
node 1: echo hello2 > helloworld
node2: cat helloworld
hello
node 1: echo hello3 > helloworld
node2: cat helloworld
hello

So this is wrong, and no amount of waiting updates the file.  What does update the file on node 2 is appending:

node1: echo hello3 >> helloworld
node2: cat helloworld
hello3
hello3

So appending always works and we get the fresh copy of the file instantly on the other node.  But rewriting the file a second time seems to cause things to become out of sync.

### Have you found a mitigation/solution?

No.

### If relevant, please share your mount command.

--file-cache-timeout-in-seconds=0 --use-adls=true

vibhansa-msft:
For small files, kernel will also cache the contents and the read operation will be served from the kernel itself, it will not even land to blobfuse. Along with file-cache-timeout=0 try cleaning up kernel cache manually and validate.

vibhansa-msft:
On node2 where you are doing cat use this before cat and see : sysctl -w vm.drop_caches=3
Also file-cache-timeout shall be 0 on all three nodes. Blobfuse caches attributes of file as well so you can disable that by removing --attr-cache cli param. However this will reduce performance a bit.

cris-b:
The docs suggest that attr-cache is false by default?

setting -o direct_io on the mount seems to resolve this problem, as this stops the kernel caching.  But from the auto_cache parameter which to an extent (supposedly) controls this, it mentions that the cache refresh is based on file size and mtime.  But blobfuse only seems to be regarding the file size.  mtime on the file is updated

.

vibhansa-msft:
- Blobfuse refresh the cache based on file-cache-timeout and its counted from last downloaded, and eviction is only possible after all handles are closed for that file. We can not rely on last modified time as it reflects the actual time when file was last updated in container, and it can be days old as well. 
- Attr-cache is disabled by default.
[...]

PR #670, opened 2021-11-18
Initial changes for fuse 3 support

cris-b:
This makes it possible to conditionally build against fuse3, prinicipally so that FUSE_CAP_AUTO_INVAL_DATA can be used (see #668)

This results in a working file system for me when building against fuse 3, although I havent extensively tested it.

if AZS_FUSE3 environment variable is defined before running build.sh fuse3 will be used.  

I have left in fuse 2 code path because it seems fuse 2 would still need to be used for some of the older supported OS's (I'm using the OS's that are included in the binary build artifacts on the release page as a guide for what is supported 'officially'):

* Debian 9 (EOL June 2022)
* Ubuntu 18.04 (EOL April 2023)

I defined FUSE_USE_VERSION to 34 if using fuse3, as Debian 10 is the oldest supported OS which contains fuse 3, and that uses 3.4.1

There are zero changes to the code when building against fuse 2.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -75,6 +75,15 @@ elseif( EXISTS "/etc/os-release")
   set(Boost_USE_STATIC_LIBS OFF)
 endif()
 
+if(AZS_FUSE3)
+  set(FUSE_LIB fuse3)
+  set(FUSE_INCLUDES /usr/include/fuse3)
+  add_definitions(-DAZS_FUSE3)
+else()
+  set(FUSE_LIB fuse)
+  set(FUSE_INCLUDES "")
+endif()
+
 if(UNIX)
   #set( CMAKE_VERBOSE_MAKEFILE on )
   project(blobfuse)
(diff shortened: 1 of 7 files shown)


vibhansa-msft:
Thanks for integrating fuse3 in blobfuse. We are close to our release cycle so will not take this PR up for the next release but will discuss internally on how to proceed with this.

cris-b:
following 1.4.2 is there any any update on this?

vibhansa-msft:
@cris-b : no plans for fuse3 so far, will keep this PR in open state only. If you need this for your use case you can continue using your forked branch.

vibhansa-msft:
Problem with this change right now is its compile time check, so going this path we need to release two binaries one with fuse2 option and one with fuse3. Moving completely to fuse3 poses another issue that many older linux distros do not support it.

cris-b:
but as mentioned in the comments

a - you already produce a different packages for different linux distros, and you dont produce 'generic' binaries.  Anyone not using one of those distros already has to compile themselves, in which case they get to choose.
b - only 2 of those distros do not support fuse 3 (debian 9, ubuntu 18.04), and ubuntu 9 is eol in six months.

so surely compiling for fuse 3 with all the distros except the above is no different to compiling a different binary for difference distros anyway?

vibhansa-msft:
As you pointed out fuse3 is not supported on ubn-18 and deb-9, so for these distros we need to compile with fuse2 and rest with fuse3. Now the customers who are on updated distros takes our latest release will suddenly start observing failure as blobfuse will look for fuse3 while customers have fuse2 installed. So we will have to request all customers also to move to fuse3. My point here is its a breaking change, this upgrade from fuse2 to fuse3 can not be seamless. Also for customers who have been using fuse2 for long, do not get any significant advantage by moving to fuse3 and hence they may not show much interest in upgrading and potentially de-stabilizing the system.

cris-b:
but as you are providing distro packages for all those distros (which specify their deps as requirements), it doesn't matter to the end user whether it pulls in fuse2 or 3 - it should be transparent.  both of them can co-exist happily and they should not notice the difference.

from the last libfuse 2 release (2.9.9) almost 3 years ago:
```Note: This is a maintenance release from the 2.9 branch of libfuse. Users are encouraged to transition to the
actively developed libfuse 3.x.```


libfuse3 was released 5 years ago.  It should also be noted that we are talking about the userspace library here, the kernel moves on regardless.

libfuse2 is essentially unmaintained at this point.

vibhansa-msft:
We have just previewed our next generation of blobfuse (https://github.com/Azure/azure-storage-fuse/releases/tag/blobfuse2-2.0.0-preview.1). It is fuse3 compliant and does enable FUSE_CAP_AUTO_INVAL_DATA by default. Closing this PR as we will not be backporting this feature to blobfuse.

Issue #675, opened 2021-12-14
segfault when using streaming=true 


jeffjjohnston:
### Which version of the blobfuse was used?

1.4.2

### Which OS (please include version) are you using?

Red Hat Enterprise Linux Server release 7.9 (Maipo)

### What problem was encountered?

blobfuse segfaults consistently when streaming is enabled

### Have you found a mitigation/solution?

Using `streaming=false` prevents the segfaults, but we'd prefer to use the new streaming feature. We are attempting to have Apache serve read-only files mounted on blobfuse. These are very large files that are accessed via Apache in small chunks via HTTP range requests. With `streaming=false`, there is a large delay due to blobfuse transferring the entire file to the local cache. Streaming appears to work for a few initial requests, but will quickly segfault.

### By default, blobfuse logs errors to syslog. If this is relevant, is there anything in the syslog that might be helpful?

When running with LOG_DEBUG, no messages precede the segfault:

```
Dec 14 20:03:17 server blobfuse[21118]: Using syscall to detect directory is already mounted or not
Dec 14 20:03:17 server blobfuse[21118]: Setting logging level to : LOG_DEBUG
Dec 14 20:03:17 server blobfuse[21118]: Disk Thresholds : 90 - 80, Cache Eviction : 1000-0, List Cancel time : 0 Retry Policy (0, 60.000000, 1.200000), Background Download 0, Evict on sync 0
Dec 14 20:03:17 server blobfuse[21118]: Streaming : 1, Stream buffer size : 1073741824, Max Blocks : 3, Block Size : 67108864
Dec 14 20:03:17 server blobfuse[21118]: Blobfuse version : 1.4.2
Dec 14 20:03:17 server blobfuse[21118]: release is 3.100000, Kernel version is 3.10.0-1160.45.1.el7.x86_64
Dec 14 20:03:17 server blobfuse[21118]: CURL Version is 7.29.0
Dec 14 20:03:17 server blobfuse[21118]: ** Delaying tls init to post fork for older libcurl version
Dec 14 20:03:17 server blobfuse[21118]: ** Delaying authentication to post fork for older curl versions
Dec 14 20:03:17 server blobfuse[21125]: ** Post fork authentication for older libcurl version
Dec 14 20:03:33 server kernel: blobfuse[21179]: segfault at 9 ip 00007f903bdbe4a9 sp 00007f902ef661f0 error 4 in libc-2.17.so[7f903bd3e000+1c4000]
```

[...]

Issue #674, opened 2021-12-08
CONTENT-TYPE is not applied on big files >= 100Mb


bo44arov:
### Which version of the blobfuse was used?
1.4.2

### Which OS (please include version) are you using?
ubuntu 18.04

### What problem was encountered?
When using option --set-content-type=true the content type is not applied for large files and still is "application/octet-stream".

I tried it on .mp4 files 10-50 Mb - it works properly and in blobstorage I see file content-type as video/mp4.
But if I use file 100 Mb and more - content-type is application/octet-stream

![image](https://user-images.githubusercontent.com/54901472/145236631-d0c83571-a688-447f-9c30-01fc097af82b.png)

and

![image](https://user-images.githubusercontent.com/54901472/145236761-c684e33c-1297-4394-b7d0-03bc64955d19.png)

### Have you found a mitigation/solution?
No

vibhansa-msft:
In cpplite sdk that we are using I see for files > 64MB in size content-type is not set in the put-block-list rest api call.

vibhansa-msft:
This issue is fixed in 1.4.3, kindly upgrade
https://github.com/Azure/azure-storage-fuse/releases/tag/blobfuse-1.4.3

bo44arov:
> This issue is fixed in 1.4.3, kindly upgrade https://github.com/Azure/azure-storage-fuse/releases/tag/blobfuse-1.4.3

Great, thank you!

PR #678, opened 2021-12-17
Vibhansa/stream crash

vibhansa-msft:
#674 : Set content type correctly for files > 64MB in size
#675 : SegFault in streaming when open and close come in parallel for same file.

### DIFF
--- a/blobfuse/include/BlobStreamer.h
+++ b/blobfuse/include/BlobStreamer.h
@@ -95,11 +95,6 @@ class StreamObject {
         int DecRefCount() {
             //std::lock_guard<std::mutex> lock(m_mutex);
             ref_count--;
-
-            if (ref_count == 0) {
-                Cleanup();
-            }
-
             return ref_count;
         }
 
--- a/blobfuse/include/BlobStreamer.h
+++ b/blobfuse/include/BlobStreamer.h
@@ -98,6 +98,10 @@ class StreamObject {
             return ref_count;
         }
 
+        int GetRefCount() {
+            return ref_count;
+        }
+
         void SetSize(uint64_t size) {
             file_size = size;
         }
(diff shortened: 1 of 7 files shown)


blobfuse/include/BlobStreamer.h
```diff
@@ -87,24 +87,18 @@ class StreamObject {
         }
 
         int IncRefCount() {
```
gapra-msft, line 89: Should we add a comment here to mention that users of this method must hold the lock before calling this? For this method, Dec and GetRefCount

blobfuse/include/BlobStreamer.h
```diff
@@ -87,24 +87,18 @@ class StreamObject {
...
             return ref_count;
         }
 
         int DecRefCount() {
-            std::lock_guard<std::mutex> lock(m_mutex);
+            //std::lock_guard<std::mutex> lock(m_mutex);
             ref_count--;
-
-            if (ref_count == 0) {
-                Cleanup();
```
gapra-msft, line 100: Was this call to Cleanup unnecessary cause we have code that handles it in BlobStreamer.cpp line 214?

gapra-msft (approved):

tasherif-msft (approved):

blobfuse/include/BlobStreamer.h
vibhansa-msft: It was redundant, in closeFile flow we were decrementing the count and also it was checking if its 0 then cleanup. Hence removed a implicit cleanup.

blobfuse/include/BlobStreamer.h
vibhansa-msft: Will add comments

PR #681, opened 2022-01-05
Vibhansa/pathmessup

vibhansa-msft:
When blob name is merged with temp cache path to form a local path, in some combination its reaching out of temp-cache directory path. Blobfuse shall invalidate such paths and return ENOENT

### DIFF
--- a/blobfuse/utilities.cpp
+++ b/blobfuse/utilities.cpp
@@ -188,8 +188,8 @@ int azs_getattr(const char *path, struct stat *stbuf)
                 char *result = realpath(mntPathString.c_str(), real_path);
                 if (result != NULL) {
                     // Path got resolved. lets check we are not moving away from temp path
-                    // If the file name was created using \..\..\..\abc.txt kind of pattern then mergin it with temp cache path
-                    // may lead the path out of the temp cache folde rand blobfuse may end up deleting those file if open is hit
+                    // If the file name was created using some special characters then merging it with temp cache path
+                    // may lead the path out of the temp cache folder and blobfuse may end up touching those files if open is hit
                     // To restrict that here we stop getattr and return enoent so that open is never called for such path
                     // any absolute path going out of temp cache path needs to be invalidated
                     std::string realPathString(real_path);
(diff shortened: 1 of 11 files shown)


gapra-msft (approved):
Some of these PRs have remnants of other PRs. But looks good to me.

blobfuse/utilities.cpp
```diff
@@ -183,6 +184,20 @@ int azs_getattr(const char *path, struct stat *stbuf)
             }
             else
             {
+                char real_path[PATH_MAX];
+                char *result = realpath(mntPathString.c_str(), real_path);
+                if (result != NULL) {
+                    // Path got resolved. lets check we are not moving away from temp path
+                    // If the file name was created using \..\..\..\abc.txt kind of pattern then mergin it with temp cache path
```
tasherif-msft, line 191:
```suggestion
                    // If the file name was created using \..\..\..\abc.txt kind of pattern then merging it with temp cache path
```

blobfuse/utilities.cpp
```diff
@@ -183,6 +184,20 @@ int azs_getattr(const char *path, struct stat *stbuf)
...
+                    // may lead the path out of the temp cache folde rand blobfuse may end up deleting those file if open is hit
```
tasherif-msft, line 192:
```suggestion
                    // may lead the path out of the temp cache folder and blobfuse may end up deleting those files if open is hit
```

blobfuse/utilities.cpp
```diff
@@ -183,6 +184,20 @@ int azs_getattr(const char *path, struct stat *stbuf)
...
+                    // To restrict that here we stop getattr and return enoent so that open is never called for such path
+                    // any absolute path going out of temp cache path needs to be invalidated
+                    std::string realPathString(real_path);
+                    if (std::string::npos == realPathString.find(config_options.absoluateTmpPath)) {
```
tasherif-msft, line 196: Let's have a proper PR description/context for this fix please!

blobfuse/utilities.cpp
vibhansa-msft: Corrected

blobfuse/utilities.cpp
vibhansa-msft: corrected

PR #697, opened 2022-02-11
Showing warnings for old versions

souravgupta-msft:
Show security warnings to the user if they are using older version of blobfuse2.


vibhansa-msft (approved):
Looks good

gapra-msft:
@souravgupta-msft can you merge main back into your branch? I messed up a little.

cmd/root.go
```diff
@@ -112,13 +109,40 @@ func beginDetectNewVersion() chan interface{} {
...
+
+			_, isPresent := vJson.SecurityWarnings[common.Blobfuse2Version]
+			if isPresent {
+				hasWarning := false
+				ctr := 1
+				for _, msg := range vJson.SecurityWarnings[common.Blobfuse2Version] {
+					msg = strings.TrimSpace(msg)
+					if len(msg) > 0 {
+						if !hasWarning {
+							fmt.Fprintf(stderr, "Following vulnerabilities were detected in your current version (%s):\n", common.Blobfuse2Version)
```
gapra-msft, line 123: The following

gapra-msft: Also should we log these?

gapra-msft (approved):

tasherif-msft (approved):

cmd/root.go
vibhansa-msft: I am not sure if logging makes sense here, its more of user needs to notice this and take action. Dumping to logs will go unnoticed. Also version check is happening before the logging lib is initialized.

souravgupta-msft: Added the warnings to log

PR #704, opened 2022-02-15
Perf and Memory Optimizations

vibhansa-msft:
Optimizations :
- Instead of returning handle ID to libfuse return the handle object address itself so that on read/write/flush/close calls we can save from map lookup to convert handl id to handle object
- Adding a new config which allows libfuse to directly read from locally cached file instead of forwarding request up in the pipeline
- various structures were not aligned properly leaving lot of padding in between, rearranged fields to minimize the memory usage
- lot of flags are used in handle/attr-cache/file-cache maps. Unified them all to use a bitmap instead of individual flags to conserve memory.
- sleep on file cache tests where we expect the file to take time to be deleted


gapra-msft (approved):

gapra-msft (approved):

souravgupta-msft (approved):

component/file_cache/file_cache_test.go
```diff
@@ -995,8 +995,14 @@ func (suite *fileCacheTestSuite) TestRenameFileNotInCache() {
 	handle, _ := suite.fileCache.CreateFile(internal.CreateFileOptions{Name: src, Mode: 0777})
 	suite.fileCache.CloseFile(internal.CloseFileOptions{Handle: handle})
 
+	_, err := os.Stat(suite.cache_path + "/" + src)
+	for !os.IsNotExist(err) {
```
tasherif-msft, line 999: my only worry for this approach is if we run into a file cache bug where the file doesn't get deleted for one reason or another down the road. So maybe we should break after a given amount of time has passed and if we broke then the test fails

component/libfuse/libfuse2_handler.go
```diff
@@ -440,8 +440,8 @@ func libfuse_create(path *C.char, mode C.mode_t, fi *C.fuse_file_info_t) C.int {
 		}
 	}
 
-	id := handlemap.Add(handle)
-	fi.fh = C.ulong(id)
+	handlemap.Add(handle)
+	fi.fh = C.ulong(uintptr(unsafe.Pointer(handle)))
```
tasherif-msft, line 444: was there a reason we didn't go with this approach from the start? why did we go with the handlemap approach first for the fuse layer?

tasherif-msft (approved):

component/libfuse/libfuse_handler.go
```diff
@@ -185,6 +185,10 @@ func libfuse_init(conn *C.fuse_conn_info_t, cfg *C.fuse_config_t) (res unsafe.Po
 	conn.want |= C.FUSE_CAP_PARALLEL_DIROPS
 	conn.want |= C.FUSE_CAP_AUTO_INVAL_DATA
 	conn.want |= C.FUSE_CAP_READDIRPLUS
+	conn.want |= C.FUSE_CAP_ASYNC_READ
+
+	conn.max_background = 128
```
tasherif-msft, line 190: could we add comments elaborating here please

tasherif-msft:
We should add a test to confirm that libfuse doesn't hit file cache when the flag is set

component/libfuse/libfuse_handler_test_wrapper.go
```diff
@@ -417,8 +411,7 @@ func testFsync(suite *libfuseTestSuite) {
 	openOptions := internal.OpenFileOptions{Name: name, Flags: flags, Mode: mode}
 	suite.mock.EXPECT().OpenFile(openOptions).Return(handle, nil)
 	libfuse_open(path, info)
-	handle.Path = name
-	handle.ID = handlemap.HandleID(info.fh)
+	handle = (*handlemap.Handle)(unsafe.Pointer(uintptr(info.fh)))
```
tasherif-msft, line 414: let's add a comment explaining what we're doing here, i.e, that we're converting the pointer to a handle and that it looks correct on the other component

component/file_cache/file_cache_test.go
vibhansa-msft: Nice suggestion. Will see if we can break it after X retries.

vibhansa-msft: corrected on all instances of notExists check

component/libfuse/libfuse_handler.go
vibhansa-msft: added comments for all the flags

component/libfuse/libfuse_handler_test_wrapper.go
vibhansa-msft: done

PR #706, opened 2022-02-15
Block level disk cache for read streaming

vibhansa-msft:
- Create disk based block level caching for streaming


component/stream/stream.go
```diff
@@ -67,11 +68,16 @@ type StreamOptions struct {
 	Policy            string `config:"policy" yaml:"policy,omitempty"`
 	CacheOnFileHandle bool   `config:"cache-on-file-handle" yaml:"cache-on-file-handle,omitempty"`
 	readOnly          bool   `config:"read-only"`
+	Persistance       bool   `config:"persistance"`
```
gapra-msft, line 71:
```suggestion
	Persistence       bool   `config:"persistence"`
```

component/stream/stream.go
```diff
@@ -159,13 +169,42 @@ func (st *Stream) Configure() error {
...
+			}
+
+			if conf.DiskPath == "" {
+				log.Err("Stream::Configure : Config error [disk-cache-path not set]")
+				return fmt.Errorf("config error in %s error [disk-cache-path not set]", st.Name())
+			}
+
+			_, err = os.Stat(conf.DiskPath)
+			if os.IsNotExist(err) {
+				log.Err("Stream::Configure : Config error [disk-cache-path does not exist. attempting to create]")
```
gapra-msft, line 187: this should be an informational log right?

component/stream/stream_cache.go
```diff
@@ -173,7 +172,13 @@ func (c *cache) getBlock(fileKey string, offset int64, fileSize int64) (*cacheBl
 			}
 			c.evictedBlock = blockKey{}
 		}
-		return newBlock, false
+
+		dataFetched := false
```
gapra-msft, line 176: Where is the code that fetches from the memory cache? Why do we initialize this to false so late?

component/stream/stream_cache.go
```diff
@@ -200,3 +210,104 @@ func (c *cache) filePurge(key, value interface{}) {
...
+func (c *cache) wipeoutDiskCache() {
+	log.Trace("streamcache::evictDiskBlock : Wipe out disk cache in progress")
+
+	dirents, err := os.ReadDir(c.diskPath)
+	if err != nil {
+		return
+	}
+
+	for _, entry := range dirents {
+		os.RemoveAll(filepath.Join(c.diskPath, entry.Name()))
```
gapra-msft, line 311: Why not just do a os.RemoveAll of c.diskPath?

component/stream/stream_cache.go
```diff
@@ -200,3 +210,104 @@ func (c *cache) filePurge(key, value interface{}) {
 	c.blocks.Remove(key)
 	c.evictedBlock = blockKey{}
 }
+
+// Using key construct a file name for persisted block
+func (c *cache) getLocalFilePath(key blockKey) string {
+	return filepath.Join(c.diskPath, key.fileKey+"__"+fmt.Sprintf("%d", key.offset)+"__")
```
gapra-msft, line 216: Will this run into issues if the fileKey is fileName? If we have multiple handles of a file?

component/stream/stream_cache.go
vibhansa-msft: By default we only allocate block and return. So this flag is init after creating the block. Then we check if this block was earlier persisted on disk or not, if so read data from there and mark this block as already having data. Just below this line we are checking to read data from disk.

component/stream/stream_cache.go
vibhansa-msft: We might not have created this directory so removing it will not make sense, so we are just iterating its children and then deleting all of them,

component/stream/stream_cache.go
vibhansa-msft: We will never have multiple handles to the file as we maintain only one block in cache irrespective of number of handles open for that file. Also the stream blocks are in memory while this persistence is done on disk so there is no collision in actual file and persisted block. If your concern is about actual file name also following similar format then that also shall not hit issue as our persisted block will now look something like A__1__.txt__0__ so while reading it back we will again do the same conversion. Also these persisted block files are not visible to user directly through blobfuse only way is to manually list them on the persist directory. Let me know if you have any concern with the naming conventions or if you foresee any issue with this.

tasherif-msft: As Vikas said this should be fine, if the file key is names then regardless of open handles don't have duplicates. If our key is the handle id its still ok since the blocks will have unique id prefix.

component/stream/stream.go
```diff
@@ -67,11 +68,16 @@ type StreamOptions struct {
 	Policy            string `config:"policy" yaml:"policy,omitempty"`
 	CacheOnFileHandle bool   `config:"cache-on-file-handle" yaml:"cache-on-file-handle,omitempty"`
 	readOnly          bool   `config:"read-only"`
+	Persistence       bool   `config:"persistence"`
```
tasherif-msft, line 71: since all disk related options are prefixed with "disk" we should do this here as well so `disk-persistence`

component/stream/stream_cache.go
```diff
@@ -200,3 +210,104 @@ func (c *cache) filePurge(key, value interface{}) {
...
+
+	if err != nil {
+		return false
+	}
+
+	log.Debug("streamcache::getBlockFromDisk : Reading block for %s offset %d from disk", key.fileKey, key.offset)
+	if time.Since(info.ModTime()).Seconds() > c.diskTimeoutSec {
+		// File exists on local disk but disk cache timeout has elapsed
+		os.Remove(localPath)
+		c.diskBlocks.Remove(key)
```
tasherif-msft, line 255: :( I hate how cache deletion is event driven at the moment - this is something we should def consider improving on in the future

component/stream/stream_cache.go
```diff
@@ -200,3 +210,104 @@ func (c *cache) filePurge(key, value interface{}) {
...
+	f, err := os.OpenFile(localPath, os.O_RDONLY, 0775)
+	if err != nil {
+		return false
+	}
+
+	f.Read(block.data)
+	f.Close()
+	os.Remove(localPath)
+
+	c.diskBlocks.Remove(key)
```
tasherif-msft, line 268: We should leave a comment here explaining:
```suggestion
        // we remove the block from disk as it is now in our in-memory cache
	c.diskBlocks.Remove(key)
```

component/stream/stream_cache.go
```diff
@@ -200,3 +210,104 @@ func (c *cache) filePurge(key, value interface{}) {
...
+
+	return true
+}
+
+// Remove this block from in-memory cache
+func (c *cache) evictBlock(key blockKey, block *cacheBlock) {
```
tasherif-msft, line 274: There is one very important unit test we should have, here consider the following:
1. our number of cached files is full
2. the cache evicts one of the files
3. this causes a purge on all blocks in memory for that file
4. this in turn persists these blocks on disk (while the fileKey and its blocks are now gone from memory)
5. another read to that evicted file comes for one of the blocks on disk
6. we need to make sure that the key is handled correctly and we end up fetching the block from disk rather than getting another copy of it (leading to duplicate data on disk with a stale block from the last file purge)

I know we're using key.offset but just for future proofing this test is important to have to make sure no memory leaks are happening since currently its event driven

component/stream/stream_cache.go
vibhansa-msft: yes other then this event of cache being full it shall be time based cleanup as well. We need a lot of changes in stream anyway. That locking based improvements also needs to be done. Can take all those things in seperate PRs.

component/stream/stream_cache.go
vibhansa-msft: There is a test in UT where we bring the block in memory and then create a eviction case where it goes back to disk. Later when the block is read its read from disk. This is similar to cache full scenario as in that UT we take cache size to be 1 block per file and simulate the UT around it.

component/stream/stream_cache.go
vibhansa-msft: done

component/stream/stream.go
vibhansa-msft: done

souravgupta-msft (approved):

PR #707, opened 2022-02-15
[BUG] Check length of slice rather than capacity

tasherif-msft:
Current we check against the capacity - this is problematic as libfuse create a slice with a capacity of the file size. What we need is the slice length

### DIFF
--- a/component/azstorage/azstorage.go
+++ b/component/azstorage/azstorage.go
@@ -326,15 +326,15 @@ func (az *AzStorage) ReadFile(options internal.ReadFileOptions) (data []byte, er
 	return az.storage.ReadBuffer(options.Handle.Path, 0, 0)
 }
 
-func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (len int, err error) {
+func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (length int, err error) {
 	//log.Trace("AzStorage::ReadInBuffer : Read %s from %d offset", h.Path, offset)
 
 	if options.Offset > options.Handle.Size {
 		return 0, syscall.ERANGE
 	}
 
-	var dataLen int64 = int64(cap(options.Data))
-	if options.Handle.Size < (options.Offset + int64(cap(options.Data))) {
+	var dataLen int64 = int64(len(options.Data))
+	if options.Handle.Size < (options.Offset + int64(len(options.Data))) {
 		dataLen = options.Handle.Size - options.Offset
 	}
 
@@ -347,7 +347,7 @@ func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (len int
 		log.Err("AzStorage::ReadInBuffer : Failed to read %s (%s)", options.Handle.Path, err.Error())
 	}
 
-	len = int(dataLen)
+	length = int(dataLen)
 	return
 }
 


component/azstorage/azstorage.go
```diff
@@ -326,15 +326,15 @@ func (az *AzStorage) ReadFile(options internal.ReadFileOptions) (data []byte, er
...
+func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (length int, err error) {
 	//log.Trace("AzStorage::ReadInBuffer : Read %s from %d offset", h.Path, offset)
 
 	if options.Offset > options.Handle.Size {
 		return 0, syscall.ERANGE
 	}
 
-	var dataLen int64 = int64(cap(options.Data))
-	if options.Handle.Size < (options.Offset + int64(cap(options.Data))) {
+	var dataLen int64 = int64(len(options.Data))
```
vibhansa-msft, line 336: I guess in options we have length field, we shall rather rely on that instead of cap/len of slice.

component/azstorage/azstorage.go
```diff
@@ -347,7 +347,7 @@ func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (len int
 		log.Err("AzStorage::ReadInBuffer : Failed to read %s (%s)", options.Handle.Path, err.Error())
 	}
 
-	len = int(dataLen)
+	length = int(dataLen)
```
vibhansa-msft, line 350: datalen is populated based on the slice len, is there any possibility that the read call came back with less data then requested, in that case sending back datalen will result in data corruption. We need to check if READ api in sdk returns back any sort of length of bytes actually read or the length of the slice needs to be recalculated post the read

component/azstorage/azstorage.go
vibhansa-msft: Ok rechecked we do not have any length fields in options here. Would it be good to add that as a fields in options and then rely on it.

component/azstorage/azstorage.go
vibhansa-msft: Rechecked this as well, the sdk API returns only error and not the length of data. In that case better to take len(options.Data) here as well instead of using the static length ?

component/azstorage/azstorage.go
```diff
@@ -326,15 +326,15 @@ func (az *AzStorage) ReadFile(options internal.ReadFileOptions) (data []byte, er
 	return az.storage.ReadBuffer(options.Handle.Path, 0, 0)
 }
 
-func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (len int, err error) {
+func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (length int, err error) {
```
vibhansa-msft, line 329: So root-cause is libfuse always converting the C buffer to 2^30 cap slice. We shall restrict that to create rather a smaller slice just in case anywhere we try to use that inside sdk also it will result into buffer overrun. 
we can convert that (1 << 30)  to (1 << 22) ~= 4MB size. As of now we know libfuse always reads in blocks of 128K only so ideal size shall be (1 << 16)

gapra-msft (approved):

vibhansa-msft (approved):

souravgupta-msft (approved):

PR #708, opened 2022-02-16
Auto update of version on release

souravgupta-msft:
Adding a stage in the release pipeline which will update the version number in the latestVersion.json file

### DIFF
--- a/blobfuse2-release.yaml
+++ b/blobfuse2-release.yaml
@@ -14,6 +14,11 @@ parameters:
     type: boolean
     default: false
 
+  - name: update_version
+    displayName: 'Update Version'
+    type: boolean
+    default: true
+
 # Do not trigger this pipeline automatically
 trigger: none
 pr: none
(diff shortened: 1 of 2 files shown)


vibhansa-msft (approved):

blobfuse2-release.yaml
```diff
@@ -527,3 +532,43 @@ stages:
...
+                  version = sys.argv[2].replace('blobfuse2-', '')
+                  if(len(version)==0):
+                      print('Incorrect Release Tag')
+                      sys.exit(1)
+                  dataDict = requests.get(url).json()
+                  dataDict['blobfuse2'] = version
+                  resp = requests.put(url, data=json.dumps(dataDict, indent=4), headers={'Content-Type': 'application/json', 'x-ms-blob-type': 'BlockBlob'})
+                  print('Status Code: ' + str(resp.status_code))
+                  sys.exit(1) if(resp.status_code<200 or resp.status_code>202) else print('SUCCESS')" \
+                  '$(LATEST_VERSION_SAS)' '${{ parameters.tag }}'
```
vibhansa-msft, line 573: We are relying on the tag here that it will always be in sync with the binary version.

souravgupta-msft:
@gapra-msft , what convention will we be using for preview releases, **-preview** or **~preview**?
As of now I see that for both tag and version we are using **-preview**. So, going forward will we be going with the same convention?

blobfuse2-release.yaml
souravgupta-msft: Done

gapra-msft:
> @gapra-msft , what convention will we be using for preview releases, **-preview** or **~preview**? As of now I see that for both tag and version we are using **-preview**. So, going forward will we be going with the same convention?

Yeah we are going with -preview

blobfuse2-release.yaml
```diff
@@ -527,3 +532,45 @@ stages:
...
+                      print('Less arguments')
+                      sys.exit(1)
+                  url = sys.argv[1]
+                  version = sys.argv[2].replace('blobfuse2-', '')
+                  if(len(version)==0):
+                      print('Incorrect Release Tag')
+                      sys.exit(1)
+                  dataDict = requests.get(url).json()
+                  dataDict['blobfuse2'] = version
+                  resp = requests.put(url, data=json.dumps(dataDict, indent=4), headers={'Content-Type': 'application/json', 'x-ms-blob-type': 'BlockBlob'})
```
gapra-msft, line 572: what kind of authorization are we doing here? Is this file not only read only for public access?

blobfuse2-release.yaml
gapra-msft: We could also somehow read it from the code

blobfuse2-release.yaml
souravgupta-msft: We are using a SAS url which will have read and write access to the file

blobfuse2-release.yaml
souravgupta-msft: Removed the dependency from the tag that we specify during the build.
Now it will install blobfuse2 and pass the version as a parameter to the python script from the **blobfuse2 version** command

blobfuse2-release.yaml
```diff
@@ -527,3 +532,57 @@ stages:
...
+                  cd $(Build.ArtifactStagingDirectory)/blobfuse2
+                  ls | grep ubuntu-18.04
+                  sudo apt-get install ./`ls | grep ubuntu-18.04` -y
+                  blobfuse2 version
+                displayName: 'Installing blobfuse2'
+
+              # Add '-preview' to pre-release versions. For example, 2.0.0-preview.1
+              # Also considering that the tag and version are same
+              - script: |
+                  python3 -c "import json; import requests; import sys;
```
tasherif-msft, line 572: We initially had python in some of our pipelines and sometimes we had dependency issues on the VMs. This seems safe enough since its a MS hosted VM however - I don't like the idea of having the python script be part of the yaml. Can we instead make this a separate pythons cript file (or preferably written in GO - refer to config-gen.go)

souravgupta-msft: The reason why I kept this inline is because we don't have access to the code in this step. So, if we were to maintain this in a separate script, we will have to clone the repo and then run the script. Let me know what are your thoughts on this.

tasherif-msft (approved):

gapra-msft (approved):

PR #709, opened 2022-02-17
Code signing through release pipeline

vibhansa-msft:


souravgupta-msft (approved):

gapra-msft (commented):
Have you tested running this pipeline yet?

gapra-msft (approved):

vibhansa-msft:
> Have you tested running this pipeline yet?

Yes I did run the pipeline and signing part is working fine. Only step not tested so far is to post these binaries on github. Other then that stage everything is green. I have added these signed binaries to our artifacts of pipeline also and there are steps to dump md5sum of each deb/rpm before and after signing just to make sure that files did change as part of image signing.

PR #711, opened 2022-02-18
[FEATURE] Support Write Stream on AzStorage Component

tasherif-msft:


component/azstorage/block_blob.go
```diff
@@ -654,6 +656,186 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
 	return nil
 }
 
+// GetFileBlockOffsets: store blocks ids and corresponding offsets
+func (bb *BlockBlob) GetFileBlockOffsets(name string) (common.BlockOffsetList, bool, error) {
+	var blockOffset int64 = 0
+	var blockList common.BlockOffsetList
+	blobURL := bb.Container.NewBlockBlobURL(filepath.Join(bb.Config.prefixPath, name))
+	storageBlockList, _ := blobURL.GetBlockList(
```
vibhansa-msft, line 664: err checking and logging missing

internal/base_component.go
```diff
@@ -307,3 +308,10 @@ func (base *BaseComponent) InvalidateObject(name string) {
 		base.next.InvalidateObject(name)
 	}
 }
+
+func (base *BaseComponent) GetFileBlockOffsets(options GetFileBlockOffsetsOptions) (common.BlockOffsetList, bool, error) {
```
gapra-msft, line 312: Can you just add a comment to specify what the bool return type means somewhere?

component/azstorage/block_blob.go
```diff
@@ -654,6 +656,186 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+			bb.downloadOptions.ClientProvidedKeyOptions)
+		if err != nil {
+			log.Err("BlockBlob::stageAndCommitModifiedBlocks : Failed to stage to blob %s at block %v (%s)", name, blockOffset, err.Error())
+			return err
+		}
+		blockOffset = blk.Size + blockOffset
+	}
+	var blockIDList []string
+	// TODO: we can probably clean up this for loop - move it to types method?
+	for _, blk := range blockList {
```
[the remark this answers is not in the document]
tasherif-msft: if we are modifying in between, we're essentially keeping the same blockIDs that already existed we're just overwriting them. Therefore, we need to maintain the exact same order of the block IDs

common/types.go
```diff
@@ -168,3 +169,40 @@ type LogConfig struct {
...
+	// size of mod block list
+	size := int64(0)
+	currentBlockOffset := offset
+	var modBlockList BlockOffsetList
+	// TODO: change this to binary search (logn) for better perf
+	for _, blk := range bol {
+		if currentBlockOffset >= blk.StartIndex && currentBlockOffset <= blk.EndIndex && currentBlockOffset <= offset+length {
+			modBlockList = append(modBlockList, blk)
+			size += blk.Size
+			currentBlockOffset = blk.EndIndex
```
gapra-msft, line 193: can you break once you know youre past the required offset?

common/types.go
```diff
@@ -168,3 +169,40 @@ type LogConfig struct {
...
+		}
+	}
+	return modBlockList, size, offset+length >= bol[len(bol)-1].EndIndex
```
gapra-msft, line 196: Can you add a comment for what the bool return is for? Just for easy reading

gapra-msft: This is if we need to create more blocks right?

common/types.go
```diff
@@ -168,3 +169,40 @@ type LogConfig struct {
...
+}
+
+// NewUUID returns a new uuid using RFC 4122 algorithm with the given length.
+func NewUUID(length int64) []byte {
```
gapra-msft, line 200: Just wondering, does the GoSDK already have a method like this that we can borrow?

tasherif-msft: they have one but it doesn't support the length variable, so I decided to just have a version we maintain

common/types.go
tasherif-msft: done

component/azstorage/block_blob.go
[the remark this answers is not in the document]
vibhansa-msft: So is that supported where you can keep the block-Id same but change the content. As per my understanding we need to reupload the modified block with a new block-id and then update block-list to contain that new block id.

component/azstorage/block_blob.go
```diff
@@ -654,6 +656,197 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+	blobURL := bb.Container.NewBlockBlobURL(filepath.Join(bb.Config.prefixPath, name))
+
+	// does the file consist of blocks?
+	multipleBlocks := true
+	// tracks the case where our offset is great than our current file size (appending only - not modifying pre-existing data)
+	appendOnly := false
+	var blockList common.BlockOffsetList
+	var dataBuffer *[]byte
+
+	// if this is not 0 then we passed a cached block ID list - from stream for example
```
gapra-msft, line 746: instead of saying something like, "if this is not 0", which is descriptive from the code, could you add like when this is 0, this means this

component/azstorage/azstorage.go
```diff
@@ -347,15 +347,20 @@ func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (len int
...
 	}
 
-	len = int(dataLen)
+	length = int(dataLen)
 	return
 }
 
 func (az *AzStorage) WriteFile(options internal.WriteFileOptions) (int, error) {
-	err := az.storage.WriteFromBuffer(options.Handle.Path, nil, options.Data)
+	err := az.storage.Write(options.Handle.Path, options.Offset, int64(len(options.Data)), options.Data, options.FileOffsets)
```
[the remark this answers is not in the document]
tasherif-msft: by libfuse definition we say "libfuse_write writes data to an open file" and then it uses write_file. I think both methods should be one in the same

common/types.go
```diff
@@ -168,3 +169,71 @@ type LogConfig struct {
...
+}
+
+// list that holds blocks containing ids and corresponding offsets
+type BlockOffsetList struct {
+	BlockOffsetList []*Block //blockId to offset mapping
+	SmallFile       bool     // does it consist of blocks
+	Cached          bool     // is it cached?
+}
+
+func (bol BlockOffsetList) binarySearch(offset int64) (bool, int) {
```
gapra-msft, line 187: Can you add some unit tests for this?

common/types.go
```diff
@@ -168,3 +169,71 @@ type LogConfig struct {
...
+			lowerBound = middleIndex + 1
+			// if the start index is larger than the offset we need to decrease our upper bound
+		} else if bol.BlockOffsetList[middleIndex].StartIndex > offset {
+			higherBound = middleIndex - 1
+		}
+	}
+	return false, 0
+}
+
+func (bol BlockOffsetList) FindBlocksToModify(offset, length int64) (*BlockOffsetList, int64, bool) {
```
gapra-msft, line 206: unit test for this too

component/azstorage/block_blob.go
```diff
@@ -653,6 +655,195 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+	}
+
+	// case 1: file consists of no blocks (small file)
+	if fileOffsets.SmallFile {
+		// get all the data
+		oldData, _ := bb.ReadBuffer(name, 0, 0)
+		// update the data with the new data
+		// if we're only overwriting existing data
+		if int64(len(oldData)) >= offset+length {
+			copy(oldData[offset:], data)
```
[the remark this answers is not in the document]
gapra-msft: maybe we can add an item to investigate this. any copying we can get rid of can help us improve perf

component/libfuse/libfuse_handler.go
```diff
@@ -583,9 +583,11 @@ func libfuse_write(path *C.char, buf *C.char, size C.size_t, off C.off_t, fi *C.
 	data := (*[1 << 30]byte)(unsafe.Pointer(buf))
 	bytesWritten, err := fuseFS.NextComponent().WriteFile(
 		internal.WriteFileOptions{
-			Handle: handle,
-			Offset: int64(offset),
-			Data:   data[:size],
+			Handle:       handle,
```
vibhansa-msft, line 586: Similar handling is missing in libfuse2_handler.go

component/azstorage/block_blob.go
```diff
@@ -653,6 +655,195 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+func (bb *BlockBlob) Write(name string, offset, length int64, data []byte, fileOffsets, modFileOffsets *common.BlockOffsetList) error {
+	defer log.TimeTrack(time.Now(), "BlockBlob::Write", name)
+	log.Trace("BlockBlob::Write : name %s offset %v", name, offset)
+	blobURL := bb.Container.NewBlockBlobURL(filepath.Join(bb.Config.prefixPath, name))
+	// tracks the case where our offset is great than our current file size (appending only - not modifying pre-existing data)
+	appendOnly := false
+	var dataBuffer *[]byte
+
+	// when the file offset mapping is cached we don't need to make a get block list call
+	if !fileOffsets.Cached {
```
vibhansa-msft, line 746: fileOffsets and modFileOffsets are of pointer type here and we shall check them against nil for safety.
As these are not populated in fuse2, when I tried running this code it was crashing.

component/azstorage/block_blob.go
```diff
@@ -653,6 +655,195 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+		fileOffsets, err = bb.GetFileBlockOffsets(name)
+		if err != nil {
+			return err
+		}
+	}
+
+	// case 1: file consists of no blocks (small file)
+	if fileOffsets.SmallFile {
+		// get all the data
+		oldData, _ := bb.ReadBuffer(name, 0, 0)
```
vibhansa-msft, line 757: If the current file we are writing to is a newly created file or file size is 0 then we shall not attempt to read the data. This will be just an extra call which we know will not get any data from container.

component/azstorage/block_blob.go
```diff
@@ -653,6 +655,195 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+		if int64(len(oldData)) >= offset+length {
+			copy(oldData[offset:], data)
+			dataBuffer = &oldData
+			// else appending and/or overwriting
+		} else {
+			// new data buffer with the size of old and new data
+			newDataBuffer := make([]byte, offset+length)
+			// copy the old data into it
+			// TODO: better way to do this?
+			copy(newDataBuffer, oldData)
```
vibhansa-msft, line 769: If current data is getting overwritten (when current offset is 0) or there was no data in file (newly created files) this extra buffer allocation and copy from one slice to other shall be ignored.

component/azstorage/block_blob.go
```diff
@@ -653,6 +655,195 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+			oldData = nil
+			// overwrite with the new data we want to add
+			copy(newDataBuffer[offset:], data)
```
vibhansa-msft, line 772: There are two copies being done to newDataBuffer first with oldData and then appending the newData, we shall think of some better solution here.

component/azstorage/block_blob.go
tasherif-msft: This would be difficult to achieve on the azstorage component as it means we would need to keep track of newly created files.

vibhansa-msft: In current implementation I was trying to upload a file of some 30MB, for every write of 128KB it first downloads the file and then append the data and uploads it back. And it keeps repeating the same for every such write operation. It will be very highly unoptimized solution with this approach until stream component kicks in. May be for POC part its fine but we asking anyone to use this for their real use-case may have issues.

component/azstorage/azstorage.go
```diff
@@ -352,10 +352,15 @@ func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (length
 }
 
 func (az *AzStorage) WriteFile(options internal.WriteFileOptions) (int, error) {
-	err := az.storage.WriteFromBuffer(options.Handle.Path, nil, options.Data)
+	err := az.storage.Write(options.Handle.Path, options.Offset, int64(len(options.Data)), options.Data, options.FileOffsets, options.ModBlockList)
```
vibhansa-msft, line 355: Here we are updating the actual content on container with every write and not wait for flush/close. We need to document this that even after doing write if you just ignore close/flush call from application data will be persisted. This does not happen with regular file-cache based flow as we upload only on flush/close.

component/azstorage/block_blob.go
```diff
@@ -653,6 +655,200 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+	log.Trace("BlockBlob::Write : name %s offset %v", name, offset)
+	blobURL := bb.Container.NewBlockBlobURL(filepath.Join(bb.Config.prefixPath, name))
+	// tracks the case where our offset is great than our current file size (appending only - not modifying pre-existing data)
+	appendOnly := false
+	var dataBuffer *[]byte
+
+	// when the file offset mapping is cached we don't need to make a get block list call
+	if fileOffsets != nil && !fileOffsets.Cached {
+		var err error
+		fileOffsets, err = bb.GetFileBlockOffsets(name)
```
vibhansa-msft, line 748: Inside this method we crate NewBlockBlobURL and just in line 740 also we created the same, Can we do something better in terms of creating it only once

component/azstorage/block_blob.go
```diff
@@ -653,6 +655,200 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+			dataBuffer = &oldData
+			// else appending and/or overwriting
+		} else {
+			// if the file is not empty then we need to combine the data
+			if len(oldData) > 0 {
+				// new data buffer with the size of old and new data
+				newDataBuffer := make([]byte, offset+length)
+				// copy the old data into it
+				// TODO: better way to do this?
+				copy(newDataBuffer, oldData)
```
vibhansa-msft, line 771: If new offset is 0 then we can skip this part of copying old data

common/types.go
```diff
@@ -168,3 +169,71 @@ type LogConfig struct {
...
+	// size of mod block list
+	size := int64(0)
+	currentBlockOffset := offset
+	modBlockList := BlockOffsetList{}
+	found, index := bol.binarySearch(offset)
```
vibhansa-msft, line 211: Block Size for a given file will be fixed as in all blocks of a given file will be of same size. If that is true then there is no need to search the blocks. We can directly convert offset to index and see how many block we are going to create/overwrite.

component/azstorage/block_blob.go
```diff
@@ -653,6 +655,200 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+	} else {
+		modifiedBlockList, oldDataSize, exceedsFileBlocks := fileOffsets.FindBlocksToModify(offset, length)
+		// keeps track of how much new data will be appended to the end of the file (applicable only to case 3)
+		newBufferSize := int64(0)
+		// case 3?
+		if exceedsFileBlocks {
+			// get length of blockID in order to generate a consistent size block ID so storage does not throw
+			existingBlockId, _ := base64.StdEncoding.DecodeString(fileOffsets.BlockOffsetList[0].Id)
+			blockIdLength := len(existingBlockId)
+			appendOnly, newBufferSize = bb.createNewBlocks(modifiedBlockList, fileOffsets, offset, length, int64(blockIdLength))
```
vibhansa-msft, line 797: We are modifying existing blocks here and then below code we modify the last block seperatly before doing stage and commit. This looks little duplicate stuff as base objective is to modify some/all data in a given block. We shall think of simplifying this code further.

component/azstorage/block_blob.go
```diff
@@ -653,6 +655,200 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+			return err
+		}
+	}
+	return nil
+}
+
+// TODO: make a similar method facing stream that would enable us to write to cached blocks then stage and commit
+func (bb *BlockBlob) stageAndCommitModifiedBlocks(name string, blobURL azblob.BlockBlobURL, data []byte, modifiedBlockList, blockList *common.BlockOffsetList) error {
+	blockOffset := int64(0)
+	for _, blk := range modifiedBlockList.BlockOffsetList {
```
vibhansa-msft, line 819: Here we pass two lists one containing all the blocks and one containing only modified blocks and we end up iterating both one by one. Can we use just one list containing all the blocks and in that have a flag per block which marks which is modified and which is "as is" then in single iteration we can create the new list.

vibhansa-msft: if the block is modified then we can just stage it in the same loop too

component/azstorage/block_blob.go
```diff
@@ -653,6 +655,200 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+	}
+	_, err := blobURL.CommitBlockList(context.Background(),
+		blockIDList,
+		azblob.BlobHTTPHeaders{ContentType: getContentType(name)},
+		nil,
+		bb.blobAccCond,
+		bb.Config.defaultTier,
+		nil, // datalake doesn't support tags here
+		bb.downloadOptions.ClientProvidedKeyOptions)
+	if err != nil {
```
vibhansa-msft, line 845: If this commit block list fails then entire write operation is failure and we loose the changes right?

common/types.go
```diff
@@ -168,3 +169,71 @@ type LogConfig struct {
 	FileCount   uint64
 	TimeTracker bool
 }
+
+type Block struct {
+	Id         string
```
vibhansa-msft, line 174: As there are many blocks stored in the system lets arrange the fields in bye order, put all 64 bit fields before the name to conserve memory.

vibhansa-msft: same applies for all structures which will be stored in bulk

common/types.go
```diff
@@ -168,3 +169,71 @@ type LogConfig struct {
...
+	StartIndex int64
+	EndIndex   int64
+	Size       int64
+}
+
+// list that holds blocks containing ids and corresponding offsets
+type BlockOffsetList struct {
+	BlockOffsetList []*Block //blockId to offset mapping
```
vibhansa-msft, line 182: Name of the structure and this field are same is this intentional?

common/types.go
```diff
@@ -168,3 +169,71 @@ type LogConfig struct {
...
+	SmallFile       bool     // does it consist of blocks
```
vibhansa-msft, line 183: If block list is empty then its implicit that file is not blocked, do we need this flag seperatly?

common/types.go
```diff
@@ -168,3 +169,71 @@ type LogConfig struct {
...
+	found, index := bol.binarySearch(offset)
+	if !found {
+		return &modBlockList, 0, true
+	}
+	// after the binary search just iterate to find the remaining blocks
+	for _, blk := range bol.BlockOffsetList[index:] {
+		if blk.StartIndex > offset+length {
+			break
+		}
+		if currentBlockOffset >= blk.StartIndex && currentBlockOffset < blk.EndIndex && currentBlockOffset <= offset+length {
```
vibhansa-msft, line 220: Instead of generating a list of blocks to be modified, as blocks are always in ascending order of offset we can just return the first block from where the modification starts and then the write method can just start from that block and keep modifying until length of write operation and doing the same it can create new block if required. This will simplify the logic a bit I feel.

component/azstorage/block_blob.go
```diff
@@ -653,6 +655,200 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+			newBlock := bb.createBlock(blockIdLength, startIndex, bb.Config.blockSize)
+			modBlockList.BlockOffsetList = append(modBlockList.BlockOffsetList, newBlock)
+			blockList.BlockOffsetList = append(blockList.BlockOffsetList, newBlock)
+			startIndex = newBlock.EndIndex
+			// reset the counter since it will help us to determine if there is leftovers at the end
+			counter = 0
+		}
+	}
+	if counter != 0 {
+		newBlock := bb.createBlock(blockIdLength, startIndex, counter)
```
vibhansa-msft, line 729: These 3 lines are duplicate code, within loop also we are doing the same. unify them.

component/azstorage/block_blob.go
vibhansa-msft: CreateNewBlocks is only generating new blocks I see then where is the loop to modify one or more blocks which are getting affected due to write operation. Below code modifies only one block.

internal/base_component.go
```diff
@@ -307,3 +308,10 @@ func (base *BaseComponent) InvalidateObject(name string) {
 		base.next.InvalidateObject(name)
 	}
 }
+
+func (base *BaseComponent) GetFileBlockOffsets(options GetFileBlockOffsetsOptions) (*common.BlockOffsetList, error) {
```
vibhansa-msft, line 312: We need to add this method prototype to component interface. BaseComponent is just implementation of that interface.

component/azstorage/azstorage.go
tasherif-msft: yes the azstorage component and this `write` code is stateless. In the stream layer we can delay flushes.

common/types.go
tasherif-msft: You can't guarantee that. Blocks in a file can be of variable lengths. If they used other SDKs etc to append to the file.

common/types.go
gapra-msft: is there a rule about this?

component/azstorage/block_blob.go
tasherif-msft: yea if the flush fails then the entire operation fails. Once again with our pipeline approach storage is not meant to maintain a state and therefore if something fails (just like any of the other operations on azstorage with two calls) then operation fails.

component/azstorage/block_blob.go
tasherif-msft: I'm not sure I follow here - we can discuss offline

component/azstorage/azstorage.go
vibhansa-msft: Yes I understood the flow but my suggestion is to document this part clearly in the README under some streaming section that 1. it will update file on every write 2. it will not be very performant if new file is getting created or file is written serially

common/types.go
vibhansa-msft: is this possible that blocks of the same file are variable length in nature ? My understanding was that all blocks have to be of same size.

component/azstorage/block_blob.go
vibhansa-msft: Sure we can discuss, base idea is we are iterating these two block lists multiple times in flow so we can think of simplifying the solution.

common/types.go
vibhansa-msft: there is no rule for that but structure and its member having same name does not make any sense logically.

vibhansa-msft: when you refer it it will be more list BlockOffsetList.BlockOffsetList[]

common/types.go
tasherif-msft: I'll look into removing this

common/types.go
tasherif-msft: Yes blocks can be of variable lengths

component/azstorage/azstorage.go
tasherif-msft: Will mention in the README.

component/azstorage/block_blob.go
```diff
@@ -653,6 +656,181 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+		}
+	}
+
+	// case 1: file consists of no blocks (small file)
+	if fileOffsets != nil && fileOffsets.SmallFile {
+		// get all the data
+		oldData, _ := bb.ReadBuffer(name, 0, 0)
+		// update the data with the new data
+		// if we're only overwriting existing data
+		if int64(len(oldData)) >= offset+length {
```
tasherif-msft, line 739: we can potentially reduce this if else if we knew the old data size before hand (file size)

internal/base_component.go
tasherif-msft: Will add to the component interface.

component/azstorage/block_blob.go
tasherif-msft: explained offline

common/types.go
```diff
@@ -163,8 +164,77 @@ func (ep *EvictionPolicy) Parse(s string) error {
 
 type LogConfig struct {
 	Level       LogLevel
```
vibhansa-msft, line 166: As this structure is created only once in one run, it shall be ok even if its unaligned. We need restructuring of fields when we allocate that structure in bulk.

common/types.go
```diff
@@ -163,8 +164,77 @@ func (ep *EvictionPolicy) Parse(s string) error {
...
+			lowerBound = middleIndex + 1
+			// if the start index is larger than the offset we need to decrease our upper bound
+		} else if bol.BlockList[middleIndex].StartIndex > offset {
+			higherBound = middleIndex - 1
+		}
+	}
+	return false, 0
+}
+
+func (bol BlockOffsetList) FindBlocksToModify(offset, length int64) (int, int64, bool, bool) {
```
vibhansa-msft, line 206: Can we add comment on top for all the return values as to what they mean.

common/types.go
```diff
@@ -163,8 +164,77 @@ func (ep *EvictionPolicy) Parse(s string) error {
...
 type LogConfig struct {
 	Level       LogLevel
-	FilePath    string
 	MaxFileSize uint64
 	FileCount   uint64
+	FilePath    string
 	TimeTracker bool
 }
+
+type Block struct {
```
vibhansa-msft, line 173: There are not very common types. Only stream and azstorage components will need it. Shall we move them to a new file and name it accordingly which clearly says its related to streaming.

common/types_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (suite *typesTestSuite) SetupTest() {
+	suite.assert = assert.New(suite.T())
+}
+
+func TestGenerateConfig(t *testing.T) {
+	suite.Run(t, new(typesTestSuite))
+}
+
+func (suite *typesTestSuite) TestBinarySearch() {
```
vibhansa-msft, line 56: We shall move cases here
- we modify just one block
- we modify multiple blocks
- we modify one block and extend
- we modify multiple blocks and extend
- offset is not found in the list
- offsets we are searching are on start/end boundaries

component/azstorage/azstorage.go
vibhansa-msft, line 355: Like we did for our component interface, shall we also pass a structure instead of so many arguments. That will make it easy to modify in the future.

component/azstorage/block_blob.go
```diff
@@ -653,6 +656,179 @@ func (bb *BlockBlob) WriteFromBuffer(name string, metadata map[string]string, da
...
+		}
+	}
+	return nil
+}
+
+// TODO: make a similar method facing stream that would enable us to write to cached blocks then stage and commit
+func (bb *BlockBlob) stageAndCommitModifiedBlocks(name string, data []byte, index int, offsetList *common.BlockOffsetList) error {
+	blobURL := bb.Container.NewBlockBlobURL(filepath.Join(bb.Config.prefixPath, name))
+	blockOffset := int64(0)
+	for _, blk := range offsetList.BlockList[index:] {
```
vibhansa-msft, line 795: 795 and 814 loop can be combined. Just iterate the entire list once. For each item append id to blockidList and if the block is modified then stage it in the same loop.

vibhansa-msft: This will also help us to reuse the exact same method when stream component kicks in as it will mostly send the blockIdList where multiple blocks are marked as modified and they may not be adjacent to each-other (random write case)

component/azstorage/connection.go
```diff
@@ -111,6 +112,8 @@ type AzConnection interface {
 
 	WriteFromFile(name string, metadata map[string]string, fi *os.File) error
 	WriteFromBuffer(name string, metadata map[string]string, data []byte) error
+	Write(name string, offset int64, len int64, data []byte, fileOffsets, modBlockList *common.BlockOffsetList) error
```
vibhansa-msft, line 115: convert all these params to structure

vibhansa-msft (approved):

component/azstorage/datalake_test.go
```diff
@@ -691,6 +692,280 @@ func (s *datalakeTestSuite) TestCreateFile() {
 	s.assert.Empty(props.XMsProperties())
 }
 
+func (s *datalakeTestSuite) TestWriteSmallFile() {
```
gapra-msft, line 695: Can we add some error cases here? Appending to an offset > file size?

gapra-msft (approved):

common/types.go
tasherif-msft: I think we should keep it in common, it would be a little strange creating a file used explicitly by stream and azstorage

common/types_test.go
tasherif-msft: I'm not sure I get what you mean, all these cases are tested across here and block blob tests, could you clarify?

component/azstorage/datalake_test.go
tasherif-msft: technically appending offset larger than file size is allowed by the FS it will just be a bunch of `\x00` and we mimic that. I can add a test for this case.

component/azstorage/block_blob.go

common/types.go
tasherif-msft: done

component/azstorage/azstorage.go
tasherif-msft: I can make that change in a separate pr to change all the azstorage methods to take in options bags, I don't think this change should be made here.

component/azstorage/connection.go
tasherif-msft: I can make that change in a separate pr to change all the azstorage methods to take in options bags, I don't think this change should be made here.

component/azstorage/datalake_test.go
vibhansa-msft: Ok so thats a valid test case I see. Do we have it in our UT just to validate that it works and does create 0 data buffers. One part I suspect here is we will create a memory chunk and copy only at certain offset leaving the other byes as is. In lang like C/C++ these untouched bytes will be garbadge and not 0. Can you validate its 0 in go otherwise we will need to fill them explicitly with 0s.

souravgupta-msft (approved):

component/azstorage/datalake_test.go
tasherif-msft: Done

PR #714, opened 2022-02-22
Fixed bug where some user provided values that can be set to 0 were s…

gapra-msft:
…et to default value

### DIFF
--- a/common/config/config_parser.go
+++ b/common/config/config_parser.go
@@ -216,22 +216,21 @@ func UnmarshalKey(key string, obj interface{}) error {
 	if err != nil {
 		return fmt.Errorf("config error: unmarshalling [%v]", err)
 	}
-	userOptions.envTree.MergeWithKey(key, obj, func(val interface{}) (interface{}, interface{}, bool) {
+	userOptions.envTree.MergeWithKey(key, obj, func(val interface{}) (interface{}, bool) {
 		envVar := val.(string)
 		res, ok := os.LookupEnv(envVar)
 		if ok {
-			return res, "", true
-		} else if res == "" {
-			return "", "", false
+			return res, true
+		} else {
+			return "", false
 		}
-		return "", "", false
 	})
-	userOptions.flagTree.MergeWithKey(key, obj, func(val interface{}) (interface{}, interface{}, bool) {
+	userOptions.flagTree.MergeWithKey(key, obj, func(val interface{}) (interface{}, bool) {
 		flag := val.(*pflag.Flag)
 		if flag.Changed {
-			return flag.Value.String(), flag.DefValue, true
+			return flag.Value.String(), true
 		} else {
-			return "", flag.DefValue, true
+			return "", false
 		}
 	})
 	return nil
@@ -244,22 +243,21 @@ func Unmarshal(obj interface{}) error {
 	if err != nil {
 		return fmt.Errorf("config error: unmarshalling [%v]", err)
 	}
-	userOptions.envTree.Merge(obj, func(val interface{}) (interface{}, interface{}, bool) {
+	userOptions.envTree.Merge(obj, func(val interface{}) (interface{}, bool) {
 		envVar := val.(string)
 		res, ok := os.LookupEnv(envVar)
 		if ok {
-			return res, "", true
-		} else if res == "" {
-			return "", "", false
+			return res, true
+		} else {
+			return "", false
 		}
-		return "", "", false
 	})
-	userOptions.flagTree.Merge(obj, func(val interface{}) (interface{}, interface{}, bool) {
+	userOptions.flagTree.Merge(obj, func(val interface{}) (interface{}, bool) {
 		flag := val.(*pflag.Flag)
 		if flag.Changed {
-			return flag.Value.String(), flag.DefValue, true
+			return flag.Value.String(), true
 		} else {
-			return "", flag.DefValue, true
+			return "", false
 		}
 	})
 
@@ -276,6 +274,10 @@ func SetBool(key string, val bool) {
 
 func IsSet(key string) bool {
 	return viper.IsSet(key)
+}
+
+func IsCLISet(key string) bool {
+	return userOptions.flags.Lookup(key) != nil && userOptions.flags.Lookup(key).Changed
 }
 
 //AttachToFlagSet is used to attach the flags in config to the cmd flags
--- a/common/config/config_parser.go
+++ b/common/config/config_parser.go
@@ -273,11 +273,18 @@ func SetBool(key string, val bool) {
 }
 
 func IsSet(key string) bool {
-	return viper.IsSet(key)
-}
-
-func IsCLISet(key string) bool {
-	return userOptions.flags.Lookup(key) != nil && userOptions.flags.Lookup(key).Changed
+	if viper.IsSet(key) {
+		return true
+	}
+	pieces := strings.Split(key, ".")
+	node := userOptions.flagTree.head
+	for _, s := range pieces {
+		node = node.children[s]
+		if node == nil {
+			return false
+		}
+	}
+	return node.value.(*pflag.Flag).Changed
 }
 
 //AttachToFlagSet is used to attach the flags in config to the cmd flags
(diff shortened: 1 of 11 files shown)


component/libfuse/libfuse.go
```diff
@@ -158,19 +158,19 @@ func (lf *Libfuse) Validate(opt *LibfuseOptions) error {
 		}
 	}
 
-	if config.IsSet(compName + ".entry-expiration") {
+	if config.IsSet(compName + ".entry-expiration-sec") {
```
vibhansa-msft, line 161: Here we check for IsSet, cant we use the same model in all components with respect to timeouts. Only if its set then use the configured value otherwise use default. File-cache/Attr-Cache/Stream/AzStorage all have some sort of timeout related configuration.

common/config/keys_tree.go
```diff
@@ -240,7 +240,9 @@ func (tree *Tree) MergeWithKey(key string, obj interface{}, getValue func(val in
 					tree.MergeWithKey(subKey, elem.Field(i).Addr().Interface(), getValue)
 				} else if elem.Field(i).Type().Kind() == reflect.Ptr {
 					subKey := key + "." + idx
-					tree.MergeWithKey(subKey, elem.Field(i).Elem().Addr().Interface(), getValue)
+					if !elem.Field(i).IsNil() {
```
vibhansa-msft, line 243: Will this work while we merge config/cli/env ? like in config if it was not there then tree will not have the element. If we merge with condition that it should have been there that might fail.

gapra-msft: Shouldnt affect anything except pointer fields

gapra-msft: Regardless if the Field is Nil, the call to Addr() will fail

component/libfuse/libfuse.go
gapra-msft: I can change this.

component/attr_cache/attr_cache.go
```diff
@@ -123,10 +123,10 @@ func (ac *AttrCache) Configure() error {
 		return fmt.Errorf("config error in %s [%s]", ac.Name(), err.Error())
 	}
 
-	if conf.Timeout != 0 {
-		ac.cacheTimeout = conf.Timeout
-	} else {
+	if conf.Timeout == nil {
```
vibhansa-msft, line 126: that IsSet() model if it works better to use that here instead of converting type to pointer

setup/baseConfig.yaml
```diff
@@ -26,9 +26,9 @@ components:
 # Libfuse configuration
 libfuse:
   default-permission: 0777|0666|0644|0444 <default permissions to be presented for block blobs>
-  attribute-expiration-sec: <time kernel can cache inode attributes (in sec)>
-  entry-expiration-sec: <time kernel can cache directory listing attributes (in sec)>
-  negative-entry-expiration-sec: <time kernel can cache attributes of non existant paths (in sec)>
+  attribute-expiration-sec: <time kernel can cache inode attributes (in sec). Default - 120>
```
vibhansa-msft, line 29: We have defined cli flags for these parameters. Can you confirm names given in that register function are correct and matching.

component/libfuse/libfuse.go
vibhansa-msft: These fields are also read from CLI and config file. Issue here shall be same as that of attr-cache then.

setup/baseConfig.yaml
gapra-msft: Yes, they match

component/attr_cache/attr_cache.go
```diff
@@ -521,7 +521,7 @@ func NewAttrCacheComponent() internal.Component {
 // On init register this component to pipeline and supply your constructor
 func init() {
 	internal.AddComponent(compName, NewAttrCacheComponent)
-	attrCacheTimeout := config.AddInt32Flag("attr-cache-timeout", int32(defaultAttrCacheTimeout), "attribute cache timeout")
+	attrCacheTimeout := config.AddUint32Flag("attr-cache-timeout", 0, "attribute cache timeout")
```
vibhansa-msft, line 524: If IsSet works do we need to change the default value here ? If user has not given anything then let it be to default value. Putting 0 here will fail the IsSet right ?

gapra-msft: This is the only way the is set works. I've tested this and if there is no timeout in attr cache or cli it will use 120

component/libfuse/libfuse.go
gapra-msft: I dont think this is an actual issue. I think I was testing the wrong thing yesterday

component/attr_cache/attr_cache.go
vibhansa-msft: I think I have some misunderstanding here. So when we specify the cli-flags we also mention a default value. So ideally if user has not provided anything in cli options then code trying to get its value shall get the default value. If it does not populate the value to default then what is the point of defining default value here? If we are saying not giving this option on cli yeilds 0 then in most cases it will be 0 only. If we are adding attr-cache in pipeline it means user wants to cache but then value 0 will just contradict it. As long as this component is there in pipeline value 0 shall be the last value that we expect. If user misses to provide the value that is where we want the default value.

gapra-msft: unfortunately if you look at the code in keys_tree it uses if the value is 0 or not to determine if it was set. So technically the CLI function will not work if we pass in --cli=0 -> this will bump it to default

vibhansa-msft: this is also wrong. User can set file-cache-timeout or fuse timeouts to 0 through cli. So if its not working that way then lets fix that as well

vibhansa-msft: "not provided", "zero value", "non zero value" shall be three different state and all possible combination of those shall be supported.

vibhansa-msft, line 524: Now we can restore this 0 to its original default value right?

component/file_cache/file_cache.go
```diff
@@ -179,7 +180,11 @@ func (c *FileCache) Configure() error {
 	}
 
 	c.createEmptyFile = conf.CreateEmptyFile
-	c.cacheTimeout = conf.Timeout
+	if config.IsSet(compName + ".timeout-sec") {
```
vibhansa-msft, line 183: file-cache-timeout is also a cli enable flag so we shall be checking both as you have done in attr-cache

common/config/config_parser.go
```diff
@@ -278,6 +276,10 @@ func IsSet(key string) bool {
 	return viper.IsSet(key)
 }
 
+func IsCLISet(key string) bool {
```
vibhansa-msft, line 279: So now we are putting this extra condition check in all components. Can we instead call this implicitly inside IsSet() itself so that component need to call just one method. I understand that cli and config param name differs but those clis are also mapped to the component config so there might be a way to do this.

tasherif-msft:
can we add some tests for this change?

component/file_cache/file_cache.go
gapra-msft: actually looks like its not enabled in CLI, I can add it

component/attr_cache/attr_cache.go
gapra-msft: done

common/config/config_parser.go
```diff
@@ -275,7 +273,18 @@ func SetBool(key string, val bool) {
 }
 
 func IsSet(key string) bool {
-	return viper.IsSet(key)
+	if viper.IsSet(key) {
```
tasherif-msft, line 276: what are we doing here?

tasherif-msft (approved):

common/config/config_parser.go
gapra-msft: first checking if its in config otherwise travel through the flag tree and see if its provided there

common/config/config_parser.go
vibhansa-msft: only issue here is the cli flag and config flag name differs so we need to see how a single call can resolve this

souravgupta-msft (approved):

common/config/config_parser.go
gapra-msft: I've addressed this

vibhansa-msft (approved):

PR #717, opened 2022-03-01
[BUG] Upload after creating does not work if > 256MB

tasherif-msft:
Currently when a file is created it uses the dfs endpoint - this means we cannot use the blob endpoint for staging and committing. Now we create using block endpoint and just do a second call for ACLs.


vibhansa-msft (approved):
looks good

PR #718, opened 2022-03-02
List blob retrying for 20 times even on 403 errors

vibhansa-msft:
In list-blob, get-attr and is-directory-empty checks blobfuse considers 404 as error while 403 status codes are retried. In most listing cases max retry of 20 is hard coded so if user does not have permission on a folder/file REST call will fail with 403 and blobfuse will always go in a loop of 20 retries affecting the performance of overall operation. As a fix to this it will now condifer 403 as non-retryable error and stop after first attempt.

### DIFF
--- a/blobfuse-artifacts.yml
+++ b/blobfuse-artifacts.yml
@@ -622,7 +622,10 @@ stages:
   jobs:
   - job: ReleaseBlobfuse
     pool:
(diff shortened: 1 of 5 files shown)


gapra-msft (approved):
lgtm. Do we need to add this to v2 as well? How do we validate that this works? Do  you just do a manual test?

PR #720, opened 2022-03-03
Redownload of read-only file fix

vibhansa-msft:
- If a file is readonly and file-cache decides to reload that file then it fails because read-only file can not be written
- fix for this is to change the mode and download and post download reset the mode

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -64,7 +64,7 @@ type FileCache struct {
 
 	createEmptyFile bool
 	allowNonEmpty   bool
-	cacheTimeout    uint32
+	cacheTimeout    float64
 	cleanupOnStart  bool
 	policyTrace     bool
 	missedChmodList sync.Map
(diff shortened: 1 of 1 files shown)


component/file_cache/file_cache.go
```diff
@@ -63,7 +63,7 @@ type FileCache struct {
 
 	createEmptyFile bool
 	allowNonEmpty   bool
-	cacheTimeout    uint32
+	cacheTimeout    float64
```
gapra-msft, line 66: Why are we making this a float?

component/file_cache/file_cache.go
```diff
@@ -697,8 +699,33 @@ func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Hand
 		// If the lock is held then we cannot update the file
 		f, err = os.OpenFile(localPath, os.O_WRONLY, options.Mode)
 		if err != nil {
-			log.Err("FileCache::OpenFile : Failed to open file %s in WR mode", localPath)
-			return nil, err
+			if os.IsPermission(err) {
```
gapra-msft, line 702: You mentioned we should figure out a more efficient way to deal with this problem right? Could you create an item for it?

tasherif-msft (approved):

component/file_cache/file_cache.go
vibhansa-msft: the time comparison part needs float so if we do not modify the base type here, every comparison will need type conversion.

component/file_cache/file_cache.go
vibhansa-msft: https://msazure.visualstudio.com/One/_backlogs/backlog/XStore-DevExp/Epics/?showParents=true&workitem=13735363

souravgupta-msft (approved):

PR #721, opened 2022-03-03
Gen1 mount using adlsgen1fuse

souravgupta-msft:
- Created a new command called mountgen1 which is hidden.
- This command is run from the mountv1 command by setting the enable-gen1 flag
- The config parameters are passed in the same way as key value pairs as in mountv1 command
- On the basis of the config parameters, a json file (adlsgen1fuse.json) is created which is passed as the parameter to the adlsgen1fuse binary


blobfuse2-gen1-artifcats.yaml
```diff
@@ -0,0 +1,189 @@
+# Pipeline to build artifact for gen1 compliant adlsgen1fuse
```
gapra-msft, line 1: typo in file name: artifacts

cmd/mountgen1.go
```diff
@@ -0,0 +1,243 @@
...
+
+		rustFuseMap["credentialtype"] = "servicePrincipal"
+	} else {
+		log.Err("mountgen1::MountRustFuse : For Gen1 account only SPN auth is supported")
+		return fmt.Errorf("for Gen1 account only SPN auth is supported")
+	}
+
+	rustFuseMap["resourceurl"] = "https://datalake.azure.net/"
+
+	if LibFuseOpt.AttributeExpiration != 0 {
```
gapra-msft, line 154: Why do we do this 0 check?

cmd/mountv1.go
```diff
@@ -118,6 +118,8 @@ var useAttrCache bool
 var useStream bool
 var useFileCache bool = true
 var convertConfigOnly bool
+var enableGen1 bool
```
gapra-msft, line 121: one consideration, maybe in azstorage if a user mistakenly passes gen1 or something and uses the mount command, can we print out an error message that says use the mount gen1 command or something?

blobfuse2-gen1-artifcats.yaml
```diff
@@ -0,0 +1,189 @@
...
+            parameters:
+              work_dir: $(work_dir)
+              root_dir: $(root_dir)
+              tags: $(tags)
+
+          # place the generated binary files & any additional files in appropriate locations
+          - script: |
+              mkdir -p pkgDir/usr/bin/
+              mkdir -p pkgDir/usr/share/blobfuse2/
+              cp azure-storage-fuse/blobfuse2 pkgDir/usr/bin/blobfuse2
```
vibhansa-msft, line 158: Can we create a sample json file also and add it out repo as well as this package

cmd/mountgen1.go
```diff
@@ -0,0 +1,243 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+	"strings"
+
+	"github.com/spf13/cobra"
+)
+
+var AzStorageOpt azstorage.AzStorageOptions
```
vibhansa-msft, line 53: As we are parsing the config into these, can we have them as local variables to the method instead of global

cmd/mountgen1.go
```diff
@@ -0,0 +1,243 @@
...
+		}
+
+		err = config.UnmarshalKey("azstorage", &AzStorageOpt)
+		if err != nil {
+			log.Err("mountgen1: AzStorage config error [invalid config attributes]")
+			return err
+		}
+
+		// not checking ClientSecret since adlsgen1fuse will be reading secret from env variable (ADL_CLIENT_SECRET)
+		if AzStorageOpt.AuthMode == "" && AzStorageOpt.ClientID != "" && AzStorageOpt.TenantID != "" {
```
vibhansa-msft, line 92: if client id, tenant-id are not given then we shall error out?

cmd/mountgen1.go
```diff
@@ -0,0 +1,243 @@
...
+				return fmt.Errorf("unable to run adlsgen1fuse binary: %s", err.Error())
+			}
+		}
+
+		return nil
+	},
+}
+
+// code to generate json file for rustfuse
+func mountRustFuse() error {
```
vibhansa-msft, line 132: This function is just generating the json config file for gen1fuse so name of the function is not correct

cmd/mountgen1_test.go
```diff
@@ -0,0 +1,126 @@
...
+	suite.Run(t, new(genOneConfigTestSuite))
+}
+
+func (suite *genOneConfigTestSuite) TestHelp() {
+	defer suite.cleanupTest()
+	_, err := executeCommandC(rootCmd, "mountgen1", "-h")
+	suite.assert.Nil(err)
+}
+
+func (suite *genOneConfigTestSuite) TestConfigCreation() {
```
vibhansa-msft, line 95: We can add some negative tests also here which fails if config is not correct

cmd/mountv1.go
vibhansa-msft: In mount also we want to support gen1 mount not just using mountv1

vibhansa-msft: mountgen1 is subcommand that can be used directly with v2 binary to do gen1 mount

blobfuse2-gen1-artifcats.yaml
souravgupta-msft: Fixed

cmd/mountgen1.go
souravgupta-msft: This is done so that we don't add a field for attributeExpiration having 0 value, in the adlsgen1fuse.json file. This json file is passed as parameter to the gen1 binary.
The same logic is used in gen1 mount for v1

cmd/mountgen1.go
souravgupta-msft: Done

cmd/mountgen1.go
souravgupta-msft: Added the condition where if either among client id, tenant id or account name is empty, then we will return an error

cmd/mountgen1_test.go
souravgupta-msft: Added. Please check it out.

cmd/mountv1.go
gapra-msft: So is mount doing the transform for the user from v2 to gen1 style config? and mountgen1 just takes in directly the gen1 configurations?

vibhansa-msft: In V1 when you mount we generate json file and internally execute the gen1fuse binary. So we are following the same model just that in v2 instead of a cli flag we have a new subcommand which converts config to json and starts gen1fuse binary. mounv1 command is also suppose to support the flag that v1 had, so in mountv1 we convert v1 config to v2 and if enable-gen1 flag is there then we execute the newly added subcommand of v2 which internally will fire gen1fuse.

vibhansa-msft (approved):

blobfuse2-gen1-artifcats.yaml
souravgupta-msft: Added. Please check

cmd/mountgen1.go
```diff
@@ -0,0 +1,256 @@
...
+var gen1ConfigFilePath string
+
+func resetGenOneOptions() {
+	azStorageOpt = azstorage.AzStorageOptions{}
+	libFuseOpt = libfuse.LibfuseOptions{}
+	fileCacheOpt = file_cache.FileCacheOptions{}
+}
+
+var gen1Cmd = &cobra.Command{
+	Use:        "mountgen1",
```
tasherif-msft, line 68: I wonder if its better to have adls in that mount command so like `moundadlsgen1`... but that might be too verbose for one word

blobfuse2-gen1-artifacts.yaml
```diff
@@ -0,0 +1,189 @@
...
+
+          # List commits consumed for this build
+          - script: |
+              echo "Below commits were done in last 12 hours : " `date` 
+              git --no-pager log --since="12 hours ago" --stat
+            displayName: "List Commits"
+            workingDirectory: $(WORK_DIR)
+            
+          # Build adlsgen1fuse binary
+          - script: |
```
tasherif-msft, line 65: why are we building the binary ourselves? can't we just take the published binary?

blobfuse2-gen1-artifacts.yaml
```diff
@@ -0,0 +1,189 @@
+# Pipeline to build artifact for gen1 compliant adlsgen1fuse
+
+# Do not trigger this pipeline automatically
+trigger: none
+pr: none
+
+stages:
+
+  # ---------------------- Build adlsgen1fuse here and copy binary to artifact ------------------------------
```
tasherif-msft, line 9: so we're always packaging adlsgen1fuse along with blobfuse? could you elaborate more on this pipeline

blobfuse2-gen1-artifacts.yaml
vibhansa-msft: gen1fuse standard binary comes packaged inside the ML workspace and that does not run independently. To solve this gen1 mount issue Nara created another workspace where we took all the required code from ML workspace and created a rust based binary out of it. So there is no standard release of datasetfuse as such. We have our own repo which we build and create the gen1fuse binary and later package it with the blobfuse binary.

blobfuse2-gen1-artifacts.yaml
vibhansa-msft: We need to create a package that has both the binaries as blobfuse will just launch the other binary. This is not the standard package that we publish on linux repo or on github this is rather privately shared with the team who needs it from our pipeline artifacts.

gapra-msft (approved):

cmd/mountgen1.go
vibhansa-msft: yes that will be too longer of an option for cli too, mountgen1 shall hold good.

tasherif-msft (approved):

PR #722, opened 2022-03-04
Remount on same path is not failing

vibhansa-msft:
- With libfuse the signature in /etc/mtab has changed and code tuned for previous fuse driver is no more working now. So correcting the signature match for libfuse
- Some errors were not in the common format which are used across different components.

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -63,7 +63,10 @@ func IsDirectoryMounted(path string) bool {
 		if strings.TrimSpace(line) != "" {
 			mntPoint := strings.Split(line, " ")[1]
 			if path == mntPoint {
-				if strings.Index(line, " fuse.") > -1 {
+				// with earlier fuse driver ' fuse.' was searched in /etc/mtab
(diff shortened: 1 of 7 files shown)


souravgupta-msft (approved):

gapra-msft (approved):

common/util.go
```diff
@@ -63,7 +63,10 @@ func IsDirectoryMounted(path string) bool {
 		if strings.TrimSpace(line) != "" {
 			mntPoint := strings.Split(line, " ")[1]
 			if path == mntPoint {
-				if strings.Index(line, " fuse.") > -1 {
+				// with earlier fuse driver ' fuse.' was searched in /etc/mtab
+				// however with libfuse entry does not have that signature
+				// if this path is already mounted using fuse then fail
+				if strings.Contains(line, "fuse") {
```
tasherif-msft, line 69: that `.` makes all the difference :P

tasherif-msft (approved):

PR #726, opened 2022-03-09
Adding version and security warnings check

souravgupta-msft:
- Have files with version name in the release container
- Checking the status code for existence of the warnings
- Updated the release pipeline to update the latest version file when we do a release

### DIFF
--- a/blobfuse2-release.yaml
+++ b/blobfuse2-release.yaml
@@ -662,5 +662,5 @@ stages:
                   createUrl = url + '/latest/' + releaseVersion + '.version?' + sasToken
                   resp = requests.put(createUrl, headers={'x-ms-blob-type': 'BlockBlob'})
                   sys.exit(1) if(resp.status_code<200 or resp.status_code>202) else print('Created release version file')" \
-                  '$(LATEST_VERSION_SAS_TOKEN)' "`blobfuse2 version`"
+                  '$(LATEST_DIR_SAS_TOKEN)' "`blobfuse2 version`"
                 displayName: 'Updating version number'
--- a/blobfuse2-release.yaml
+++ b/blobfuse2-release.yaml
@@ -645,7 +645,7 @@ stages:
                   # list latest version file in the container
                   containerUrl = sasUrl.split('?')[0]
                   sasToken = sasUrl.split('?')[1]
-                  listUrl = containerUrl + '?restype=container&comp=list&prefix=latest/'
+                  listUrl = sasUrl + '&restype=container&comp=list&prefix=latest/'
                   resp = requests.get(listUrl)
                   sys.exit(1) if(resp.status_code<200 or resp.status_code>202) else print('Listed latest version file')
                   listData = minidom.parseString(resp.content)
@@ -660,7 +660,7 @@ stages:
                   resp = requests.delete(deleteUrl)
                   sys.exit(1) if(resp.status_code<200 or resp.status_code>202) else print('Deleted latest version file')
                   # create release version file in the container
-                  createUrl = containerUrl + '/latest/' + releaseVersion + '.version?' + sasToken
+                  createUrl = containerUrl + '/latest/' + releaseVersion + '?' + sasToken
                   resp = requests.put(createUrl, headers={'x-ms-blob-type': 'BlockBlob'})
                   sys.exit(1) if(resp.status_code<200 or resp.status_code>202) else print('Created release version file')" \
                   '$(VERSION_CNT_SAS_URL)' "`blobfuse2 version`"
--- a/blobfuse2-release.yaml
+++ b/blobfuse2-release.yaml
@@ -607,6 +607,8 @@ stages:
           vmImage: 'ubuntu-18.04'
         variables:
           - group: NightlyBlobFuse
+          - name: root_dir
+            value: '$(System.DefaultWorkingDirectory)'
         
         steps:
           - checkout: none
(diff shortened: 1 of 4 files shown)


cmd/root.go
```diff
@@ -66,70 +59,40 @@ var rootCmd = &cobra.Command{
...
 
+// check if the version file exists in the container
+func checkVersionExists(versionUrl string) bool {
+	resp, err := http.Get(versionUrl)
+	if err != nil {
+		log.Err("checkVersionExists: error getting version file from container [%s]", err)
+		return false
+	}
+
+	return (resp.StatusCode >= 200 && resp.StatusCode <= 202)
```
vibhansa-msft, line 70: Shall we explicitly check for 404 to say file exists or not?

releaseVersionUpdate.py
```diff
@@ -0,0 +1,36 @@
...
+listUrl = sasUrl + '&restype=container&comp=list&prefix=latest/'
+resp = requests.get(listUrl)
+sys.exit(1) if(resp.status_code<200 or resp.status_code>202) else print('Listed latest version file')
+listData = minidom.parseString(resp.content)
+name = listData.getElementsByTagName('Name')
+if(len(name)!=1):
+    print('Latest version file not present')
+    sys.exit(1)
+latestVersion = name[0].firstChild.data
+print('Latest version file: ' + latestVersion)
```
vibhansa-msft, line 26: Last release version

releaseVersionUpdate.py
```diff
@@ -0,0 +1,36 @@
...
+
+# delete latest version file in the container
+deleteUrl = containerUrl + '/' + latestVersion + '?' + sasToken
+resp = requests.delete(deleteUrl)
+sys.exit(1) if(resp.status_code<200 or resp.status_code>202) else print('Deleted latest version file')
```
vibhansa-msft, line 31: Deleted last release file

cmd/root.go
souravgupta-msft: Using 404 status code to check the existence of the file

cmd/root.go
```diff
@@ -47,9 +47,14 @@ import (
 	"github.com/spf13/cobra"
 )
 
-type VNextJson struct {
-	Blobfuse2        string              `json:"blobfuse2"`
-	SecurityWarnings map[string][]string `json:"securityWarnings"`
+type ContainerList struct {
```
vibhansa-msft, line 50: I think this represents file list and not container list.

cmd/root.go
```diff
@@ -66,34 +71,58 @@ var rootCmd = &cobra.Command{
...
-			log.Err("beginDetectNewVersion: error getting version txt from container [%s]", err)
-			return
-		}
 
-		body, err := ioutil.ReadAll(resp.Body)
+		latestVersionUrl := common.Blobfuse2ListContainerURL + "?restype=container&comp=list&prefix=latest/"
+		remoteVersion, err := getRemoteVersion(latestVersionUrl)
 		if err != nil {
-			log.Err("beginDetectNewVersion: error reading body of txt result [%s]", err)
+			log.Err("beginDetectNewVersion: error getting remote version [%s]", err)
```
vibhansa-msft, line 122: error getting latest version

blobfuse2-release.yaml
```diff
@@ -632,23 +634,14 @@ stages:
...
-                  if(len(version)==0):
-                      print('Incorrect Release Tag')
-                      sys.exit(1)
-                  dataDict = requests.get(url).json()
-                  dataDict['blobfuse2'] = version
-                  resp = requests.put(url, data=json.dumps(dataDict, indent=4), headers={'Content-Type': 'application/json', 'x-ms-blob-type': 'BlockBlob'})
-                  print('Status Code: ' + str(resp.status_code))
-                  sys.exit(1) if(resp.status_code<200 or resp.status_code>202) else print('SUCCESS')" \
-                  '$(LATEST_VERSION_SAS)' "`blobfuse2 version`"
+                  wget https://raw.githubusercontent.com/Azure/azure-storage-fuse/`echo $(Build.SourceBranch) | cut -d "/" -f 1,2 --complement`/releaseVersionUpdate.py
```
gapra-msft, line 638: Why do we need to do this wget?

cmd/root.go
```diff
@@ -66,34 +71,58 @@ var rootCmd = &cobra.Command{
...
+	if err != nil {
+		log.Err("getRemoteVersion: error unmarshalling xml response [%s]", err)
+		return "", err
+	}
+
+	if len(versionList.Blobs) != 1 {
+		return "", fmt.Errorf("the latest version container is empty")
+	}
+
+	versionName := strings.Split(versionList.Blobs[0].Name, "/")[1]
```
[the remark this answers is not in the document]
gapra-msft: Should we add a warning? Do we expect there to be 1 here always?

cmd/root.go
```diff
@@ -113,23 +142,13 @@ func beginDetectNewVersion() chan interface{} {
...
-						ctr++
-					}
-				}
+			warningsUrl := common.Blobfuse2ListContainerURL + "/securitywarnings/" + common.Blobfuse2Version
+			hasWarnings := checkVersionExists(warningsUrl)
+
+			if hasWarnings {
+				warningsPage := common.BlobFuse2WarningsURL + "#" + strings.ReplaceAll(common.Blobfuse2Version, ".", "")
+				fmt.Fprintf(stderr, "Vist %s to see the list of vulnerabilities assosiated with your current version (%s)\n", warningsPage, common.Blobfuse2Version)
+				log.Info("Vist %s to see the list of vulnerabilities associated with your current version (%s)\n", warningsPage, common.Blobfuse2Version)
```
gapra-msft, line 151: Should this be log.Warn instead?

blobfuse2-release.yaml
gapra-msft: nvm

releaseVersionUpdate.py
```diff
@@ -0,0 +1,36 @@
...
+print('Release Version: ' + releaseVersion)
+if(len(releaseVersion)==0):
+    print('Incorrect Release Version')
+    sys.exit(1)
+
+containerUrl = sasUrl.split('?')[0]
+sasToken = sasUrl.split('?')[1]
+
+# list latest version file in the container
+listUrl = sasUrl + '&restype=container&comp=list&prefix=latest/'
```
gapra-msft, line 17: why not use the Python SDK for some of this code?

cmd/root.go
```diff
@@ -66,34 +71,58 @@ var rootCmd = &cobra.Command{
...
+	if err != nil {
+		log.Err("checkVersionExists: error getting version file from container [%s]", err)
+		return false
+	}
+
+	return resp.StatusCode != 404
+}
+
+func getRemoteVersion(req string) (string, error) {
+	resp, err := http.Get(req)
```
gapra-msft, line 86: why did we choose to manually write the http request code instead of just using the GoSDK?

cmd/root.go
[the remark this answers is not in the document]
vibhansa-msft: our pipeline will make sure to delete the older file and create a new file so ideally it shall be always one as we can have only one latest version of release. But just in case by any mistake if there exists multiple files we shall add an error log that we have failed to determine the latest version and shall not proceed with the checks as it might lead to invalid conclusions.

releaseVersionUpdate.py
vibhansa-msft: then we need to clone/build that as well on build machine?

cmd/root.go
vibhansa-msft: we can use the sdk as well, but as this is happening before our pipeline starts we might not have all the config here. so creating a pipeline for gosdk doing some auth and all will be overkill. Thats what I feel, if there is an easier way to access public files in container easily using sdk then we shall explore that.

cmd/root.go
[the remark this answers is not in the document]
souravgupta-msft: Ideally there should only be one file for the latest version in the container.
But if there are more than one file, then we can add some error logs and use the logic that Vikas suggested

cmd/root.go
souravgupta-msft: Renamed to VersionFilesList

cmd/root.go
souravgupta-msft: Done

cmd/root.go
souravgupta-msft: Updated

cmd/root.go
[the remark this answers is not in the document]
souravgupta-msft: Added the logic where it will check that if the number of files in the latest version container is not equal to 1, then it will return an error and also log that. Please check.

vibhansa-msft (approved):

tasherif-msft (approved):

cmd/root.go
```diff
@@ -66,34 +71,58 @@ var rootCmd = &cobra.Command{
...
+
+	var versionList ContainerList
+	err = xml.Unmarshal(body, &versionList)
+	if err != nil {
+		log.Err("getRemoteVersion: error unmarshalling xml response [%s]", err)
+		return "", err
+	}
+
+	if len(versionList.Blobs) != 1 {
+		return "", fmt.Errorf("the latest version container is empty")
```
vibhansa-msft: for size != 1 version container is empty is not a correct error message, we shall rather say not able to validate or some sort of generic error message.

souravgupta-msft: Done

PR #727, opened 2022-03-09
chmod fix to preserve additional principals added to to ACL

vibhansa-msft:
…t ACL to preserve the existing principals added to the file
context:
In the portal if "Manage ACL" to any file/folder in container there are two things in it. One is regular Linux style triplet of 'rwx' and additionally you can provide specific access to various principals and identities as well. If we call SetACL it just overwrites all of them because we pass only the 'rwx' triplet and not the other principal related info. While we set permissions it just overwrites the triplets and leave the principals as is.

### DIFF
--- a/component/azstorage/utils.go
+++ b/component/azstorage/utils.go
@@ -475,6 +475,25 @@ func getAccessTierType(name string) azblob.AccessTierType {
 }
 
 // Called by x method
+func getACLPermissions(mode os.FileMode) string {
+	// Format for ACL and Permission string is different
+	// ACL:"user::rwx,user:<id>:rwx,group::rwx,mask::rwx,other::rwx"
+	// Permissions:"rwxrwxrwx+"
+	// If we call the set ACL without giving user then all other principals will be removed.
+	var sb strings.Builder
+	writePermission(&sb, mode&(1<<8) != 0, 'r')
+	writePermission(&sb, mode&(1<<7) != 0, 'w')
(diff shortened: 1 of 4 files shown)


souravgupta-msft (approved):

gapra-msft (approved):

component/azstorage/utils.go
```diff
@@ -474,6 +474,25 @@ func getAccessTierType(name string) azblob.AccessTierType {
 	return azblob.AccessTierNone
 }
 
+// Called by x method
+func getACLPermissions(mode os.FileMode) string {
+	// Format for ACL and Permission string is different
+	// ACL:"user::rwx,user:<id>:rwx,group::rwx,mask::rwx,other::rwx"
+	// Permissions:"rwxrwxrwx+"
+	// If we call the set ACL without giving user then all other principals will be removed.
```
tasherif-msft, line 482: "If we call the set ACL without giving user then all other principals will be removed." could you explain what this means

vibhansa-msft: From portal if you do "Manage ACL" to any file/folder in container there are two things in it. one is regular linux style triplet of 'rwx' and additionally you can provide specific access to various principals and identities as well. If we call SetACL it just overwrites all of them because we pass only the 'rwx' triplet and not the other principal related info. While we set permissions it just overwrites the triplets and leave the principals as is.

tasherif-msft (approved):

PR #730, opened 2022-03-11
Fixed bug where endpoint is populated incorrectly when passed as envi…

gapra-msft:
…ronment variable

### DIFF
--- a/cmd/mountv1.go
+++ b/cmd/mountv1.go
@@ -194,6 +194,15 @@ var generateConfigCmd = &cobra.Command{
 
 		// Set the endpoint if not explicitly provided
 		if bfv2StorageConfigOptions.Endpoint == "" {
+			accountName := bfv2StorageConfigOptions.AccountName
+			if accountName == "" {
+				res, ok := os.LookupEnv(azstorage.EnvAzStorageAccount)
+				if !ok {
+					return fmt.Errorf("invalid account name")
+				} else {
+					accountName = res
+				}
+			}
 			http := "https"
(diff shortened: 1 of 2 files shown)


tasherif-msft (approved):

cmd/mountv1_test.go
```diff
@@ -441,13 +441,17 @@ func (suite *generateConfigTestSuite) TestConfigCLILogLevel() {
 func (suite *generateConfigTestSuite) TestCLIParamLogging() {
 	defer suite.cleanupTest()
 	name := generateFileName()
+	v1ConfigFile, _ := ioutil.TempFile("", name+".tmp.cfg")
+	defer os.Remove(v1ConfigFile.Name())
+	v1ConfigFile.WriteString("accountName myAccountName")
```
vibhansa-msft, line 446: Are we trying to force put the account name in config file? If in v1 it was given through env-var then it shall go as env-var to v2 as well.

cmd/mountv1.go
```diff
@@ -194,6 +194,15 @@ var generateConfigCmd = &cobra.Command{
 
 		// Set the endpoint if not explicitly provided
 		if bfv2StorageConfigOptions.Endpoint == "" {
+			accountName := bfv2StorageConfigOptions.AccountName
+			if accountName == "" {
```
vibhansa-msft, line 198: Container name can also be given on cli param in v1, can we just confirm that part work fine with mountv1

souravgupta-msft (approved):

vibhansa-msft (approved):

PR #731, opened 2022-03-12
[PERF] ResNet50 Weekly Perf Runs

tasherif-msft:


blobfuse2-perf.yaml
```diff
@@ -0,0 +1,107 @@
...
+        displayName: "Create Perf Config File"
+        env:
+          NIGHTLY_STO_ACC_NAME: $(PERF_WEEKLY_STO_BLOB_ACC_NAME)
+          NIGHTLY_STO_ACC_KEY: $(PERF_WEEKLY_STO_BLOB_ACC_KEY)
+          ACCOUNT_TYPE: 'block'
+          ACCOUNT_ENDPOINT: 'https://$(PERF_WEEKLY_STO_BLOB_ACC_NAME).blob.core.windows.net'
+        continueOnError: false
+
+      - script:
+        go run $(WORK_DIR)/blobfuse2 mount $(MOUNT_DIR) --config-file=$(BLOBFUSE2_CFG)
```
vibhansa-msft, line 85: why do we have these two runs?

blobfuse2-perf.yaml
```diff
@@ -0,0 +1,107 @@
...
+        python3 resnet50_classify.py --dataset-path='' --job-name=''
+      displayName: "Blobfuse2 ResNet50 on Main"
+
+      - script:
+        blobfuse2 mount $(MOUNT_DIR) --config-file=$(BLOBFUSE2_CFG)
+        python3 resnet50_classify.py --dataset-path='' --job-name=''
```
vibhansa-msft, line 91: Shall we also capture some sort of output in terms of how much time it took or any perf numbers and just show them in pipeline stage like dumping the log file or something

tasherif-msft: its still a draft - I haven't completed the pipeline work. I am planning to compare binary to main branch and pushing to log file. Once this pr is ready I will unmark it from draft.

resnet50_classify.py
```diff
@@ -0,0 +1,89 @@
...
+    log_file_path = args['log']
+    job_name = args['job']
+    procs = args['procs']
+    p = Pool(processes=procs)
+    tic = time.time()
+
+    sys.stdout.write('collecting images at time: {}\n'.format(time.strftime('%Y-%m-%d %H:%M:%S', time.localtime(tic))))
+    # get list of files and split them in batches of 10k to be classified
+    images = [os.path.join(dp, f) for dp, dn, filenames in os.walk(dataset_path) for f in filenames]
+    image_subsets = list(chunks(images[1:10], 10000))
```
tasherif-msft, line 66: will remove 1:10 once im done testing the pipeline is behaving correctly

resnet50_classify.py
```diff
@@ -0,0 +1,89 @@
+import os
```
tasherif-msft, line 1: create a perf dir

blobfuse2-perf.yaml
```diff
@@ -0,0 +1,140 @@
...
+      demands:
+        - Agent.Name -equals $(AgentName)
+
+    variables:
+      - group: NightlyBlobFuse
+      - name: MOUNT_DIR
+        value: "/home/vsts/workv2/blob_mnt"
+      - name: TEMP_DIR
+        value: "/home/vsts/workv2/blobfuse2tmp"
+      - name: BLOBFUSE2_CFG
```
tasherif-msft, line 34: turn on read through

testdata/config/azure_key_lfu.yaml
```diff
@@ -20,6 +20,7 @@ file_cache:
   max-size-mb: 2048
   allow-non-empty-temp: true
   cleanup-on-start: true
+  direct-read: true
```
vibhansa-msft, line 23: If we enable direct read in all these test files, then all our pipeline tests will not hit the file-cache. May be this will end up overlooking some potential bugs in file-cache. Does this flag make good amount of difference in perf for resnet50 ? if so we can create another file for resNet50 test and leave these others as is.

test/perf_test/generate_perf_report.py
```diff
@@ -0,0 +1,40 @@
+# Python program to read
+# json file
+
```
vibhansa-msft, line 3: Either we create a folder under perf_test for resNet50 or have that somewhere in file-name as later if we add similar test cases for other scenarios it will be useful.

blobfuse2-perf.yaml
```diff
@@ -0,0 +1,140 @@
...
+          sudo apt-get install fuse3
+          blobfuse2 mount $(MOUNT_DIR) --config-file=$(BLOBFUSE2_CFG) --default-working-dir=$(WORK_DIR)
+          cd $(WORK_DIR)
+          python3 $(WORK_DIR)/test/perf_test/resnet50_classify.py --dataset='$(MOUNT_DIR)/data' --job='binary' --log=$(ROOT_DIR)/blobfuse2-perf.log
+        displayName: "Blobfuse2 ResNet50 Binary"
+        continueOnError: false
+
+      - script: |
+          cd $(WORK_DIR)
+          python3 $(WORK_DIR)/test/perf_test/generate_perf_report.py --metrics=images/second --log=$(ROOT_DIR)/blobfuse2-perf.log
```
vibhansa-msft, line 118: Report generated by this step shall clearly tell with test1 result X, with test2 result Y and deviation Z. Also publish the same report in artifact so we dont need to go and check the results from pipeline stages.
Also we can fail this pipeline if the new numbers are falling beyond 2-3% compared to previous numbers (or test1 numbers)

blobfuse2-perf.yaml
```diff
@@ -0,0 +1,140 @@
...
+          root_dir: $(ROOT_DIR)
+          mount_dir: $(MOUNT_DIR)
+          temp_dir: $(TEMP_DIR)
+          gopath: $(GOPATH)
+          container: cont1
+          skip_ut: true
+
+      - script: |
+          cd $(WORK_DIR)
+          $(WORK_DIR)/blobfuse2 gen-test-config --config-file=azure_key.yaml --container-name=cont1 --temp-path=$(TEMP_DIR) --output-file=$(BLOBFUSE2_CFG)
```
vibhansa-msft, line 88: Here we are comparing v2 preview build vs v2 main branch. Shall we run the same test with v1 binary also. Just to compare how much deviation we are having from there as well. Relying on old numbers of v1 may not be a good idea because based on network/storage/azure and all factors these numbers may get skewed some times. At then end we can generate a report showing v1 numbers, preview binary numbers and main branch numbers.

test/perf_test/generate_perf_report.py
tasherif-msft: this script will be called on all perf apps including resnet - you basically pass it a list of metrics and the two jobs to compare

blobfuse2-perf.yaml
tasherif-msft: We can add v1 in a separate pr - for now let's try to get this in

blobfuse2-perf.yaml
tasherif-msft: Done

vibhansa-msft (approved):

blobfuse2-perf.yaml
```diff
@@ -0,0 +1,144 @@
...
+          root_dir: $(ROOT_DIR)
+          mount_dir: $(MOUNT_DIR)
+          temp_dir: $(TEMP_DIR)
+          gopath: $(GOPATH)
+          container: cont1
+          skip_ut: true
+
+      - script: |
+          cd $(ROOT_DIR)
+          wget https://github.com/Azure/azure-storage-fuse/releases/download/blobfuse2-2.0.0-preview.1/blobfuse2-2.0.0-preview.1-ubuntu-20.04-x86-64.deb
```
souravgupta-msft, line 88: Why is this hardcoded?

test/perf_test/resnet50_classify.py
```diff
@@ -0,0 +1,91 @@
...
+import sys
+import time
+import json
+import argparse
+import numpy as np
+from multiprocessing import Pool
+from tensorflow.keras.applications import resnet50
+from tensorflow.keras.preprocessing.image import load_img
+from tensorflow.keras.preprocessing.image import img_to_array
+from tensorflow.keras.applications.imagenet_utils import decode_predictions
```
souravgupta-msft, line 11: Should we add a stage in the perf pipeline to install these libraries?

blobfuse2-perf.yaml
tasherif-msft: I don't believe there is a way to get the latest blobfuse2 as its currently in beta, is there?

gapra-msft (approved):

PR #739, opened 2022-03-16
Preserve metadata when reuploading files

gapra-msft:


component/azstorage/azstorage.go
```diff
@@ -352,7 +352,7 @@ func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (length
 }
 
 func (az *AzStorage) WriteFile(options internal.WriteFileOptions) (int, error) {
-	err := az.storage.Write(options.Handle.Path, options.Offset, int64(len(options.Data)), options.Data, options.FileOffsets, options.ModBlockList)
+	err := az.storage.Write(options.Handle.Path, options.Offset, int64(len(options.Data)), options.Data, options.FileOffsets, options.ModBlockList, options.Metadata)
```
vibhansa-msft, line 355: From options we are passing down everything then why not pass the option structure itself ?

component/azstorage/azstorage.go
```diff
@@ -387,7 +387,7 @@ func (az *AzStorage) TruncateFile(options internal.TruncateFileOptions) error {
 		}
 	}
 
-	err = az.storage.WriteFromBuffer(options.Name, nil, data)
+	err = az.storage.WriteFromBuffer(options.Name, attr.Metadata, data)
```
vibhansa-msft, line 390: If attr-cache is not in the pipeline, it shall come down as nil. We may need to make changes in libfuse layer for this.

vibhansa-msft (approved):

component/azstorage/azstorage.go
gapra-msft: Yeah that is true. This is just how it was before. I can change it

component/azstorage/azstorage.go
gapra-msft: Is there a problem with it being nil? We were passing nil before. The GO SDK can handle it

vibhansa-msft: No what I mean here is shall we explicitly pass nil where this call originates like libfuse or in file-cache or we want to rely on the default value being nil

component/azstorage/block_blob.go
```diff
@@ -753,7 +753,7 @@ func (bb *BlockBlob) Write(name string, offset, length int64, data []byte, fileO
 			}
 		}
 		// WriteFromBuffer should be able to handle the case where now the block is too big and gets split into multiple blocks
-		err := bb.WriteFromBuffer(name, nil, *dataBuffer)
+		err := bb.WriteFromBuffer(name, metadata, *dataBuffer)
```
vibhansa-msft, line 756: In stage block and put block list also we need to take care of this metadata update?

tasherif-msft (approved):

tasherif-msft:
how does the `noSyminks` option come into play with these changes?

component/azstorage/azstorage.go
gapra-msft: Oh I see, we can pass nil thats fine

component/attr_cache/attr_cache.go
```diff
@@ -349,6 +349,16 @@ func (ac *AttrCache) RenameFile(options internal.RenameFileOptions) error {
 
 // WriteFile : Mark the file invalid
 func (ac *AttrCache) WriteFile(options internal.WriteFileOptions) (int, error) {
+
+	// GetAttr on cache hit will serve from cache, on cache miss will serve from next component.
+	attr, err := ac.GetAttr(internal.GetAttrOptions{Name: options.Handle.Path, RetrieveMetadata: true})
+	if err != nil {
```
vibhansa-msft, line 355: If file does not exists in container then ignore the error.

gapra-msft: Can you review this once before I merge?

component/azstorage/azstorage.go
```diff
@@ -352,7 +352,7 @@ func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (length
 }
 
 func (az *AzStorage) WriteFile(options internal.WriteFileOptions) (int, error) {
-	err := az.storage.Write(options.Handle.Path, options.Offset, int64(len(options.Data)), options.Data, options.FileOffsets, options.ModBlockList)
+	err := az.storage.Write(options)
```
gapra-msft, line 355: Can you review this too? @vibhansa-msft

component/azstorage/block_blob.go
```diff
@@ -709,12 +709,15 @@ func (bb *BlockBlob) createNewBlocks(blockList *common.BlockOffsetList, offset,
 }
 
 // Write : write data at given offset to a blob
-func (bb *BlockBlob) Write(name string, offset, length int64, data []byte, fileOffsets, modFileOffsets *common.BlockOffsetList) error {
-	defer log.TimeTrack(time.Now(), "BlockBlob::Write", name)
+func (bb *BlockBlob) Write(options internal.WriteFileOptions) error {
+	name := options.Handle.Path
```
gapra-msft, line 713: modFileOffsets isnt used in this method, not sure why it was passed in in the first place

component/azstorage/azstorage.go
vibhansa-msft: looks good (both the changes getattr and passing params)

vibhansa-msft (approved):

PR #750, opened 2022-03-28
Codespell fix

gapra-msft:


vibhansa-msft (approved):

vibhansa-msft:
With you other spell fixes already in, does this action report any more issues which we need to address ? if so we can fix them as part of this PR itself?

gapra-msft:
> With you other spell fixes already in, does this action report any more issues which we need to address ? if so we can fix them as part of this PR itself?

The codespell is passing in this PR

souravgupta-msft (approved):

Issue #733, opened 2022-03-12
[blobfuse2] cannot install deb file inside a container image


andyzhangx:
### Which version of the blobfuse was used?
blobfuse v2 preview

### Which OS (please include version) are you using?
Ubuntu 18.04

### What problem was encountered?

 - Dockerfile is like following, it works for blobfuse v1(`packages-microsoft-prod.deb`)
```
FROM k8s.gcr.io/build-image/debian-base:bullseye-v1.1.0

ARG ARCH=amd64
COPY ./_output/${ARCH}/blobplugin /blobplugin

RUN apt update && apt-mark unhold libcap2
RUN clean-install ca-certificates uuid-dev util-linux mount udev wget e2fsprogs nfs-common netbase
# install updated packages to fix CVE issues
RUN clean-install libgmp10 bsdutils
RUN mkdir /blobfuse-proxy/
COPY ./_output/blobfuse-proxy.deb /blobfuse-proxy/
ARG ARCH=amd64
RUN if [ "$ARCH" = "amd64" ] ; then \
  clean-install libcurl4-gnutls-dev && \
  wget -O /blobfuse-proxy/packages-microsoft-prod.deb https://packages.microsoft.com/config/ubuntu/18.04/packages-microsoft-prod.deb && \
  wget -O /blobfuse-proxy/blobfuse2.deb https://github.com/Azure/azure-storage-fuse/releases/download/blobfuse2-2.0.0-preview.1/blobfuse2-2.0.0-preview.1-ubuntu-18.04-x86-64.deb && \
  dpkg -i /blobfuse-proxy/packages-microsoft-prod.deb /blobfuse-proxy/blobfuse2.deb && apt update && apt install blobfuse fuse -y && apt remove wget -y; fi
LABEL maintainers="andyzhangx"
LABEL description="Azure Blob Storage CSI driver"

ENTRYPOINT ["/blobplugin"]
```

 - when building image with the above Dockerfile, it failed with following error:
```
#12 4.802 Selecting previously unselected package blobfuse2.
#12 4.809 (Reading database ... 7134 files and directories currently installed.)
#12 4.809 Preparing to unpack /blobfuse-proxy/blobfuse2.deb ...
#12 4.817 /var/lib/dpkg/tmp.ci/preinst: 1: sudo: not found
#12 4.817 /var/lib/dpkg/tmp.ci/preinst: 2: sudo: not found
[...]

PR #751, opened 2022-03-28
Fixed bug where blobfuse2 could not be installed in docker container

gapra-msft:
The postinstall script was generating various problems so we now default to ship the postinstall script in /usr/share/blobfuse2/
Fixes : #733

### DIFF
--- a/blobfuse2-nightly.yaml
+++ b/blobfuse2-nightly.yaml
@@ -888,7 +888,7 @@ jobs:
               sudo rm -rf [0-9]
             displayName: 'Clean Agent Directories'
             condition: always() 
- - job: Set_8
+  - job: Set_8
     timeoutInMinutes: 30
 
     strategy:
--- a/blobfuse2-nightly.yaml
+++ b/blobfuse2-nightly.yaml
@@ -893,7 +893,7 @@ jobs:
 
     strategy:
       matrix:
-        Ubuntu-18:
+        Ubuntu-18-CSI:
           imageName: 'ubuntu-18.04'
           containerName: 'test-cnt-ubn-18'
           adlsSas: $(UBUNTU-18-ADLS-SAS)
--- a/blobfuse2-nightly.yaml
+++ b/blobfuse2-nightly.yaml
@@ -888,4 +888,85 @@ jobs:
               sudo rm -rf [0-9]
             displayName: 'Clean Agent Directories'
             condition: always() 
-  
+ - job: Set_8
+    timeoutInMinutes: 30
+
+    strategy:
+      matrix:
+        Ubuntu-18:
+          imageName: 'ubuntu-18.04'
+          containerName: 'test-cnt-ubn-18'
+          adlsSas: $(UBUNTU-18-ADLS-SAS)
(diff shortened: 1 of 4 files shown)


gapra-msft:
Resolves #733

blobfuse2-nightly.yaml
```diff
@@ -888,4 +888,85 @@ jobs:
...
+          mkdir -p pkgDir/usr/bin/
+          mkdir -p pkgDir/usr/share/blobfuse2/
+          cp azure-storage-fuse/blobfuse2 pkgDir/usr/bin/blobfuse2
+          cp azure-storage-fuse/setup/baseConfig.yaml pkgDir/usr/share/blobfuse2/
+          cp azure-storage-fuse/sampleFileCacheConfig.yaml pkgDir/usr/share/blobfuse2/
+          cp azure-storage-fuse/sampleStreamingConfig.yaml pkgDir/usr/share/blobfuse2/
+          cp azure-storage-fuse/tools/postinstall.sh pkgDir/usr/share/blobfuse2/
+          mkdir -p pkgDir/etc/rsyslog.d
+          mkdir -p pkgDir/etc/logrotate.d
+          cp azure-storage-fuse/setup/11-blobfuse2.conf pkgDir/etc/rsyslog.d
```
vibhansa-msft, line 954: Does this work at install time without sudo for /etc/rsyslog.d

tools/postinstall.sh
```diff
@@ -28,10 +28,3 @@ then
   sudo service rsyslog restart
   echo "Finished syslog configuration!"
 fi
-
```
vibhansa-msft, line 31: in postinstall.sh we have not done any change only we have removed it from install steps. So we are expecting user to run this explicitly right ?

blobfuse2-nightly.yaml
```diff
@@ -888,4 +888,85 @@ jobs:
               sudo rm -rf [0-9]
             displayName: 'Clean Agent Directories'
             condition: always() 
-  
\ No newline at end of file
+  - job: Set_8
```
vibhansa-msft, line 891: This step tries to create a package and then install it on a container. Shall we do e2e tests inside the container just to make sure other functionalities are also working. If so we shall change the container name as other pipeline jobs will also be using the same (for regular ubn-18 jobs)

vibhansa-msft (approved):
approved with minor comments.

souravgupta-msft (approved):

tools/postinstall.sh
gapra-msft: Yes

blobfuse2-nightly.yaml
gapra-msft: You still have to use sudo but this works on the Docker container

blobfuse2-nightly.yaml
gapra-msft: We can create a separate issue to run E2E tests since that might be more work - figuring out how to get the env variables passed through and everything

vibhansa-msft (approved):

PR #752, opened 2022-03-30
Native (C) pread/pwrite to improve performance

vibhansa-msft:
- Add dynamic profiler to monitor profiling info on the go
- Add slice option in libfuse only if kernel supports it
- Add native pread/pwrite support using CGo
- Reduce number of read calls to disk by caching data in buffer to serve read requests.


component/file_cache/file_cache.go
```diff
@@ -69,7 +69,7 @@ type FileCache struct {
 	missedChmodList sync.Map
 	mountPath       string
 	allowOther      bool
-	directRead      bool
+	offloadIO       bool
```
gapra-msft, line 72: Should we just make this default behavior?

component/file_cache/file_cache.go
```diff
@@ -890,8 +894,11 @@ func (fc *FileCache) ReadFile(options internal.ReadFileOptions) ([]byte, error)
 func (fc *FileCache) ReadInBuffer(options internal.ReadInBufferOptions) (int, error) {
 	//defer exectime.StatTimeCurrentBlock("FileCache::ReadInBuffer")()
 	// The file should already be in the cache since CreateFile/OpenFile was called before and a shared lock was acquired.
-	localPath := filepath.Join(fc.tmpPath, options.Handle.Path)
-	fc.policy.CacheValid(localPath)
+	options.Handle.OptCnt++
+	if (options.Handle.OptCnt % 1000) == 0 {
```
gapra-msft, line 898: Should we add a comment here about why we are only doing this every 1000 calls?

component/file_cache/file_cache_test.go
```diff
@@ -1413,9 +1413,9 @@ func (suite *fileCacheTestSuite) TestCachePathSymlink() {
 	suite.assert.EqualValues(data, d)
 }
 
-func (suite *fileCacheTestSuite) TestZZDirectRead() {
+func (suite *fileCacheTestSuite) TestZZOffloadIO() {
 	defer suite.cleanupTest()
-	configuration := fmt.Sprintf("file_cache:\n  path: %s\n  direct-read: true\n  timeout: 0\n\nloopbackfs:\n  path: %s",
+	configuration := fmt.Sprintf("file_cache:\n  path: %s\n  timeout: 0\n\nloopbackfs:\n  path: %s",
```
gapra-msft, line 1418: What is this test testing now?

component/file_cache/file_cache.go
vibhansa-msft: yes its default only, offload here means offload to GO so default is false which means its native IO

component/file_cache/file_cache.go
vibhansa-msft: As such there is no reason for selecting the value 1K. But updating cache invovles locking and all so doing it on every call has its own cost. Also post this PR default will be to do read/write from native code and there for read m not updating cache at all for write we will update every 1K calls. here 1K calls translates to 128K * 1000 ~= 128MB so every 128MB read we will update the cache which shall ideally be well over few seconds.

component/file_cache/file_cache_test.go
vibhansa-msft: direct-read has been replaced with offload-io and default is to use native code (= direct-read) so test case remains the same but this config is not needed anymore.

component/file_cache/file_cache.go
gapra-msft: Why do we need this option at all?

component/file_cache/file_cache.go
gapra-msft: Can we add this somewhere in the source as a comment?

component/file_cache/lfu_policy.go
```diff
@@ -157,8 +157,10 @@ func (l *lfuPolicy) clearItemFromCache(path string) {
 	if err == nil {
 		// File was deleted so try clearing its parent directory
 		dirPath := filepath.Dir(path)
-		if dirPath != l.tmpPath {
-			os.Remove(dirPath)
+		for ; dirPath != l.tmpPath; dirPath = filepath.Dir(dirPath) {
```
gapra-msft, line 160: See https://github.com/Azure/azure-storage-fuse/pull/765

gapra-msft, line 160: See https://github.com/Azure/azure-storage-fuse/pull/765

gapra-msft: This fix won't work either. We will have to work on something more complex than this to prevent synchronization issues.

gapra-msft: This fix won't work either. We will have to work on something more complex than this to prevent synchronization issues.

component/libfuse/libfuse_wrapper.h
```diff
@@ -130,6 +131,11 @@ extern int libfuse_chown(char *path, uid_t uid, gid_t gid, fuse_file_info_t *fi)
 extern int libfuse_utimens(char *path, timespec_t tv[2], fuse_file_info_t *fi);
 #endif
 
+extern int blobfuse_cache_update(char* path);
+static int native_read_file(char *path, char *buf, size_t size, off_t, fuse_file_info_t *fi);
+static int native_write_file(char *path, char *buf, size_t size, off_t, fuse_file_info_t *fi);
+static int native_flush_file(char *path, fuse_file_info_t *fi);
+static int native_close_file(char *path, fuse_file_info_t *fi);
```
gapra-msft, line 138: does this need to be defined?

gapra-msft (commented):
left some comments

component/file_cache/file_cache.go
vibhansa-msft: Just in case we want to disable this logic of doing read/write from libfuse. As not sure what kind of implication it will have on cache policy. So far we have not seen any issue but just in case in future we find some big hole there is a way to disable this feature.

component/file_cache/lfu_policy.go
vibhansa-msft: Ya got your point from your PR. Will remove this deletion of parent directory logic completly.

component/libfuse/libfuse_wrapper.h
vibhansa-msft: C standards, you need to declare the function before you use. As functions bodies are at the end of the file while the registration is before that we need to declare the functions. Other way is pull up all those definitions before the registration logic. Wanted code to be clean so defined them towards the end of the file.

gapra-msft: I mean I see no native_close_file method defined, so do we need to define it here?

cmd/mount.go
```diff
@@ -382,6 +394,51 @@ func sigusrHandler(pipeline *internal.Pipeline, ctx context.Context) daemon.Sign
 	}
 }
 
+func setGOConfig() {
+	// Ensure we always have more than 1 OS thread running goroutines, since there are issues with having just 1.
```
tasherif-msft, line 398: why do we need more than 1? is this so that the profiler runs on a separate thread?

component/file_cache/file_cache.go
tasherif-msft: I think if the default behavior is on - its important to point this out in the README, not only in the baseConfig file. We should create a new file that contains critical things that customers should be aware of and this would be one of them.

component/file_cache/file_cache.go
```diff
@@ -890,8 +894,14 @@ func (fc *FileCache) ReadFile(options internal.ReadFileOptions) ([]byte, error)
 func (fc *FileCache) ReadInBuffer(options internal.ReadInBufferOptions) (int, error) {
```
tasherif-msft, line 894: So, with the new changes, how do we update a cache hit on a read call? or is this just gonna be completely skipped? if so how are we evicting? purely on expiry time?

component/file_cache/lfu_policy.go
```diff
@@ -153,14 +152,7 @@ func (l *lfuPolicy) clearItemFromCache(path string) {
...
 	// There are no open handles for this file so its safe to remove this
-	err := deleteFile(path)
-	if err == nil {
-		// File was deleted so try clearing its parent directory
-		dirPath := filepath.Dir(path)
-		if dirPath != l.tmpPath {
-			os.Remove(dirPath)
-		}
-	}
+	deleteFile(path)
```
tasherif-msft, line 155: this is related to the fio issue - could we mention in the comments here that we would just have left empty directories that will be cleaned up on `stop()`

component/libfuse/libfuse_wrapper.h
```diff
@@ -286,4 +294,187 @@ static int fill_dir_entry(fuse_fill_dir_t filler, void *buf, char *name, stat_t
...
+        handle_obj->dirty = 0;
+    }
+
+    return ret;
+}
+
+
+#ifdef ENABLE_READ_AHEAD
+// read_ahead_handler : Method to serve read call from read-ahead buffer if possible
+static int read_ahead_handler(char *path, char *buf, size_t size, off_t offset, file_handle_t* handle_obj)
```
tasherif-msft, line 407: so we're not shipping this, correct?

component/file_cache/file_cache.go
vibhansa-msft: This is something totally internal, customer does not care what is our internal implementation for read/write or any optimizations. So, I don't any reason to document this in readme. For our own internal knowledge if we wish we can dump it in code comments or in some form of document. Again, this knob I am keeping just as a fail-safe, in case the CGO wrapper reading/writing file create any issue, otherwise code is tuned to take that path only by default.

cmd/mount.go
vibhansa-msft: We don't need more then 1. If you see the earlier implementation, it runs only when you run blobfuse2 in foreground (which is purely for debugging and not for production). This profiler will run for both background and foreground and it runs as a separate thread. Earlier profiler dumps report only on successful completion and bring down of binary (i.e., unmount in our case), while this is more dynamic and on a running system anytime you wish to take a snapshot of profiling info, you can do it through browser or using "go tool pprof" command. So technically both are different things, meant for different purpose. When in our regular use-case only this will be available (as other one runs only in foreground mode).

component/file_cache/lfu_policy.go
vibhansa-msft: As Gauri is fixing this in a separate PR, just to be consistent in my test case have removed this code from here.

component/libfuse/libfuse_wrapper.h
vibhansa-msft: Yes read-ahead code is removed. However, I left this function in code to just preserve the logic used here for read-ahead. This is disabled as of now and the function is not called from anywhere. Just preserved the logic that's all, if suggested, I can remove this piece as well.

gapra-msft (approved):

component/libfuse/native_file_io.h
```diff
@@ -0,0 +1,229 @@
...
+
+// native_pread :  Do pread on file directly without involving any Go code
+static int native_pread(char *path, char *buf, size_t size, off_t offset, file_handle_t* handle_obj)
+{
+    errno = 0;
+    int res = pread(handle_obj->fd, buf, size, offset);
+    if (res == -1)
+        res = -errno;
+    
+    #if 0
```
gapra-msft, line 84: Why is this behind an #if 0? Doesnt that mean the code won't be called. We should updated the cache on reads too right?

component/libfuse/native_file_io.h
```diff
@@ -0,0 +1,229 @@
...
+    if (ret == 0) {
+        // As file is flushed and uploaded, reset the dirty bit here
+        handle_obj->dirty = 0;
+    }
+
+    return ret;
+}
+
+
+#ifdef ENABLE_READ_AHEAD
```
gapra-msft, line 154: can we explicitly mark this code as still in progress?

component/libfuse/native_file_io.h
vibhansa-msft: yes "#if 0" means code is disabled. Last run that i had with this change was successful so thinking to keep it in disabled mode only.

component/libfuse/native_file_io.h
vibhansa-msft: This code is disabled as of now so it will never hit. Just wanted to preserve some code done as part of read-ahead so that in future if we want to add it back, we have some reference. But the code is disabled so just treat this like some comments only.

cmd/mount.go
```diff
@@ -76,6 +79,9 @@ type mountOptions struct {
 	MemProfile        string     `config:"mem-profile"`
 	PassPhrase        string     `config:"passphrase"`
 	SecureConfig      bool       `config:"secure-config"`
+	DynamicProfiler   bool       `config:"dynamic-profile"`
+	ProfilerPort      int        `config:"profiler-port"`
+	ProfilerIP        string     `config:"profiler-ip"`
```
souravgupta-msft, line 84: Should we add the above three in baseConfig.yaml?

souravgupta-msft (approved):

tasherif-msft (approved):

PR #754, opened 2022-03-31
Fixed bug where remount on empty containers was not throwing error when the mount path has trailing '/'

souravgupta-msft:
Let's say we mount an empty container using blobfuse2. Now if we try to remount it again on the same path, it should ideally throw an error. But if the path contains trailing '/', it doesn't. So, added a line where we trim the trailing '/' from the mount path.

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -59,6 +59,9 @@ func IsDirectoryMounted(path string) bool {
 		return false
 	}
 
+	// removing trailing / from the path
+	path = strings.TrimRight(path, "/")
+
 	for _, line := range strings.Split(string(mntList), "\n") {
 		if strings.TrimSpace(line) != "" {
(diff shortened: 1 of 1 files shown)


vibhansa-msft (approved):

common/util.go
```diff
@@ -59,6 +59,9 @@ func IsDirectoryMounted(path string) bool {
 		return false
 	}
 
+	// removing trailing / from the path
+	path = strings.TrimRight(path, "/")
```
gapra-msft, line 63: Do we need a test for this?

gapra-msft (approved):

common/util.go
souravgupta-msft: Creating one. Will raise a new PR for this.

PR #756, opened 2022-04-04
Support . and .. in ls

gapra-msft:


component/libfuse/libfuse2_handler.go
```diff
@@ -428,6 +428,13 @@ func libfuse2_readdir(_ *C.char, buf unsafe.Pointer, filler C.fuse_fill_dir_t, o
 			}
 		}
 
+		if off_64 == 0 {
+			flags := internal.NewDirBitMap()
+			flags.Set(internal.PropFlagModeDefault)
+			attrs = append(attrs, &internal.ObjAttr{Flags: flags, Name: "."})
```
vibhansa-msft, line 434: We are still adding it in between here. Attr slice has already elements in place, and we are appending to that so . and .. will come towards the end of this slice, while there might be follow up calls to read this same directory

component/libfuse/libfuse2_handler.go
```diff
@@ -428,6 +428,13 @@ func libfuse2_readdir(_ *C.char, buf unsafe.Pointer, filler C.fuse_fill_dir_t, o
 			}
 		}
 
+		if off_64 == 0 {
+			flags := internal.NewDirBitMap()
```
vibhansa-msft, line 432: these flags are static in nature, we can define them as class variables and use the same all the time instead of creating on each readdir.

gapra-msft: just updated

souravgupta-msft (approved):

vibhansa-msft (approved):

PR #759, opened 2022-04-07
[BUG] Fixed block size max blocks exceeded bug when blocksize not set

tasherif-msft:
If the blocksize is not set - let's set it dynamically based on file size to avoid max blocks exceeded errors. 
changed default block size to 0.

### DIFF
--- a/component/azstorage/block_blob.go
+++ b/component/azstorage/block_blob.go
@@ -35,7 +35,6 @@ package azstorage
 
 import (
 	"blobfuse2/common"
-	"blobfuse2/common/config"
 	"blobfuse2/common/log"
 	"blobfuse2/internal"
 	"bytes"
@@ -614,7 +613,7 @@ func (bb *BlockBlob) WriteFromFile(name string, metadata map[string]string, fi *
 	defer log.TimeTrack(time.Now(), "BlockBlob::WriteFromFile", name)
 	blockSize := bb.Config.blockSize
 	// if the block size is not set then we configure it based on file size
-	if !config.IsSet(compName + ".block-size-mb") {
+	if bb.Config.blockSize == 0 {
 		// get the size of the file
 		stat, err := fi.Stat()
 		if err != nil {
@@ -724,12 +723,16 @@ func (bb *BlockBlob) createBlock(blockIdLength, startIndex, size int64) *common.
 }
 
 func (bb *BlockBlob) createNewBlocks(blockList *common.BlockOffsetList, offset, length, blockIdLength int64) int64 {
+	blockSize := bb.Config.blockSize
 	prevIndex := blockList.BlockList[len(blockList.BlockList)-1].EndIndex
+	if blockSize == 0 {
+		blockSize = (16 * 1024 * 1024)
+	}
 	// BufferSize is the size of the buffer that will go beyond our current blob (appended)
 	var bufferSize int64
-	for i := prevIndex; i < offset+length; i += bb.Config.blockSize {
+	for i := prevIndex; i < offset+length; i += blockSize {
 		// create a new block if we hit our block size
-		blkSize := int64(math.Min(float64(bb.Config.blockSize), float64((offset+length)-i)))
+		blkSize := int64(math.Min(float64(blockSize), float64((offset+length)-i)))
 		newBlock := bb.createBlock(blockIdLength, i, blkSize)
 		blockList.BlockList = append(blockList.BlockList, newBlock)
 		// reset the counter since it will help us to determine if there is leftovers at the end
--- a/component/azstorage/block_blob.go
+++ b/component/azstorage/block_blob.go
@@ -612,7 +612,6 @@ func (bb *BlockBlob) WriteFromFile(name string, metadata map[string]string, fi *
 	blockSize := bb.Config.blockSize
 	// if the block size is not set then we configure it based on file size
 	if !config.IsSet(compName + ".block-size-mb") {
-		fmt.Println("not here")
 		// get the size of the file
 		stat, err := fi.Stat()
 		if err != nil {
--- a/component/azstorage/block_blob.go
+++ b/component/azstorage/block_blob.go
@@ -72,7 +72,7 @@ type BlockBlob struct {
 var _ AzConnection = &BlockBlob{}
 
 const (
-	MaxBlocks = azblob.BlockBlobMaxStageBlockBytes * azblob.BlockBlobMaxBlocks
+	MaxBlocksSize = azblob.BlockBlobMaxStageBlockBytes * azblob.BlockBlobMaxBlocks
 )
 
 func (bb *BlockBlob) Configure(cfg AzStorageConfig) error {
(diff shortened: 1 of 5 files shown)


component/azstorage/block_blob.go
```diff
@@ -606,10 +607,35 @@ func (bb *BlockBlob) WriteFromFile(name string, metadata map[string]string, fi *
 	//defer exectime.StatTimeCurrentBlock("WriteFromFile::WriteFromFile")()
 
 	blobURL := bb.Container.NewBlockBlobURL(filepath.Join(bb.Config.prefixPath, name))
-
 	defer log.TimeTrack(time.Now(), "BlockBlob::WriteFromFile", name)
+	blockSize := bb.Config.blockSize
+	// if the block size is not set then we configure it based on file size
+	if !config.IsSet(compName + ".block-size-mb") {
```
vibhansa-msft, line 613: We could have set the value of blockSize field in config to 0 and just checked that instead of checking IsSet in each upload call. That will be more efficient.

component/azstorage/block_blob.go
```diff
@@ -606,10 +607,35 @@ func (bb *BlockBlob) WriteFromFile(name string, metadata map[string]string, fi *
...
+		// get the size of the file
+		stat, err := fi.Stat()
```
vibhansa-msft, line 615: Do we need to do this fi.Stat() before upload. Just check if we can pass down the file size in upload request itself from file-cache or something.

component/azstorage/block_blob.go
```diff
@@ -606,10 +607,35 @@ func (bb *BlockBlob) WriteFromFile(name string, metadata map[string]string, fi *
...
+		if err != nil {
+			log.Err("BlockBlob::WriteFromFile : Failed to get file size %s (%s)", name, err.Error())
+			return err
+		}
+		fileSize := stat.Size()
+		// If bufferSize > (BlockBlobMaxStageBlockBytes * BlockBlobMaxBlocks), then error
+		if fileSize > azblob.BlockBlobMaxStageBlockBytes*azblob.BlockBlobMaxBlocks {
```
vibhansa-msft, line 622: These are constants so instead of multiplying them on every upload, we can just do it once on startup and keep the value ready to be compared, will save some cycles on every upload.

component/azstorage/block_blob.go
```diff
@@ -606,10 +607,35 @@ func (bb *BlockBlob) WriteFromFile(name string, metadata map[string]string, fi *
...
+			err = errors.New("buffer is too large to upload to a block blob")
+			log.Err("BlockBlob::WriteFromFile : buffer is too large to upload to a block blob %s", name)
```
vibhansa-msft, line 624: Shall we say "block blob" or just "blob".

component/azstorage/block_blob.go
gapra-msft: We shouldnt assume anyone can give us the file size except for libfuse since a customer can one day just use libfuse azstorage right?

component/azstorage/block_blob.go
gapra-msft: I think block blob is more accurate and informative

component/azstorage/block_blob.go
tasherif-msft: if we do that then for all other calls our default block size would be 0. And in that case, we would be relying on the GO sdk logic to determine the block size for us. In our meeting we discussed we don't want to do that. Hence adding this logic to the CopyFromFile ourselves and leaving the rest as 16MB default. 
I don't think its that inefficient given that this method is `WriteFromFile` which uploads the entire file so I think its ok.

component/azstorage/block_blob.go
tasherif-msft: block blob is fine I think.

component/azstorage/block_blob.go

component/azstorage/block_blob.go
vibhansa-msft: I did not mean that. I am just suggesting instead of checking config.IsSet, the variable is already read into a field in config we can check that. It will be more efficient to check variable instead of searching into config map on every upload.

vibhansa-msft: Also, when we are initializing the class object, we are setting it to 16MB remove that and set it to 0, as there is no default block size anymore.

tasherif-msft: ah ok yea I thought we wanted to maintain that 16MB default but if we don't care then I can just have it as 0

component/azstorage/block_blob.go
```diff
@@ -606,10 +610,35 @@ func (bb *BlockBlob) WriteFromFile(name string, metadata map[string]string, fi *
 	//defer exectime.StatTimeCurrentBlock("WriteFromFile::WriteFromFile")()
 
 	blobURL := bb.Container.NewBlockBlobURL(filepath.Join(bb.Config.prefixPath, name))
-
 	defer log.TimeTrack(time.Now(), "BlockBlob::WriteFromFile", name)
+	blockSize := bb.Config.blockSize
+	// if the block size is not set then we configure it based on file size
+	if bb.Config.blockSize == 0 {
```
vibhansa-msft, line 616: blockSize is already populated with same value, shall we change the check also to if blocksize == 0

vibhansa-msft (approved):

component/azstorage/block_blob.go
```diff
@@ -71,6 +71,10 @@ type BlockBlob struct {
 // Verify that BlockBlob implements AzConnection interface
 var _ AzConnection = &BlockBlob{}
 
+const (
+	MaxBlocks = azblob.BlockBlobMaxStageBlockBytes * azblob.BlockBlobMaxBlocks
```
vibhansa-msft, line 75: While reading config if user has given config for block size, we shall check there if it exceeds the max allowed block size or not. Also, we can add the same test caes to our UT as well.

component/azstorage/block_blob.go
```diff
@@ -606,10 +610,35 @@ func (bb *BlockBlob) WriteFromFile(name string, metadata map[string]string, fi *
...
+		}
+		fileSize := stat.Size()
+		// If bufferSize > (BlockBlobMaxStageBlockBytes * BlockBlobMaxBlocks), then error
+		if fileSize > MaxBlocks {
+			err = errors.New("buffer is too large to upload to a block blob")
+			log.Err("BlockBlob::WriteFromFile : buffer is too large to upload to a block blob %s", name)
+			return err
+		}
+		// If bufferSize <= BlockBlobMaxUploadBlobBytes, then Upload should be used with just 1 I/O request
+		if fileSize <= azblob.BlockBlobMaxUploadBlobBytes {
```
vibhansa-msft, line 631: Shall we move this block size calculation to a seperate method in utils where you give blocksize and filesize and it returns back the chosen blocksize. That way this code can be reused in future if required and also then you can add UT to try all possible combinations and validate thta we do return correct blocksize.

gapra-msft (approved):

PR #763, opened 2022-04-11
Optimize Read Streaming Map Lookups

tasherif-msft:
optimized streaming by storing file buffers on handle level only and avoiding file name lookups on a map

### DIFF
--- a/component/stream/stream.go
+++ b/component/stream/stream.go
@@ -300,7 +300,9 @@ func (st *Stream) WriteFile(options internal.WriteFileOptions) (int, error) {
 func (st *Stream) CloseFile(options internal.CloseFileOptions) error {
 	log.Trace("Stream::CloseFile : name=%s, handle=%d", options.Handle.Path, options.Handle.ID)
 	st.NextComponent().CloseFile(options)
-	st.streamCache.removeCachedHandle(options.Handle)
+	if !st.streamOnly {
+		st.streamCache.removeCachedHandle(options.Handle)
+	}
 	return nil
 }
 
--- a/component/stream/stream.go
+++ b/component/stream/stream.go
@@ -45,7 +45,6 @@ import (
 	"io"
 	"math"
 	"os"
-	"strconv"
 	"strings"
 
 	"github.com/bluele/gcache"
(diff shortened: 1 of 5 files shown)


component/stream/stream.go
```diff
@@ -55,23 +54,20 @@ import (
 
 type Stream struct {
 	internal.BaseComponent
-	streamCache       *cache
-	streamOnly        bool // parameter used to check if its pure streaming
-	fileKeyLocks      *common.LockMap
-	cacheOnFileHandle bool // much better option for prefetching
+	streamCache *cache
+	streamOnly  bool // parameter used to check if its pure streaming
```
vibhansa-msft, line 58: so the major change here is to just always use file-handle level cache only ?

component/stream/stream_cache.go
```diff
@@ -80,104 +80,61 @@ type cache struct {
...
+	handle.DataBuffer = fc
 }
 
 // try to retrieve the block - return missing if it is not cached
-func (c *cache) getBlock(fileKey string, offset int64, fileSize int64) (*cacheBlock, bool) {
+func (c *cache) getBlock(handle *handlemap.Handle, offset int64) (*cacheBlock, bool) {
 	blockSize := c.blockSize
-	blockKeyObj := blockKey{fileKey: fileKey, offset: offset}
+	blockKeyObj := blockKey{handle: handle, offset: offset}
 	c.Lock()
```
vibhansa-msft, line 103: Do we still need this lock? as blocks are within the handle itself policy level locks are no more needed I guess.

component/stream/stream_cache.go
```diff
@@ -190,17 +147,26 @@ func (c *cache) teardown() {
...
 	}
-
 	// Cleanup block residing on disk path
 	if c.diskPersistence {
 		c.wipeoutDiskCache()
 	}
 }
 
+func (c *cache) removeCachedHandle(handle *handlemap.Handle) {
+	c.Lock()
```
vibhansa-msft, line 163: we can have handle level lock instead of entire cache policy lock here.

component/stream/stream_cache.go
```diff
@@ -46,8 +47,8 @@ import (
 )
 
 type blockKey struct {
-	offset  int64
-	fileKey string
+	offset int64
+	handle *handlemap.Handle
```
vibhansa-msft, line 51: cacheFile struct is no more in use we can remove that

component/stream/stream.go
```diff
@@ -289,10 +266,10 @@ func (st *Stream) copyCachedBlock(fileKey string, handle *handlemap.Handle, offs
 		// round all offsets to the specific blocksize offsets
 		cachedBlockStartIndex := (offset - (offset % st.streamCache.blockSize))
 		// Lock on requested block and fileName to ensure it is not being rerequested or manipulated
-		block, exists, err := st.fetchBlock(fileKey, handle, cachedBlockStartIndex)
+		block, exists, err := st.fetchBlock(handle, cachedBlockStartIndex)
```
vibhansa-msft, line 269: When write stream comes this offset based calculation of figuring out the block will not work as different blocks can be of different sizes. We need to reorg this code where block search is based on whether its read-only or read-write streaming.

component/stream/stream_cache.go
tasherif-msft: Its not possible to remove those locks :/ it would cause a deadlock in some cases if eviction needs to happen on the same handle that we're adding a new block to. And the solution around that deadlock would mean the possibility of duplicate blocks in the buffer.

component/stream/stream_cache.go
tasherif-msft: same as above. One thing to note - perf did not change much when I removed this lock.

component/stream/stream.go
tasherif-msft: yea when putting the pr for write stream I will create a base class where read and write stream inherit from - read stream will have a different block definition from write stream. This will not be in this PR.

component/stream/stream_cache.go
gapra-msft: @tasherif-msft can you add a comment in the relevant areas to mention that we cant remove this lock due to potential deadlock issues? Just for future code maintainability?

component/stream/stream_cache.go
vibhansa-msft: we can have a handle level lock and a block level lock. So before any addition/deletion of block lock the handle -> lock thta particular block and then do the operation. For regular read-write lock the handle, get the block, lock the block, unlock the handle. this shall not create any deadlock I suppose.

PR #765, opened 2022-04-13
Temporarily fixed fio error caused due to race condition between OpenFile and CloseFile

gapra-msft:
This will require some more thought into how to implement directory deletion properly. 

Note: This is fine to do temporarily for now since we already have a flaw in our logic where if we cache a file a/b/c/d/file.txt, then on close we will only delete directory d, leaving a/b/c still in the path. 

There is a race condition between https://github.com/Azure/azure-storage-fuse/blob/main/component/file_cache/lru_policy.go#L442 and https://github.com/Azure/azure-storage-fuse/blob/main/component/file_cache/file_cache.go#L728

OpenFile expects the directories to be present so we call MkDirAll. In between those two calls another CloseFile can come and cause that dir to be deleted

### DIFF
--- a/component/file_cache/lfu_policy.go
+++ b/component/file_cache/lfu_policy.go
@@ -36,7 +36,6 @@ package file_cache
 import (
 	"blobfuse2/common/log"
 	"os"
-	"path/filepath"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


component/file_cache/lfu_policy.go
```diff
@@ -158,7 +158,8 @@ func (l *lfuPolicy) clearItemFromCache(path string) {
 		// File was deleted so try clearing its parent directory
 		dirPath := filepath.Dir(path)
 		if dirPath != l.tmpPath {
-			os.Remove(dirPath)
```
vibhansa-msft, line 161: Lets remove this entire condition/ You can just add the comment saying we need to cleanup empty directories at some point in time.

vibhansa-msft (approved):

component/file_cache/lfu_policy.go
gapra-msft: Done

vibhansa-msft (approved):

souravgupta-msft (approved):

PR #769, opened 2022-04-19
Deprecate using 3rd party eviction policies

tasherif-msft:
Removed disk persistence, file name caching to optimize perf and improve logic cleanliness until we gauge customer needs on GA.


component/stream/stream.go
```diff
@@ -55,25 +54,38 @@ import (
...
+type cache struct {
+	sync.RWMutex
+
+	evictionPolicy      common.EvictionPolicy
+	blocks              handlemap.LRUCache // blocks stored: {blockKey(off1, fileKey1): cacheBlock1, blockKey(off1, fileKey2): cacheBlock2, ...}
+	blockSize           int64
+	bufferSizePerHandle uint64 // maximum number of blocks allowed to be stored for a file
+	cacheSize           uint64 // maximum allowed configured number of blocks
+
+	diskBlocks      gcache.Cache // blocks stored on disk when persistence is on
```
vibhansa-msft, line 82: if we still use gcache then its eviction logic and locks will still kick in right ?

internal/handlemap/lru_policy.go
```diff
@@ -0,0 +1,254 @@
...
+	Path       string
+}
+
+//LRUCache definition for Least Recently Used Cache implementation.
+type LRUCache struct {
+	Capacity        int64                      //defines a cache object of the specified capacity.
+	List            *list.List                 //DoublyLinkedList for backing the cache value.
+	Elements        map[BlockKey]*list.Element //blockKey: KeyPair
+	Occupied        int64
+	DiskPersistence bool
```
vibhansa-msft, line 36: we don't need this flag on per handle basis, it's a global thing at policy level.

tasherif-msft: yea I had issues with imports so I placed this file here for now, we can discuss in technical meeting

setup/baseConfig.yaml
```diff
@@ -39,7 +39,6 @@ stream:
   blocks-per-file: <maximum number of blocks, per file, that can be cached at any given time. Default - 0>
   cache-size-mb: <total amount of memory that streaming can consume on the system. Default - 0 MB>
   policy: lfu|lru|arc <cache eviction policy for cached blocks. Default - arc>
```
gapra-msft, line 41: remove this?

internal/handlemap/lru_policy.go
```diff
@@ -0,0 +1,266 @@
...
+	"container/list"
+	"fmt"
+	"os"
+	"path/filepath"
+	"sync"
+)
+
+type BlockKey struct {
+	StartIndex int64
+	Handle     *Handle
```
vibhansa-msft, line 13: Instead of handle object we can use the handle id here.

setup/baseConfig.yaml
```diff
@@ -35,17 +35,11 @@ libfuse:
...
-  cache-size-mb: <total amount of memory that streaming can consume on the system. Default - 0 MB>
-  policy: lfu|lru|arc <cache eviction policy for cached blocks. Default - arc>
-  cache-on-file-handle: true|false <cache block per file handle instead of file-name. (better performance but increases memory consumption)>
-  disk-persistence: <persist blocks on disk when evicted from memory>
-  disk-cache-path: <path to perist evicted blocks>
-  disk-size-mb: <maximum disk space to be consumed by disk-persistence>
-  disk-timeout-sec: <number of seconds a persisted block is considered valid>
+  block-size-mb: <for read only mode - size of each block to be cached in memory while streaming (in MB). Default - 0 MB>
+  handle-limit: <total amount of memory that streaming can consume on the system. Default - 0 MB>
+  handle-buffer-size-mb: <capacity of cached data for per file handle. Default - 0>
```
vibhansa-msft, line 40: handle-buffer-size-mb is not required here. We are reading only total cache capacity and number of handles

component/stream/stream.go
```diff
@@ -245,57 +176,74 @@ func (st *Stream) OpenFile(options internal.OpenFileOptions) (*handlemap.Handle,
...
-func (st *Stream) copyCachedBlock(fileKey string, handle *handlemap.Handle, offset int64, data []byte) (int, error) {
+func (st *Stream) copyCachedBlock(handle *handlemap.Handle, offset int64, data []byte) (int, error) {
 	dataLeft := int64(len(data))
 	// counter to track how much we have copied into our request buffer thus far
 	dataRead := 0
 	// covers the case if we get a call that is bigger than the file size
 	for dataLeft > 0 && offset < handle.Size {
 		// round all offsets to the specific blocksize offsets
-		cachedBlockStartIndex := (offset - (offset % st.streamCache.blockSize))
+		cachedBlockStartIndex := (offset - (offset % st.cache.blockSize))
```
vibhansa-msft, line 238: this logic will not work for write streaming so when we break the code into smaller parts this getBlocka and copyCachedBlock shall be part of the implementation of that piece.

component/stream/stream.go
```diff
@@ -43,35 +43,29 @@ import (
...
-	"go.uber.org/atomic"
 )
 
 type Stream struct {
 	internal.BaseComponent
-	streamCache       *cache
-	streamOnly        bool // parameter used to check if its pure streaming
-	fileKeyLocks      *common.LockMap
-	cacheOnFileHandle bool // much better option for prefetching
+	cache      *cache
```
vibhansa-msft, line 53: Why we need this as pointer, could have been a regular member as well.

component/stream/stream.go
```diff
@@ -134,104 +117,52 @@ func (st *Stream) Configure() error {
...
-		st.streamCache.teardown()
+	handleMap := handlemap.GetHandles()
+	handles := map[string]interface{}{}
+	handleMap.Range(func(key, value interface{}) bool {
+		handles[fmt.Sprint(key)] = value
+		return true
+	})
+	for _, v := range handles {
+		handle := v.(*handlemap.Handle)
+		handle.CacheObj.Lock()
```
vibhansa-msft, line 152: not a safe assumption that all handle objects have cacheobj as we create handle objects for readDir operation as well.

component/stream/stream.go
```diff
@@ -134,104 +117,52 @@ func (st *Stream) Configure() error {
...
 }
 
 // Stop : Stop the component functionality and kill all threads started
 func (st *Stream) Stop() error {
 	log.Trace("Stopping component : %s", st.Name())
-	if !st.streamOnly {
-		st.streamCache.teardown()
+	handleMap := handlemap.GetHandles()
+	handles := map[string]interface{}{}
+	handleMap.Range(func(key, value interface{}) bool {
```
vibhansa-msft, line 146: Why there are two iterations one to convert handle map to a slice and then iterate a slice. This could have been done in single iteration as well.

component/stream/stream.go
```diff
@@ -245,57 +176,74 @@ func (st *Stream) OpenFile(options internal.OpenFileOptions) (*handlemap.Handle,
...
-
-		flock := st.fileKeyLocks.Get(fileKey)
-		flock.Lock()
-		st.streamCache.addFileKey(fileKey)
-		block, exists, _ := st.fetchBlock(fileKey, handle, 0)
+		atomic.AddInt32(&st.cache.openHandles, 1)
+		cacheObj := handlemap.Cache{
+			DataBuffer: common.NewLRUCache(int64(st.cache.bufferSizePerHandle)),
+		}
+		handle.CacheObj = &cacheObj
```
vibhansa-msft, line 188: Shall we have a method in handlemap to do this where we can just call handle.CreateCacheObject and pass on the size ?

vibhansa-msft: just in case later we need to add more stuff related to policy or other things, it will be easier to have it as a seperate method.

component/stream/stream.go
```diff
@@ -245,57 +176,74 @@ func (st *Stream) OpenFile(options internal.OpenFileOptions) (*handlemap.Handle,
...
 }
 
-func (st *Stream) fetchBlock(fileKey string, handle *handlemap.Handle, offset int64) (*cacheBlock, bool, error) {
-	block, exists := st.streamCache.getBlock(fileKey, offset, handle.Size)
-	if !exists {
+func (st *Stream) getBlock(handle *handlemap.Handle, offset int64) (*common.CacheBlock, bool, error) {
+	blockSize := st.cache.blockSize
+	blockKeyObj := offset
+	handle.CacheObj.Lock()
+	block, found := handle.CacheObj.DataBuffer.Get(blockKeyObj)
```
vibhansa-msft, line 200: handle object has a cache pointer and that internally has a lru object so if we use lru object as unname variable this indirections will look more easier to understand. like handle.CacheObj.Get()/Put()

component/stream/stream.go
```diff
@@ -245,57 +176,74 @@ func (st *Stream) OpenFile(options internal.OpenFileOptions) (*handlemap.Handle,
...
+	if !found {
+		if (offset + blockSize) > handle.Size {
```
vibhansa-msft, line 202: Can we remove all this block allocations related stuff from here. May be just like handle and policy we can have a cache object file and allocation of block and all can be done inside. That simplifies the outside logic where you just call Get()/Put() on handle and if required it can allocate the required block inside and return back a locked block object.

vibhansa-msft: This will also help us keep a finer control over locks so that in no code change we accidently mess it up here.

vibhansa-msft: Even the unlock of block can be moved inside the cache layer instead of having a wrapper function here.

component/stream/stream.go
```diff
@@ -245,57 +176,74 @@ func (st *Stream) OpenFile(options internal.OpenFileOptions) (*handlemap.Handle,
...
+			Data:   block.Data,
 		}
 		_, err := st.NextComponent().ReadInBuffer(options)
 		if err != nil && err != io.EOF {
-			return nil, exists, err
+			return nil, false, err
 		}
+		return block, false, nil
+	} else {
+		handle.CacheObj.Unlock()
```
vibhansa-msft, line 226: Rlock() is done inside the Get() method but exclusive Lock() is done here in stream layer. Better to unify it somehow so that all are done in one place.

common/lru_policy.go
```diff
@@ -0,0 +1,113 @@
...
+	return &LRUCache{
+		Capacity: capacity,
+		List:     new(list.List),
+		Elements: make(map[int64]*list.Element),
+	}
+}
+
+//Get: returns the cache value stored for the key, cache hits the handle and moves the list pointer to front of the list
+func (cache *LRUCache) Get(bk int64) (*CacheBlock, bool) {
+	found := false
```
vibhansa-msft, line 43: Here we are assuming that we have a start index of the buffer and using that as key we are searching. This might be true in case of read stream but when write stream comes into play this logic will be very specific to read stream alone. So better to rename this method to GetBlockByStartIndex or something like this so that later we can add another method GetBlockByOffset() which write stream can use.

common/lru_policy.go
```diff
@@ -0,0 +1,113 @@
...
+		found = true
+		cb.RLock()
+	}
+	return cb, found
+}
+
+//Put: Inserts the key,value pair in LRUCache.
+func (cache *LRUCache) Put(key int64, value *CacheBlock) {
+	if cache.Occupied >= cache.Capacity {
+		pair := cache.List.Back().Value.(*list.Element).Value.(KeyPair)
```
vibhansa-msft, line 57: We cannot blindly remove this block here. There might be someone reading from this block so better to take a lock and then remove it to ensure consistency.

common/lru_policy.go
```diff
@@ -0,0 +1,113 @@
...
+
+func (cache *LRUCache) RecentlyUsed() *CacheBlock {
+	return cache.List.Front().Value.(*list.Element).Value.(KeyPair).value
+}
+
+//Remove: removes the entry for the respective key
+func (cache *LRUCache) Remove(key int64) {
+	// get the keyPair associated with the blockKey
+	if node, ok := cache.Elements[key]; ok {
+		node.Value.(*list.Element).Value.(KeyPair).value.Lock()
```
vibhansa-msft, line 94: difficult to read as there are multiple indirection and type conversion happening here.

common/lru_policy.go
```diff
@@ -0,0 +1,113 @@
...
+		node.Value.(*list.Element).Value.(KeyPair).value.Data = nil
+		delete(cache.Elements, key)
+		cache.List.Remove(node)
+	}
+}
+
+//Purge: clears LRUCache
+func (cache *LRUCache) Purge() {
+	for _, bk := range cache.Keys() {
+		cache.Remove(bk)
```
vibhansa-msft, line 108: Here as well we shall try to lock each node and then remove just to ensure consistency.

setup/baseConfig.yaml
tasherif-msft: I thought we agreed in the meeting that its more user friendly to have it in the context of handle since we already have `handle-limit` its easier to understand handle buffer size.

component/stream/stream.go
tasherif-msft: sounds good - that will be in the write PR.

common/lru_policy.go
tasherif-msft: why would it be specific? the way I see this is, by the time we get the call here with the startIndex we have already found exactly which block offset from the block list we're looking for, here we are simply seeing if that block is cached or not. But even in write stream we will use start indices.

common/lru_policy.go
tasherif-msft: we do lock when we call the `Remove()` method.

common/lru_policy.go
tasherif-msft: locking happens in remove method, is there a reason you don't want it there?

setup/baseConfig.yaml
```diff
@@ -35,17 +35,11 @@ libfuse:
...
-  blocks-per-file: <maximum number of blocks, per file, that can be cached at any given time. Default - 0>
-  cache-size-mb: <total amount of memory that streaming can consume on the system. Default - 0 MB>
-  policy: lfu|lru|arc <cache eviction policy for cached blocks. Default - arc>
-  cache-on-file-handle: true|false <cache block per file handle instead of file-name. (better performance but increases memory consumption)>
-  disk-persistence: <persist blocks on disk when evicted from memory>
-  disk-cache-path: <path to perist evicted blocks>
-  disk-size-mb: <maximum disk space to be consumed by disk-persistence>
-  disk-timeout-sec: <number of seconds a persisted block is considered valid>
+  block-size-mb: <for read only mode - size of each block to be cached in memory while streaming (in MB). Default - 0 MB>
+  handle-limit: <total amount of memory that streaming can consume on the system. Default - 0 MB>
```
gapra-msft, line 39: This is total number of handles right?

tasherif-msft: yea open handles at once

component/stream/stream.go
tasherif-msft: I think youre describing the read.go and read_write.go, where stream -> read.go/read_write.go -> lru_policy.go. Where implementation of block etc is gonna be specific to read only or read write mode. This shall be in the pr for write which im working on. In the meantime this is sufficient changes in one PR. For the locking of block ill move it outside of lru_policy

component/stream/stream.go
tasherif-msft: done

common/lru_policy.go
tasherif-msft: I made a common method that makes it more readable.

common/lru_policy.go
```diff
@@ -0,0 +1,113 @@
...
+	value *CacheBlock
+}
+
+//LRUCache definition for Least Recently Used Cache implementation
+type LRUCache struct {
+	sync.RWMutex
+	Capacity int64
+	List     *list.List              //DoublyLinkedList: node1->node2.... node:=KeyPair
+	Elements map[int64]*list.Element //blockKey:KeyPair
+	Occupied int64
```
gapra-msft, line 29: What is this variable? To keep track of the number of blocks in the cache?

component/stream/stream.go
tasherif-msft: I added a check to make sure its not an empty cache{}

common/lru_policy.go
tasherif-msft: to keep track of the occupied space/capacity

setup/baseConfig.yaml
vibhansa-msft: Yes, we agreed to get handle limit and total memory and in code as well you divide the total memory by handle to get per handle buffer size. Then why we are reading this "handle-buffer-size-mb" from config. Are we like two modes where either user can give total memory or per handle buffer size?

common/lru_policy.go
vibhansa-msft: as block sizes are not fixed in case of write streaming, how do you plan to calculate start index from offset?

common/lru_policy.go
vibhansa-msft: Wil rechcek the remove code.

common/lru_policy.go
tasherif-msft: it won't be done on the policy side. It would be done in read_write.go where we will call the binary search on block offset etc when `Get()` in lru policy returns block not found. Then we simply cache the missing block with its block offset list offset/startIndex.

common/lru_policy.go
```diff
@@ -0,0 +1,118 @@
...
+
+//KeyPair: the list node containing both block key and cache block values
+type KeyPair struct {
+	key   int64
+	value *CacheBlock
+}
+
+//LRUCache definition for Least Recently Used Cache implementation
+type LRUCache struct {
+	sync.RWMutex
```
vibhansa-msft, line 25: class holding this cache is also having a lock, we can remove one as they are duplicates

vibhansa-msft, line 25: class holding this cache is also having a lock, we can remove one as they are duplicates

common/lru_policy.go
```diff
@@ -0,0 +1,118 @@
...
+		},
+	}
+	pointer := cache.List.PushFront(node)
+	cache.Occupied += (node.Value.(KeyPair).value.EndIndex - node.Value.(KeyPair).value.StartIndex)
+	cache.Elements[key] = pointer
+}
+
+func (cache *LRUCache) Print() {
+	for _, value := range cache.Elements {
+		fmt.Printf("Key:%+v,Value:%+v\n", getKeyPair(value).value.StartIndex, getKeyPair(value).value.EndIndex)
```
vibhansa-msft, line 72: convert this to log.debug

vibhansa-msft, line 72: convert this to log.debug

tasherif-msft: issue with logger and importing it. Will keep as print for now as it is only used for debugging.

vibhansa-msft (approved):

common/lru_policy.go
gapra-msft: whats the issue with importing it?

common/lru_policy.go
```diff
@@ -0,0 +1,118 @@
...
+type CacheBlock struct {
+	sync.RWMutex
+	StartIndex int64
+	EndIndex   int64
+	Data       []byte
+	Last       bool //last block in the file?
+}
+
+//KeyPair: the list node containing both block key and cache block values
+type KeyPair struct {
```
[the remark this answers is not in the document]
tasherif-msft: I think KeyPair is more clear with respect to the LRU policy than `CacheItem`

gapra-msft (approved):

PR #771, opened 2022-04-19
Truncate file performance optimization

gapra-msft:


gapra-msft:
Will reopen this PR if it becomes an actual issue

PR #777, opened 2022-04-30
Write Stream Caching Support

tasherif-msft:


common/cache_policy/lru_policy.go
```diff
@@ -50,11 +44,30 @@ func (cache *LRUCache) Get(bk int64) (*CacheBlock, bool) {
...
+func (cache *LRUCache) findCleanBlockToEvict() bool {
+	node := cache.List.Back()
+	pair := getKeyPair(node)
+	for i := 0; i < cache.List.Len(); i++ {
+		if !pair.value.Dirty {
+			fmt.Println("somehow not dirty")
+			cache.Remove(pair.key)
+			return false
+		} else {
+			node = node.Next()
```
vibhansa-msft, line 57: We started with cache.list.Back() (last node in list) so for iteration we shall be going node.Prev() and not .Next() right ?

common/types.go
```diff
@@ -173,17 +174,20 @@ type LogConfig struct {
 }
 
 type Block struct {
+	sync.RWMutex
 	StartIndex int64
 	EndIndex   int64
-	Size       int64
+	Data       []byte
+	Dirty      bool
```
vibhansa-msft, line 181: In common we have a bitmap we can use it here instead of maintaining multiple flags. That will conserve memory and also addition of new flags in future will not require additional variables in struct. See HandleMap implementation, it was converted to same model.

common/types.go
```diff
@@ -173,17 +174,20 @@ type LogConfig struct {
 }
 
 type Block struct {
+	sync.RWMutex
 	StartIndex int64
 	EndIndex   int64
-	Size       int64
+	Data       []byte
```
vibhansa-msft, line 180: move slice to end of struct for better packing.

common/types.go
```diff
@@ -173,17 +174,20 @@ type LogConfig struct {
...
+	Last       bool //last block in the file?
 	Id         string
-	Modified   bool
 }
 
 // list that holds blocks containing ids and corresponding offsets
 type BlockOffsetList struct {
 	BlockList []*Block //blockId to offset mapping
 	Cached    bool     // is it cached?
+	SmallFile bool
```
vibhansa-msft, line 190: same here, we can use bitmap flag.

common/types.go
```diff
@@ -225,9 +249,9 @@ func (bol BlockOffsetList) FindBlocksToModify(offset, length int64) (int, int64,
 		}
 		if currentBlockOffset >= blk.StartIndex && currentBlockOffset < blk.EndIndex && currentBlockOffset <= offset+length {
 			appendOnly = false
-			blk.Modified = true
+			blk.Dirty = true
```
vibhansa-msft, line 252: Block which are marked dirty, once uploaded shall be reset to non-dirty mode.

component/stream/read_write.go
```diff
@@ -0,0 +1,197 @@
...
+		if err != nil && err != io.EOF {
+			log.Err("Stream::WriteFile : error failed to write data for %s: [%s]", options.Handle.Path, err.Error())
+		}
+		return data, err
+	}
+	data, err := rw.readWriteBlocks(options.Handle, options.Offset, options.Data, true)
+	return data, err
+}
+
+func (rw *ReadWriteCache) CloseFile(options internal.CloseFileOptions) error {
```
vibhansa-msft, line 86: Just like our main component we shall have both FlushFile and CloseFile seperatly. In flush just upload and in close call flush and then close the handle, may be delete all cached buffers and all. However in flush we keep our buffers and just upload the current changes.

internal/handlemap/handle_map.go
```diff
@@ -56,6 +56,8 @@ const (
 
 type Cache struct {
 	*cache_policy.LRUCache
+	*common.BlockOffsetList
+	StreamOnly bool
```
vibhansa-msft, line 60: do we need this per handle ? streamonly or cache-and-stream is a stream component level decision ?

tasherif-msft: yes it's per handle since we want to dictate if a handle limit is reached then the handle should stream only until we have a free spot again

internal/base_component.go
```diff
@@ -282,6 +282,13 @@ func (base *BaseComponent) GetAttr(options GetAttrOptions) (*ObjAttr, error) {
 	return &ObjAttr{}, nil
 }
 
+func (base *BaseComponent) GetFileBlockOffsets(options GetFileBlockOffsetsOptions) (*common.BlockOffsetList, error) {
```
vibhansa-msft, line 285: This method is added to MockComponent ?

component/stream/stream.go
```diff
@@ -163,117 +128,40 @@ func (st *Stream) unlockBlock(block *cache_policy.CacheBlock, exists bool) {
...
-		}
-		return data, err
-	}
-	return st.copyCachedBlock(options.Handle, options.Offset, options.Data)
+	return st.cache.ReadInBuffer(options)
 }
 
 func (st *Stream) WriteFile(options internal.WriteFileOptions) (int, error) {
-	return st.NextComponent().WriteFile(options)
+	return st.cache.WriteFile(options)
```
vibhansa-msft, line 144: If we are in streaming only mode, then we need to forward the write call to next component?

internal/base_component.go
tasherif-msft: yep

component/stream/stream.go
tasherif-msft: yes, if its stream only mode it will go to next comp

common/types.go
```diff
@@ -172,22 +174,51 @@ type LogConfig struct {
 	TimeTracker bool
 }
 
+// Flags represented in BitMap for various flags in the handle
+const (
+	BlockFlagUnknown uint16 = iota
```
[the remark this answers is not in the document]
gapra-msft: oh nvm its just like a null flag

common/types.go
```diff
@@ -172,22 +174,51 @@ type LogConfig struct {
 	TimeTracker bool
 }
 
+// Flags represented in BitMap for various flags in the handle
+const (
```
gapra-msft, line 178: can you comment this as Flags for Block, Flags for Block Offset List, etc

component/azstorage/block_blob.go
```diff
@@ -731,33 +731,39 @@ func (bb *BlockBlob) GetFileBlockOffsets(name string) (*common.BlockOffsetList,
...
+	newBlockId := base64.StdEncoding.EncodeToString(common.NewUUIDWithLength(blockIdLength))
 	newBlock := &common.Block{
 		Id:         newBlockId,
 		StartIndex: startIndex,
 		EndIndex:   startIndex + size,
-		Size:       size,
-		Modified:   true,
 	}
+	// mark truncated since it is a new empty block
+	newBlock.Flags.Set(common.TruncatedBlock)
```
gapra-msft, line 760: is truncated flag for just empty blocks?

component/azstorage/block_blob.go
```diff
@@ -766,16 +772,98 @@ func (bb *BlockBlob) createNewBlocks(blockList *common.BlockOffsetList, offset,
...
+	// log.Trace("BlockBlob::TruncateFile : name=%s, size=%d", name, size)
+	attr, err := bb.GetAttr(name)
+	if err != nil {
+		log.Err("BlockBlob::TruncateFile : Failed to get attributes of file %s (%s)", name, err.Error())
+		if err == syscall.ENOENT {
+			return err
+		}
+	}
+	if size == 0 || attr.Size == 0 {
+		err := bb.WriteFromBuffer(name, nil, make([]byte, size))
```
gapra-msft, line 813: Do we need to worry about size being huge and allocating a huge array?

component/azstorage/block_blob.go
gapra-msft: Null block or Zeroed block might be better description

component/azstorage/block_blob.go
```diff
@@ -766,16 +772,98 @@ func (bb *BlockBlob) createNewBlocks(blockList *common.BlockOffsetList, offset,
...
+			log.Err("BlockBlob::TruncateFile : Failed to set the %s to 0 bytes (%s)", name, err.Error())
+		}
+		return err
+	}
+	bol, err := bb.GetFileBlockOffsets(name)
+	if err != nil {
+		log.Err("BlockBlob::TruncateFile : Failed to get block list of file %s (%s)", name, err.Error())
+		return err
+	}
+	if !bol.SmallFile() {
```
gapra-msft, line 824: Can you comment some of these sections

component/azstorage/block_blob.go
```diff
@@ -766,16 +772,98 @@ func (bb *BlockBlob) createNewBlocks(blockList *common.BlockOffsetList, offset,
...
+		if err != nil {
+			log.Err("BlockBlob::TruncateFile : Failed to truncate file %s", name, err.Error())
+			return err
+		}
+	}
+	return nil
+}
+
 // Write : write data at given offset to a blob
 func (bb *BlockBlob) Write(options internal.WriteFileOptions) error {
```
gapra-msft, line 868: We shouldnt enter a situation where a block that is being written is a TruncatedBlock right? Since we immediately StageAndCommit on Truncate?

component/azstorage/block_blob.go
```diff
@@ -895,6 +974,50 @@ func (bb *BlockBlob) stageAndCommitModifiedBlocks(name string, data []byte, inde
 	return nil
 }
 
+func (bb *BlockBlob) StageAndCommit(name string, bol *common.BlockOffsetList) error {
+	blobURL := bb.Container.NewBlockBlobURL(filepath.Join(bb.Config.prefixPath, name))
```
gapra-msft, line 978: Whats the difference between this and the method above?

common/types.go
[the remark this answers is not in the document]
tasherif-msft: just a null flag

component/azstorage/block_blob.go
tasherif-msft: We shouldn't, since `write` is used when we're streaming directly, and the truncate operation is also essentially being streamed.

component/azstorage/block_blob.go
tasherif-msft: the one above is intended for write streams (w/o caching) where we need to have a window on the data buffer breaking it into blocks. I can look into cleaning it up further.

component/azstorage/block_blob.go
tasherif-msft: yes, that is a possibility - it's one of the improvement items post this PR. Did not want to have too many changes in this PR.

component/stream/connection.go
```diff
@@ -0,0 +1,42 @@
...
+)
+
+type StreamConfig struct {
+	blockSize           int64
+	bufferSizePerHandle uint64 // maximum number of blocks allowed to be stored for a file
+	handleLimit         int32
+	openHandles         int32
+}
+
+type StreamConnection interface {
```
gapra-msft, line 15: looks so neat! I love clean code

tasherif-msft: ty ty C:

vibhansa-msft (approved):

common/cache_policy/lru_policy.go
```diff
@@ -50,11 +43,26 @@ func (cache *LRUCache) Get(bk int64) (*CacheBlock, bool) {
 	return cb, found
 }
 
-//Put: Inserts the key,value pair in LRUCache.
-func (cache *LRUCache) Put(key int64, value *CacheBlock) {
+//Resize: resizes a cached block and adjusts occupied size
+func (cache *LRUCache) Resize(bk, newEndIndex int64) bool {
+	var cb *common.Block
+	if node, ok := cache.Elements[bk]; ok {
```
gapra-msft, line 49: is this only for the last block in a file?

common/cache_policy/lru_policy.go
```diff
@@ -116,3 +129,21 @@ func getKeyPair(node *list.Element) KeyPair {
 	//uncast the keypair
 	return node.Value.(*list.Element).Value.(KeyPair)
 }
+
+// return true if no eviction happened/cache full, return false otherwise
+func (cache *LRUCache) evict() bool {
```
gapra-msft, line 134: I think we can think of a better name for this since the boolean return semantics is the opposite of what you'd expect

component/azstorage/block_blob.go
```diff
@@ -731,33 +731,39 @@ func (bb *BlockBlob) GetFileBlockOffsets(name string) (*common.BlockOffsetList,
 		log.Err("BlockBlob::GetFileBlockOffsets : Failed to get block list %s ", name, err.Error())
 		return &common.BlockOffsetList{}, err
 	}
+	// if block list empty its a small file
+	if len(blockList.BlockList) == 0 {
```
gapra-msft, line 735: Dont we have a SmallFile flag for this?

gapra-msft: Oh you're setting it below

component/stream/read_write.go
```diff
@@ -0,0 +1,458 @@
...
+		}
+	}
+	return handle, err
+}
+
+func (rw *ReadWriteCache) ReadInBuffer(options internal.ReadInBufferOptions) (int, error) {
+	// log.Trace("Stream::ReadInBuffer : name=%s, handle=%d, offset=%d", options.Handle.Path, options.Handle.ID, options.Offset)
+	options.Handle.CacheObj.Lock()
+	defer options.Handle.CacheObj.Unlock()
+	if !rw.StreamOnly && options.Handle.CacheObj.StreamOnly {
```
gapra-msft, line 70: if the handle is stream only why do we need to create a cache handle?

common/cache_policy/lru_policy.go
tasherif-msft: currently im using it for that, but this can be a resize for any block in the cache.

component/stream/read_write.go
tasherif-msft: when we convert a handle to stream only it's because we're at capacity, so in subsequent calls we can still check if we're at capacity and if we're not then we can convert to caching on that handle - rather than stream it.

common/cache_policy/lru_policy.go
tasherif-msft: yea I can change to true being a successful eviction and false being unsuccessful

common/cache_policy/lru_policy.go
gapra-msft: it has to be the end though right? otherwise a block can overlap

tasherif-msft: yes, this policy is separate from streaming - streaming maintains its own state of how the blocks should be placed. This simply caches/resizes whatever blocks are passed down.

gapra-msft (approved):

souravgupta-msft (approved):

common/cache_policy/lru_policy.go
```diff
@@ -1,23 +1,16 @@
 package cache_policy
 
 import (
+	"blobfuse2/common"
 	"blobfuse2/common/log"
```
vibhansa-msft, line 5: copyright statements are missing in all files.

PR #780, opened 2022-05-03
Added cleanup to test that was missing call to cleanup

gapra-msft:


tasherif-msft (approved):

tasherif-msft:
let's also delete the files in the container that we left behind, otherwise LGTM

souravgupta-msft (approved):

gapra-msft:
> let's also delete the files in the container that we left behind, otherwise LGTM

What do you mean? Deleting the container should delete the files as well

tasherif-msft:
> > let's also delete the files in the container that we left behind, otherwise LGTM
> 
> What do you mean? Deleting the container should delete the files as well

oh yes meant to say delete the existing container

PR #787, opened 2022-05-13
[BUG] Fix directory deletes with `rm -rf` with files in dir succeeds

tasherif-msft:


souravgupta-msft (approved):

component/libfuse/libfuse2_handler.go
```diff
@@ -714,6 +714,10 @@ func libfuse2_rename(src *C.char, dst *C.char) C.int {
 	srcPath = common.NormalizeObjectName(srcPath)
 	dstPath := trimFusePath(dst)
 	dstPath = common.NormalizeObjectName(dstPath)
+	if strings.Contains(dstPath, ".fuse_hidden") {
```
vibhansa-msft, line 717: why is this file getting created. We shall stop it from creation otherwise we are uploading extra data to container which is not intended by user. There is a list of ignoreFiles in libfuse component we shall add this name/pattern there.

gapra-msft: The ignorefiles is more of a word for word matching, this is a if it starts with this pattern, ignore

tasherif-msft:
user can set `-o hard_remove` to avoid this libfuse behaviour

PR #796, opened 2022-05-26
Publish logs from nightly pipeline

gapra-msft:


tasherif-msft:
The reason we don't publish is because after enough runs we run of credits for our devops subscription, hence printing it out instead.

gapra-msft:
oh let me add a flag that defaults to not printing - we can enable it when needed

gapra-msft:
this is still useful when we need to debug

vibhansa-msft (approved):

souravgupta-msft (approved):

tasherif-msft:
just to confirm we're defaulting to `false` on publishing here, correct?

tasherif-msft:
just to confirm we're defaulting to `false` on publishing here, correct?

gapra-msft:
> just to confirm we're defaulting to `false` on publishing here, correct?

working on doing this

tasherif-msft (approved):

PR #801, opened 2022-05-31
[BUG] df showed root mount stats against blobfuse rather than file cache stats

tasherif-msft:


component/file_cache/file_cache.go
```diff
@@ -190,6 +192,7 @@ func (c *FileCache) Configure() error {
 	c.cleanupOnStart = conf.CleanupOnStart
 	c.policyTrace = conf.EnablePolicyTrace
 	c.offloadIO = conf.OffloadIO
+	c.cacheSize = conf.MaxSizeMB
```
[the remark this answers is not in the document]
tasherif-msft: yep good point

component/file_cache/file_cache.go
```diff
@@ -266,9 +269,24 @@ func (c *FileCache) OnConfigChange() {
...
+func (c *FileCache) StatFs() (*syscall.Statfs_t, bool, error) {
+	// cache_size = f_blocks * f_frsize/1024
+	// cache_size - used = f_frsize * f_bavail/1024
+	// cache_size - used = vfs.f_bfree * vfs.f_frsize / 1024
+	usage := getUsage(c.tmpPath)
+	available := (uint64(c.cacheSize*MB) - uint64(usage*MB))
+	statfs := &syscall.Statfs_t{}
+	statfs.Frsize = 4096
+	statfs.Blocks = uint64(c.cacheSize*MB) / uint64(statfs.Frsize)
+	statfs.Bavail = available / uint64(statfs.Frsize)
```
gapra-msft, line 285: should we just do this computation once? - same with cacheSize*MB?

gapra-msft (approved):
couple small comments

component/libfuse/libfuse2_handler.go
```diff
@@ -342,6 +342,28 @@ func libfuse2_getattr(path *C.char, stbuf *C.stat_t) C.int {
...
+	name = common.NormalizeObjectName(name)
+	log.Trace("Libfuse::libfuse_statfs : %s", name)
+
+	attr, populated, err := fuseFS.NextComponent().StatFs()
+	if err != nil {
+		log.Err("Libfuse::libfuse_statfs: Failed to get stats %s (%s)", name, err.Error())
+		return -C.EIO
+	}
+	C.populate_statfs(path, buf)
+	if populated {
```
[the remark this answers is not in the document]
tasherif-msft: this is just to say that we need to overwrite the root attributes, I will add a comment!

component/file_cache/file_cache.go
```diff
@@ -266,9 +270,27 @@ func (c *FileCache) OnConfigChange() {
...
 	c.offloadIO = conf.OffloadIO
+	c.maxCacheSize = conf.MaxSizeMB
 	c.policy.UpdateConfig(c.GetPolicyConfig(conf))
 }
 
+func (c *FileCache) StatFs() (*syscall.Statfs_t, bool, error) {
+	// cache_size = f_blocks * f_frsize/1024
+	// cache_size - used = f_frsize * f_bavail/1024
+	// cache_size - used = vfs.f_bfree * vfs.f_frsize / 1024
+	usage := getUsage(c.tmpPath) * MB
```
vibhansa-msft, line 281: If user has configured 0 (unlimited) cache limit then how are we calculating it ?

tasherif-msft: in that case just use root mount stats since availability will depend on it.

component/file_cache/file_cache.go
```diff
@@ -266,9 +270,31 @@ func (c *FileCache) OnConfigChange() {
...
+	// if cache size is set to 0 then we have the root mount usage
+	cacheSize := c.maxCacheSize * MB
+	if cacheSize == 0 {
+		return nil, false, nil
+	}
+	usage := getUsage(c.tmpPath) * MB
+	available := cacheSize - usage
+	statfs := &syscall.Statfs_t{}
+
+	statfs.Frsize = 4096
```
vibhansa-msft, line 290: this part is generally 4k but depends on file system so on certain system we may end up showing wrong stats. Better to get cache paths stats from system and get this value from those returned values. we are doing "/" statfs in libfuse wrapper we can do the same here in golang for cache directory (mount point of cache path) and then take this value from there.

tasherif-msft: done

gapra-msft:
Merge main and add the changelog entry for this

vibhansa-msft (approved):

souravgupta-msft (approved):

PR #807, opened 2022-06-07
Streaming small fixes and minor perf improvement

tasherif-msft:
- delay flush on appended data
- fixed nil pointer issue in pure streaming
- lock on stage+commit to avoid storage throwing on race
- added more tests

### DIFF
--- a/component/azstorage/block_blob.go
+++ b/component/azstorage/block_blob.go
@@ -983,8 +983,9 @@ func (bb *BlockBlob) stageAndCommitModifiedBlocks(name string, data []byte, offs
 
 func (bb *BlockBlob) StageAndCommit(name string, bol *common.BlockOffsetList) error {
 	// lock on the blob name so that no stage and commit race condition occur causing failure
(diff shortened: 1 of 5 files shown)


component/azstorage/block_blob.go
```diff
@@ -980,6 +982,9 @@ func (bb *BlockBlob) stageAndCommitModifiedBlocks(name string, data []byte, offs
 }
 
 func (bb *BlockBlob) StageAndCommit(name string, bol *common.BlockOffsetList) error {
+	// lock on the blob name so that no stage and commit race condition occur causing failure
+	unlock := bb.blockLocks.Lock(name)
```
vibhansa-msft, line 986: Ambiguous in name "unlock := l....lock()". Rather we can just return the mutex from the method and let the caller decide whether to lock or unlock. Will be more readable code,

gapra-msft:
Can you add an item to the changelog

souravgupta-msft (approved):

PR #809, opened 2022-06-12
Added doc generation command

tasherif-msft:


gapra-msft:
Can you show us an example of what this looks like?

gapra-msft (approved):

tasherif-msft:
> Can you show us an example of what this looks like?

You can run it locally to check it out what it outputs. This will be used primarily by the docs team.

souravgupta-msft:
Should we add a test case for this command?

gapra-msft:
> > Can you show us an example of what this looks like?
> 
> You can run it locally to check it out what it outputs. This will be used primarily by the docs team.

Just tried it out, it looks good. There is some improvement we could make on the source code end eventually that we can create an item for.

souravgupta-msft (approved):

PR #813, opened 2022-06-15
[Bug] Directory level sas fix and sas key correction

vibhansa-msft:


gapra-msft:
Should we add some tests for this?

gapra-msft:
Like add an e2e test with Dir SAS?

gapra-msft:
Let's add a CHANGELOG item for this under preview 3

gapra-msft (approved):

tasherif-msft (approved):

PR #820, opened 2022-06-20
Migrated Nightly pipeline to 1ES Hosted Pools

gapra-msft:
TODO : Debian 9


blobfuse2-nightly.yaml
```diff
@@ -48,7 +42,7 @@ parameters:
     default: false
 
 jobs:
-  # Ubuntu based test suite
+  # Ubuntu Tests
   - job: Set_1
```
vibhansa-msft, line 46: We need to add Ubn-22 here.

vibhansa-msft: For Debb-9 we can add one more set and have our self-hosted agent for now? or we want to move completely to 1ES and see it later?

gapra-msft: yes I can add that

gapra-msft: Done

vibhansa-msft (approved):

souravgupta-msft (approved):

tasherif-msft (approved):

PR #825, opened 2022-06-24
Adding fix for file append error

souravgupta-msft:
In writeback-cache mode (enabled by the FUSE_WRITEBACK_CACHE flag) writes go to
the cache only, which means that the write(2) syscall can often complete very
fast.  Dirty pages are written back implicitly (background writeback or page
reclaim on memory pressure) or explicitly (invoked by close(2), fsync(2) and
when the last ref to the file is being released on munmap(2)).  This mode
assumes that all changes to the filesystem go through the FUSE kernel module
(size and atime/ctime/mtime attributes are kept up-to-date by the kernel), so
it's generally not suitable for network filesystems.  If a partial page is
written, then the page needs to be first read from userspace.  This means, that
even for files opened for O_WRONLY it is possible that READ requests will be
generated by the kernel.

Reference: https://www.kernel.org/doc/Documentation/filesystems/fuse-io.txt


component/libfuse/libfuse_handler.go
```diff
@@ -272,11 +272,17 @@ func libfuse_init(conn *C.fuse_conn_info_t, cfg *C.fuse_config_t) (res unsafe.Po
...
-		log.Info("Libfuse::libfuse_init : Enable Capability : FUSE_CAP_WRITEBACK_CACHE")
-		conn.want |= C.FUSE_CAP_WRITEBACK_CACHE
-	}
+	/*
+		FUSE_CAP_WRITEBACK_CACHE flag is not suitable for network filesystems.  If a partial page is
+		written, then the page needs to be first read from userspace.  This means, that
+		even for files opened for O_WRONLY it is possible that READ requests will be
+		generated by the kernel. This will result error in file cache
+	*/
+	// if (conn.capable & C.FUSE_CAP_WRITEBACK_CACHE) != 0 {
```
gapra-msft, line 281: Should we keep all this code in comments? Why not just leave a comment above and remove this code. Same with file_cache code

vibhansa-msft: We are not sure will there be any perf impact due to this so wanted to preserve the code.
Keeping the code along with comments just helps you remember what you did and why you reverted, so I am kind of ok with that. Agree that over a period of time it just adds a lot of dead code but trivial things shall be kept. File-cache part we can cleanup, but lets keep this one.

vibhansa-msft (approved):

component/libfuse/libfuse_handler.go
gapra-msft: Yeah, I guess. Even if we remove this dead code (and leave the comment), worst case someone can always go back to the commit or PR in git history and see what was removed, so I was recommending to remove the code. Either way it's not a big deal

gapra-msft (approved):

tasherif-msft (approved):

PR #826, opened 2022-06-27
File-cache co-existence with streaming

vibhansa-msft:
- File-Cache and Streaming both can co-exist in pipeline
- To cache or to stream is based on these 3 criteria
    - user configures a max file size and any size beyond that size is always streamed
    - file being opened cannot be hold in remaining disk/cache space then stream
    - if user has configured max file-cache size and 90% of it is already full then stream


testdata/config/azure_key_huge.yaml
```diff
@@ -5,6 +5,7 @@ logging:
 
 components:
   - libfuse
+  - stream
```
gapra-msft, line 8: Should we also keep a file cache only test as well?

vibhansa-msft: Yes its a mix, some of the config files in testdata have stream while some dont. for e.g. key based regular test is file-cache only while sas is with stream+file-cache. Huge file tests are with stream+

PR #827, opened 2022-06-28
Files auth + testing

meganliu-msft:
no access to pipeline (nightlyblobfuse) to add fileshare variables and keys so tests might not work


vibhansa-msft (approved):

gapra-msft (commented):
In this PR, also implement any methods that should call into getAzAuth - this includes SetupPipeline and TestPipeline, that is how we validate the storage creds.

gapra-msft (approved):

PR #830, opened 2022-07-06
Update readme to point to v2

vibhansa-msft:
Updating Readme to point to blobfuse2 code base and preview binaries.


souravgupta-msft (approved):

gapra-msft (approved):
Should we add a note on how to "migrate" to v2?

PR #832, opened 2022-07-07
Code coverage threshold check

souravgupta-msft:
- Fixed code coverage pipeline for Ubuntu-20
- Check overall code coverage
- Check file level code coverage


vibhansa-msft:
As a result of this pipeline last step lists all the files that have less then 80% coverage. So, in a way we are creating a list where scope of testing is still there. Maybe we can just list them all in a work item and keep removing them as it increases.

vibhansa-msft (approved):

gapra-msft (approved):

tasherif-msft (approved):

PR #833, opened 2022-07-07
Files mvp interface methods

meganliu-msft:
GetAttr(), Exists(), ListShares(), List()

### DIFF
--- a/component/azstorage/file_share.go
+++ b/component/azstorage/file_share.go
@@ -194,6 +194,7 @@ func (fs *FileShare) CreateFile(name string, mode os.FileMode) error {
 func (fs *FileShare) CreateDirectory(name string) error {
 	return syscall.ENOTSUP
 }
+
 func (fs *FileShare) CreateLink(source string, target string) error {
 	return syscall.ENOTSUP
 }
@@ -216,16 +217,25 @@ func (fs *FileShare) RenameDirectory(string, string) error {
 func (fs *FileShare) GetAttr(name string) (attr *internal.ObjAttr, err error) {
 	log.Trace("FileShare::GetAttr : name %s", name)
 
+	// retrieve DirectoryURL of file/directory path (everything except file name)
+	// covers case where name param includes subdirectories and not just the file name
 	path := filepath.Join(fs.Config.prefixPath, name)
 	splitPath := strings.Split(path, "/")
 	splitPath = splitPath[:len(splitPath)-1]
 	joinedPath := strings.Join(splitPath, "/")
 
 	fileURL := fs.Share.NewDirectoryURL(joinedPath).NewFileURL(filepath.Base(name))
-	log.Trace("FileShare::GetAttr : getting fileeeeeeeeeeee properties for %s, %s, %s", fs.Config.prefixPath, name, fileURL)
-	prop, err := fileURL.GetProperties(context.Background())
+	prop, fileerr := fileURL.GetProperties(context.Background())
 
-	if err == nil { // file
+	if fileerr == nil { // file
+		ctime, err := time.Parse(time.RFC1123, prop.FileChangeTime())
+		if err != nil {
+			ctime = prop.LastModified()
+		}
+		crtime, err := time.Parse(time.RFC1123, prop.FileCreationTime())
+		if err != nil {
+			crtime = prop.LastModified()
+		}
 		attr = &internal.ObjAttr{
 			Path:   name, // We don't need to strip the prefixPath here since we pass the input name
 			Name:   filepath.Base(name),
@@ -233,8 +243,8 @@ func (fs *FileShare) GetAttr(name string) (attr *internal.ObjAttr, err error) {
 			Mode:   0,
 			Mtime:  prop.LastModified(),
 			Atime:  prop.LastModified(),
-			Ctime:  prop.LastModified(),
-			Crtime: prop.LastModified(),
+			Ctime:  ctime,
+			Crtime: crtime,
 			Flags:  internal.NewFileBitMap(),
 		}
 		parseMetadata(attr, prop.NewMetadata())
@@ -242,12 +252,19 @@ func (fs *FileShare) GetAttr(name string) (attr *internal.ObjAttr, err error) {
 		attr.Flags.Set(internal.PropFlagModeDefault)
 
 		return attr, nil
-	} else { // directory
+	} else if storeFileErrToErr(fileerr) == ErrFileNotFound { // directory
 		dirURL := fs.Share.NewDirectoryURL(filepath.Join(fs.Config.prefixPath, name))
-		log.Trace("FileShare::GetAttr : getting directoryyyyyyyyyyyyyy properties for %s, %s, %s", fs.Config.prefixPath, name, dirURL)
-		prop, err := dirURL.GetProperties(context.Background())
+		prop, direrr := dirURL.GetProperties(context.Background())
 
-		if err == nil {
+		if direrr == nil {
+			ctime, err := time.Parse(time.RFC1123, prop.FileChangeTime())
+			if err != nil {
+				ctime = prop.LastModified()
+			}
+			crtime, err := time.Parse(time.RFC1123, prop.FileCreationTime())
+			if err != nil {
+				crtime = prop.LastModified()
+			}
 			attr = &internal.ObjAttr{
 				Path:   name,
 				Name:   filepath.Base(name),
@@ -255,8 +272,8 @@ func (fs *FileShare) GetAttr(name string) (attr *internal.ObjAttr, err error) {
 				Mode:   0,
 				Mtime:  prop.LastModified(),
 				Atime:  prop.LastModified(),
-				Ctime:  prop.LastModified(),
-				Crtime: prop.LastModified(),
+				Ctime:  ctime,
+				Crtime: crtime,
 				Flags:  internal.NewDirBitMap(),
 			}
 			parseMetadata(attr, prop.NewMetadata())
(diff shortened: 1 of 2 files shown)


component/azstorage/file_share.go
```diff
@@ -150,86 +154,259 @@ func (fs *FileShare) TestPipeline() error {
...
+		marker = resp.NextMarker
+	}
+
+	return cntList, nil
 }
 
 // This is just for test, shall not be used otherwise
 func (fs *FileShare) SetPrefixPath(string) error {
-	return nil
+	return syscall.ENOTSUP
```
gapra-msft, line 180: Go ahead and implement this as part of your next PR - should be fairly simple, just setting a prefixPath variable

component/azstorage/file_share.go
```diff
@@ -150,86 +154,259 @@ func (fs *FileShare) TestPipeline() error {
...
+			Flags:  internal.NewFileBitMap(),
+		}
+		parseMetadata(attr, prop.NewMetadata())
+		attr.Flags.Set(internal.PropFlagMetadataRetrieved)
+		attr.Flags.Set(internal.PropFlagModeDefault)
+
+		return attr, nil
+	} else { // directory
+		dirURL := fs.Share.NewDirectoryURL(filepath.Join(fs.Config.prefixPath, name))
+		log.Trace("FileShare::GetAttr : getting directoryyyyyyyyyyyyyy properties for %s, %s, %s", fs.Config.prefixPath, name, dirURL)
```
gapra-msft, line 247: remove this log too

component/azstorage/file_share.go
```diff
@@ -150,86 +154,259 @@ func (fs *FileShare) TestPipeline() error {
...
+	log.Trace("FileShare::GetAttr : getting fileeeeeeeeeeee properties for %s, %s, %s", fs.Config.prefixPath, name, fileURL)
+	prop, err := fileURL.GetProperties(context.Background())
+
+	if err == nil { // file
+		attr = &internal.ObjAttr{
+			Path:   name, // We don't need to strip the prefixPath here since we pass the input name
+			Name:   filepath.Base(name),
+			Size:   prop.ContentLength(),
+			Mode:   0,
+			Mtime:  prop.LastModified(),
```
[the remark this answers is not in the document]
gapra-msft: same for Directory Properties

component/azstorage/file_share.go
```diff
@@ -150,86 +154,259 @@ func (fs *FileShare) TestPipeline() error {
...
+				return attr, err
+			}
+		}
+
+	}
+
 }
 
-// Standard operations to be supported by any account type
+// List : Get a list of blobs matching the given prefix
```
gapra-msft, line 281: list of files

gapra-msft: /directories

component/azstorage/file_share.go
```diff
@@ -150,86 +154,259 @@ func (fs *FileShare) TestPipeline() error {
...
+	listPath := filepath.Join(fs.Config.prefixPath, prefix)
+
+	// Get a result segment starting with the file indicated by the current Marker.
+	var listFile *azfile.ListFilesAndDirectoriesSegmentResponse
+	var err error
+	if listPath != "" {
+		listFile, err = fs.Share.NewDirectoryURL(listPath).ListFilesAndDirectoriesSegment(context.Background(), azfile.Marker{Val: marker},
+			azfile.ListFilesAndDirectoriesOptions{MaxResults: count})
+	} else {
+		if (prefix != "" && prefix[len(prefix)-1] == '/') || (prefix == "" && fs.Config.prefixPath != "") {
```
gapra-msft, line 308: is there a point to doing this check and changing listPath? I think you might be able to just use the NewDirectoryURL code above and get rid of this if else. NewDirectoryURL("") should just return whatever NewRootDirectoryURL does

component/azstorage/file_share.go
```diff
@@ -150,86 +154,259 @@ func (fs *FileShare) TestPipeline() error {
...
+			// TODO: support times when Azure SDK is updated
+			Mtime:  time.Now(),
+			Atime:  time.Now(),
+			Ctime:  time.Now(),
+			Crtime: time.Now(),
+			Flags:  internal.NewFileBitMap(),
+		}
+
+		// parseMetadata(attr, fileInfo.Metadata)
+		// attr.Flags.Set(internal.PropFlagMetadataRetrieved)
```
gapra-msft, line 337: you can probably just remove this code

component/azstorage/file_share.go
```diff
@@ -150,86 +154,259 @@ func (fs *FileShare) TestPipeline() error {
...
+			Crtime: time.Now(),
+			Flags:  internal.NewDirBitMap(),
+		}
+
+		// parseMetadata(attr, dirInfo.Metadata)
+		// attr.Flags.Set(internal.PropFlagMetadataRetrieved)
+		attr.Flags.Set(internal.PropFlagModeDefault)
+		fileList = append(fileList, attr)
+
+		if attr.IsDir() {
```
gapra-msft, line 366: you're already covering this above by setting Size to 4096 so this is redundant

component/azstorage/file_share.go
```diff
@@ -150,86 +154,259 @@ func (fs *FileShare) TestPipeline() error {
...
+	return syscall.ENOTSUP
 }
 
+// GetAttr : Retrieve attributes of a file or directory
 func (fs *FileShare) GetAttr(name string) (attr *internal.ObjAttr, err error) {
-	return nil, nil
+	log.Trace("FileShare::GetAttr : name %s", name)
+
+	path := filepath.Join(fs.Config.prefixPath, name)
+	splitPath := strings.Split(path, "/")
```
vibhansa-msft, line 220: What are we doing with this splits and join, can you explain. Better to explain the logic in comments.

component/azstorage/utils.go
```diff
@@ -354,6 +354,21 @@ func storeDatalakeErrToErr(err error) uint16 {
 	return ErrNoErr
 }
 
+// Convert datalake storage error to common errors
+func storeFileErrToErr(err error) uint16 {
+	if serr, ok := err.(azfile.StorageError); ok {
```
vibhansa-msft, line 359: There is a similar method already for blobs, may be we can extend the same to support this as well.

component/azstorage/file_share.go
```diff
@@ -150,86 +154,259 @@ func (fs *FileShare) TestPipeline() error {
...
+			Ctime:  prop.LastModified(),
+			Crtime: prop.LastModified(),
+			Flags:  internal.NewFileBitMap(),
+		}
+		parseMetadata(attr, prop.NewMetadata())
+		attr.Flags.Set(internal.PropFlagMetadataRetrieved)
+		attr.Flags.Set(internal.PropFlagModeDefault)
+
+		return attr, nil
+	} else { // directory
```
souravgupta-msft, line 245: Try to match the error here if it is a directory.
For example, let's say that we have a file and the GetProperties() returns error for the file due to some unexpected reasons. The current logic will then do the GetProperties() call for the directory which will surely return error since we have a file. So, in this case we are calling GetProperties() twice.
Instead, what we can try is to match the error for fileURL.GetProperties(). If it's a directory, we can call the GetProperties() for directory, otherwise it will go to the final else block.

component/azstorage/file_share.go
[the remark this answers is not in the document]
meganliu-msft: the file/directory response has methods for ChangeTime and CreationTime but their return value is of type string rather than time.Time (seen in zz_generated_models.go); am i looking in the wrong place though?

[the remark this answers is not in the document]
gapra-msft: maybe try to see if you can look into how to convert a string to a time.time

component/azstorage/utils.go
meganliu-msft: datalake has a separate storeDatalakeErrToErr() method; i think it'd be better to have a separate method for files too since it references azfile (and the others reference azblob and azbfs)

component/azstorage/file_share.go
gapra-msft: What do you mean by try to match the error?

component/azstorage/utils.go
```diff
@@ -354,6 +354,21 @@ func storeDatalakeErrToErr(err error) uint16 {
...
+		switch serr.ServiceCode() {
+		case azfile.ServiceCodeShareAlreadyExists:
+			return ErrFileAlreadyExists
+		case azfile.ServiceCodeShareNotFound:
```
gapra-msft, line 363: I think you need to be matching these errors with the FileAlreadyExists/DirectoryAlreadyExists, NotFound ones too

component/azstorage/file_share.go
gapra-msft: oh I think I get what you mean. @meganliu-msft check if the error is FileNotExists or something like that. Only then should you check for the Directory

component/azstorage/file_share.go
```diff
@@ -150,86 +154,253 @@ func (fs *FileShare) TestPipeline() error {
...
+			parseMetadata(attr, prop.NewMetadata())
+			attr.Flags.Set(internal.PropFlagMetadataRetrieved)
+			attr.Flags.Set(internal.PropFlagModeDefault)
+
+			return attr, nil
+		}
+		return attr, syscall.ENOENT
+	}
+	// error
+	log.Err("FileShare::GetAttr : Failed to get file/directory properties for %s (%s)", name, err.Error())
```
souravgupta-msft, line 290: I think this should be
```suggestion
	log.Err("FileShare::GetAttr : Failed to get file/directory properties for %s (%s)", name, fileerr.Error())
```

vibhansa-msft (approved):

gapra-msft (approved):

PR #835, opened 2022-07-12
Block mounted directory list calls

souravgupta-msft:
Adding code for blocking list calls for mounted directory based on block-list-on mount-sec variable set in config


vibhansa-msft (approved):

gapra-msft (approved):

tasherif-msft:
did this config param just not work before?

tasherif-msft (approved):

PR #837, opened 2022-07-14
Dont delete directory in case of gen2 account if directory is not empty

vibhansa-msft:
Commands like "git stash" try attempting a non-empty directory. On regular filesystem it fails with error "directory on empty"
In case of blobfuse, for block blob account we check for non-empty directory condition but for gen2 account we go ahead and delete the directory recursively on container as well, this results into "git stash" command deleting ".git" directory and resulting into a corrupting git repository.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,15 +1,15 @@
 cmake_minimum_required(VERSION 3.6)
 
 # Change this while bumping up the version, needed for RHEL package
-set(VERSION "1.4.4")
+set(VERSION "1.4.5-preview")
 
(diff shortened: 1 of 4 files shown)


gapra-msft (approved):

gapra-msft:
This shouldn't be an issue in v2 right? We have this check before DeleteDir https://github.com/Azure/azure-storage-fuse/blob/main/component/libfuse/libfuse_handler.go#L505

souravgupta-msft (approved):

tasherif-msft (approved):

PR #838, opened 2022-07-14
V2 : Dont delete directory based on local cache status

vibhansa-msft:
Commands like "git stash" try attempting a non-empty directory. On regular filesystem it fails with error "directory on empty"
In case of blobfuse, for block blob account we check for non-empty directory condition but for gen2 account we go ahead and delete the directory recursively on container as well, this results into "git stash" command deleting ".git" directory and resulting into a corrupting git repository.
In V2 we take decision based on local directory status. If local directory is empty then we allow deletion of directory. This is a wrong decision as local directory might be empty due to file-cache-eviction as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Fixed a bug in append file flow resolving data corruption
 - Fixed a bug in MSI auth to send correct resource string
 - Fixed a bug in OAuth token parsing when expires_on denotes numbers of seconds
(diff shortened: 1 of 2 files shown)


gapra-msft (approved):
Can you add a Changelog entry for this?

souravgupta-msft (approved):

tasherif-msft (approved):

PR #845, opened 2022-07-21
Background auth validation fix

gapra-msft:
… twice

### DIFF
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -293,13 +293,17 @@ var mountCmd = &cobra.Command{
 
 		config.Set("mount-path", options.MountPath)
 
-		log.Crit("Starting Blobfuse2 Mount : %s on (%s)", common.Blobfuse2Version, common.GetCurrentDistro())
-		log.Crit("Logging level set to : %s", logLevel.String())
-		pipeline, err := internal.NewPipeline(options.Components)
-		if err != nil {
-			log.Err("Mount: error initializing new pipeline [%v]", err)
-			fmt.Println("failed to mount :", err)
-			Destroy(1)
+		var pipeline *internal.Pipeline
+
+		if !daemon.WasReborn() {
+			log.Crit("Starting Blobfuse2 Mount : %s on (%s)", common.Blobfuse2Version, common.GetCurrentDistro())
+			log.Crit("Logging level set to : %s", logLevel.String())
+			pipeline, err = internal.NewPipeline(options.Components)
+			if err != nil {
+				log.Err("Mount: error initializing new pipeline [%v]", err)
+				fmt.Println("failed to mount :", err)
+				Destroy(1)
+			}
 		}
 
 		if !options.Foreground {
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -295,15 +295,13 @@ var mountCmd = &cobra.Command{
 
 		var pipeline *internal.Pipeline
 
-		if !daemon.WasReborn() {
-			log.Crit("Starting Blobfuse2 Mount : %s on (%s)", common.Blobfuse2Version, common.GetCurrentDistro())
-			log.Crit("Logging level set to : %s", logLevel.String())
-			pipeline, err = internal.NewPipeline(options.Components)
-			if err != nil {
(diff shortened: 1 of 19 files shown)


cmd/mount.go
```diff
@@ -293,13 +293,17 @@ var mountCmd = &cobra.Command{
...
-			log.Err("Mount: error initializing new pipeline [%v]", err)
-			fmt.Println("failed to mount :", err)
-			Destroy(1)
+		var pipeline *internal.Pipeline
+
+		if !daemon.WasReborn() {
+			log.Crit("Starting Blobfuse2 Mount : %s on (%s)", common.Blobfuse2Version, common.GetCurrentDistro())
+			log.Crit("Logging level set to : %s", logLevel.String())
+			pipeline, err = internal.NewPipeline(options.Components)
+			if err != nil {
```
vibhansa-msft, line 302: newPipeline we need to do in background as well. Only the auth part we wanted to shut down. How does this work ?

gapra-msft: Why? Doesnt the old pipeline variable get inherited?

gapra-msft: nvm youre right. Let me look into this again

internal/base_component.go
```diff
@@ -62,7 +62,7 @@ func (base *BaseComponent) SetName(name string) {
 	base.compName = name
 }
 
-func (base *BaseComponent) Configure() error {
+func (base *BaseComponent) Configure(validateCredentials bool) error {
```
vibhansa-msft, line 65: validateCredential does not make sense for all other components, so better to call this something which says "its running in background/daemon mode"

internal/pipeline.go
```diff
@@ -52,7 +52,7 @@ type NewComponent func() Component
 var registeredComponents map[string]NewComponent
 
 // NewPipeline : Using a list of strings holding name of components, create and configure the component objects
-func NewPipeline(components []string) (*Pipeline, error) {
+func NewPipeline(components []string, validateCredentials bool) (*Pipeline, error) {
```
vibhansa-msft, line 55: same here, no validate-creds just isDaemon or something like that.

component/libfuse/libfuse_handler_test_wrapper.go
```diff
@@ -77,7 +77,7 @@ func newTestLibfuse(next internal.Component, configuration string) *Libfuse {
 	config.ReadConfigFromReader(strings.NewReader(configuration))
 	libfuse := NewLibfuseComponent()
 	libfuse.SetNextComponent(next)
-	libfuse.Configure()
+	libfuse.Configure(true)
```
vibhansa-msft, line 80: Similar changes are needed for fuse2 as well, it's not building right now.

gapra-msft: good catch

internal/pipeline.go
gapra-msft: renamed to isParent

internal/base_component.go
gapra-msft: done

souravgupta-msft (approved):

internal/pipeline.go
```diff
@@ -63,7 +63,7 @@ func NewPipeline(components []string) (*Pipeline, error) {
 			comp := compInit()
 
 			// request component to parse and validate config of its interest
-			err := comp.Configure()
+			err := comp.Configure(isParent)
```
vibhansa-msft, line 66: Does isParent makes more sense in terms of name of isDaemon makes it more obvious.

vibhansa-msft (approved):

internal/pipeline.go
gapra-msft: that would make it a negative bool cause we only want to run it in the Parent code not the Daemon code

gapra-msft: Hopefully adding this comment in azstorage where we consume isParent bool helps
// The daemon runs all pipeline Configure code twice. isParent allows us to only validate credentials in parent mode, preventing a second unnecessary REST call.

PR #847, opened 2022-07-26
Module name fix

souravgupta-msft:
The name of the module in go mod is in incorrect format. This restricts us to import our code in other go code. So, changed the module name from **blobfuse2** to **github.com/Azure/azure-storage-fuse/v2.**
This needs binary to be rebuilt with different name using,
`go build -o blobfuse2`


vibhansa-msft (approved):

gapra-msft (approved):
lgtm

tasherif-msft (approved):

PR #848, opened 2022-07-26
Adding extra auth validation cases

gapra-msft:


azure-pipeline-templates/build.yml
```diff
@@ -114,6 +114,8 @@ steps:
       echo "\"block-key\"": "\"$(AZTEST_BLOCK_KEY)\"", >> $cnfFile
       echo "\"adls-key\"": "\"$(AZTEST_ADLS_KEY)\"", >> $cnfFile
       echo "\"block-sas\"": "\"$(AZTEST_BLOCK_SAS)\"", >> $cnfFile
+      echo "\"block-cont-sas-ubn-18\"": "\"$(AZTEST_BLOCK_CONT_SAS_UBN_18)\"", >> $cnfFile
+      echo "\"block-cont-sas-ubn-20\"": "\"$(AZTEST_BLOCK_CONT_SAS_UBN_20)\"", >> $cnfFile
```
vibhansa-msft, line 118: UBN 22 is also there with similar naming convention.

blobfuse2-code-coverage.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
           stressParallel: 3
         Ubuntu-20:
           imageName: 'ubuntu-20.04'
-          containerName: 'test-cnt-coverage'
+          containerName: 'test-cnt-ubn-20'
```
vibhansa-msft, line 23: This was explicitly done to avoid same container being used by both nightly and coverage.

vibhansa-msft (approved):
Left few minor comments.

blobfuse2-code-coverage.yaml
gapra-msft: the likelihood of them being run at the same time is kinda low right?

souravgupta-msft (commented):
The binary file **azure-storage-fuse** is checked in. Can you add this file to .gitignore?

souravgupta-msft (approved):

blobfuse2-code-coverage.yaml
vibhansa-msft: Code coverage for now is on demand basis but once we make it run periodically then likelihood might increase. Any specific reason you wanted to change the name here ?

gapra-msft: mostly cause it runs container sas tests now

gapra-msft: also why do we allow the ubn18 container to overlap

azure-pipeline-templates/build.yml
gapra-msft: I dont see a SAS for Block UBN22 in the Nightly Variables

tasherif-msft (approved):

blobfuse2-code-coverage.yaml
gapra-msft: it also makes managing the SASs hard since we need one per container

tasherif-msft: I think we should minimize maintaining SASs and we can just rerun in the odd case

PR #855, opened 2022-08-02
Miscellaneous Tests

gapra-msft:
Secure encrypt 
Mount with secure config
Removed some unused code in types
Encrypt, DecryptData functions
DirectoryEmpty function
Libfuse Default Permissions cases

### DIFF
--- a/blobfuse2-code-coverage.yaml
+++ b/blobfuse2-code-coverage.yaml
@@ -374,6 +374,28 @@ jobs:
         workingDirectory: $(WORK_DIR)
         displayName: "CLI : Config change simulator"
 
+      # Secure Config, fine to use insecure passphrase as this is just for testing
+      - script: |
+          rm -rf $(MOUNT_DIR)/*
+          rm -rf $(TEMP_DIR)/*
+          ./blobfuse2 unmount all
+          
+          ./blobfuse2.test -test.v -test.coverprofile=$(WORK_DIR)/mount_secure.cov secure encrypt --config-file=$(BLOBFUSE2_CFG) --output-file=$(Pipeline.Workspace)/blobfuse2.azsec --passphrase=123123123123123123123123 
+          if [ $? -ne 0 ]; then
+            exit 1
+          fi
+          ./blobfuse2.test -test.v -test.coverprofile=$(WORK_DIR)/mount_secure.cov mount $(MOUNT_DIR) --config-file=$(Pipeline.Workspace)/blobfuse2.azsec --passphrase=123123123123123123123123 &
+          sleep 10
+          ps -aux | grep blobfuse2
+          rm -rf $(MOUNT_DIR)/*
+          cd test/e2e_tests
+          go test -v -timeout=7200s ./... -args -mnt-path=$(MOUNT_DIR) -adls=true -tmp-path=$(TEMP_DIR)
+          cd -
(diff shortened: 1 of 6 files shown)


common/types.go
```diff
@@ -116,56 +116,6 @@ func (l *LogLevel) Parse(s string) error {
 	return err
 }
 
-type FileType int
```
gapra-msft, line 119: Was there a purpose for these types? They arent used anywhere

vibhansa-msft: Might be during jacobsa days were used to detect the file-type, with fuse might not be of much use.

vibhansa-msft: With libfuse these checks for filetypes are done as part of method "fillStat" which is relying on attribute structure.

blobfuse2-code-coverage.yaml
```diff
@@ -374,6 +374,28 @@ jobs:
...
+      - script: |
+          rm -rf $(MOUNT_DIR)/*
+          rm -rf $(TEMP_DIR)/*
+          ./blobfuse2 unmount all
+          
+          ./blobfuse2.test -test.v -test.coverprofile=$(WORK_DIR)/mount_secure.cov secure encrypt --config-file=$(BLOBFUSE2_CFG) --output-file=$(Pipeline.Workspace)/blobfuse2.azsec --passphrase=123123123123123123123123 
+          if [ $? -ne 0 ]; then
+            exit 1
+          fi
+          ./blobfuse2.test -test.v -test.coverprofile=$(WORK_DIR)/mount_secure.cov mount $(MOUNT_DIR) --config-file=$(Pipeline.Workspace)/blobfuse2.azsec --passphrase=123123123123123123123123 &
```
vibhansa-msft, line 387: Do not use the same coverage file name, otherwise data from last test will be lost. "/mount_secure.cov"

component/libfuse/libfuse_handler_test.go
```diff
@@ -60,7 +60,7 @@ func (suite *libfuseTestSuite) TestDefault() {
 func (suite *libfuseTestSuite) TestConfig() {
 	defer suite.cleanupTest()
 	suite.cleanupTest() // clean up the default libfuse generated
-	config := "allow-other: true\nread-only: true\nlibfuse:\n  default-permission: 0777\n  attribute-expiration-sec: 60\n  entry-expiration-sec: 60\n  negative-entry-expiration-sec: 60\n  fuse-trace: true\n"
```
vibhansa-msft, line 63: any reason to remove this default-permission thing ?

gapra-msft: I added a test below for default permission  TestConfigDefaultPermission.

I was trying to hit the different permissions cases with these 3 tests., setting default permissions with allow other is actually not useful based on our logic https://github.com/Azure/azure-storage-fuse/blob/main/component/libfuse/libfuse.go#L161

common/types.go
gapra-msft: ok, so I'm good to remove them then

blobfuse2-code-coverage.yaml
gapra-msft: oh i see. will fix

vibhansa-msft (approved):

souravgupta-msft (approved):

PR #858, opened 2022-08-05
Retry on timeout error from server

vibhansa-msft:
- Updates retry policy to retry on time-out errors from server
- Nightly restructure to run data validation tests as a seprate stage
- As git clone is taking time, merge all git related test cases into a single test case along fir tar to save time on nightly
- disable data validation test on all distros and run it only on ubn as a new stage with block and gen2 accounts with storage key auth.
- Remove ubn-18 from pipelines as devops is giving depricated warnings for that, instead use ubn-20 for fuse2 testing and ubn-22 for fuse3
- for release pipeline keep ubn-18 as fuse2 binary built on ubn-20 gives glibc version error on other distros using that .deb and .rpms.


azure-pipeline-templates/verbose-tests.yml
```diff
@@ -123,7 +123,7 @@ steps:
         adls: ${{ parameters.adls }}
         idstring: ${{ parameters.service }} with Key Credentials
         distro_name: ${{ parameters.distro_name }}
-        clone: "true"
+        # clone: true
```
gapra-msft, line 126: why keep this commented? Why not remove it?

gapra-msft (approved):

azure-pipeline-templates/verbose-tests.yml
vibhansa-msft: will cleanup.

souravgupta-msft (approved):

PR #859, opened 2022-08-09
File read write methods + some unit testing 

meganliu-msft:


component/azstorage/file_share.go
```diff
@@ -592,3 +908,44 @@ func getFileAndDirFromPath(completePath string) (fileName string, dirPath string
 
 	return fileName, dirPath
 }
+
+// calculateFileSize : calulates range size of the file based on file size
+func (fs *FileShare) calculateRangeSize(name string, fileSize int64) (rangeSize int64, err error) {
+	if fileSize > FileMaxSizeInBytes {
+		log.Err("FileShare::calculateBlockSize : buffer is too large to upload to an Azure file %s", name)
```
gapra-msft, line 915: fix the method name everywhere, keep it consistent

component/azstorage/file_share.go
```diff
@@ -355,6 +415,7 @@ func (fs *FileShare) RenameDirectory(source string, target string) error {
 			err = fs.RenameDirectory(filepath.Join(source, dirInfo.Name), filepath.Join(target, dirInfo.Name))
 			if err != nil {
 				log.Err("FileShare::RenameDirectory : Failed to move subdirectories to new directory  %s", err.Error())
+				fs.DeleteDirectory(target)
```
vibhansa-msft, line 418: If rename fails, why we want to delete targer directory. It might be the case where targer already exists and this move command is more like moving some more files in that directory.

component/azstorage/file_share.go
```diff
@@ -508,28 +569,162 @@ func (fs *FileShare) List(prefix string, marker *string, count int32) ([]*intern
...
+	return nil
 }
 
+// ReadBuffer : Downloads a file to a buffer
 func (fs *FileShare) ReadBuffer(name string, offset int64, len int64) ([]byte, error) {
-	return nil, syscall.ENOTSUP
+	log.Trace("FileShare::ReadBuffer : name %s", name)
+	var buff []byte
+
+	if offset != 0 {
```
vibhansa-msft, line 606: ReadBuffer method is suppose to read based on offset and length so checking for offset != 0 might not be correct here.

component/azstorage/file_share.go
```diff
@@ -550,29 +745,150 @@ func (fs *FileShare) WriteFromBuffer(name string, metadata map[string]string, da
...
+	}
+
+	return nil
 }
-func (fs *FileShare) StageAndCommit(name string, bol *common.BlockOffsetList) error {
-	return syscall.ENOTSUP
+
+// GetFileBlockOffsets : store file range list and corresponding offsets
+func (fs *FileShare) GetFileBlockOffsets(name string) (shareFileRangeList *common.BlockOffsetList, err error) {
+	log.Trace("FileShare::GetFileBlockOffsets : name %s", name)
```
vibhansa-msft, line 851: Do we need these methods here ? It shall only be required if we wish to support file-share over streaming.

component/azstorage/file_share.go
gapra-msft: We are checking since the SDK doesnt support setting an offset.

component/azstorage/file_share.go
meganliu-msft: I added those lines to delete the target directory because one of the unit tests failed without it. The test wanted to rename a nonexistent src directory to some target dir. Since the src dir didn't exist, it would error out and keep the target directory, which caused an assertion to fail. But you made a good point; I'll add an if statement to catch this edge case where the target dir already exists.

component/azstorage/file_share.go
gapra-msft: Should be fine to implement. There's no harm in supporting it is there?

component/azstorage/file_share.go
vibhansa-msft: if source directory does not exists then why rename will create the destination directory, sorry but dint get the context here.

component/azstorage/file_share.go
```diff
@@ -550,29 +745,150 @@ func (fs *FileShare) WriteFromBuffer(name string, metadata map[string]string, da
...
+	fileURL := fs.Share.NewDirectoryURL(dirPath).NewFileURL(fileName)
+
+	storageRangeList, err := fileURL.GetRangeList(
+		context.Background(), 0, 0)
+	if err != nil {
+		log.Err("FileShare::GetFileBlockOffsets : Failed to get range list %s ", name, err.Error())
+		return &common.BlockOffsetList{}, err
+	}
+
+	if len(rangeList.BlockList) == 0 {
```
gapra-msft, line 864: This should be storageRangeList.Ranges

component/azstorage/file_share.go
vibhansa-msft: that is true even in case of block blob as well, either we upload entire file or it overwrites from offset 0, just check what we have done in there and let's keep it symmetric to that.

component/azstorage/file_share.go
vibhansa-msft: There is no harm as such but if read/write streaming is not possible for files then we can just have dummy implementation.

gapra-msft: I think writes streaming is actually more suited to file shares since they have very convenient Upload APIs

component/azstorage/file_share.go
meganliu-msft: If the source directory doesn't exist, the rename will create the destination directory, because in the current code for this method, the second line in the method (after the first line of logging) is to create the target directory. I put it there because it's a recursive method, so it's handy for the subdirectories. 
But to address the cases where the source doesn't exist and the target already exists, I'll put an if statement before calling CreateDirectory() so that it won't create the destination directory unnecessarily (and delete the DeleteDirectory() calls). The if-statement would check if GetAttr() on the source directory has no errors to validate whether it's a valid directory or not.

meganliu-msft: I meant GetProperties() rather than GetAttr()

component/azstorage/file_share.go
```diff
@@ -508,28 +573,162 @@ func (fs *FileShare) List(prefix string, marker *string, count int32) ([]*intern
...
+	return buff, nil
 }
 
+// ReadInBuffer : Download specific range from a file to a user provided buffer
 func (fs *FileShare) ReadInBuffer(name string, offset int64, len int64, data []byte) error {
-	return syscall.ENOTSUP
+	log.Trace("FileShare::ReadInBuffer : name %s", name)
+
+	if offset != 0 {
+		log.Err("FileShare::ReadToFile : offset is not 0")
```
tasherif-msft, line 651: This is the `ReadInBuffer` method not `ReadToFile`. Also, ReadInBuffer supports offsets that are not 0. what's the reason for this check?

component/azstorage/file_share.go
tasherif-msft: what was the conclusion of the conversation regarding file creating and pre-specifying the size of the file? the setting we set on the top just dictates the max size of the file, but the file can be of size 0 and we can keep adding to it, correct?

gapra-msft: we're going to have to resize when we append basically.

component/azstorage/file_share.go
gapra-msft: Blob takes in an offset so we pass it

component/azstorage/file_share.go
meganliu-msft: Oops I'll fix that. 
ReadInBuffer() is supposed to support offsets that are not 0 but the SDK doesn't. ReadInBuffer() calls DownloadAzureFileToBuffer() in the SDK, which doesn't take offset as one of the arguments (or as part of any of the arguments).

component/azstorage/file_share.go
```diff
@@ -266,6 +294,36 @@ func (fs *FileShare) DeleteDirectory(name string) (err error) {
 	log.Trace("FileShare::DeleteDirectory : name %s", name)
 
 	dirURL := fs.Share.NewDirectoryURL(filepath.Join(fs.Config.prefixPath, name))
+
```
gapra-msft, line 297: Does DeleteDirectory work recursively? Does it give you an error if the directory is not empty?

component/azstorage/file_share.go
```diff
@@ -329,6 +387,20 @@ func (fs *FileShare) RenameFile(source string, target string) error {
 func (fs *FileShare) RenameDirectory(source string, target string) error {
```
gapra-msft, line 387: To get the rename code, in our repo run `go get github.com/Azure/azure-storage-file-go/azfile@f37a99d62e3f9ed290ff5c7b20cb184bab71367d `

vibhansa-msft (approved):

component/azstorage/file_share.go
meganliu-msft: Yes, DeleteDirectory works recursively. No, it doesn't give an error if the directory is not empty because for any directory it deletes, it'll delete all the files and recursively delete any subdirectories.

gapra-msft: I think I meant if the file share SDK API works recursively. I'm wondering if we can replace all this with one call to the SDK basically

meganliu-msft: Oh gotcha. Yes it does! I deleted all the recursive stuff and just called the SDK in the most recent commits

component/file_cache/file_cache.go
```diff
@@ -780,7 +780,7 @@ func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Hand
 		}
 
 		// Open the file in write mode.
-		f, err = os.OpenFile(localPath, os.O_CREATE|os.O_WRONLY, options.Mode)
+		f, err = os.OpenFile(localPath, os.O_CREATE|os.O_RDWR, options.Mode) // need these permissions for file share to work
```
gapra-msft, line 783: @vibhansa-msft Is this safe? The file share SDK expects Read access to the file as well in the WriteToFile method

component/azstorage/file_share.go
meganliu-msft: *Sorry got this mixed up with RenameDirectory(). No, the SDK doesn't work recursively; the directory must be empty before being deleted by the SDK call.

component/azstorage/file_share.go
```diff
@@ -593,3 +875,41 @@ func getFileAndDirFromPath(completePath string) (fileName string, dirPath string
...
+			if (rangeSize & (-8)) != 0 {
+				// EXTRA : round off the range size to next higher multiple of 8.
+				// No reason to do so; assuming odd numbers in range size will not be good on server end
+				rangeSize = (rangeSize + 7) & (-8)
+			}
+
+			if rangeSize > azfile.FileMaxUploadRangeBytes {
+				// After rounding off the rangeSize has become bigger then max allowed range size.
+				log.Err("FileShare::calculateRangeSize : rangeSize exceeds max allowed range size for %s", name)
+				err = errors.New("ragnge size is too large to upload to a file")
```
gapra-msft, line 907:
```suggestion
				err = errors.New("range size is too large to upload to a file")
```

component/azstorage/file_share_test.go
```diff
@@ -128,6 +133,734 @@ func (s *fileTestSuite) cleanupTest() {
 	log.Destroy()
 }
 
+// func (s *fileTestSuite) TestCleanupShares() {
```
gapra-msft, line 136: This method is to clean up any shares in case the account has some - due to cancelling a test early or something during debug mode

gapra-msft (approved):

component/file_cache/file_cache.go
vibhansa-msft: This is ok, when we are downloading the file we can keep this permissions. Anyway this handle will be closed as soon as download is complete and then we will open another handle with permissions user requested.

tasherif-msft (approved):

souravgupta-msft (approved):

PR #860, opened 2022-08-09
Added more storage tests

gapra-msft:


vibhansa-msft (approved):

component/azstorage/block_blob.go
```diff
@@ -507,9 +507,10 @@ func (bb *BlockBlob) List(prefix string, marker *string, count int32) ([]*intern
 		if !dirList[blobInfo.Name] {
 			//log.Info("BlockBlob::List : meta file does not exists for dir %s", blobInfo.Name)
 			// For these dirs we get only the name and no other properties so hardcoding time to current time
+			name := strings.TrimSuffix(blobInfo.Name, "/")
 			attr := &internal.ObjAttr{
-				Path:  split(bb.Config.prefixPath, blobInfo.Name),
-				Name:  filepath.Base(blobInfo.Name),
+				Path:  split(bb.Config.prefixPath, name),
```
vibhansa-msft, line 512: This is a working code, do we have some bug here ?

gapra-msft: I think we expect names to not have a trailing / so I am trimming it. Storage returns these dir prefix names with a / at the end.

gapra-msft: This is to make the test case TestReadDirNoVirtualDirectory get the expected result

component/azstorage/azstorage.go
```diff
@@ -429,6 +430,10 @@ func (az *AzStorage) Chown(options internal.ChownOptions) error {
 
 func (az *AzStorage) FlushFile(options internal.FlushFileOptions) error {
 	log.Trace("AzStorage::FlushFile : Flush file %s", options.Handle.Path)
+	if options.Handle.CacheObj.BlockOffsetList.SmallFile() {
```
gapra-msft, line 433: confirmed with Tamer that this is not a valid path to hit

vibhansa-msft (approved):

tasherif-msft (approved):

PR #861, opened 2022-08-10
Streaming small fixes + more tests

tasherif-msft:
added e2e tests for streaming
invalidate attr cache on flush
don't flush on close, instead flush when flush is called


component/attr_cache/attr_cache.go
```diff
@@ -505,6 +505,19 @@ func (ac *AttrCache) CreateLink(options internal.CreateLinkOptions) error {
...
 
+// FlushFile : flush file
+func (ac *AttrCache) FlushFile(options internal.FlushFileOptions) error {
+	log.Trace("AttrCache::FlushFile : %s", options.Handle.Path)
+	err := ac.NextComponent().FlushFile(options)
+	if err == nil {
+		ac.cacheLock.RLock()
+		defer ac.cacheLock.RUnlock()
+
+		ac.invalidatePath(options.Handle.Path)
```
gapra-msft, line 516: are we invalidating here since the Size can change?

component/stream/read.go
```diff
@@ -184,9 +184,8 @@ func (r *ReadCache) ReadInBuffer(options internal.ReadInBufferOptions) (int, err
 	return r.copyCachedBlock(options.Handle, options.Offset, options.Data)
 }
 
-func (r *ReadCache) CloseFile(options internal.CloseFileOptions) error {
```
gapra-msft, line 187: Shouldnt we still implement CloseFile?

component/stream/read_write.go
```diff
@@ -208,19 +209,21 @@ func (rw *ReadWriteCache) RenameFile(options internal.RenameFileOptions) error {
...
 }
 
-func (rw *ReadWriteCache) CloseFile(options internal.CloseFileOptions) error {
-	log.Trace("Stream::CloseFile : name=%s, handle=%d", options.Handle.Path, options.Handle.ID)
+func (rw *ReadWriteCache) FlushFile(options internal.FlushFileOptions) error {
+	// log.Trace("Stream::FlushFile : name=%s, handle=%d", options.Handle.Path, options.Handle.ID)
 	if !rw.StreamOnly && !options.Handle.CacheObj.StreamOnly {
 		err := rw.purge(options.Handle, -1, true)
 		if err != nil {
 			log.Err("Stream::CloseFile : failed to flush and purge handle cache %s [%s]", options.Handle.Path, err.Error())
```
gapra-msft, line 217: FlushFile

component/attr_cache/attr_cache.go
tasherif-msft: yea

component/attr_cache/attr_cache.go
```diff
@@ -505,6 +505,19 @@ func (ac *AttrCache) CreateLink(options internal.CreateLinkOptions) error {
 	return err
 }
 
+// FlushFile : flush file
+func (ac *AttrCache) FlushFile(options internal.FlushFileOptions) error {
+	log.Trace("AttrCache::FlushFile : %s", options.Handle.Path)
```
vibhansa-msft, line 510: In file-cache case, flushFile was not being forwarded as flush was resulting into uploadFromFile and that was taking care of invalidating data. Are we invalidating here because stream layer forwards the flushFile call ?
Also, if so we shall have code in streaming layer that if close is called without  a flush call and fir handle is dirty then implicitly call flush on stream layer (same is done in file-cache as well)

component/stream/read_write.go
```diff
@@ -208,12 +210,23 @@ func (rw *ReadWriteCache) RenameFile(options internal.RenameFileOptions) error {
 	return nil
 }
 
+func (rw *ReadWriteCache) FlushFile(options internal.FlushFileOptions) error {
+	// log.Trace("Stream::FlushFile : name=%s, handle=%d", options.Handle.Path, options.Handle.ID)
+	err := rw.NextComponent().FlushFile(options)
+	if err != nil {
+		log.Err("Stream::FlushFile : error flushing file %s [%s]", options.Handle.Path, err.Error())
```
vibhansa-msft, line 217: If error was there in flush then we shall not clean up the dirty flag, and on close we shall reattempt flush.

component/stream/read_write.go
```diff
@@ -326,15 +339,15 @@ func (rw *ReadWriteCache) Stop() error {
...
+func (rw *ReadWriteCache) purge(handle *handlemap.Handle, size int64) error {
 	handle.CacheObj.Lock()
 	defer handle.CacheObj.Unlock()
-	if flush {
-		err := rw.NextComponent().FlushFile(internal.FlushFileOptions{Handle: handle})
-		if err != nil {
-			return err
-		}
-	}
+	// if flush {
```
vibhansa-msft, line 345: As this flag itself is removed from function call, we can remove this dead code.

test/e2e_tests/dir_test.go
```diff
@@ -394,27 +394,29 @@ func (suite *dirTestSuite) TestDirRenameFull() {
 }
 
 func (suite *dirTestSuite) TestTarDir() {
-	dirName := suite.testPath + "/tar"
-	tarName := suite.testPath + "/tardir.tar.gz"
+	if clonePtr == "true" || clonePtr == "True" {
```
vibhansa-msft, line 397: TarDir test case works with streaming ?

component/attr_cache/attr_cache.go
tasherif-msft: yes, we are invalidating here because stream layer forwards to flush call. How can we repro a case of close without flush? I was not able to do that. For example, in Python if you close a handle flush is always called first on the kernel, even if we just read without writing.
If no flush call comes then we should not flush.

component/stream/read_write.go
tasherif-msft: that sounds like strange behavior, is there a reason we follow such an approach? retry logic should be handled by storage, it seems odd to have a retry on close.

test/e2e_tests/dir_test.go
tasherif-msft: TarDir test has a git clone operation, therefore we should add the clonePtr check since stream does not support git clone.

component/stream/read_write.go
tasherif-msft: sure

azure-pipeline-templates/verbose-tests.yml
```diff
@@ -76,6 +80,23 @@ steps:
   - script: cat ${{ parameters.config }}
     displayName: Print config file
 
+# Stream e2e
```
gapra-msft, line 83: Should we also make a test case for read only config? Maybe one where we upload a bunch of files (with python SDK or something) then read all those files and validate their checksum or something?

component/stream/read_write_test.go
```diff
@@ -286,6 +347,20 @@ func (suite *streamTestSuite) TestStreamOnly() {
 	assertNumberOfCachedFileBlocks(suite, 0, handle)
 	// confirm new handle is stream only since limit is exceeded
 	assertHandleStreamOnly(suite, handle)
+
+	suite.mock.EXPECT().OpenFile(openFileOptions).Return(handle, syscall.ENOENT)
+	_, err := suite.stream.OpenFile(openFileOptions)
+	suite.assert.NotEqual(nil, err)
+
+	// append new block and confirm old gets evicted
```
gapra-msft, line 355: this comment doesnt make sense to me, this is just testing error case of WriteFile right?

gapra-msft (approved):

component/attr_cache/attr_cache.go
vibhansa-msft: yes, application can do a close without flush however implicitly its converted to flush and close. Just to ensure on some system in case we directly get a close, in file-cache we have handled this explicitly. I remember there was a test case where we were able to simulate this. If you do a scp from outside to directly a mounted directory, I guess. Not able to recollect the exact case but there was one condition in which we were able to hit this, and it was fixed in v1 as well. So better to have that check in streaming as well.

component/stream/read_write.go
vibhansa-msft: in case of any timeout or any other error if flush fails this just gives us a second retry at flush during close. So in flush case if it fails we dont clear the flag and in close we always check if file is still dirty then first call flush and then do close handling. You can refer file-cache close code for this.

test/e2e_tests/dir_test.go
vibhansa-msft: I have moved all these cases into single git test case for optimization. Once my pr is merged this all will be fixed.

component/attr_cache/attr_cache.go
tasherif-msft: done.

component/stream/read_write.go
tasherif-msft: done.

azure-pipeline-templates/verbose-tests.yml
tasherif-msft: we can do that in a separate pr

component/stream/read.go
```diff
@@ -201,6 +204,16 @@ func (r *ReadCache) WriteFile(options internal.WriteFileOptions) (int, error) {
 	return 0, syscall.ENOTSUP
 }
 
+func (r *ReadCache) FlushFile(options internal.FlushFileOptions) error {
+	// log.Trace("Stream::FlushFile : name=%s, handle=%d", options.Handle.Path, options.Handle.ID)
+	err := r.NextComponent().FlushFile(options)
```
vibhansa-msft, line 209: In ReadCache we can skip calling flush for next component, as file-system is read-only so there will not be any change of contents. We can just return back nil here.

component/stream/read_write.go
```diff
@@ -146,6 +145,7 @@ func (rw *ReadWriteCache) WriteFile(options internal.WriteFileOptions) (int, err
 	if err != nil {
 		log.Err("Stream::WriteFile : error failed to write data to %s: [%s]", options.Handle.Path, err.Error())
 	}
+	options.Handle.Flags.Set(handlemap.HandleFlagDirty)
```
vibhansa-msft, line 148: Why in open file we are setting the handle dirty unconditionally ?

vibhansa-msft: I guess this shall be done in CreateFile rather.

component/stream/read_write.go
```diff
@@ -173,9 +173,8 @@ func (rw *ReadWriteCache) TruncateFile(options internal.TruncateFileOptions) err
 	err := rw.NextComponent().TruncateFile(options)
```
vibhansa-msft, line 173: In TruncateFile we are changing the file on container, shouldn't we invalidate our local caches here ?

component/stream/read_write.go
```diff
@@ -203,25 +202,41 @@ func (rw *ReadWriteCache) RenameFile(options internal.RenameFileOptions) error {
 	err := rw.NextComponent().RenameFile(options)
 	if err != nil {
 		log.Err("Stream::RenameFile : error renaming file %s [%s]", options.Src, err.Error())
+	}
+	return err
+}
+
+func (rw *ReadWriteCache) FlushFile(options internal.FlushFileOptions) error {
+	// log.Trace("Stream::FlushFile : name=%s, handle=%d", options.Handle.Path, options.Handle.ID)
```
vibhansa-msft, line 210: FlushFile shall work only on dirty file handles, for clean handles we can just return back success.

component/stream/read_write.go
```diff
@@ -203,25 +202,41 @@ func (rw *ReadWriteCache) RenameFile(options internal.RenameFileOptions) error {
...
+
+	options.Handle.Flags.Clear(handlemap.HandleFlagDirty)
 	return nil
 }
 
 func (rw *ReadWriteCache) CloseFile(options internal.CloseFileOptions) error {
-	log.Trace("Stream::CloseFile : name=%s, handle=%d", options.Handle.Path, options.Handle.ID)
+	// log.Trace("Stream::CloseFile : name=%s, handle=%d", options.Handle.Path, options.Handle.ID)
+	// try to flush again to make sure it's cleaned up
+	err := rw.FlushFile(internal.FlushFileOptions(options))
```
vibhansa-msft, line 224: Either put a dirty flag check here or inside flushfile.

component/stream/read_write.go
```diff
@@ -203,25 +202,41 @@ func (rw *ReadWriteCache) RenameFile(options internal.RenameFileOptions) error {
 	err := rw.NextComponent().RenameFile(options)
 	if err != nil {
 		log.Err("Stream::RenameFile : error renaming file %s [%s]", options.Src, err.Error())
+	}
+	return err
+}
+
+func (rw *ReadWriteCache) FlushFile(options internal.FlushFileOptions) error {
```
vibhansa-msft, line 209: For direct streaming, data is already uploaded so we can just return back success here in that case.

component/stream/read_write.go
tasherif-msft: I thought we had agreed no modifying other handles on read_write?

component/stream/read_write.go
tasherif-msft: storage component will just not do anything since all blocks are clean. However, I guess it would save time for the extra operations

vibhansa-msft (approved):

souravgupta-msft (approved):

PR #863, opened 2022-08-12
MD5 Sum updation on upload and validation on download

vibhansa-msft:
- Azstorage config to enable/disable this feature
- compute md5 locally and set it while uploading
- compute md5 locally post download and validate it with container, if no match fail the file download.


setup/baseConfig.yaml
```diff
@@ -105,6 +105,8 @@ azstorage:
   sdk-trace: true|false <enable storage sdk logging>
   fail-unsupported-op: true|false <for block blob account return failure for unsupported operations like chmod and chown>
   auth-resource: <resource string to be used during OAuth token retrieval>
+  update-md5: true|false <set md5 sum on upload. Impacts performance>
```
gapra-msft, line 108: This is only possible with file cache right? Should we add that as a note here?

gapra-msft (approved):

setup/baseConfig.yaml
vibhansa-msft: True, will add comment here.

component/azstorage/block_blob.go
```diff
@@ -575,6 +577,31 @@ func (bb *BlockBlob) ReadToFile(name string, offset int64, count int64, fi *os.F
...
+			prop, err := blobURL.GetProperties(context.Background(), bb.blobAccCond, bb.blobCPKOpt)
+			if err != nil {
+				log.Warn("BlockBlob::ReadToFile : Failed to get properties of blob %s (%s)", name, err.Error())
+			} else {
+				blobMD5 := prop.ContentMD5()
+				if blobMD5 == nil {
+					log.Warn("BlockBlob::ReadToFile : Failed to get MD5 Sum for blob %s", name)
+				} else {
+					// compare md5 and fail is not match
+					if !reflect.DeepEqual(fileMD5, prop.ContentMD5()) {
```
souravgupta-msft, line 596: Reuse the variable blobMD5 created above
```suggestion
					if !reflect.DeepEqual(fileMD5, blobMD5) {
```

souravgupta-msft (approved):

tasherif-msft (approved):

PR #865, opened 2022-08-16
Added Azurite Test 

gapra-msft:


azure-pipeline-templates/verbose-tests.yml
```diff
@@ -70,6 +76,7 @@ steps:
       NIGHTLY_STO_ACC_KEY: ${{ parameters.account_key }}
       ACCOUNT_ENDPOINT: ${{ parameters.account_endpoint }}
       VERBOSE_LOG: ${{ parameters.verbose_log }}
+      USE_HTTP: false
```
vibhansa-msft, line 79: most places we are passing it down as "false" only then we can define with some default value and pass it only when true is required.

azure-pipeline-templates/verbose-tests.yml
```diff
@@ -112,6 +119,25 @@ steps:
...
 
+  # Create azurite config file if we need to test it
+  - script: |
+      cd ${{ parameters.working_dir }}
+      ${{ parameters.working_dir }}/blobfuse2 gen-test-config --config-file=azure_key.yaml --container-name=${{ parameters.container }} --temp-path=${{ parameters.temp_dir }} --output-file=${{ parameters.azurite_config }}
+    displayName: Create Key Config File
+    # Note: It is fine to hardcode these credentials since they are publicly documented fake credentials for Azurite
+    env:
+      ACCOUNT_TYPE: ${{ parameters.account_type }}
+      NIGHTLY_STO_ACC_NAME: 'devstoreaccount1'
```
vibhansa-msft, line 130: lets not checkin this part where account name and key are mentioned in plain text, we can define new parameters in lib and use them instead.

vibhansa-msft: oh ok just saw the comment above now,

azure-pipeline-templates/verbose-tests.yml
```diff
@@ -112,6 +119,25 @@ steps:
...
+      NIGHTLY_STO_ACC_KEY: 'Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=='
+      ACCOUNT_ENDPOINT: 'http://127.0.0.1:10000/devstoreaccount1'
```
vibhansa-msft, line 132: we need to run azurite service on the vm, where are we doing that ?

azure-pipeline-templates/verbose-tests.yml
gapra-msft: how do you make env as a default?

azure-pipeline-templates/verbose-tests.yml
gapra-msft: yeah its fine, Azurite also documents them publicly

azure-pipeline-templates/verbose-tests.yml
```diff
@@ -174,6 +200,34 @@ steps:
           displayName: 'E2E Test: Mount with SPN Credential Configuration'
           timeoutInMinutes: 3
           continueOnError: false
+
+  - ${{ if eq(parameters.test_azurite, true) }}:
+    - bash: |
+        sudo apt-get install azure-cli -y
+        sudo npm install -g azurite
+        sudo mkdir azurite
+        sudo azurite --silent --location azurite --debug azurite\debug.log &
```
gapra-msft, line 209: running azurite here

azure-pipeline-templates/verbose-tests.yml
vibhansa-msft: but we are sending everywhere false only right ? then why to send at all

azure-pipeline-templates/verbose-tests.yml
```diff
@@ -112,6 +119,25 @@ steps:
...
+      cd ${{ parameters.working_dir }}
+      ${{ parameters.working_dir }}/blobfuse2 gen-test-config --config-file=azure_key.yaml --container-name=${{ parameters.container }} --temp-path=${{ parameters.temp_dir }} --output-file=${{ parameters.azurite_config }}
+    displayName: Create Key Config File
+    # Note: It is fine to hardcode these credentials since they are publicly documented fake credentials for Azurite
+    env:
+      ACCOUNT_TYPE: ${{ parameters.account_type }}
+      NIGHTLY_STO_ACC_NAME: 'devstoreaccount1'
+      NIGHTLY_STO_ACC_KEY: 'Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw=='
+      ACCOUNT_ENDPOINT: 'http://127.0.0.1:10000/devstoreaccount1'
+      USE_HTTP: 'true'
```
vibhansa-msft, line 133: ok I see its this place where we send as true. does that mean azurite test is only over http and not https ?

azure-pipeline-templates/verbose-tests.yml
vibhansa-msft, line 132: So how does this work, it's like all REST calls will go to this endpoint (azurite endpoint mostly) instead of standard paths ?

vibhansa-msft (approved):

azure-pipeline-templates/verbose-tests.yml
gapra-msft: yeah. They have support for https, but you need to turn some other flags on and pass in a ca cert and stuff, so it seemed not necessary

vibhansa-msft (approved):

souravgupta-msft (approved):

PR #867, opened 2022-08-17
Upgrade version to preview 3

gapra-msft:


common/types.go
```diff
@@ -48,7 +48,7 @@ import (
 const (
 	DefaultMaxLogFileSize = 512
 	DefaultLogFileCount   = 10
-	Blobfuse2Version      = "2.0.0-preview.2"
+	Blobfuse2Version      = "2.0.0-preview.3"
```
vibhansa-msft, line 51: updating to preview.3 will still have issues. Those who build locally or get a private build from us vs those who took official preview.3 release, there is no way to differentiate between. In v1 case we used to put preview in version string before the release and at release time we just remove preview string, we might need to do something different here as official releases themselves are preview.

souravgupta-msft (approved):

common/types.go
gapra-msft: I guess that's true but its better to step on an unreleased preview version than a released one right?

gapra-msft: This is how we do it for the SDKs too. If we give binaries to a customer we just add something to the version string describing the feature. If the customer builds on their own, they are on their own

tasherif-msft (approved):

common/types.go
vibhansa-msft: I just feel there shall be some way to know its private build / pre-release build vs an official preview release that we have done.

vibhansa-msft (approved):

PR #868, opened 2022-08-17
Health monitor

souravgupta-msft:


.gitignore
```diff
@@ -15,8 +15,8 @@ __debug_bin
...
 cpplite/
 build/
-tools/
 test/manual_scripts/create1000.go
 test/manual_scripts/cachetest.go
 lint.log
-azure-storage-fuse
\ No newline at end of file
+azure-storage-fuse
+healthmon
```
vibhansa-msft, line 22: We need some cool name for this binary, healthmon just sounds too boring.

azure-pipeline-templates/build-release.yml
```diff
@@ -44,6 +44,22 @@ steps:
...
+    inputs:
+      command: 'build'
+      arguments: "-o healthmon ./tools/health-monitor/"
+      workingDirectory: ${{ parameters.work_dir }}
+    displayName: 'Building health-monitor binary'
+  
+  # Verifying whether built binary is correct
+  - script: |
+      sudo chmod +x ./healthmon
+      ./healthmon --version
```
vibhansa-msft, line 58: We need to package this binary as well by copying it to respective folder. Don't see that step in release pipeline.

azure-pipeline-templates/build-release.yml
```diff
@@ -44,6 +44,22 @@ steps:
     displayName: "Test Binary built"
     continueOnError: false
 
+  # Build the health-monitor binary  
+  - task: Go@0
+    inputs:
+      command: 'build'
+      arguments: "-o healthmon ./tools/health-monitor/"
```
vibhansa-msft, line 51: we can instead use ./build.sh script for both build steps so that later if we need any other options its just a change in the script rather than pipelines

cmd/health-monitor.go
```diff
@@ -0,0 +1,179 @@
...
+	var cliParams []string
+
+	cliParams = append(cliParams, "--pid="+pid)
+	if hmonOptions.BfsPollInterval != 0 {
+		cliParams = append(cliParams, fmt.Sprintf("--blobfuse2-poll-interval=%v", hmonOptions.BfsPollInterval))
+	}
+	if hmonOptions.StatsPollinterval != 0 {
+		cliParams = append(cliParams, fmt.Sprintf("--stats-poll-interval=%v", hmonOptions.StatsPollinterval))
+	}
+	cliParams = append(cliParams, "--cache-path="+cacheMonitorOptions.TmpPath)
```
vibhansa-msft, line 147: Do we need to sanitize these paths. It may happen that blobfuse failed on mount and statsmon just still kicks in.

cmd/health-monitor.go
```diff
@@ -0,0 +1,179 @@
...
+
+var healthMonCmd = &cobra.Command{
+	Use:               "health-monitor",
+	Short:             "Monitor blobfuse2 mount",
+	Long:              "Monitor blobfuse2 mount",
+	SuggestFor:        []string{"healthmon", "monitor health"},
+	Args:              cobra.ExactArgs(0),
+	Hidden:            true,
+	FlagErrorHandling: cobra.ExitOnError,
+	RunE: func(cmd *cobra.Command, args []string) error {
```
vibhansa-msft, line 72: How do we turn down healthmon when blobfuse unmounts. As we start it on mount it shall automatically be turn down as well when we unmount blobfuse. May be send a signal to it on shutdown to take it down gracefully. To send signal we might use the same channel as well saying "I am going down"

component/azstorage/azstorage.go
```diff
@@ -63,6 +63,8 @@ const compName = "azstorage"
 //Verification to check satisfaction criteria with Component Interface
 var _ internal.Component = &AzStorage{}
 
+var AzStatsCollector *internal.StatsCollector
```
vibhansa-msft, line 66: Do we need this to be public and a global var, it can be part of the azstorage object structure.

component/azstorage/azstorage.go
```diff
@@ -162,12 +164,17 @@ func (az *AzStorage) Start(ctx context.Context) error {
 	// On mount block the ListBlob call for certain amount of time
 	az.startTime = time.Now()
 	az.listBlocked = true
+
+	// create stats collector for azstorage
+	AzStatsCollector = internal.NewStatsCollector(az.Name(), nil)
```
vibhansa-msft, line 169: we shall put an error log just in case this was nil

component/azstorage/azstorage.go
```diff
@@ -276,6 +295,8 @@ func (az *AzStorage) StreamDir(options internal.StreamDirOptions) ([]*internal.O
 
 	log.Debug("AzStorage::StreamDir : Retrieved %d objects with %s marker for Path %s", len(new_list), options.Token, path)
 
+	AzStatsCollector.AddStats(az.Name(), "StreamDir", path, true, map[string]interface{}{"Count": len(new_list)})
```
vibhansa-msft, line 298: for streamdir we shall also put the name of the directory being listed.

component/azstorage/azstorage.go
```diff
@@ -335,13 +363,24 @@ func (az *AzStorage) CloseFile(options internal.CloseFileOptions) error {
...
 }
 
 func (az *AzStorage) RenameFile(options internal.RenameFileOptions) error {
 	log.Trace("AzStorage::RenameFile : %s to %s", options.Src, options.Dst)
 
-	return az.storage.RenameFile(options.Src, options.Dst)
+	err := az.storage.RenameFile(options.Src, options.Dst)
+
+	if err == nil {
+		AzStatsCollector.AddStats(az.Name(), "RenameFile", options.Src, true, map[string]interface{}{"Dest": options.Dst})
```
vibhansa-msft, line 381: give both source and destination name for rename in event

component/azstorage/azstorage.go
```diff
@@ -386,7 +425,12 @@ func (az *AzStorage) GetFileBlockOffsets(options internal.GetFileBlockOffsetsOpt
 
 func (az *AzStorage) TruncateFile(options internal.TruncateFileOptions) error {
 	log.Trace("AzStorage::TruncateFile : %s to %d bytes", options.Name, options.Size)
-	return az.storage.TruncateFile(options.Name, options.Size)
+	err := az.storage.TruncateFile(options.Name, options.Size)
+
+	if err == nil {
+		AzStatsCollector.AddStats(az.Name(), "TruncateFile", options.Name, true, map[string]interface{}{"Size": options.Size})
```
vibhansa-msft, line 431: name fo the file being truncated as part of event

component/azstorage/block_blob.go
```diff
@@ -649,6 +674,16 @@ func (bb *BlockBlob) calculateBlockSize(name string, fileSize int64) (blockSize
 	return blockSize, nil
 }
 
+func trackUpload(name string, bytesTransferred int64, count int64, uploadPtr *int64) {
+	if bytesTransferred >= (*uploadPtr)*100*1024*1024 || bytesTransferred == count {
+		(*uploadPtr)++
+		log.Debug("BlockBlob::trackUpload : Upload: Blob = %v, Bytes transferred = %v, Size = %v", name, bytesTransferred, count)
+
+		// send upload progress as event
+		AzStatsCollector.AddStats(compName, "UploadProgress", name, true, map[string]interface{}{"Bytes Transferred": bytesTransferred, "Size": count})
```
vibhansa-msft, line 683: We need to check how much perf hit we get due to these new stats and upload/download tracking

setup/baseConfig.yaml
```diff
@@ -114,3 +114,16 @@ mountall:
     - <list of containers to be mounted>
   container-denylist:
     - <list of containers not to be mounted>
+
+# health-monitor configuration
+health-monitor:
+  enable-monitoring: true|false <enable health-monitor>
+  blobfuse2-poll-interval: <Blobfuse2 stats polling interval (in sec). Default - 5 sec>
+  stats-poll-interval: <CPU, memory and network usage polling interval (in sec). Default - 10 sec>
```
vibhansa-msft, line 122: this name sounds incorrect, its not the stats polling rather process level health polling, stats sounds more like the first one. May be "blobfuse2-poll-interval" shall be called "stats-poll-interval" and "stats-poll-interval" shall be "process-moniitor-interval" [Suggestion]

vibhansa-msft:
In the demo part there were two enable config one in common section and one under healthmonitor. but the base config mentions one. If there are two required, then its redundancy and we shall remove it.

vibhansa-msft:
As per demo, memory usage output just gives some number but no suffix like MB/GB etc. It shall be added

vibhansa-msft:
In upload/download progress tracking it mentions only bytes transferred so far, if we can show the % as well it will be easier to understand the nature,

vibhansa-msft:
Output file (json) rotation is not correct. Generally we always keep the latest file as _0 and rest all are renamed starting from last. This way you never have to check the file time to figure out which one is latest.

.gitignore
souravgupta-msft: Agree, we can think of something better

cmd/health-monitor.go
souravgupta-msft: When we do unmount operation, the CPU and memory monitors automatically turns off, since they are monitoring on the blobfuse2 pid.
For the blobfuse stats monitor, after unmounting one end of the pipe at the blobfuse2 binary side closes. This results in broken pipe error at the health monitor side causing the blobfuse2 stats monitor to turn down as well.

cmd/health-monitor.go
souravgupta-msft: In this case, even if the monitor is triggered it won't start since the blobfuse2 pid that we passed to the monitor binary is not active.

component/azstorage/azstorage.go
souravgupta-msft: The **path** will indicate the directory being listed. If the listing operation is done on the root, path will be an empty string. Otherwise, it will be the relative path of the folder being listed.

component/azstorage/azstorage.go
souravgupta-msft: Already have `options.Name` here which will be the relative path of the file being truncated.

souravgupta-msft:
> In the demo part there were two enable config one in common section and one under healthmonitor. but the base config mentions one. If there are two required, then its redundancy and we shall remove it.

You can ignore the one in the common section. It's not being used. I'll remove it.
All the config related to monitor is under the health-monitor section.

setup/baseConfig.yaml
gapra-msft: Since these values are in sec we should add -sec to the end of the name like for the other Blobfuse options.

component/azstorage/block_blob.go
```diff
@@ -533,13 +533,33 @@ func (bb *BlockBlob) List(prefix string, marker *string, count int32) ([]*intern
 	return blobList, listBlob.NextMarker.Val, nil
 }
 
+// track the progress of downloads of blobs
+func trackDownload(name string, bytesTransferred int64, count int64, downloadPtr *int64) {
+	if bytesTransferred >= (*downloadPtr)*100*1024*1024 || bytesTransferred == count {
```
gapra-msft, line 538: nit: use the MB constant here

gapra-msft: what is this bytesTransferred check for? Can we add a comment about that?

azure-pipeline-templates/build-release.yml
```diff
@@ -44,6 +44,22 @@ steps:
...
+      workingDirectory: ${{ parameters.work_dir }}
+    displayName: 'Building health-monitor binary'
+  
+  # Verifying whether built binary is correct
+  - script: |
+      sudo chmod +x ./healthmon
```
tasherif-msft, line 57: why `healthmon` instead of just `healthmoniter`

setup/baseConfig.yaml
tasherif-msft: I feel like the default values should be a little longer, maybe every 60s?

tools/health-monitor/monitor/file_cache/cache_monitor.go
```diff
@@ -0,0 +1,311 @@
...
+	fc.ExportStats(time.Now().Format(time.RFC3339), e)
+}
+
+func (fc *FileCache) chmodEvent(event *watcher.Event) {
+	if !event.IsDir() {
+		delete(fc.cacheObj.fileRemovedMap, event.Path)
+		fileSize := fc.cacheObj.fileCreatedMap[event.Path]
+
+		if fileSize == event.Size() {
+			return
```
gapra-msft, line 224: does this mean we skip lines 232+ if the file size didnt change with the mode? I don't understand why we do that.

gapra-msft: same for the writeEvent method below

tools/health-monitor/monitor/file_cache/cache_monitor.go
```diff
@@ -0,0 +1,311 @@
...
+		}
+
+		fc.cacheObj.cacheSize += event.Size() - fileSize
+		fc.cacheObj.fileCreatedMap[event.Path] = event.Size()
+		fc.cacheObj.cacheConsumed = (float64)(fc.cacheObj.cacheSize*100) / (fc.maxSizeMB * common.MbToBytes)
+	} else {
+		return
+	}
+
+	// not logging write events since there are multiple events in case of uploads
```
gapra-msft, line 255: so we will never log this and just let the blobfuse stats take care of the file download progress?

tools/health-monitor/monitor/file_cache/cache_monitor.go
souravgupta-msft: Nice observation. I corrected the part for chmod events where we will export those events.
For write events, I am not logging and exporting it because there can be lot of write events for large files. On the other hand, blobfuse is send the upload and download progress to the monitor binary for every 100MB of data transferred.

tools/health-monitor/monitor/file_cache/cache_monitor.go
souravgupta-msft: That's right. For large files upload/download, there can lots of write events. So, I am neither logging nor exporting the write events. The upload/download progress part is taken care by the blobfuse binary where it sends an event to the monitor binary for every 100MB of data transferred.
The reason why we have the writeEvent() method is because for upload operation we don't get the complete file size during the create event. So, writeEvent() method takes care of updating the file size and cache consumption stats.

setup/baseConfig.yaml
souravgupta-msft: I think 60s will be long for both blobfuse stats polling as well as collecting system stats (cpu, mem) for the pid. We can discuss on what should be the ideal default values

component/azstorage/block_blob.go
souravgupta-msft: This check is basically for tracking every 100 MB of data transferred, in this case downloaded.
The condition `bytesTransferred == count` checks if the download has completed or not. So, we are also tracking the download completion here. I will add some comments for this.

azure-pipeline-templates/build-release.yml
souravgupta-msft: I kept the name of the binary as healthmon for now. In one of the above comments, @vibhansa-msft suggested to come up with some better name. So, we can think of what name should we have for the binary

gapra-msft (approved):

tasherif-msft (approved):

cmd/health-monitor.go
```diff
@@ -0,0 +1,178 @@
...
+		resetMonitorOptions()
+
+		err := validateMonOptions()
+		if err != nil {
+			log.Err("health-monitor: [%v]", err)
```
vibhansa-msft, line 76: Where are we initializing logging lib here ?

vibhansa-msft (approved):

cmd/health-monitor.go
souravgupta-msft: This is part of the blobfuse cmd package. So, we don't need to initialize logging library separately here.

azure-pipeline-templates/build-release.yml
souravgupta-msft: Done

PR #871, opened 2022-08-18
Stream Filename Caching Support

tasherif-msft:


component/stream/read_write.go
```diff
@@ -343,6 +343,11 @@ func (rw *ReadWriteCache) Stop() error {
 	return nil
 }
 
+func (rw *ReadWriteCache) GetAttr(options internal.GetAttrOptions) (*internal.ObjAttr, error) {
+	// log.Trace("AttrCache::GetAttr : %s", options.Name)
```
gapra-msft, line 347: Same comment as above

component/stream/read.go
```diff
@@ -196,11 +196,16 @@ func (r *ReadCache) CloseFile(options internal.CloseFileOptions) error {
...
 		options.Handle.CacheObj.StreamOnly = true
-		atomic.AddInt32(&r.CachedHandles, -1)
+		atomic.AddInt32(&r.CachedObjects, -1)
 	}
 	return nil
 }
 
+func (r *ReadCache) GetAttr(options internal.GetAttrOptions) (*internal.ObjAttr, error) {
+	// log.Trace("AttrCache::GetAttr : %s", options.Name)
+	return r.NextComponent().GetAttr(options)
```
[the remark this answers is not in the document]
gapra-msft: oh I see. Stream calls into read or read/write or readwritefile and readwritefile does something special

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,478 @@
...
+	rw.BlockSize = int64(conf.BlockSize) * mb
+	rw.BufferSize = conf.BufferSize * mb
+	rw.CachedObjLimit = int32(conf.CachedObjLimit)
+	rw.fileCache = make(map[string]*handlemap.Cache)
+	rw.CachedObjects = 0
+	return nil
+}
+
+func (rw *ReadWriteFilenameCache) CreateFile(options internal.CreateFileOptions) (*handlemap.Handle, error) {
+	// log.Trace("Stream::CreateFile : name=%s, mode=%s", options.Name, options.Mode)
```
gapra-msft, line 72: why not leave the trace log?

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,478 @@
...
+type ReadWriteFilenameCache struct {
+	sync.RWMutex
+	*Stream
+	StreamConnection
+	fileCache map[string]*handlemap.Cache
+}
+
+func (rw *ReadWriteFilenameCache) Configure(conf StreamOptions) error {
+	if conf.BufferSize <= 0 || conf.BlockSize <= 0 || conf.CachedObjLimit <= 0 {
+		rw.StreamOnly = true
```
vibhansa-msft, line 61: In all three modes we have this streamingOnly support, some time in future we shall think on making this generic and outside the streaming sub-layers.

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,478 @@
...
+			log.Err("Stream::OpenFile : error failed to create cache object %s [%s]", options.Name, err.Error())
+		}
+	}
+	return handle, err
+}
+
+func (rw *ReadWriteFilenameCache) ReadInBuffer(options internal.ReadInBufferOptions) (int, error) {
+	// log.Trace("Stream::ReadInBuffer : name=%s, handle=%d, offset=%d", options.Handle.Path, options.Handle.ID, options.Offset)
+	if !rw.StreamOnly && options.Handle.CacheObj.StreamOnly {
+		err := rw.createFileCache(options.Handle)
```
vibhansa-msft, line 105: If handle is in streaming only mode, why do we need to create the fileCache. It shall just get required data from upstream component and return it back.

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,478 @@
...
+	return written, err
+}
+
+// TODO: truncate in cache
+func (rw *ReadWriteFilenameCache) TruncateFile(options internal.TruncateFileOptions) error {
+	// log.Trace("Stream::TruncateFile : name=%s, size=%d", options.Name, options.Size)
+	if !rw.StreamOnly {
+		rw.purge(options.Name, false)
+	}
+	err := rw.NextComponent().TruncateFile(options)
```
vibhansa-msft, line 158: Shouldn't the order be reverse, like you update on container first and if it's successful we shall make changes in local cache.

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,478 @@
...
+	if err != nil {
+		log.Err("Stream::TruncateFile : error truncating file %s [%s]", options.Name, err.Error())
+		return err
+	}
+	return nil
+}
+
+func (rw *ReadWriteFilenameCache) RenameFile(options internal.RenameFileOptions) error {
+	// log.Trace("Stream::RenameFile : name=%s", options.Src)
+	if !rw.StreamOnly {
```
vibhansa-msft, line 168: Same reverse order comment as above.

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,478 @@
...
+		log.Err("Stream::RenameFile : error renaming file %s [%s]", options.Src, err.Error())
+		return err
+	}
+	return nil
+}
+
+func (rw *ReadWriteFilenameCache) CloseFile(options internal.CloseFileOptions) error {
+	// log.Trace("Stream::CloseFile : name=%s, handle=%d", options.Handle.Path, options.Handle.ID)
+	// try to flush again to make sure it's cleaned up
+	err := rw.FlushFile(internal.FlushFileOptions(options))
```
vibhansa-msft, line 182: Only when handle is dirty we shall call the flush file

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,478 @@
...
+	}
+	err := rw.NextComponent().DeleteFile(options)
+	if err != nil {
+		log.Err("Stream::DeleteFile : error deleting file %s [%s]", options.Name, err.Error())
+		return err
+	}
+	return nil
+}
+
+// TODO: any cached file to be purged without flushing
```
vibhansa-msft, line 222: Are these todos still valid, or they have been already taken care of

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,478 @@
...
+			buffer.Purge()
+			atomic.AddInt32(&rw.CachedObjects, -1)
+		}
+	}
+	return nil
+}
+
+// GetAttr : Try to serve the request from the attribute cache, otherwise cache attributes of the path returned by next component
+func (rw *ReadWriteFilenameCache) GetAttr(options internal.GetAttrOptions) (*internal.ObjAttr, error) {
+	// log.Trace("AttrCache::GetAttr : %s", options.Name)
```
vibhansa-msft, line 273: We need to handle the dir listing operation as well? there might be case where a new file is created but its not uploaded to container and in such case list operation on dir shall be showing those files as well. If this is a miss, we shall recheck rw-streaming layer also for the same miss.

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,478 @@
...
+		return attrs, err
+	}
+	attrs.Mtime = buffer.Mtime
+	attrs.Size = buffer.Size
+	return attrs, nil
+}
+
+func (rw *ReadWriteFilenameCache) purge(fileName string, close bool) {
+	// check if this file is cached
+	rw.Lock()
```
vibhansa-msft, line 292: here we are locking the entire streaming layer which might hurt the performance. We might need to rethink on this.

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,478 @@
...
+			buffer.StreamOnly = true
+			atomic.AddInt32(&rw.CachedObjects, -1)
+		}
+	}
+}
+
+//TODO: Store block offset list
+func (rw *ReadWriteFilenameCache) createFileCache(handle *handlemap.Handle) error {
+	// check if file is cached
+	rw.Lock()
```
vibhansa-msft, line 316: This function is doing a lot of operations, do we really need this lock to be up for entire lifespan of this method.

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,478 @@
...
+			}
+			offsets, err := rw.NextComponent().GetFileBlockOffsets(opts)
+			if err != nil {
+				return err
+			}
+			handle.CacheObj.BlockOffsetList = offsets
+			atomic.StoreInt64(&handle.CacheObj.Size, handle.Size)
+			handle.CacheObj.Mtime = handle.Mtime
+			if handle.CacheObj.SmallFile() {
+				if uint64(atomic.LoadInt64(&handle.Size)) > memory.FreeMemory() {
```
vibhansa-msft, line 340: As per below code for small files we are downloading entire data and in our cache layer we represent this as a single block holding entire data along with a block-id assigned to it. At the end small-file flag is cleared, won't this create issue when we upload the file back assuming we have now split the file into blocks.

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,478 @@
...
+	read, err := rw.readWriteBlocks(options.Handle, options.Offset, options.Data, false)
+	if err != nil {
+		log.Err("Stream::ReadInBuffer : error failed to download requested data for %s: [%s]", options.Handle.Path, err.Error())
+	}
+	return read, err
+}
+
+func (rw *ReadWriteFilenameCache) WriteFile(options internal.WriteFileOptions) (int, error) {
+	// log.Trace("Stream::WriteFile : name=%s, handle=%d, offset=%d", options.Handle.Path, options.Handle.ID, options.Offset)
+	if !rw.StreamOnly && options.Handle.CacheObj.StreamOnly {
```
gapra-msft, line 130: same here as above

component/stream/read_write_filename.go
tasherif-msft: The reason is if the handle is read only but the mount is not read only then that means the handle became read only because we hit our cache limit. So we basically should check if now we can create a cache object (i.e cache cleared out a little)

component/stream/read_write_filename.go
tasherif-msft: I'm leaving it up to storage component to make that check, do you think we should just do it here?

component/stream/read_write_filename.go
tasherif-msft: that's the issue with filename caching, synchronization and perf hits, if we're removing an item from the cache and another handle tries to look it up, we will have a race. Hence the lock

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,478 @@
...
+		if err != nil {
+			log.Err("Stream::OpenFile : error failed to create cache object %s [%s]", options.Name, err.Error())
+		}
+	}
+	return handle, err
+}
+
+func (rw *ReadWriteFilenameCache) ReadInBuffer(options internal.ReadInBufferOptions) (int, error) {
+	// log.Trace("Stream::ReadInBuffer : name=%s, handle=%d, offset=%d", options.Handle.Path, options.Handle.ID, options.Offset)
+	if !rw.StreamOnly && options.Handle.CacheObj.StreamOnly {
```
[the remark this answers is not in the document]
tasherif-msft: this is saying if we're not in read only mode (on mount) but our handle is stream only - then we ran out of space and hence the handle was set as stream only. So this basically tries to create a cache object for it if memory has cleared up

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,478 @@
...
+	return nil
+}
+
+// TODO: any cached file to be purged without flushing
+func (rw *ReadWriteFilenameCache) DeleteFile(options internal.DeleteFileOptions) error {
+	// log.Trace("Stream::DeleteFile : name=%s", options.Name)
+	if !rw.StreamOnly {
+		rw.purge(options.Name, false)
+	}
+	err := rw.NextComponent().DeleteFile(options)
```
gapra-msft, line 214: should we swap the order here too?

component/stream/read_write_filename.go
vibhansa-msft: storage component we do not make any dirty check (atleast for file-cache). As handle or block is marked dirty by streaming ideally respective checks shall be done by the layer itself and not upstream component. Storage shall be dumb component, if someone asks it to upload it shall just do it.

component/stream/read_write_filename.go
tasherif-msft: Why would it be an issue?

component/stream/read_write_filename.go
tasherif-msft: Done

component/stream/read_write_filename.go
tasherif-msft: I have added all those changes will remove the todos

vibhansa-msft (approved):

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,476 @@
...
+			defer buffer.Unlock()
+			buffer.Purge()
+			atomic.AddInt32(&rw.CachedObjects, -1)
+		}
+	}
+	return nil
+}
+
+// GetAttr : Try to serve the request from the attribute cache, otherwise cache attributes of the path returned by next component
+func (rw *ReadWriteFilenameCache) GetAttr(options internal.GetAttrOptions) (*internal.ObjAttr, error) {
```
gapra-msft, line 271: Should we also add SyncFile impl? this should also be applied to the other read write impl too right?

gapra-msft: IsDirEmpty, StreamDir, ReadDir these should be implemented here too right? same for other read write impl

tasherif-msft: I think syncFile migh tneed to be implemented - however the dir ones should be passed to storage. Is there a reason to implement them?

gapra-msft: In case we need to update the file data - size for example. We wont hit a case where the file exists in the stream cache but isnt in the container right?

component/stream/read_write_filename.go
```diff
@@ -0,0 +1,476 @@
...
+			buffer.Lock()
+			defer buffer.Unlock()
+			buffer.Purge()
+			buffer.StreamOnly = true
+			atomic.AddInt32(&rw.CachedObjects, -1)
+		}
+	}
+}
+
+func (rw *ReadWriteFilenameCache) createFileCache(handle *handlemap.Handle) error {
```
gapra-msft, line 312: can we add a few more comments here for each of the cases?

component/stream/read_write_filename.go
tasherif-msft: we won't have a case of file in cache but not in container so "isDirEmpty" should not be implemented. As for the other two - I guess they would be useful but only if the developer wrote to the file (and we updated the LMT in cache only) and then listed. This however would probably only apply to filename caching (where we do update the LMT). But it would not apply to handle level we dont adjust the LMT since getAttr is not on a handle level

tasherif-msft: will add in a separate pr

component/stream/read_write_filename.go
tasherif-msft: done

PR #872, opened 2022-08-18
Added scripts to generate output conveniently

gapra-msft:
Covers git clone and upload download cases so far. Next PR to include fio


vibhansa-msft (approved):

vibhansa-msft:
Our ResNet50 pipeline shall be taken as a perftest pipeline and may be these kinds of tests we can add there so that on weekly basis from our main branch this perftest and comparison against v1 is done. Periodically or on any major change we can just take the results from here and see if anything has gone up or down.

test/scripts/pwrite.sh
```diff
@@ -1,9 +1,14 @@
...
+thread=$1
+count=$2
+size=$3 
+mntPath=$4 
+outputPath=$5 
+sed_line=$6
 
 start_time=`date +%s`
-time seq 1 $2 | parallel --will-cite -j $1 -I{} dd if=/dev/zero of=$4/$3_{}.tst bs=1M count=$3
+time seq 1 $count | parallel --will-cite -j $thread -I{} dd if=/dev/zero of=$mntPath/$size_{}.tst bs=1M count=$size
```
vibhansa-msft, line 11: Here replace /dev/zero with /dev/random. Write now this test case is generating a file which has only 0, which will be very fast in terms of generation and disk writing and does not qualify to be a real-life test case. /dev/random will generate random data.

test/scripts/git_clone.sh
```diff
@@ -0,0 +1,93 @@
+#!/bin/bash
+
+# To run this script from your workspace execute this command
+#   ./test/scripts/run.sh /mnt/blob_mnt /mnt/blobfusetmp ./config.yaml ./v1.cfg 2&> results.txt
```
vibhansa-msft, line 4: we need to correct these steps here as this might not require these many inputs for git clone.

vibhansa-msft, line 4: "run.sh" commands in comments is not correct

gapra-msft: fixed, will push in a bit

gapra-msft: it does need the same inputs

souravgupta-msft (approved):

PR #873, opened 2022-08-23
fix bug: delete or rename directory

Lyndon1994:
**There are 2 bugs: cannot delete or rename directory**

Demo:
image

**Bug Reason:**

1. delete bug:
    When delete directory, http request headers cannot have `x-ms-delete-snapshots`, see: https://docs.microsoft.com/en-us/rest/api/storageservices/delete-blob 
image

2. rename bug:
    When using blob mount to rename directory, `m_blob_client->start_copy` cannot copy directory, so I add `copy_directory` function to copy recursively.

**Result:**
image
image

### DIFF
--- a/blobfuse/include/BlockBlobBfsClient.h
+++ b/blobfuse/include/BlockBlobBfsClient.h
@@ -132,6 +132,10 @@ protected:
     /// Helper function - Authenticates with spn
     ///</summary>
     std::shared_ptr<blob_client_wrapper> authenticate_blob_spn();
+    ///< summary>
+    /// Helper function - Copy directory.
+    ///</summary>
+    bool copy_directory(const std::string sourcePath, const std::string destinationPath);
 private:
     ///<summary>
     /// Helper function - Renames single file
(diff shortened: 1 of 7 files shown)


vibhansa-msft:
@Lyndon1994: Thanks for the contribution. As we are working on the next generation of blobfuse (aka blobfuse2), new changes or features are not accepted in legacy code. Kindly give your test case a try with 'blobfuse2' and validate it works fine. If not, we will be happy to fix it there. 

Here is the preview-2 link for blobfuse2: https://github.com/Azure/azure-storage-fuse/releases/tag/blobfuse2-2.0.0-preview2
It has more features and better performance compared to legacy blobfuse. It resides in the 'main' branch of this repo. Here is a link to the readme: https://github.com/Azure/azure-storage-fuse/blob/main/README.md

vibhansa-msft:
Regarding Delete request having "x-ms-delete-snapshots" : blobfuse does not work on snapshots and always tries to delete the base blob only. In that case how does having this field results into failure ?

vibhansa-msft:
/usr/blob_mnt$ mkdir a
/usr/blob_mnt$ mv a b

/usr/blob_mnt$ ls a
ls: cannot access 'a': No such file or directory

/usr/blob_mnt$ ls b

/usr/blob_mnt$ rm -r a
rm: cannot remove 'a': No such file or directory

/usr/blob_mnt$ rm -r b
/blob_mnt$ ls b
ls: cannot access 'b': No such file or directory


I do not see any issue in my local setup with either rename or delete.

vibhansa-msft:
Can you add more details on what kind of accoun it is, what auth mode you are using and under what condition this delete or rename fails.

vibhansa-msft:
Closing this as there are no updates on this thread. If you are still facing the issue can you give blobfuse2 a try and validate things work fine there or not.

PR #887, opened 2022-08-31
Make writeback cache default

gapra-msft:


tasherif-msft:
please make sure to run a nightly run after the data validation step for stream is merged in. Just so we make sure that nothing broke there

component/libfuse/libfuse_handler.go
```diff
@@ -612,12 +611,29 @@ func libfuse_open(path *C.char, fi *C.fuse_file_info_t) C.int {
 	if fi.flags&C.O_SYNC != 0 || fi.flags&C.__O_DIRECT != 0 {
 		log.Err("Libfuse::libfuse_open : Reset flags for open %s, fi.flags %X", name, fi.flags)
 		// Blobfuse2 does not support the SYNC or DIRECT flag. If a user application passes this flag on to blobfuse2
-		// and we open the file with this flag, subsequent write operations wlil fail with "Invalid argument" error.
+		// and we open the file with this flag, subsequent write operations wikk fail with "Invalid argument" error.
```
vibhansa-msft, line 614: still a typo here.

component/libfuse/libfuse_handler.go
```diff
@@ -612,12 +611,29 @@ func libfuse_open(path *C.char, fi *C.fuse_file_info_t) C.int {
...
+			log.Debug("Libfuse::libfuse_open : Flag O_WRONLY (%X) for open %s when write back cache is on.", fi.flags, name)
+			// 	fi.flags = fi.flags &^ C.O_ACCMODE
+			// 	fi.flags = fi.flags | C.O_RDWR
+		}
+
+		if fi.flags&C.O_APPEND != 0 {
+			log.Err("Libfuse::libfuse_open : Unsupported flag O_APPEND (%X) for open %s when write back cache is on. Pass --disable-writeback-cache via CLI or libfuse.disable-writeback-cache: true via config file", fi.flags, name)
+			// TODO: Potentially add a configuration option to support "unreliable append"
+			if fuseFS.ignoreAppendFlag {
+				fi.flags = fi.flags &^ C.O_APPEND
```
vibhansa-msft, line 631: We want to disable append flag or return back with error in case append is sent ?

vibhansa-msft: If we want to disable that then here we shall add WR flag so that atleast file can be written.

setup/baseConfig.yaml
```diff
@@ -38,7 +38,7 @@ libfuse:
   negative-entry-expiration-sec: <time kernel can cache attributes of non existent paths (in sec). Default - 120 sec>
   fuse-trace: true|false <enable libfuse api trace logs for debugging>
   extension: <physical path to extension library>
-  enable-writeback-cache: true|false <allow libfuse to buffer write requests. do not turn this on if you plan to open files in append-only mode or write-only mode as it can result in inconsistent behavior. Default - true for read-only mode, false otherwise>
+  disable-writeback-cache: true|false <disallow libfuse to buffer write requests>
```
vibhansa-msft, line 41: append ignore missing here.

component/libfuse/libfuse_handler.go
```diff
@@ -612,12 +611,29 @@ func libfuse_open(path *C.char, fi *C.fuse_file_info_t) C.int {
...
 		fi.flags = fi.flags &^ C.__O_DIRECT
 	}
+	if !fuseFS.disableWritebackCache {
+		if fi.flags&C.O_ACCMODE == C.O_WRONLY {
+			log.Debug("Libfuse::libfuse_open : Flag O_WRONLY (%X) for open %s when write back cache is on.", fi.flags, name)
+			// 	fi.flags = fi.flags &^ C.O_ACCMODE
+			// 	fi.flags = fi.flags | C.O_RDWR
+		}
+
+		if fi.flags&C.O_APPEND != 0 {
```
vibhansa-msft, line 627: lets add test cases for all these combinations.

component/libfuse/libfuse_handler.go
gapra-msft: append should always come with write I think

gapra-msft: we will return failure by default. If a customer opts into ignoring the append mode, we mask it

gapra-msft: the WR or RWR flag will always be turned on with Append mode. one of these is a must O_RDONLY, O_WRONLY, or O_RDWR

component/libfuse/libfuse_handler.go
```diff
@@ -612,12 +611,29 @@ func libfuse_open(path *C.char, fi *C.fuse_file_info_t) C.int {
...
+			log.Err("Libfuse::libfuse_open : Unsupported flag O_APPEND (%X) for open %s when write back cache is on. Pass --disable-writeback-cache via CLI or libfuse.disable-writeback-cache: true via config file", fi.flags, name)
```
tasherif-msft, line 628: shouldn't this log be in the else statement? since the line after we check if ignore-append is on

component/libfuse/libfuse_handler.go
```diff
@@ -612,12 +611,29 @@ func libfuse_open(path *C.char, fi *C.fuse_file_info_t) C.int {
...
+			if fuseFS.ignoreAppendFlag {
```
tasherif-msft, line 629: so ignore-append is ignoring append, read only and write only?

component/libfuse/libfuse_handler.go
gapra-msft: its still useful in both situations I think

souravgupta-msft (approved):

vibhansa-msft (approved):

tasherif-msft (approved):

PR #888, opened 2022-08-31
Run scripts in performance pipeline

gapra-msft:


test/scripts/fio.sh
```diff
@@ -76,13 +91,14 @@ do
...
-	rm -rf $mntPath/*
-	sudo fusermount3 -u $mntPath
+	rm $mntPath/testfile4G$i
 
 	(( sed_line++ ))
     blobfuse_write_average=$(( $blobfuse_write_average + $write_iops ))
     blobfuse_read_average=$(( $blobfuse_read_average + $read_iops ))
+    echo "========================================================="
 done
+sudo fusermount3 -u $mntPath
```
vibhansa-msft, line 101: FIO results as per pipeline output shall be reverse, when v2 is doing better it shall show negative or in case we want to show improvement in terms of positive values then git clone shall also show the same. As of now if you check pipeline results are different for git and fio.

vibhansa-msft: May be we can start showing the difference as perf improvement and show the numbers as positive values only

vibhansa-msft (approved):

souravgupta-msft (approved):

test/scripts/fio.sh
gapra-msft: Yeah I can reverse the negation for time based tests. The reason for the discrepancy rn is cause time we want to be smaller but iops we want to be bigger.

gapra-msft: done

tasherif-msft:
before merging - the pipeline is now fixed - please make sure this still runs resnet successfully before merging

tasherif-msft (approved):

test/scripts/fio.sh
vibhansa-msft: Upload/Download test show aggregate results as "%write improvement" and "%read improvement" that is a positive number and makes sense. Gitclone results shows % difference and "-ve" is better. FIO results shows % difference but "+ve" is better. So I still do not see all results represented in same way. "% improvement" sounds like a good term and lets put all the results in +ve terms only as they show improvement.

vibhansa-msft: Once you make this change and validate, there are all approvals there so you can go ahead and merge it.

gapra-msft:
Validated in the perf pipeline, everything is looking good.

Issue #894, opened 2022-09-07
blobfuse2 failed to mount when specify `AZURE_STORAGE_BLOB_ENDPOINT` explicitly


cvvz:
### Which version of the blobfuse was used?
blobfuse2

### What problem was encountered?
blobfuse2 failed to mount when specify `AZURE_STORAGE_BLOB_ENDPOINT` explicitly, while blobfuse v1 can mount successfully.

### Have you found a mitigation/solution?
unset `AZURE_STORAGE_BLOB_ENDPOINT` or change the endpoint to `account.dfs.core.windows.net`

### If relevant, please share your mount command.
```shell
export AZURE_STORAGE_ACCESS_KEY=xxx
export AZURE_STORAGE_ACCOUNT=acount
export AZURE_STORAGE_BLOB_ENDPOINT=acount.blob.core.windows.net

blobfuse2 mount /test --cancel-list-on-mount-seconds=0 --file-cache-timeout-in-seconds=120 --use-adls=true -o allow_other --pre-mount-validate=true --use-https=true --empty-dir-check=false --tmp-path=/mnt/tmp-test --container-name=xxxx
failed to mount : failed to authenticate credentials for azstorage
```

gapra-msft:
Resolved as the PR has been merged.

PR #898, opened 2022-09-08
Fixed parsing of endpoint to support blob endpoints for ADLS accounts

gapra-msft:
Resolves https://github.com/Azure/azure-storage-fuse/issues/894

### DIFF
--- a/component/azstorage/config.go
+++ b/component/azstorage/config.go
@@ -321,7 +321,6 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 	az.stConfig.authConfig.Endpoint = formatEndpointAccountType(az.stConfig.authConfig.Endpoint, az.stConfig.authConfig.AccountType)
 
 	az.stConfig.authConfig.ActiveDirectoryEndpoint = opt.ActiveDirectoryEndpoint
-	az.stConfig.authConfig.ActiveDirectoryEndpoint = formatEndpointProtocol(az.stConfig.authConfig.ActiveDirectoryEndpoint, false)
 
 	// If subdirectory is mounted, take the prefix path
 	az.stConfig.prefixPath = opt.PrefixPath
--- a/component/azstorage/config.go
+++ b/component/azstorage/config.go
@@ -321,6 +321,7 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 	az.stConfig.authConfig.Endpoint = formatEndpointAccountType(az.stConfig.authConfig.Endpoint, az.stConfig.authConfig.AccountType)
 
 	az.stConfig.authConfig.ActiveDirectoryEndpoint = opt.ActiveDirectoryEndpoint
+	az.stConfig.authConfig.ActiveDirectoryEndpoint = formatEndpointProtocol(az.stConfig.authConfig.ActiveDirectoryEndpoint, false)
 
 	// If subdirectory is mounted, take the prefix path
 	az.stConfig.prefixPath = opt.PrefixPath
--- a/component/azstorage/config.go
+++ b/component/azstorage/config.go
@@ -208,25 +208,42 @@ func RegisterEnvVariables() {
 
 // formatEndpointProtocol : add the protocol and missing "/" at the end to the endpoint
 func formatEndpointProtocol(endpoint string, http bool) string {
-	correctedEP := endpoint
+	correctedEndpoint := endpoint
 
 	// If the pvtEndpoint does not have protocol mentioned in front, pvtEndpoint parsing will fail while
 	// creating URI also the string shall end with "/"
-	if correctedEP != "" {
-		if !strings.Contains(correctedEP, "://") {
+	if correctedEndpoint != "" {
+		if !strings.Contains(correctedEndpoint, "://") {
 			if http {
-				correctedEP = "http://" + correctedEP
+				correctedEndpoint = "http://" + correctedEndpoint
 			} else {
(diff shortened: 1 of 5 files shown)


component/azstorage/config.go
```diff
@@ -300,10 +317,11 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 		}
 	}
 	az.stConfig.authConfig.Endpoint = opt.Endpoint
-	az.stConfig.authConfig.Endpoint = formatEndPoint(az.stConfig.authConfig.Endpoint, opt.UseHTTP)
+	az.stConfig.authConfig.Endpoint = formatEndpointProtocol(az.stConfig.authConfig.Endpoint, opt.UseHTTP)
+	az.stConfig.authConfig.Endpoint = formatEndpointAccountType(az.stConfig.authConfig.Endpoint, az.stConfig.authConfig.AccountType)
 
 	az.stConfig.authConfig.ActiveDirectoryEndpoint = opt.ActiveDirectoryEndpoint
-	az.stConfig.authConfig.ActiveDirectoryEndpoint = formatEndPoint(az.stConfig.authConfig.ActiveDirectoryEndpoint, false)
+	az.stConfig.authConfig.ActiveDirectoryEndpoint = formatEndpointProtocol(az.stConfig.authConfig.ActiveDirectoryEndpoint, false)
```
vibhansa-msft, line 324: I have a doubt here, formatting the account endpoint based on selected protocol or account type is fine but the AAD endpoint shouldn't we keep just what user has given as is. We may have a setup where accont access needs to go over http but AAD endpoint is over https.

component/azstorage/utils_test.go
```diff
@@ -262,6 +262,29 @@ func (s *utilsTestSuite) TestBfsProxyOptions() {
...
+type endpointAccountType struct {
+	endpoint string
+	account  AccountType
+	result   string
+}
+
+func (s *utilsTestSuite) TestFormatEndpointAccountType() {
+	assert := assert.New(s.T())
+	var inputs = []endpointAccountType{
+		{endpoint: "https://account.blob.core.windows.net", account: EAccountType.BLOCK(), result: "https://account.blob.core.windows.net"},
```
vibhansa-msft, line 275: Let add more cases to validate we are not modifying the wrong strings such as .blobx or xblob., sort of negative test cases.

component/azstorage/utils_test.go
```diff
@@ -262,6 +262,29 @@ func (s *utilsTestSuite) TestBfsProxyOptions() {
...
+		{endpoint: "https://account.dfs.core.windows.net", account: EAccountType.BLOCK(), result: "https://account.blob.core.windows.net"},
+		{endpoint: "https://account.blob.core.windows.net", account: EAccountType.ADLS(), result: "https://account.dfs.core.windows.net"},
+		{endpoint: "https://account.dfs.core.windows.net", account: EAccountType.ADLS(), result: "https://account.dfs.core.windows.net"},
```
vibhansa-msft, line 278: Shall we have a test case to mimic our actual code where we first modify based on protocol and then based on account type.

vibhansa-msft (approved):

souravgupta-msft (approved):

component/azstorage/utils_test.go
gapra-msft: Yep Good point, I will also add some other cloud endpoints too

component/azstorage/utils_test.go
gapra-msft: We can add a unit test in Azstorage

component/azstorage/config.go
gapra-msft: Here we are saying always go over HTTPS - since we pass false. I don't think you can get an AAD token over http anyway

gapra-msft: Actually I think we should keep whatever the user provides as is

tasherif-msft:
looks good

tasherif-msft (approved):

component/azstorage/config.go
```diff
@@ -300,10 +317,10 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 		}
 	}
 	az.stConfig.authConfig.Endpoint = opt.Endpoint
-	az.stConfig.authConfig.Endpoint = formatEndPoint(az.stConfig.authConfig.Endpoint, opt.UseHTTP)
+	az.stConfig.authConfig.Endpoint = formatEndpointProtocol(az.stConfig.authConfig.Endpoint, opt.UseHTTP)
+	az.stConfig.authConfig.Endpoint = formatEndpointAccountType(az.stConfig.authConfig.Endpoint, az.stConfig.authConfig.AccountType)
 
 	az.stConfig.authConfig.ActiveDirectoryEndpoint = opt.ActiveDirectoryEndpoint
-	az.stConfig.authConfig.ActiveDirectoryEndpoint = formatEndPoint(az.stConfig.authConfig.ActiveDirectoryEndpoint, false)
```
vibhansa-msft, line 306: Lets keep this, if we look closely we check if the string contains :// or not, if its not there we add it based on selected protocol, if its there we do not perform any transformations. So in case of AAD endpoint if user has given fully qualified URL it will not touch it, if its not given then anyway we are not sure whether to take http or https path and based on selected protocol we will tweak it which shall be ok.

vibhansa-msft: Or other way to look at this is as user has not given fully qualified URI, it may not work and we can ask user to correct his input. if we wish to take that path then removing this link is fine.

PR #904, opened 2022-09-14
Correct Migration guide installation instructions

gapra-msft:


vibhansa-msft (approved):

souravgupta-msft (approved):

jimhavrilla:
Is this in the new release?  I still get this error in 2.0.0-preview.4

gapra-msft:
@jimhavrilla this PR was simply a documentation change. Please install the fuse3 dependency onto your system and it should work.

jimhavrilla:
I have installed both `sudo dnf install https://vault.centos.org/centos/8/BaseOS/x86_64/os/Packages/fuse3-libs-3.2.1-12.el8.x86_64.rpm` and `sudo dnf install fuse3` and neither has made a difference.

gapra-msft:
Do either of those commands succeed when you run them?

jimhavrilla:
Both commands succeed when running.

gapra-msft:
@jimhavrilla you mentioned you are using an Azure VM? Could you share the exact image so we can try to reproduce the issue?

gapra-msft:
@jimhavrilla yes, this helps. I was able to create a VM with the same image. I will attempt to repro the issue now.

gapra-msft:
@jimhavrilla  Just so I reproduce this correctly, are these the commands you ran?

1. sudo dnf install fuse3
2.  wget https://github.com/Azure/azure-storage-fuse/releases/download/blobfuse2-2.0.0-preview.4/blobfuse2-2.0.0-preview.4-CentOS-8.0-x86-64.rpm
3. rpm -i blobfuse2-2.0.0-preview.4-CentOS-8.0-x86-64.rpm
4. attempt to mount using blobfuse2

gapra-msft:
> Is this in the new release? I still get this error in `2.0.0-preview.4`. In addition, trying to `go build` the latest on a CentOS 8 Azure VM gets me (using `dnf install go` for golang): image

Also I just re-read this and to build blobfuse2 you also need the libfuse3-dev library

jimhavrilla:
How do I get that?  Only libfuse3 version I could find for CentOS8 was `sudo dnf install https://vault.centos.org/centos/8/BaseOS/x86_64/os/Packages/fuse3-libs-3.2.1-12.el8.x86_64.rpm`. I then ran `sudo dnf install fuse3`, and I tried both
git clone this repo, go get (gives fuse.h error) and go build --tags=fuse3 (gives same error).  Forget trying to mount, blobfuse2 doesn't even install.

jimhavrilla:
I don't know if you're aware (in case you're looking in emails) but I keep going back and editing these for more detail which I hope helps.

gapra-msft:
@jimhavrilla I was able to successfully install blobfuse2 with the following steps

1. sudo dnf install fuse fuse3 fuse3-devel
2. wget https://github.com/Azure/azure-storage-fuse/releases/download/blobfuse2-2.0.0-preview.4/blobfuse2-2.0.0-preview.4-CentOS-8.0-x86-64.rpm
3. sudo rpm -i blobfuse2-2.0.0-preview.4-CentOS-8.0-x86-64.rpm
4. Validate install using `blobfuse2 --version`

I will note that it is unusual that you need to install fuse to work. I will create an issue on Github for us to look into that. But I hope this helps unblock you for now. Please let me know if this works for you.

Note: I dont think dnf install blobfuse2 will work unless you configure the Microsoft repo following this link, I havent tried this yet. https://learn.microsoft.com/en-us/windows-server/administration/Linux-Package-Repository-for-Microsoft-Software

jimhavrilla:
image
Thanks for unblocking me 👍 

P.S., FYI, with `dnf` you can just do `sudo dnf install https://github.com/Azure/azure-storage-fuse/releases/download/blobfuse2-2.0.0-preview.4/blobfuse2-2.0.0-preview.4-CentOS-8.0-x86-64.rpm` without the `wget` step.

Now that I have you...can you show me what I am doing wrong to mount (I had tried this on blobfuse 1.4 which was why I upgraded to 2)?  I have 32GB of RAM and may need to move a file up to 16GB so I want to make sure my blob has sufficient temp storage.  I am running:
`blobfuse2 mount /mnt/informatics --tmp-path=/mnt/temp --config-file=connection.cfg` which tells me a `.cfg` file is no longer used and instead a yaml?

From your site I understand you're using YAML now but I don't need such a large YAML, right?  What is the most base YAML?  
```
azstorage:
# Required
  type: block|adls <type of storage account to be connected. Default - block>
  account-name: <name of the storage account>
  container: <name of the storage container to be mounted>
  endpoint: <storage account endpoint (example - https://account-name.blob.core.windows.net)>
  mode: key|sas|spn|msi <kind of authentication to be used>
  account-key: <storage account key>
# Disk cache related configuration
file_cache:
  # Required
  path: <path to local disk cache>
```
Is that the bare minimum?

gapra-msft:
@jimhavrilla Yep that looks correct for the bare minimum.

jimhavrilla:
So then is the command right minus replacing cfg with yaml?

jimhavrilla:
For anyone else I had to create an extremely detailed endpoint containing the name of the account itself (copy your container URL minus the container) and use the following command `blobfuse2 mount /mnt/informatics --config-file=config.yaml`.

gapra-msft:
Yep that is correct!

PR #907, opened 2022-09-15
Health Monitor Readme + Generated CLI Docs

tasherif-msft:


doc/blobfuse2.md
```diff
@@ -0,0 +1,29 @@
+## blobfuse2
```
gapra-msft, line 1: Should we add this to the release pipeline so it automatically does this for us?

tasherif-msft: Seems unnecessary. These parameters are not going to be changing on each release

vibhansa-msft, line 1: Do we really need to checkin all these md files ?

vibhansa-msft (approved):

gapra-msft (approved):

doc/blobfuse2.md
tasherif-msft: Yea doesn't hurt to keep them. Customers might also just go through these if they didn't access the MS Docs

PR #911, opened 2022-09-22
Script to test attr calls after writes + perf vm fix

tasherif-msft:


test/scripts/readwrite.c
```diff
@@ -0,0 +1,48 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <time.h>
+#include <unistd.h>
+
+// To build this just run : gcc readwrite.c -o readwrite
+// To execute this run : ./readwrite
+
+int main()
```
gapra-msft, line 9: Should we run this script in the pipeline or something?

tasherif-msft: No - we shouldn't. This is only for manual testing purposes. We can replicate this with a unit test or something. But I wanted to dump this script here anyway.

vibhansa-msft (approved):

gapra-msft (approved):

PR #914, opened 2022-09-25
Docker installation and container image build

vibhansa-msft:
- Added steps to install docker
- Build blobfuse2 container image based on latest ubuntu


docker/dockerinstall.sh
```diff
@@ -0,0 +1,37 @@
...
+
+
+# Install docker 
+sudo apt-get update
+sudo apt-get install docker-ce docker-ce-cli containerd.io docker-compose-plugin -y
+sudo apt-get update
+
+# Start docker service
+sudo service docker start
+
```
gapra-msft, line 28: nit remove new line

docker/buildandruncontainer.sh
```diff
@@ -0,0 +1,57 @@
+
+# Build blobfuse2 binary
+cd ..
+if [ "$1" == "fuse2" ]
```
gapra-msft, line 4: Should we also build the health monitor here?

docker/buildandruncontainer.sh
```diff
@@ -0,0 +1,57 @@
...
+then
+	echo "Building blobfuse2 with libfuse"
+	./build.sh fuse2
+else
+	echo "Building blobfuse2 with libfuse3"
+	./build.sh
+fi
+
+# As docker build can not go out of scope os this directory copy the binary here
```
gapra-msft, line 13:
```suggestion
# As docker build can not go out of scope of this directory copy the binary here
```

docker/buildandruncontainer.sh
```diff
@@ -0,0 +1,57 @@
...
+ 
+# Image build is executed so we can clean up temp executable from here
+rm -rf ./blobfuse2
+rm -rf 11-blobfuse2.conf blobfuse2-logrotate
+
+# If build was successful then launch a container instance
+status=`docker images | grep $tag`
+
+if [ $? = 0 ]; then
+	echo " **** Build successful, trying container now ******"
```
gapra-msft, line 43:
```suggestion
	echo " **** Build successful, running container now ******"
```

docker/Dockerfile
```diff
@@ -0,0 +1,50 @@
...
+# Copy blobfuse2 binary to executable path
+COPY ./blobfuse2 /usr/local/bin/
+COPY ./config.yaml /usr/share/blobfuse2/
+
+# Install fuse library
+RUN \
+	apt update && \
+	apt-get install -y ca-certificates vim rsyslog
+
+RUN if [ "$FUSE2" = "TRUE" ] ; then apt-get install -y libfuse-dev fuse ; else apt-get install -y libfuse3-dev fuse3 ; fi
```
gapra-msft, line 19: Are we building inside the container? or just running? if just running then we just need fuse/fuse3 right? We dont need libfuse-dev stuff

gapra-msft (commented):
left few comments

docker/buildandruncontainer.sh
vibhansa-msft: Just trying to have a very basic container which can mount blobfuse so avoiding all unnecessary stuff. If we combine health monitor here then config has to be setup for that and if we enable that by default then on container spawn it will be turned on which takes more memory and resources. So to conserve all that just keeping it very simple.

docker/Dockerfile
vibhansa-msft: we are just running inside the container, will check removing libfuse(3)-dev works fine or not.

vibhansa-msft: done, it works without this dependency as well.

README.md
```diff
@@ -156,6 +156,8 @@ az cli has a command to generate a sas token. Open a command prompt and make sur
 az storage container generate-sas --account-name <account name ex:myadlsaccount> --account-key <accountKey> -n <container name> --permissions dlrwac --start <today's date ex: 2021-03-26> --expiry <date greater than the current time ex:2021-03-28>
 - Why do I get EINVAL on opening a file with WRONLY or APPEND flags?
 To improve performance, Blobfuse2 by default enables writeback caching, which can produce unexpected behavior for files opened with WRONLY or APPEND flags, so Blobfuse2 returns EINVAL on open of a file with those flags. Either use disable-writeback-caching to turn off writeback caching (can potentiallu result in degraded performance) or ignore-open-flags (replace WRONLY with RDWR and ignore APPEND) based on your workload. 
+- How to mount blobfuse inside a container?
```
souravgupta-msft, line 159:
```suggestion
- How to mount blobfuse2 inside a container?

```

souravgupta-msft (approved):

gapra-msft (approved):

PR #916, opened 2022-09-26
Added subcommands to health monitor

gapra-msft:


tasherif-msft:
Is there a way we can add tests for this?

vibhansa-msft (approved):
We need to add UT/test cases for these commands.

souravgupta-msft:
> We need to add UT/test cases for these commands.

We can take this up separately. There are other health monitor code for which we'll have to add the tests as well.

cmd/health-monitor_stop.go
```diff
@@ -0,0 +1,117 @@
...
+// Attempts to get pid of the health monitor
+func getPid(blobfuse2Pid string) (string, error) {
+	psAux := exec.Command("ps", "aux")
+	out, err := psAux.Output()
+	if err != nil {
+		return "", err
+	}
+	processes := strings.Split(string(out), "\n")
+	for _, process := range processes {
+		if strings.Contains(process, fmt.Sprintf("bfusemon --pid=%s", blobfuse2Pid)) {
```
souravgupta-msft, line 84: This may not always be correct because `pid` is a flag and its order can be anything wrt to other flags. So, what we can do here is something like,
`strings.Contains(process, "bfusemon") && strings.Contains(process, fmt.Sprintf("--pid=%s", blobfuse2Pid))`

cmd/health-monitor_stop.go
```diff
@@ -0,0 +1,117 @@
...
+			re := regexp.MustCompile(`[-]?\d[\d,]*[\.]?[\d{2}]*`)
+			pids := re.FindAllString(process, 1)
+			if pids == nil {
+				return "", fmt.Errorf("failed to process PID from %s", process)
+			}
+			pid := pids[0]
+			fmt.Printf("Successfully got health monitor PID %s.\n", pid)
```
souravgupta-msft, line 91: Do we need both print statements here and at line 60?

souravgupta-msft (approved):

cmd/health-monitor_stop.go
gapra-msft: I'll remove the one on line 60

cmd/health-monitor_stop.go
gapra-msft: True, will fix this

tasherif-msft (approved):

PR #924, opened 2022-09-30
Mount using /etc/fstab

### LINKED ISSUE #921: Blobfuse2 fstab not working
2.0.0-preview.3

### Which OS (please include version) are you using?
Ubuntu Server 20.04.5 LTS

### What problem was encountered?
One-time mounting works fine. Can't get persistent mount working with fstab. No problems with blobfuse1.

No

### By default, blobfuse logs errors to syslog. If this is relevant, is there anything in the syslog that might be helpful?
No

In fstab I have the following entry:
blobfuse2 /mnt/blobcontainer fuse defaults,_netdev,--config-file=/home/user/blobfuse_config.yaml 0 0

sudo mount -a results in:
Error: unknown command "/mnt/blobcontainer" for "blobfuse2"

Adding "mount" after blobfuse2 in fstab and then trying sudo mount -a results in:
mount: /etc/fstab: parse error at line 5 -- ignored

Nothing related being output to varlog.

souravgupta-msft:
Fixes #921


vibhansa-msft (approved):

common/types.go
```diff
@@ -65,8 +65,12 @@ const (
 
 	MbToBytes  = 1024 * 1024
 	BfuseStats = "blobfuse_stats"
+
+	FuseAllowedFlags = "invalid FUSE options. Allowed FUSE configurations are: `-o attr_timeout=TIMEOUT`, `-o negative_timeout=TIMEOUT`, `-o entry_timeout=TIMEOUT` `-o allow_other`, `-o allow_root`, `-o umask=PERMISSIONS -o default_permissions`, `-o ro`"
 )
 
+var FuseIgnoredFlags = []string{"default_permissions", "rw", "dev", "nodev", "suid", "nosuid", "delay_connect", "uid", "gid", "auto", "noauto", "user", "nouser", "exec", "noexec"}
```
vibhansa-msft: This shall be converted to function to make it constant.

blobfuse2-release.yaml
```diff
@@ -104,7 +104,6 @@ stages:
               cp azure-storage-fuse/setup/baseConfig.yaml pkgDir/usr/share/blobfuse2/
               cp azure-storage-fuse/sampleFileCacheConfig.yaml pkgDir/usr/share/blobfuse2/
               cp azure-storage-fuse/sampleStreamingConfig.yaml pkgDir/usr/share/blobfuse2/
-              cp azure-storage-fuse/tools/postinstall.sh pkgDir/usr/share/blobfuse2/
```
gapra-msft, line 107: Why did we remove this?

cmd/unmount.go
```diff
@@ -51,20 +51,13 @@ var unmountCmd = &cobra.Command{
...
+	RunE: func(_ *cobra.Command, args []string) error {
 		if strings.Contains(args[0], "*") {
 			mntPathPrefix := args[0]
 
-			lstMnt, err := common.ListMountPoints()
-			if err != nil {
-				return fmt.Errorf("failed to list mount points [%s]", err.Error())
-			}
-
+			lstMnt, _ := common.ListMountPoints()
```
gapra-msft, line 58: Why are we not returning errors on failure to complete these operations?

cmd/root.go
```diff
@@ -177,7 +177,60 @@ func VersionCheck() error {
...
+				return true
+			}
+		}
+	}
+	return false
+}
+
+// parse arguments passed
+func parseArgs(cmdArgs []string) []string {
+	cmdArgs = cmdArgs[1:]
```
gapra-msft, line 195: can we add a comment about why we need this here?

gapra-msft: in general the parseArgs function

gapra-msft (commented):
made some comments

blobfuse2-release.yaml
vibhansa-msft: postinstall is basically installing the bash scripts for auto-completion. For fstab to work, we had to make mount as a default command option which creates ambiguity in commands like "blobfuse2 abc" = "blobfuse mount abc" and hence autocompletion scripts are failing due to this. Untill we find a way to fix those we are removing these scripts, fstab being a must have feature and auto-completion can be ignored.

cmd/unmount.go
vibhansa-msft: List mount fails only in case we are not able to read through /etc/mtab, which shall never happen so its most likely a dead code which we will never hit and even our code coverage will report its never hitting. To avoid such cases we have ignored those error handlings.

cmd/root_test.go
```diff
@@ -153,6 +159,36 @@ func (suite *rootCmdSuite) testExecute() {
...
+		{input: "--help", output: "--help"},
+		{input: "version", output: "version"},
+		{input: "--version", output: "--version"},
+		{input: "version --check=true", output: "version --check=true"},
+		{input: "mount abc --config-file=./config.yaml -o ro", output: "mount abc --config-file=./config.yaml -o ro"},
+		{input: "abc", output: "mount abc"},
+		{input: "-o", output: ""},
+		{input: "", output: ""},
+
+		{input: "/home/mntdir -o rw,--config-file=config.yaml,dev,suid", output: "mount /home/mntdir -o rw,dev,suid --config-file=config.yaml"},
```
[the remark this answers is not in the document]
vibhansa-msft: In case of mount using fstab we observed these parameters are coming as input. In case of v1 as well these are coming but there command parsing is done by libfuse and it ignores whatever is not required. In case of v2 we parse and if there are unsupported options we throw back failure, so we have added some of these options as ignorable options.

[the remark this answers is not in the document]
gapra-msft: I mean does this code handle CLI like this ? -o rw,suid,dev as well as -o rw -o suid -o dev

gapra-msft (approved):

cmd/root_test.go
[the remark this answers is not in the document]
vibhansa-msft: yes it does. "-o rw,suid,dev" format we receive when fstab is mounting using blobfuse.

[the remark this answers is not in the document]
vibhansa-msft: even we can receive "-o rw,suid,--config-file=config.yaml,dev,allow_other" which we will convert to "-o rw,suid,dev,allow_other --config-file=config.yaml"

PR #926, opened 2022-09-30
MSI Object ID Credential Support

tasherif-msft:
Tested obj id on perf vm - succeeded


component/azstorage/azauthmsi.go
```diff
@@ -116,25 +114,25 @@ type azAuthBfsMSI struct {
...
+	token, err := azmsi.fetchToken()
 	if err != nil {
 		return nil
 	}
 
 	// Using token create the credential object, here also register a call back which refreshes the token
-	tc := azbfs.NewTokenCredential(spt.Token().AccessToken, func(tc azbfs.TokenCredential) time.Duration {
-		err := spt.Refresh()
+	tc := azbfs.NewTokenCredential(token.AccessToken, func(tc azbfs.TokenCredential) time.Duration {
+		newTOken, err := token.Refresh(context.Background())
```
gapra-msft, line 124:
```suggestion
		newToken, err := token.Refresh(context.Background())
```

component/azstorage/config.go
```diff
@@ -390,12 +390,26 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 		az.stConfig.authConfig.SASKey = sanitizeSASKey(opt.SaSKey)
 	case EAuthType.MSI():
 		az.stConfig.authConfig.AuthMode = EAuthType.MSI()
-		if opt.ApplicationID == "" && opt.ResourceID == "" {
+		if opt.ApplicationID == "" && opt.ResourceID == "" && opt.ObjectID == "" {
```
gapra-msft, line 393: this should be or now right? and change the error string below

component/azstorage/config.go
```diff
@@ -390,12 +390,26 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
...
 			//lint:ignore ST1005 ignore
 			return errors.New("Application ID and Resource ID not provided")
 		}
+		v := make(map[string]bool, 3)
+		if opt.ClientID != "" {
+			v[opt.ClientID] = true
+		}
+		if opt.ObjectID != "" {
+			v[opt.ObjectID] = true
+		}
```
vibhansa-msft, line 403: What is use of these flags?

component/azstorage/config.go
```diff
@@ -406,6 +420,8 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 		az.stConfig.authConfig.ClientID = opt.ClientID
 		az.stConfig.authConfig.ClientSecret = opt.ClientSecret
 		az.stConfig.authConfig.TenantID = opt.TenantID
+		az.stConfig.authConfig.ObjectID = opt.ObjectID
```
vibhansa-msft, line 423: This shall be part of MSI not SPN.

component/azstorage/config.go
tasherif-msft: no this is just saying if theyre all nil and we're msi then throw

component/azstorage/config.go
tasherif-msft: this is just a validation step to confirm only one of them is provided

component/azstorage/config.go
```diff
@@ -390,10 +390,23 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 		az.stConfig.authConfig.SASKey = sanitizeSASKey(opt.SaSKey)
 	case EAuthType.MSI():
 		az.stConfig.authConfig.AuthMode = EAuthType.MSI()
-		if opt.ApplicationID == "" && opt.ResourceID == "" {
+		if opt.ApplicationID == "" && opt.ResourceID == "" && opt.ObjectID == "" {
 			//lint:ignore ST1005 ignore
 			return errors.New("Application ID and Resource ID not provided")
 		}
+		v := make(map[string]bool, 3)
```
vibhansa-msft, line 397: if opt.ApplicationID == "" && opt.ResourceID == "" && opt.ObjectID == "" {
this check is already validating the same so I think this is just duplicate check and can be removed.

component/azstorage/config.go
```diff
@@ -406,6 +419,8 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 		az.stConfig.authConfig.ClientID = opt.ClientID
 		az.stConfig.authConfig.ClientSecret = opt.ClientSecret
 		az.stConfig.authConfig.TenantID = opt.TenantID
+		az.stConfig.authConfig.ObjectID = opt.ObjectID
```
vibhansa-msft, line 422: This is still under SPN, we need to move it to MSI section or it shall be here ?

component/azstorage/config.go
tasherif-msft: no that check is validating if nothing got passed and we're type msi the other is validating if we passed more than 1 type of msi credential. I will combine both but they do different checks

component/azstorage/config.go
```diff
@@ -391,13 +391,22 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
...
+		if opt.ApplicationID != "" {
+			v[opt.ApplicationID] = true
+		}
+		if opt.ObjectID != "" {
+			v[opt.ObjectID] = true
+		}
+		if opt.ResourceID != "" {
+			v[opt.ResourceID] = true
+		}
+		if len(v) != 1 {
```
tusharg1993, line 404: This should be `> 1` instead of `!= 1`. Not specifying any of the three values (AppId, ObjId, or ResId) is also a valid scenario for MSI with System assigned identity. In that case, the IMDS endpoint gets a token for VM's identity. 

Here is a [reference](https://learn.microsoft.com/en-us/azure/active-directory/managed-identities-azure-resources/managed-identities-faq#what-identity-will-imds-default-to-if-dont-specify-the-identity-in-the-request). For example [oAuthTokenManager.go](https://github.com/Azure/azure-storage-azcopy/blob/main/common/oauthTokenManager.go#L656)

component/azstorage/config.go
vibhansa-msft: did not get that, here in if condition we were checking either of them shall be present. In the below code where we are adding items to map, we are checking exactly one of them shall be there then they both are the same. If so, if condition is a better check, then constructing a map and add values.
If condition is expected to check multiple values can be provided out of these three things, then also if condition will pass and eventually, we will end up passing down all three to the sdk and it will decide which one to prioritize.
Not able to understand the difference between if and the map length checks.

component/azstorage/config.go
```diff
@@ -391,13 +391,22 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
...
+			v[opt.ApplicationID] = true
+		}
+		if opt.ObjectID != "" {
+			v[opt.ObjectID] = true
+		}
+		if opt.ResourceID != "" {
+			v[opt.ResourceID] = true
+		}
+		if len(v) > 1 {
+			return errors.New("client ID, object ID and MSI resource ID are mutually exclusive and one need to be provided")
```
gapra-msft, line 405:
```suggestion
			return errors.New("client ID, object ID and MSI resource ID are mutually exclusive and zero or one of the inputs need to be provided")
```

component/azstorage/config.go
gapra-msft: I think Vikas's comment makes sense, instead of creating a map, we can just convert this check into some if statements

gapra-msft (commented):
left a few comments

component/azstorage/config.go
tasherif-msft: I fail to see the point still. why is multiple if conditions (since we can't XOR here) better than the map? I actually think the if condition(s) will be not as readable.

gapra-msft: you could create a function that takes in 3 strings as input and returns whether only the first is set so like a != "" && b == ""  && c== "" and then call the function like this (func(a, b, c) || func(b, a, c) || func(c, a, b))

tasherif-msft: why is that better than a map

component/azstorage/config.go
```diff
@@ -391,13 +408,13 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 		az.stConfig.authConfig.SASKey = sanitizeSASKey(opt.SaSKey)
 	case EAuthType.MSI():
 		az.stConfig.authConfig.AuthMode = EAuthType.MSI()
-		if opt.ApplicationID == "" && opt.ResourceID == "" {
-			//lint:ignore ST1005 ignore
-			return errors.New("Application ID and Resource ID not provided")
+		err := validateMsiCOnfig(opt)
```
gapra-msft, line 411:
```suggestion
		err := validateMsiConfig(opt)
```

component/azstorage/utils.go
```diff
@@ -575,7 +575,7 @@ func getMD5(fi *os.File) ([]byte, error) {
 }
 
 func autoDetectAuthMode(opt AzStorageOptions) string {
-	if opt.ApplicationID != "" || opt.ResourceID != "" {
+	if opt.ApplicationID != "" || opt.ResourceID != "" || opt.ObjectID != "" {
```
vibhansa-msft, line 578: We are saying if nothing is set then also MSI can work. Do we want to use msi as default or we can skip that case here and ask user to explicitly configure the auth mode to be msi in such case (where no config is available and vm is having the identity assigned that can access the storage account).

vibhansa-msft (approved):

component/azstorage/utils.go
tasherif-msft: Default to msi if nothing is passed has been committed

gapra-msft (approved):

souravgupta-msft (approved):

PR #930, opened 2022-10-06
SAS token format explanation in systemd example

karol-gro:
Adding explanation on how to escape SAS token in systemd unit file.
Things to keep in mind with SAS tokens and systemd:
- variable value should be in double quotes "" due to equals sign inside token
- percent sign is a special characted in systemd, so should be escaped as double percent (% -> %%)
- SAS token for blobfuse starts with question mark (?), even though Azure Portal generates it without it

### DIFF
--- a/systemd/without-config-file/blobfuse.service
+++ b/systemd/without-config-file/blobfuse.service
@@ -32,8 +32,9 @@ Environment=AZURE_STORAGE_ACCESS_KEY=<account.key>
 
 # Uncomment the block below when using SAS token auth
 # Environment=AZURE_STORAGE_AUTH_TYPE=SAS
-# Specifies the storage account key to use for authentication.
-# Environment=AZURE_STORAGE_SAS_TOKEN=<sas.token>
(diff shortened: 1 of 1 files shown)


vibhansa-msft:
We are not taking any change requests in legacy code at this time. We are working on the next generation of blobfuse aka blobfuse2. I have taken your comments and edited service file there. Blobfuse2 is in preview mode, kindly give it a try and let us know the feedback.

karol-gro:
Actually, I'm already using blobfuse2, but I couldn't find its systemd examples (now I can see it's on another branch in setup dir), so I was adapting this file...

Anyway, thanks for improving that file. So far I had no troubles with blobfuse2

vibhansa-msft:
Yes, blobfuse2 code base resides in "main" branch of this repo. In that branch under "setup" folder we have similar service file and I have updated it as per your comments.
It's great to hear that blobfuse2 is working fine for you so far.  Such user feedback helps us plan better for the GA of blobfuse2.
Thanks for the feedback.

PR #939, opened 2022-10-13
Fixed a bug where mount all required a config file

gapra-msft:


vibhansa-msft (approved):

souravgupta-msft (approved):

tasherif-msft (approved):

cmd/mount.go
```diff
@@ -558,6 +558,12 @@ func init() {
 	mountCmd.PersistentFlags().StringVar(&options.PassPhrase, "passphrase", "",
 		"Key to decrypt config file. Can also be specified by env-variable BLOBFUSE2_SECURE_CONFIG_PASSPHRASE.\nKey length shall be 16 (AES-128), 24 (AES-192), or 32 (AES-256) bytes in length.")
 
+	mountCmd.PersistentFlags().String("log-type", "syslog", "Type of logger to be used by the system. Set to syslog by default. Allowed values are silent|syslog|base.")
```
vibhansa-msft, line 561: If we start taking logging type as input param then we need to take file-path for base logger also as input param.

vibhansa-msft (approved):

cmd/mount.go
gapra-msft: this is already added

souravgupta-msft (approved):

PR #947, opened 2022-10-19
Retry listing if server returned back empty list but a new marker was given

vibhansa-msft:
Potential fix for #942


TSG.md
```diff
@@ -211,6 +217,20 @@ In such cases we have seen mounting through /etc/fstab works, because that execu
...
+By default `writeback-cache` is enabled for libfuse3 and this may result in append/write operations to fail. Either you can disable writeback-cache, which might hurt the performance or you can configure blobfuse2 to ignore open flags given by user and make it work with writeback-cache.
+
+To disable writeback-cache : Add `disable-writeback-cache: true` under libfuse section in your config file.
+
+To make it work with writeback-cache : Add `ignore-open-flags: true` under libfuse section in your config file. 
+
+
+**9. Unable to list files/directories for non-HNS (flat-namespace) accounts**
+
+For non-HNS accounts blobfuse expects special directory marker files to exist in container to identify a directory. If these files do not exists then `virtual-directory: true` in `azstorage` section is required.
```
souravgupta-msft, line 232:
```suggestion
For non-HNS accounts blobfuse expects special directory marker files to exist in container to identify a directory. If these files do not exist then `virtual-directory: true` in `azstorage` section is required.
```

TSG.md
```diff
@@ -1,22 +1,28 @@
...
 **1. Error: fusermount: failed to open /etc/fuse.conf: Permission denied**
 
 Only the users that are part of the group fuse, and the root user can run fusermount command. In order to mitigate this add your user to the fuse group.
 
 ```sudo addgroup <user> fuse```
 
-**2. failed to mount : failed to authenticate credentials for azstorage**
+**2. Error: mount command successful but log shows 'Failed to init fuse'**
+
+If are you using 'allow-other: true' config then make sure  user_allow_other is enabled in /etc/fuse.conf file. By default /etc/fuse.conf will have this option disabled we just need to enable it and save the file.
```
souravgupta-msft, line 23:
```suggestion
If are you using 'allow-other: true' config then make sure `user_allow_other` is enabled in /etc/fuse.conf file. By default /etc/fuse.conf will have this option disabled we just need to enable it and save the file.
```

setup/baseConfig.yaml
```diff
@@ -1,4 +1,27 @@
...
+#   3. 'stream' and 'file_cache' can not co-exist and config file shall have only one of them based on your use case.
+#   4. By default log level is set to 'log_warning' level and are redirected to syslog. 
+#      Either use 'base' logging or syslog filters to redirect logs to separate file.
+#      To install syslog filter follow below steps:        
+#         sudo cp setup/11-blobfuse2.conf /etc/rsyslog.d/
+#         sudo cp setup/blobfuse2-logrotate /etc/logrotate.d/
+#         sudo service rsyslog restart
+#   5. For non-HNS (flat namespace) accounts blobfuse expects special directory marker files to 
+#      exists in container to identify a directory. 
+#      If these files do not exists in container then 'virtual-directory: true' in 'azstorage' section is must
```
souravgupta-msft, line 15:
```suggestion
#      If these files do not exist in container, then 'virtual-directory: true' in 'azstorage' section is required
```

setup/baseConfig.yaml
```diff
@@ -1,4 +1,27 @@
...
+#   6. By default 'writeback-cache' is enabled for libfuse3 and this may result in append/write operations to fail.
+#      Either you can disable 'writeback-cache', which might hurt the performance
+#      or you can configure blobfuse2 to ignore open flags given by user and make it work with ''writeback-cache'.
+#      'libfuse' sectoin below has both the configurations.
+#   7. If are you using 'allow-other: true' config then make user_allow_other is enable in /etc/fuse.conf file as
```
souravgupta-msft, line 20:
```suggestion
#   7. If are you using 'allow-other: true' config then make sure user_allow_other is enabled in /etc/fuse.conf file as 
```

component/azstorage/azstorage.go
```diff
@@ -298,6 +298,21 @@ func (az *AzStorage) StreamDir(options internal.StreamDirOptions) ([]*internal.O
 
 	log.Debug("AzStorage::StreamDir : Retrieved %d objects with %s marker for Path %s", len(new_list), options.Token, path)
 
+	if new_marker != nil && *new_marker != "" {
+		log.Debug("AzStorage::StreamDir : next-marker %s for Path %s", *new_marker, path)
+		if len(new_list) == 0 {
+			/* In some customer sceanio we have seen that new_list is empty but marker is not empty
```
gapra-msft, line 304:
```suggestion
			/* In some customer scenario we have seen that new_list is empty but marker is not empty
```

component/azstorage/datalake.go
```diff
@@ -461,17 +462,31 @@ func (dl *Datalake) List(prefix string, marker *string, count int32) ([]*interna
...
+		var mode fs.FileMode
+		if pathInfo.Permissions != nil {
+			mode, err = getFileMode(*pathInfo.Permissions)
+			if err != nil {
+				log.Err("Datalake::List : Failed to get file mode for %s [%s]", *pathInfo.Name, err.Error())
+				m := ""
+				return pathList, &m, err
+			}
+		} else {
+			// This happens when a blob account is mounted with type:adls
```
gapra-msft, line 474: why are we adding this condition?

gapra-msft (approved):

component/azstorage/datalake.go
vibhansa-msft: In case of wrong configuration blobfuse2 binary is crashing (coring out). In any case we shall just go down gracefully or generate error logs. Crash is not a good thing so trying to avoid it here.

PR #953, opened 2022-10-28
Check github release page to get latest version

vibhansa-msft:
…ntainer

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -37,7 +37,6 @@ import (
 	"encoding/xml"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"strings"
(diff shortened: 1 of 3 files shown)


common/version.go
```diff
@@ -39,9 +39,11 @@ import (
 	"strings"
 )
 
-const Blobfuse2ListContainerURL = "https://blobfuse2.blob.core.windows.net/release"
-
-const BlobFuse2WarningsURL = "https://github.com/Azure/azure-storage-fuse/wiki/9.-Security-Warnings"
+const (
+	Blobfuse2LatestReleaseURL      = "https://github.com/Azure/azure-storage-fuse/releases/latest"
+	Blobfuse2HasSecurityWarningURL = "https://github.com/Azure/azure-storage-fuse/securitywarnings/"
```
souravgupta-msft, line 44: This should be either be one of these:
1. `https://github.com/Azure/azure-storage-fuse/blob/main/securitywarnings/`
2. `https://raw.githubusercontent.com/Azure/azure-storage-fuse/main/securitywarnings/`

cmd/root.go
```diff
@@ -92,30 +91,44 @@ func checkVersionExists(versionUrl string) bool {
 
 // getRemoteVersion : From public container get the latest blobfuse version
 func getRemoteVersion(req string) (string, error) {
-	resp, err := http.Get(req)
+	// 'release/latest' page redirects URL to latest release tag
```
gapra-msft, line 94: This code won't work if we just released a blobfuse v1 patch right?

vibhansa-msft: we do not see any clear solution for this problem.

vibhansa-msft: either there is api to get the latest version string or there is api to list all tags/releases which we need to sort in code in descending order and figure out the latest v2 release.

Issue #964, opened 2022-11-04
blobfuse2 renaming file sometimes throws an error (but still renames).


Mmdixon:
### Which version of blobfuse was used?
(284001979dc16cc942790b68548591e7e70b8294)
`blobfuse2 version: 2.0.0-preview.4` 

### Which OS distribution and version are you using?
`Debian GNU/Linux 11 (bullseye)`

### If relevant, please share your mount command.
`blobfuse2 mount /blob2/cache --config-file=/tmp/config.yaml`
```
default-working-dir: /tmp/.blobfuse2

logging:
  level: log_debug
  file-path: /tmp/.blobfuse2/blobfuse2.log

components:
  - libfuse
  - file_cache
  - attr_cache
  - azstorage

libfuse:
  attribute-expiration-sec: 120
  entry-expiration-sec: 120
  negative-entry-expiration-sec: 240
  disable-writeback-cache: true

file_cache:
  path: /tmp/blob2/cache
  timeout-sec: 86400
  allow-non-empty-temp: true
  cleanup-on-start: true

attr_cache:
  timeout-sec: 86400

azstorage:
  type: block
  mode: key
  max-retries: 30
  max-retry-timeout-sec: 3600
  retry-backoff-sec: 2
  max-retry-delay-sec: 30
  virtual-directory: true
```

### What was the issue encountered?
I have a hard to reproduce issue where when I create a file, write to it, then rename it, I get an error saying that the file could not be renamed.
e.g. `OSError: [Errno 5] Input/output error: '/blob2/cache/CC/dedupe/v0.2.1/cc-6/extract_documents/fr/2021-39/filter_v2/partition.4271.ndjson.incomplete' -> '/blob2/cache/CC/dedupe/v0.2.1/cc-6/extract_documents/fr/2021-39/filter_v2/partition.4271.ndjson'`
However, the file actually was successfully renamed and the error was more about it couldn't delete the original file because it no longer exists (because it was renamed). See logs for details.

### Have you found a mitigation/solution?
I just try again, and it sometimes makes progress.

### Please share logs if available.
The failure case
```
Thu Nov  3 14:50:39 UTC 2022 : blobfuse2[91430] : LOG_TRACE [file_cache.go (1149)]: FileCache::GetAttr : CC/dedupe/v0.2.1/cc-6/extract_documents/fr/2021-39/filter_v2/partition.4271.ndjson.incomplete
[...]

PR #965, opened 2022-11-04
Add retry in rename call when deletion fails and correct retry policy

vibhansa-msft:
Potential fix for #964


gapra-msft (approved):

gapra-msft:
Please add the change in retry defaults to the baseConfig and the Changelog as a "breaking change"

PR #966, opened 2022-11-04
fixed download size will get MAX if start offset > object size

heiaun-jack:
Hi Blobfuse team, 
  Found a case that might be missed: when calling BlobStreamer::GetBlock if start_offset > obj->GetSize(), download_size will get a maximum of download_size's type.
 It's from Blobfuse 1.4.5, if you gonna deprecate the v1 version?

I hope it helps thx


vibhansa-msft:
Thanks for the contribution and identifying the bug. We are about to deprecate blobfuse 1.x as we have next generation of blobfuse aka blobfuse2 in preview mode already. As part of migration plan we are not taking any feature requests or PRs in 1.x. Can you try blobfuse2 and check it works well for your scenario or not.

PR #967, opened 2022-11-04
[fixed] unexpected maximum might incur security issues.

heiaun-jack:
Hi Blobfuse team,
Found a case that might be missed: 
when calling `BlobStreamer::GetBlock` if `start_offset `> `obj->GetSize()`, `download_size `will get a maximum of `download_size`'s type.

Like when `block_size` equals 512 * 1024 * 1024, `offset` equals 512 * 1024 * 1024 + 100, `obj->GetSize()` equals 10. 
`start_offset ` will get 512 * 1024 * 1024 - 100 and it gonna execute` download_size = obj->GetSize() - start_offset;` .`obj->GetSize() - start_offset` logically would get negative. 
But, if we assign the negative value to `download_size `of type uint64_t, `download_size ` will be converted to UINT64_MAX. 

I think this should be an unexpected behavior. The risk of memory leak might incur security issues.

I hope it helps thx,
jack

### DIFF
--- a/blobfuse/src/BlobStreamer.cpp
+++ b/blobfuse/src/BlobStreamer.cpp
@@ -102,13 +102,14 @@ BlobBlock* BlobStreamer::GetBlock(const char* file_name, uint64_t offset, Stream
             obj->GetSize() > 0) 
         {
             if ((start_offset + download_size) > obj->GetSize()) {
+                download_size = obj->GetSize() - start_offset;
+                
                 if (download_size == 0 || obj->GetSize() <= start_offset) {
                     obj->UnLock();
                     syslog(LOG_ERR, "Failed to download block of %s with offset %lu.  Errno : (Out of range).\n", file_name, start_offset);
                     errno = 416;
                     return NULL;
                 }
-                download_size = obj->GetSize() - start_offset;
             }
         }
 
--- a/blobfuse/src/BlobStreamer.cpp
+++ b/blobfuse/src/BlobStreamer.cpp
@@ -101,8 +101,8 @@ BlobBlock* BlobStreamer::GetBlock(const char* file_name, uint64_t offset, Stream
         if (download_size > MAX_BLOCK_SIZE_FOR_SINGLE_READ && 
             obj->GetSize() > 0) 
         {
-            if ((start_offset + download_size) > obj->GetSize()) {                
-                if (obj->GetSize() <= start_offset) {
+            if ((start_offset + download_size) > obj->GetSize()) {
+                if (download_size == 0 || obj->GetSize() <= start_offset) {
                     obj->UnLock();
                     syslog(LOG_ERR, "Failed to download block of %s with offset %lu.  Errno : (Out of range).\n", file_name, start_offset);
                     errno = 416;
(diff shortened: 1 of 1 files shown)


vibhansa-msft (approved):

vibhansa-msft:
Are you hitting this case or just found this as a safety check?

blobfuse/src/BlobStreamer.cpp
```diff
@@ -101,14 +101,14 @@ BlobBlock* BlobStreamer::GetBlock(const char* file_name, uint64_t offset, Stream
...
-                download_size = obj->GetSize() - start_offset;
-                if (download_size == 0) {
+            if ((start_offset + download_size) > obj->GetSize()) {                
+                if (obj->GetSize() <= start_offset) {
                     obj->UnLock();
                     syslog(LOG_ERR, "Failed to download block of %s with offset %lu.  Errno : (Out of range).\n", file_name, start_offset);
                     errno = 416;
                     return NULL;
                 }
+                download_size = obj->GetSize() - start_offset;
```
vibhansa-msft, line 111: If download size becomes 0 post this calculation, then also we need to error out as done in original code.

heiaun-jack: Added a condition `download_size ==0.`, then two cases to make `download size ` zero will give out-of-range errors

heiaun-jack:
> Are you hitting this case or just found this as a safety check?

Want to tune the performance via the option (storage/ cache/ stream), and want to know better how the options work, then start reading the code. 

I came across the issue and think it is necessary to point it out.

blobfuse/src/BlobStreamer.cpp
vibhansa-msft: if (download_size == 0 || obj->GetSize() <= start_offset) {
is in correct check as dowload-size is being calculated at line 111. We need to put similar condition check and return failure post line 111

vibhansa-msft:
> > Are you hitting this case or just found this as a safety check?
> 
> Want to tune the performance via the option (storage/ cache/ stream) and want to know better how the options work, then start reading the code.
> 
> I came across the issue and think it is necessary to point it out.

Nice catch. We can put the fix once review comments have been addressed. However, one thing to note here is that we are working on next generation of blobfuse aka blobfuse2 and we are not planning to publish any new release for our legacy code, unless there is a high severity bug. Feature requests we stopped taking around a year back, so slowly we want all our customers to move to blobfuse2.

heiaun-jack:
> > > Are you hitting this case or just found this as a safety check?
> > 
> > 
> > Want to tune the performance via the option (storage/ cache/ stream) and want to know better how the options work, then start reading the code.
> > I came across the issue and think it is necessary to point it out.
> 
> Nice catch. We can put the fix once review comments have been addressed. However, one thing to note here is that we are working on next generation of blobfuse aka blobfuse2 and we are not planning to publish any new release for our legacy code, unless there is a high severity bug. Feature requests we stopped taking around a year back, so slowly we want all our customers to move to blobfuse2.

done, thanks for sharing this

blobfuse/src/BlobStreamer.cpp
heiaun-jack: done

PR #969, opened 2022-11-07
Added support for file share

gapra-msft:


vibhansa-msft (approved):

andyzhangx:
As I know, azure file share only supports smb and nfs v4.1 protocols now, if this file share feature is supported by blobfuse2 mount, do you know whether managed identity would be supported by using blobfuse2 mount. @vibhansa-msft

vibhansa-msft:
This PR is only to add support for Azure Files using SAS and Key. OAuth support is not there in Go-SDK itself as of now and Blobfuse is using a pretty old Go-SDK. Once we move to Go 1.19 there will be effort to move to new Go-SDK as well in future and once we are there and by then if Files SDK adds OAuth support as well we will get it. However, this will take some time. Is there a need in AKS to have files support from Blobfuse end ? If so, I can put across this to our PM and we can think on this on priority.

andyzhangx:
> This PR is only to add support for Azure Files using SAS and Key. OAuth support is not there in Go-SDK itself as of now and Blobfuse is using a pretty old Go-SDK. Once we move to Go 1.19 there will be effort to move to new Go-SDK as well in future and once we are there and by then if Files SDK adds OAuth support as well we will get it. However, this will take some time. Is there a need in AKS to have files support from Blobfuse end ? If so, I can put across this to our PM and we can think on this on priority.

@vikama-microsoft thanks for the clarification, I think it's not urgent, AKS team are exploring non key way(e.g. using kerberos auth) to mount azure file share, since this PR is still using the same password auth, then it's the same way as smb driver now.

PR #975, opened 2022-11-14
Subdirectory as cli option and support for nonempty libfuse option

vibhansa-msft:
- Allow mounting on a nonempty directory
- Allow subdirectory to be configured using cli option


cmd/mountv1.go
```diff
@@ -293,6 +297,8 @@ func parseFuseConfig(config []string) error {
 			bfv2AllowOtherOption = true
 		} else if v == "allow_other=false" {
 			bfv2AllowOtherOption = false
+		} else if v == "nonempty" {
```
gapra-msft, line 300: Should we add this to the mountv1 test as well?

gapra-msft (approved):

setup/baseConfig.yaml
```diff
@@ -29,6 +29,7 @@ foreground: true|false <run blobfuse2 in foreground or background>
 # Common configurations
 read-only: true|false <mount in read only mode - used for Streaming and FUSE>
 allow-other: true|false <allow other users to access the mounted directory - used for FUSE and File Cache>
+nonempty: true|false <allow mounting on non-empty directory - used for FUSE>
```
souravgupta-msft, line 32: We should mention this somewhere in the readme or migration.md files

souravgupta-msft (approved):

setup/baseConfig.yaml
vibhansa-msft: this was not a seperate config in v1, it was a "-o' suboption only and even in v2 we plan to keep it that way instead of explicitly saying you put this string in config. For that reason, have not made any explicit mention of this in other places.

cmd/mountv1.go
vibhansa-msft: In v1 test we do not have any mount related tests its only the config file related tests, so I skipped it. Let me see if we can add a test case just to create config with this option.

PR #1029, opened 2023-01-11
Set ignore-open-flags to true by default

### LINKED ISSUE #1015: mount failed on ubuntu 22.04(aks 1.25.2)
blobfuse2 version 2.0.1

Ubuntu 22.04

`blobfuse2 mount /var/lib/kubelet/plugins/kubernetes.io/csi/blob.csi.azure.com/a45a1958a856c3aa56d78a5b485feb218d61673d0ed773a2367adaa6bf90fc4c/globalmount -o allow_other --file-cache-timeout-in-seconds=120 --use-attr-cache=true --cancel-list-on-mount-seconds=10 -o attr_timeout=120 -o entry_timeout=120 -o negative_timeout=120 --log-level=LOG_WARNING --cache-size-mb=1000 --container-name=pvc-45e0990f-bd6b-4cbe-baed-2e927981c7db --pre-mount-validate=true --use-https=true --empty-dir-check=false --tmp-path=/mnt/mc_weizhichen_weizhichen125_eastus#fuse74833b66893742a894d#pvc-45e0990f-bd6b-4cbe-baed-2e927981c7db##default#`

Can't write to the file in the mountpoint, error: 
```shell
root@aks-nodepool1-16062421-vmss000001:/# ls /var/lib/kubelet/plugins/kubernetes.io/csi/blob.csi.azure.com/a45a1958a856c3aa56d78a5b485feb218d61673d0ed773a2367adaa6bf90fc4c/globalmount
test
root@aks-nodepool1-16062421-vmss000001:/# echo "123" >> /var/lib/kubelet/plugins/kubernetes.io/csi/blob.csi.azure.com/a45a1958a856c3aa56d78a5b485feb218d61673d0ed773a2367adaa6bf90fc4c/globalmount/test
bash: /var/lib/kubelet/plugins/kubernetes.io/csi/blob.csi.azure.com/a45a1958a856c3aa56d78a5b485feb218d61673d0ed773a2367adaa6bf90fc4c/globalmount/test: Invalid argument
```

Mount using the same mount options can **succeed on Ubuntu 18.04(aks < 1.25)**

```Jan  4 07:03:49 aks-nodepool1-16062421-vmss000001 blobfuse2[47882]: LOG_ERR [libfuse_handler.go (633)]: Libfuse::libfuse_open : Flag (8401) not supported to open test when write back cache is on. Pass --disable-writeback-cache=true or --ignore-open-flags=true via CLI```

### PR DESCRIPTION
**Pull request by souravgupta-msft**
Fixes #1015 

By default, on fuse3 we enable write-back-cache at kernel level. To populate kernel cache, kernel issues a read command on your file. Now assume you opened a file handle in WRONLY or APPEND mode, which means handle does not have read permissions. When kernel tries to read file using such handle the read operation fails and in turn users' Open operation returns back with error. To fix this (as write-back-cache boosts performance) we added a flag to ignore the open flag. This means even if user opens a file in write/append mode we always add read permissions as well.
So, if you are not able to open a file in write only or append only mode but want to use cache you need to now give ignore-open-flags=true always. Otherwise disable the write-back cache. We made this a configurable flag because if user goes and checks the permission on the handle it will not be the same as he expects (as we added read permissions). Just to make the expectation right we ask user to explicitly set this flag.
Issues comes from CSI driver which mostly goes with default config. write-back-cache enabled and ignore-open-flag set to false. Which means append file or write only open file are bound to fail. And not just CSI driver but any other user by default will hit failure with this default setting.
To resolve this, we are now turning ignore-open-flags always true.

### DIFF
--- a/component/libfuse/libfuse.go
+++ b/component/libfuse/libfuse.go
@@ -220,7 +220,7 @@ func (lf *Libfuse) Configure(_ bool) error {
 	log.Trace("Libfuse::Configure : %s", lf.Name())
 
 	// >> If you do not need any config parameters remove below code and return nil
-	conf := LibfuseOptions{}
+	conf := LibfuseOptions{IgnoreOpenFlags: true}
 	err := config.UnmarshalKey(lf.Name(), &conf)
 	if err != nil {
 		log.Err("Libfuse::Configure : config error [invalid config attributes]")
@@ -300,6 +300,6 @@ func init() {
 	config.BindPFlag(compName+".fuse-trace", debug)
 	debug.Hidden = true
 
-	ignoreOpenFlags := config.AddBoolFlag("ignore-open-flags", false, "Ignore unsupported open flags (APPEND, WRONLY) by blobfuse when writeback caching is enabled.")
+	ignoreOpenFlags := config.AddBoolFlag("ignore-open-flags", true, "Ignore unsupported open flags (APPEND, WRONLY) by blobfuse when writeback caching is enabled.")
 	config.BindPFlag(compName+".ignore-open-flags", ignoreOpenFlags)
 }
--- a/setup/baseConfig.yaml
+++ b/setup/baseConfig.yaml
@@ -1,6 +1,6 @@
 # MUST READ : 
 #   If you are creating a blobfuse2 config file using this kindly take care of below points 
-#   1. All boolean configs (true|false config) are set to 'false' by default. 
+#   1. All boolean configs (true|false config) (except ignore-open-flags) are set to 'false' by default. 
 #      No need to mention them in your config file unless you are setting them to true.
 #   2. 'loopbackfs' is purely for testing and shall not be used in production configuration.
 #   3. 'stream' and 'file_cache' can not co-exist and config file shall have only one of them based on your use case.
@@ -62,8 +62,8 @@ libfuse:
   negative-entry-expiration-sec: <time kernel can cache attributes of non existent paths (in sec). Default - 120 sec>
   fuse-trace: true|false <enable libfuse api trace logs for debugging>
   extension: <physical path to extension library>
-  disable-writeback-cache: true|false <disallow libfuse to buffer write requests if you must strictly open files in O_WRONLY or O_APPEND mode. alternatively, you can ignore-open-flags.>
-  ignore-open-flags: true|false <ignore the append and write only flag since O_APPEND and O_WRONLY is not supported with writeback caching. alternatively, you can disable-writeback-cache.>
+  disable-writeback-cache: true|false <disallow libfuse to buffer write requests if you must strictly open files in O_WRONLY or O_APPEND mode. alternatively, you can set ignore-open-flags.>
+  ignore-open-flags: true|false <ignore the append and write only flag since O_APPEND and O_WRONLY is not supported with writeback caching. alternatively, you can disable-writeback-cache. Default value is true>
  
   # Streaming configuration
 stream:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 ## 2.0.2 (WIP)
+**Breaking Changes**
+- Default value of `ignore-open-flags` config parameter changed to `true`
+
 **Bug Fixes**
 - [#999](https://github.com/Azure/azure-storage-fuse/issues/999) Upgrade dependencies to resolve known CVEs.
 - [#1002](https://github.com/Azure/azure-storage-fuse/issues/1002) In case version check fails to connect to public container, dump a log to check network and proxy settings.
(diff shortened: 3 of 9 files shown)

### CODE REVIEW

File: component/libfuse/libfuse.go [hunk 1]
```diff
@@ -220,7 +220,7 @@ func (lf *Libfuse) Configure(_ bool) error {
 	log.Trace("Libfuse::Configure : %s", lf.Name())
 
 	// >> If you do not need any config parameters remove below code and return nil
-	conf := LibfuseOptions{}
+	conf := LibfuseOptions{IgnoreOpenFlags: true}
```
gapra-msft, line 223: Why are we setting to true here? Will this cause the customers decision to be overridden?

souravgupta-msft: The default value of the `ignore-open-flags` will be `true` when it is not specified in the config as well as the CLI.
It is set to `false` when the flag is either specified as false in the config or the CLI. I have tested the combinations and it is working as expected. I will try to add some test cases for the different combinations around this.

setup/baseConfig.yaml
```diff
@@ -62,8 +62,8 @@ libfuse:
   negative-entry-expiration-sec: <time kernel can cache attributes of non existent paths (in sec). Default - 120 sec>
   fuse-trace: true|false <enable libfuse api trace logs for debugging>
   extension: <physical path to extension library>
-  disable-writeback-cache: true|false <disallow libfuse to buffer write requests if you must strictly open files in O_WRONLY or O_APPEND mode. alternatively, you can ignore-open-flags.>
-  ignore-open-flags: true|false <ignore the append and write only flag since O_APPEND and O_WRONLY is not supported with writeback caching. alternatively, you can disable-writeback-cache.>
+  disable-writeback-cache: true|false <disallow libfuse to buffer write requests if you must strictly open files in O_WRONLY or O_APPEND mode. alternatively, you can set ignore-open-flags.>
```
tasherif-msft, line 65: can we mention the default here

setup/baseConfig.yaml
```diff
@@ -62,8 +62,8 @@ libfuse:
...
+  ignore-open-flags: true|false <ignore the append and write only flag since O_APPEND and O_WRONLY is not supported with writeback caching. alternatively, you can disable-writeback-cache. Default value is true>
```
tasherif-msft, line 66: so we're not effectively always ignoring the flags that come in with the open call, do you remember what the reasoning was for why we had to introduce this?

setup/baseConfig.yaml
souravgupta-msft: The default is false for all the boolean flags except the `ignore-open-flags` which we are changing to true.

setup/baseConfig.yaml
souravgupta-msft: This has mainly to do with the writeback caching in fuse3. When enabled (which is by default), it causes append/write operations to fail. So, the workaround this is to set the `disable-writeback-cache: true` in config so that libfuse doesn't buffer the write requests.
Alternate option is to set `ignore-open-flags: true` where we ignore the append and write only flags. Refer [this](https://github.com/Azure/azure-storage-fuse/blob/sourav/v2/ignoreOpenFlags/component/libfuse/libfuse_handler.go#L625).

vibhansa-msft: Just to share more context on why we added and now why we are changing the default value
Bydefault on fuse3 we enable write-back-cache at kernel level. To populate kernel cache, kernel issues a read command on your file. Now assume you opened a file handle in WRONLY or APPEND mode, which means handle does not have read permissions. When kernel tries to read file using such handle the read operation fails and in turn users Open operation returns back with error. To fix this (as write-back-cache boosts performance) we added a flag to ignore the open flag. This means even if user opens a file in write/append mode we always add read permissions as well.
So if you are not able to open a file in write only or append only mode but want to use cache you need to now give ignore-open-flags=true always. Otherwise disable the write-back cache. We made this a configurable flag because if user goes and checks the permission on the handle it will not be the same as he expects (as we added read permissions). Just to make the expecation right we ask user to explicitly set this flag.
Issues comes from CSI driver which mostly goes with default config. write-back-cache enable and ignore-open-flag set to false. Which means append file or write only open file are bound to fail. And not just CSI driver but any other user by default will hit failure with this default setting.
To resolve this we are now turning ignore-open-flags always true.

tasherif-msft: @vibhansa-msft perfect I added this comment in the pr description for future referenece

vibhansa-msft (approved):

tasherif-msft (approved):

component/libfuse/libfuse.go
souravgupta-msft: Added relevant test cases.

PR #1052, opened 2023-02-07
Create work dir if it does not exists

vibhansa-msft:
Create work dir if it does not exists


tasherif-msft:
was this related to a github issue?

tasherif-msft (approved):

gapra-msft (approved):

vibhansa-msft:
> was this related to a github issue?

Not a github issue, but recently we made a change related to $HOME directory and because of that change we observed some errors in Healthmon where this directory does not existsand mount fails. Looking at that we realised earlier logging.Init() used to create the directory if it does not exist, but due to recent changes it was removed so added it back to validate() method.

PR #1071, opened 2023-02-24
Modify config files

scegg:
Removing useless environment line in systemd file.
Add notes for allow-other and nonempty in yaml file.

### DIFF
--- a/setup/baseConfig.yaml
+++ b/setup/baseConfig.yaml
@@ -31,8 +31,8 @@ foreground: true|false <run blobfuse2 in foreground or background>
 
 # Common configurations
 read-only: true|false <mount in read only mode - used for Streaming and FUSE>
-allow-other: true|false <allow other users to access the mounted directory - used for FUSE and File Cache>
-nonempty: true|false <allow mounting on non-empty directory - used for FUSE>
+allow-other: true|false <allow other users to access the mounted directory - used for FUSE and File Cache> #NOTE: Make sure user_allow_other is set in /etc/fuse.conf when set this option to true.
+#nonempty: true|false <allow mounting on non-empty directory - used for FUSE> NOTE: This function is discontinued currently. When this option specified, the value should be false always.
 
 # Dynamic profiler related configuration. This helps to root-cause high memory/cpu usage related issues.
 dynamic-profile: true|false <allows to turn on dynamic profiler for cpu/memory usage monitoring. Only for debugging, shall not be used in production>


setup/blobfuse2.service
```diff
@@ -14,39 +14,14 @@ Environment=BlobTmp=/home/AzureUser/tmpblobfuse
 # Config file path
 Environment=BlobConfigFile=/home/AzureUser/azure-storage-fuse/blobfuse2.yaml
 
-# Specifies the storage account blobfuse targets.
-Environment=AZURE_STORAGE_ACCOUNT=<account.name>
```
vibhansa-msft, line 18: These are required, as not all customers provide their credentials in config file.

setup/baseConfig.yaml
```diff
@@ -31,8 +31,8 @@ foreground: true|false <run blobfuse2 in foreground or background>
 
 # Common configurations
 read-only: true|false <mount in read only mode - used for Streaming and FUSE>
-allow-other: true|false <allow other users to access the mounted directory - used for FUSE and File Cache>
-nonempty: true|false <allow mounting on non-empty directory - used for FUSE>
+allow-other: true|false <allow other users to access the mounted directory - used for FUSE and File Cache> #NOTE: Make sure user_allow_other is set in /etc/fuse.conf when set this option to true.
+#nonempty: true|false <allow mounting on non-empty directory - used for FUSE> NOTE: This function is discontinued currently. When this option specified, the value should be false always.
```
vibhansa-msft, line 35: baseconfig just lists all the possible inputs and their meanings so there is no point in commenting this out here. Also default value of each flag is false only so we can keep the comment and remove the # from here.

setup/blobfuse2.service
```diff
@@ -14,39 +14,14 @@ Environment=BlobTmp=/home/AzureUser/tmpblobfuse
...
-Environment=AZURE_STORAGE_ACCESS_KEY=<account.key>
-
-# Uncomment the block below when using SAS token auth
-# Environment=AZURE_STORAGE_AUTH_TYPE=SAS
-# Specifies the SAS token to use for authentication. Due to systemd file specification, SAS token should be in double quotes, start with question mark (?) and every percent sign should be doubled (% -> %%)
-# Example: Environment=AZURE_STORAGE_SAS_TOKEN="?sp=rl&st=2022-10-06T11:53:25Z&se=2029-09-20T19:53:25Z&spr=https&sv=2021-06-08&sr=c&sig=AAAAAA%%2FYAAAA%%3D"
-# Environment=AZURE_STORAGE_SAS_TOKEN="<sas.token>"
-
-# Uncomment the line below and the parameter used in this block, when using Managed Identity auth
-# Environment=AZURE_STORAGE_AUTH_TYPE=MSI
```
vibhansa-msft, line 32: These are listed for reference in case you wish to configure env variables and choose a particular auth mode and hence these shall not be removed. All these files are for reference and not be used as is. Customer always needs to be aware of what they want and modify these files based on their needs in their local environment.

PR #1084, opened 2023-03-15
Adding support to mount entire account in one blobfuse instnace

vibhansa-msft:


tasherif-msft (commented):
Should we add a new e2e test for this? seems like a significant change

vibhansa-msft:
Not to be merged for now so its ok to ignore this PR.

Issue #1081, opened 2023-03-13
Bug: Can't mount with `-o allow_other` on WSL, and blobfuse2 print no error message and errno is 0


cvvz:
### Which version of blobfuse was used?
blobfuse2 2.0.2

### Which OS distribution and version are you using?
WSL2.0

### If relevant, please share your mount command.

blobfuse2 mount ~/myblob        **-o allow_other**  --file-cache-timeout-in-seconds=120     --use-attr-cache=true   --cancel-list-on-mount-seconds=10       -o attr_timeout=120         -o entry_timeout=120    -o negative_timeout=120         --log-level=LOG_WARNING         --cache-size-mb=1000    --container-name=pvc-0e85db3b-d632-49ab-b4ba-cc93d332a132       --pre-mount-validate=true   --use-https=true        --empty-dir-check=false         --tmp-path=/mnt/tmp

### What was the issue encountered?
mount failed, but blobfuse2 print no error message, and the errno is 0.
![image](https://user-images.githubusercontent.com/44308864/224633636-24978f6d-0e93-42e4-922e-d6341cbcc409.png)

blobfuse mount failed as well, but it prints concrete error message which help me to solve this problem.
![image](https://user-images.githubusercontent.com/44308864/224635905-62cdb054-acf9-471a-ab3b-3abeb5be04be.png)

1. There is definitely a bug in blobfuse2, since it doesn't print error message and errno is 0 when the mount operation actually failed.

2. Another question is, why mount with `-o allow_other` on WSL would fail whereas Linux won't? I checked `/etc/fuse.conf`, the content of them is the same:
```shell
cat  /etc/fuse.conf   
                                                                                                                          15:26:36  羽3m 15.082s
# The file /etc/fuse.conf allows for the following parameters:
#
# user_allow_other - Using the allow_other mount option works fine as root, in
# order to have it work as user you need user_allow_other in /etc/fuse.conf as
# well. (This option allows users to use the allow_other option.) You need
# allow_other if you want users other than the owner to access a mounted fuse.
# This option must appear on a line by itself. There is no value, just the
# presence of the option.

#user_allow_other


# mount_max = n - this option sets the maximum number of mounts.
# Currently (2014) it must be typed exactly as shown
# (with a single space before and after the equals sign).

#mount_max = 1000
```

### Have you found a mitigation/solution?
mount without **-o allow_other** on WSL

### Please share logs if available.

```shell
	cat /var/log/blobfuse2.log
[...]

PR #1087, opened 2023-03-17
Exit with non-zero status code if allow_other is used but not enabled in fuse config

vibhansa-msft:
Fix for #1081

### DIFF
--- a/component/libfuse/libfuse.go
+++ b/component/libfuse/libfuse.go
@@ -34,8 +34,10 @@
 package libfuse
 
 import (
+	"bytes"
 	"context"
 	"fmt"
+	"os/exec"
 
 	"github.com/Azure/azure-storage-fuse/v2/common"
 	"github.com/Azure/azure-storage-fuse/v2/common/config"
@@ -204,6 +206,20 @@ func (lf *Libfuse) Validate(opt *LibfuseOptions) error {
 		lf.negativeTimeout = defaultNegativeEntryExpiration
 	}
 
+	if lf.allowOther {
+		// Validate allow_other is enabled in fuse.conf file or not
+		cmd := exec.Command("bash", "-c", "cat /etc/fuse.conf | grep -i user_allow_other$")
+
+		var errb bytes.Buffer
+		cmd.Stderr = &errb
+		cliOut, err := cmd.Output()
+		data := string(cliOut)
+		if err == nil && data[0] == '#' {
+			log.Err("Libfuse::Validate : config error [user_allow_other is not enabled in /etc/fuse.conf]")
+			return fmt.Errorf("user_allow_other is not enabled in /etc/fuse.conf")
+		}
+	}
+
 	var err error
 	lf.ownerUID, lf.ownerGID, err = common.GetCurrentUser()
 	if err != nil {
@@ -254,7 +270,7 @@ func (lf *Libfuse) Configure(_ bool) error {
 	err = lf.Validate(&conf)
 	if err != nil {
 		log.Err("Libfuse::Configure : config error [invalid config settings]")
-		return fmt.Errorf("config error in %s [invalid config settings]", lf.Name())
+		return fmt.Errorf("%s config error %s", lf.Name(), err.Error())
 	}
 
 	log.Info("Libfuse::Configure : read-only %t, allow-other %t, default-perm %d, entry-timeout %d, attr-time %d, negative-timeout %d, ignore-open-flags: %t, nonempty %t",
--- a/component/libfuse/libfuse.go
+++ b/component/libfuse/libfuse.go
@@ -38,6 +38,7 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
+	"strings"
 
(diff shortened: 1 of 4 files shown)


component/libfuse/libfuse.go
```diff
@@ -204,6 +206,20 @@ func (lf *Libfuse) Validate(opt *LibfuseOptions) error {
...
 
+	if lf.allowOther {
+		// Validate allow_other is enabled in fuse.conf file or not
+		cmd := exec.Command("bash", "-c", "cat /etc/fuse.conf | grep -i user_allow_other$")
+
+		var errb bytes.Buffer
+		cmd.Stderr = &errb
+		cliOut, err := cmd.Output()
+		data := string(cliOut)
+		if err == nil && data[0] == '#' {
```
souravgupta-msft: Add case for when `user_allow_other` is not present in /etc/fuse.conf

vibhansa-msft: done

souravgupta-msft (approved):

component/libfuse/libfuse.go
```diff
@@ -204,6 +207,23 @@ func (lf *Libfuse) Validate(opt *LibfuseOptions) error {
 		lf.negativeTimeout = defaultNegativeEntryExpiration
 	}
 
+	if lf.allowOther {
+		// Validate allow_other is enabled in fuse.conf file or not
+		cmd := exec.Command("bash", "-c", "cat /etc/fuse.conf | grep -i user_allow_other$")
```
andyzhangx, line 212: default value is disabled now, so using `allow_other` option would always return error now?
```
# cat /etc/fuse.conf
# /etc/fuse.conf - Configuration file for Filesystem in Userspace (FUSE)

# Set the maximum number of FUSE mounts allowed to non-root users.
# The default is 1000.
#mount_max = 1000

# Allow non-root users to specify the allow_other or allow_root mount options.
#user_allow_other
```

andyzhangx: cc @cvvz

gapra-msft (approved):

component/libfuse/libfuse.go
vibhansa-msft: even today its failing, it just fails silently. With this change it will start returning non-zero error code.

andyzhangx: the above is the default setting in AKS node, and user could set `-o allow_other` mount option, and it won't return error on AKS node today

cvvz: `-o allow_other` only failed when mount by non-root user. Root user can mount successfully even without configuring `/etc/fuse.conf`.

cvvz:
Please take a look at https://github.com/Azure/azure-storage-fuse/pull/1088. I think it should be the right way to solve this kind of problem.

component/libfuse/libfuse.go
vibhansa-msft: If user_allow_other is not enabled in /etc/fuse.conf then it shall not work irrespective of you execute mount with sudo or non-sudo or root user.
In current implementation it fails without giving any error and with this change we are pre-validating the file and returning error if mount is suppose to fail later due to this check.
One question here for the AKS team, if we always mount with allow_other option in AKS why dont we enable user_allow_other in /etc/fuse.conf by default when pod is deployed ?

vibhansa-msft: Ok I verified this and appears root user can mount with allow_other even then /etc/fuse.conf does not have this option set. So I can add a check for uid != 0 while checking the things.

vibhansa-msft: Corrected code to skip this check for root user

vibhansa-msft:
> Please take a look at #1088. I think it should be the right way to solve this kind of problem, instead of handling the `-o allow_other` option alone.

That other PR will not solve this issue as post those changes blobfuse binary will exit with non zero status code but still does not give any hint on what went wrong.

cvvz:
Since this PR has been merged and I'm not the expert of blobfuse V2, so I'm not going to oppose to this pr. 

However, I have to say that with #1088, we can sure get the error code and error message, which is **directly returned from libfuse**.
image

The other concern about this PR is that: if we decide to upgrade libfuse someday and some behavior changed, or if there are some differences from fuse v2 and fuse v3, we may still encounter some problem and have to add more similar changes in blobfuse2 validation logic.

andyzhangx:
@vibhansa-msft pls hold on releasing any new blobfuse2 version, we are setting up a daily e2e test validation on blobfuse2 main branch, to make sure there is no regression after integration with blob csi driver.

cvvz:
@vibhansa-msft Can you test with https://github.com/Azure/azure-storage-fuse/pull/1088 again to see if it can solve this issue? I've tested and it can provide error message from libfuse. Please rethink about this PR. Considering the risk I just raised, I'd prefer we solve this problem by   https://github.com/Azure/azure-storage-fuse/pull/1088, rather than this PR. Thanks

PR #1085, opened 2023-03-17
Wait for child process to complete the mount before exit

vibhansa-msft:
Fixes [#1079](https://github.com/Azure/azure-storage-fuse/issues/1079) Shell returns before daemon process mounts the container and if user tries to bind the mount it leads to inconsistent state. Adding cli flag to wait after demonization to ensure child process has completed the mount.

PR #1088, opened 2023-03-19
Fix: async (background) mount may cause problems

### LINKED ISSUE #1079: Can not mount blob to AKS cluster via Python image using blobfuse2.
blobfuse2 version 2.0.2

AKS kernel version 1.23.12
Docker: FROM python:3.10.4

root@mydocker:/home# cat /proc/version 
Linux version 5.4.0-1091-azure (buildd@lcy02-amd64-023) (gcc version 7.5.0 (Ubuntu 7.5.0-3ubuntu1~18.04)) #96~18.04.1-Ubuntu SMP Tue Aug 30 19:15:32 UTC 2022

root@mydocker:/home# cat /etc/os-release 
PRETTY_NAME="Debian GNU/Linux 11 (bullseye)"
NAME="Debian GNU/Linux"
VERSION_ID="11"
VERSION="11 (bullseye)"
VERSION_CODENAME=bullseye
ID=debian
HOME_URL="https://www.debian.org/"
SUPPORT_URL="https://www.debian.org/support"
BUG_REPORT_URL="https://bugs.debian.org/"

## install blobfuse2
wget https://packages.microsoft.com/config/debian/11/packages-microsoft-prod.deb
dpkg -i packages-microsoft-prod.deb
apt-get update
apt-get install libfuse3-dev fuse3 -y
apt-get install blobfuse2

## mount command
blobfuse2 mount /home/blobfmount --config-file=/home/blobfuse2/config.yaml --container-name=mycontainer --log-level=log_debug --log-file-path=./bobfuse2b.log

# here is config.yaml below:

allow-other: true

logging:
  type: syslog
  level: log_debug

components:
  - libfuse
  - file_cache
  - attr_cache
  - azstorage

libfuse:
  attribute-expiration-sec: 120
  entry-expiration-sec: 120
  negative-entry-expiration-sec: 240

file_cache:
  path: /home/blobfuse/tempcache
  timeout-sec: 120
  max-size-mb: 4096

attr_cache:
  timeout-sec: 7200

azstorage:
  type: block
  account-name: ***
  account-key: ***
  endpoint:***
  container: ***
 

After I ran the blobfuse2 mount command, I cannot check the files in the destination blob or create test file and check in blob.

Not yet.

root@mydocker:/home/blobfmount# cat bobfuse2b.log
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_CRIT [mount.go (384)]: Starting Blobfuse2 Mount : 2.0.2 on [Debian GNU/Linux 11 (bullseye)]
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_CRIT [mount.go (385)]: Logging level set to : LOG_DEBUG
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_TRACE [libfuse.go (220)]: Libfuse::Configure : libfuse
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_INFO [libfuse.go (260)]: Libfuse::Configure : read-only false, allow-other true, default-perm 511, entry-timeout 120, attr-time 120, negative-timeout 240, ignore-open-flags: true, nonempty false
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_TRACE [file_cache.go (197)]: FileCache::Configure : file_cache
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_INFO [file_cache.go (272)]: FileCache::Configure : Using default eviction policy
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_INFO [file_cache.go (291)]: FileCache::Configure : create-empty false, cache-timeout 120, tmp-path /home/blobfuse/tempcache, max-size-mb 4096, high-mark 80, low-mark 60
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_TRACE [attr_cache.go (121)]: AttrCache::Configure : attr_cache
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_INFO [attr_cache.go (145)]: AttrCache::Configure : cache-timeout 7200, symlink false, cache-on-list true
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_TRACE [azstorage.go (83)]: AzStorage::Configure : azstorage
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_TRACE [config.go (270)]: ParseAndValidateConfig : Parsing config
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_INFO [config.go (372)]: ParseAndValidateConfig : using the following proxy address from the config file:
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_INFO [config.go (376)]: ParseAndValidateConfig : sdk logging from the config file: false
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_TRACE [config.go (479)]: ParseAndReadDynamicConfig : Reparsing config
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_DEBUG [config.go (383)]: ParseAndValidateConfig : Getting auth type
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_INFO [config.go (467)]: ParseAndValidateConfig : Account: mystorageaccount, Container: mycontainer, AccountType: BLOCK, Auth: KEY, Prefix: , Endpoint: https://mystorageaccount.blob.core.chinacloudapi.cn/, ListBlock: 0, MD5 : false false, Virtual Directory: true
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_INFO [config.go (471)]: ParseAndValidateConfig : Retry Config: Retry count 5, Max Timeout 900, BackOff Time 4, Max Delay 60
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_TRACE [block_blob.go (173)]: BlockBlob::SetupPipeline : Setting up
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_TRACE [block_blob.go (135)]: BlockBlob::getCredential : Getting credential
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_DEBUG [azauth.go (79)]: azAuth::getAzAuth : Account: mystorageaccount, AccountType: BLOCK, Protocol: https, Endpoint: https://mystorageaccount.blob.core.chinacloudapi.cn/
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_TRACE [block_blob.go (260)]: BlockBlob::SetPrefixPath : path
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_TRACE [block_blob.go (209)]: BlockBlob::TestPipeline : Validating
Wed Mar  8 06:31:11 UTC 2023 : blobfuse2[607] : LOG_INFO [mount.go (392)]: mount: Mounting blobfuse2 on /home/blobfmount
root@mydocker:/home/blobfmount# cd ..

## I tried to do an unmount command and returned this: 
root@mydocker:/home# blobfuse2 unmount all
Error: failed to list mount points [open /etc/mtab: no such file or directory]
root@mydocker:/home# blobfuse2 mount /home/blobfmount --config-file=/home/blobfuse2/config.yaml --container-name=mycontainer --log-level=log_debug --log-file-path=./bobfuse2b.log
Error: mount directory is not empty
root@mydocker:/home# rm -f ./blobfuse2/*
rm: cannot remove './blobfuse2/config.yaml': Device or resource busy
root@mydocker:/home#

cvvz:
# Fix
#1079 
Pod deletion will cause global mount point to disappear. The root cause is csi driver will bind mount the global mount point into Pod volume immediately after `blobfuse2 mount` return successfully. However, fuse is still mounting at background, so actually bind mount and fuse mount happened at the same time, this will lead to a problem that unmount the bind mount (when deleting a Pod) will cause the original mount point unmounted as well.
#1081 
Mount is actually failed, but the errno returned by blobfuse2 is 0 and no error log (both terminal and log file).

# Why
1. libfuse provide `-f` option to decide whether running the process in foreground or as a daemon process. [Daemonizaion is based on `fork`](https://github.com/libfuse/libfuse/blob/master/lib/helper.c#L253-L305). 
2. blobfuse v2 uses cgo to compile blobfuse v2 and libfuse. However, [golang cannot handle damonization well](https://github.com/golang/go/issues/227), so v2 uses [go-daemon](https://github.com/sevlyar/go-daemon) as a workaround and always [pass `-f` option to avoid damonizing in libfuse](https://github.com/Azure/azure-storage-fuse/blob/main/component/libfuse/libfuse2_handler.go#L211-L217).
3. [The daemonization in libfuse actually happens **AFTER** mount completed](https://github.com/libfuse/libfuse/blob/master/lib/helper.c#L351-L359), so the errorno and error message during mount stage can be returned successfully. Blobfuse v1 is written in C++, which can leverage the daemonization in libfuse, so blobfuse v1 can work normally.
4. The difference in blobfuse v2 is that it uses go-daemon to daemonize **BEFORE** [fuse_main](https://github.com/libfuse/libfuse/blob/master/lib/helper.c#L307) (the entry point of libfuse), which makes [the mount operation and all the other operations before libfuse daemonization](https://github.com/libfuse/libfuse/blob/master/lib/helper.c#L316-L354) in background now, thus makes the mount error (and the other operations errors) invisible from user/csi driver. 

# How
1. The `libfuse_init` callback can be an indicator that the mount operation has been successfully completed. At this point child process can send a signal to parent process to notify it that mount has been completed successfully and it can exit normally.
2. When parent process receives `sigchild` signal, it means the child process exited unexpectedly, and the parent process will exit with error code and error message sent by child. 
3. Libfuse(child) print error log to stderr, ~~so I redirect the child process stderr to a temp file~~. By reading this file before exiting, parent process can print error message in terminal. BTW, `go-daemon` uses `forkExec` to fork and exec a new child process, so it's impossible to use PIPE for communicating since after exec, they won't share any opened file descriptor. 
4. Updated: We can pass the opened fd as arguments when exec the new child process. so, it is still possible to communicate between child and parent process through PIPE. The problem is that go-daemon does not support to set child stderr as an opened PIPE fd very well, I've filed a [PR](https://github.com/sevlyar/go-daemon/pull/90/files) to the project.

# Test
I've done test manually and it can solve the problem in both #1079 and #1081.

# csi driver
To be safe, blob csi driver has to use `IsLikelyNotMountPoint` or `MountedFast` after executing `blobfuse2 mount` to make sure the mount point is really mounted successfully.  Refer: https://github.com/kubernetes-sigs/blob-csi-driver/pull/852


component/libfuse/libfuse_handler.go
```diff
@@ -353,6 +359,7 @@ func (lf *Libfuse) fillStat(attr *internal.ObjAttr, stbuf *C.stat_t) {
 // otherwise, perform necessary permission checking
 
 // libfuse_getattr gets file attributes
+//
```
vibhansa-msft, line 362: What are these changes about ?

component/libfuse/libfuse_handler.go
```diff
@@ -241,6 +241,12 @@ func (lf *Libfuse) destroyFuse() error {
 
 //export libfuse_init
 func libfuse_init(conn *C.fuse_conn_info_t, cfg *C.fuse_config_t) (res unsafe.Pointer) {
+	// fuse started processing requests, no error happened during mounting
+	log.Trace("Libfuse::libfuse_init : notifying parent")
+	ppid := syscall.Getppid()
```
vibhansa-msft, line 246: This call will not come back to blobfuse if libfuse decided to fail to init and in such case parent process will go in wait forever.

cmd/mount.go
```diff
@@ -509,7 +534,7 @@ func startMonitor(pid int) {
 	}
 }
 
-func sigusrHandler(pipeline *internal.Pipeline, ctx context.Context) daemon.SignalHandlerFunc {
+func sigusrHandler() daemon.SignalHandlerFunc {
```
vibhansa-msft, line 537: pipeline and context are needed here. Context is taken for future use where we can terminate the ongoing processing.

cmd/mount.go
```diff
@@ -391,32 +392,56 @@ var mountCmd = &cobra.Command{
 
 		log.Info("mount: Mounting blobfuse2 on %s", options.MountPath)
 		if !options.Foreground {
+			// redirect libfuse stderr to a temp file
+			f, err := ioutil.TempFile("", "libfuse_stderr_")
```
vibhansa-msft, line 396: There are customers who mount multiple blobfuse instances on single vm, this means we are creating those many temp files. And in any case if this creation fails we are failing the mount which breaks the current user expectatoins. In cases where it was working fine now there is a probability of that failing.

component/libfuse/libfuse_handler.go
cvvz: I think in that case child would exit and cause parent to exit.

cmd/mount.go
cvvz: Seems like they are not used in this function. That's why I deleted them.

cmd/mount.go
cvvz: We can sure clean it when child exit. However, I think lots of temp file won't be a problem.

cvvz: It's rarely that create a temp file would fail. And I think retry can solve the problem?

component/libfuse/libfuse_handler.go
vibhansa-msft: but parent is waiting for SIGUSR2 and as child exited parent will not receive that?

cmd/mount.go
vibhansa-msft: kept for future use so lets keep them as is.

cmd/mount.go
vibhansa-msft: agree that probability is low, but failing the mount just because some book-keeping stuff dint go well is not a good idea here.

cmd/mount.go
cvvz: ok, will make no difference

component/libfuse/libfuse_handler.go
cvvz: parent will receive SIGCHILD then and exit. Related code is in [here](https://github.com/Azure/azure-storage-fuse/pull/1088/files#diff-57454978dcd397e797f4c48e4b1f80b4b249655d5fc975ad55c54d560ba01de1R446-R448).

cmd/mount.go
cvvz: I've made it won't exit when failing to create temp file. And will remove tmp file once parent exit.

component/libfuse/libfuse_handler.go
```diff
@@ -241,6 +241,12 @@ func (lf *Libfuse) destroyFuse() error {
 
 //export libfuse_init
 func libfuse_init(conn *C.fuse_conn_info_t, cfg *C.fuse_config_t) (res unsafe.Pointer) {
+	// fuse started processing requests, no error happened during mounting
```
vibhansa-msft, line 244: 1. Any pipeline specific handling or common handlings are not done in any of our components so doing this in libfuse component itself is against the base architecture of blobfuse2.
2. this needs to be handled for both fuse2 and fuse3 solutions so doing it here is half baked solution only.

cmd/mount.go
```diff
@@ -399,6 +400,17 @@ var mountCmd = &cobra.Command{
 				Umask:       022,
 			}
 
+			if !daemon.WasReborn() {
+				// redirect libfuse stderr to a temp file
+				f, err := ioutil.TempFile("", "libfuse_stderr_")
```
vibhansa-msft, line 405: Instead of using tempFile can we convert this logic to use unix pipe instead. Be it success or failure in either case child writes down the message in pipe and parents reads from there. As soon as message is received parent can exit. If message was about failure parent can dump it on the screen as well.

cvvz: go-daemon uses forkExec to fork and exec a new child process, so it's impossible to use PIPE for communicating since they do not share opened file descriptor. 
We can use named PIPE or socket for communication, but I think that's no big difference with using a temp file.

component/libfuse/libfuse_handler.go
cvvz: 1. IMHO, maybe we sacrificed little code cleanliness, but the benefits are significant. And I did not find any better solution for the issue.
2. Thanks for pointing it out, I've handled in fuse2 as weel.

vibhansa-msft: Let me see if we can do this in bit cleaner way. I was thinking of adding that sleep as plan-b so that in case this devlops some issue we are not completly blocked and user can move to a sleep based solution instead.

cvvz: Parent can sleep to wait, but it is just a workaround to avoid bind mount problem, which we can do it in csi driver or script rather than use `--wait-for-mount`. Actually, I think nearly no customer would use `--wait-for-mount` explicitly unless they are the expert of blobfuse. 
Also, it can't resolve the problem that when libfuse failed, blobfuse2 exit without error code and any error message in terminal or log file.

vibhansa-msft: I am not denying the fact that this PR solves problem in a better way. My concern here is just in some corner case we hit a situation where parent process is waiting for ever in some customer setup, there is no knob to disable this feature. In such case if that other sleep option is added instead of making a new release later where this option is configurable we can just suggest customer to use that mode. Default will be this signaling mode here in the PR and alternative will be to use sleep. So only on SOS basis we suggest customer to use sleep mode.

cmd/mount.go
```diff
@@ -415,8 +427,27 @@ var mountCmd = &cobra.Command{
...
 				err = runPipeline(pipeline, ctx)
 				if err != nil {
+					errMsg, _ := ioutil.ReadFile(dmnCtx.LogFileName)
+					log.Err("libfuse exit with error message: %s", errMsg)
 					return err
 				}
+			} else {
+				// set signal handler
+				sigusr2 := make(chan os.Signal, 1)
+				signal.Notify(sigusr2, syscall.SIGUSR2)
```
vibhansa-msft, line 437: We shall register this before forking the child as it may happen that by the time code reaches here, child has already started and succeeded with sending a singla. As parent has not registered yet that signal will be lost and then parent will wait till child completes. You can test this scenario out by just adding a sleep of 1-2 seconds as first statement in the else condition.

cvvz: Nice catch!

component/libfuse/libfuse_handler.go
cvvz: I understand your concern, I just merged #1085 to this PR, could you take another look to see if that is ok for you?

cmd/mount.go
```diff
@@ -400,7 +403,27 @@ var mountCmd = &cobra.Command{
...
+				}
+
+				// set signal handler
+				sigusr2 = make(chan os.Signal, 1)
+				signal.Notify(sigusr2, syscall.SIGUSR2)
+				sigchild = make(chan os.Signal, 1)
+				signal.Notify(sigchild, syscall.SIGCHLD)
+			} else { // execute in child only
+				daemon.SetSigHandler(sigusrHandler(pipeline, ctx), syscall.SIGUSR1, syscall.SIGUSR2)
+				go daemon.ServeSignals()
```
cvvz, line 424: I've separated signal registration for child and parent process, also add `daemon.ServeSignals` for child process, otherwise signal registration won't take effect in child process.

component/libfuse/libfuse2_handler.go
```diff
@@ -240,6 +240,13 @@ func (lf *Libfuse) destroyFuse() error {
 
 //export libfuse2_init
 func libfuse2_init(conn *C.fuse_conn_info_t) (res unsafe.Pointer) {
+	// fuse started processing requests, no error happened during mounting
+	log.Trace("Libfuse::libfuse2_init : notifying parent")
```
vibhansa-msft, line 244: SA Issue:
Below issues are found in SA
component/libfuse/libfuse2_handler.go:247:15: Error return value of `syscall.Kill` is not checked (errcheck)
		syscall.Kill(ppid, syscall.SIGUSR2)
		            ^
cmd/mount.go:424:27: Error return value of `daemon.ServeSignals` is not checked (errcheck)
				go daemon.ServeSignals()

cmd/mount.go
```diff
@@ -400,7 +403,27 @@ var mountCmd = &cobra.Command{
...
+			var sigusr2, sigchild chan os.Signal
+			if !daemon.WasReborn() { // execute in parent only
+				// redirect libfuse stderr to a temp file
+				f, err := ioutil.TempFile("", "libfuse_stderr_")
+				if err != nil {
+					log.Err("failed to create temp file for storing libfuse error log [%v]", err)
+				} else {
+					defer os.Remove(f.Name())
+					dmnCtx.LogFileName = f.Name()
+				}
```
vibhansa-msft, line 415: This code is resulting into a panic in my env. Let me create a branch out of this and see what is causing it.

cmd/mount.go
cvvz: I just found out how to use PIPE for communication between child and parent, no more temp file.

cmd/mount.go
cvvz: I just used PIPE instead of temp file for communication between child and parent, so no more temp file, please take another look.

vibhansa-msft: I guess temp file shall be fine, PIPE will have this issue of one side getting blocked if another side goes away in between. Leave to you on what might be better approach. I am more concerned about the panic that I hit in my testing.

cvvz: Thanks, I've tested that it won't be blocked from parent side once child exited abnormally, child will always put something to stderr, which is the writing side of pipe. Also, parent will always exit when timeout. So, I think pipe won't be a problem.

cvvz: > I am more concerned about the panic that I hit in my testing. 

You could put the panic log here if you want and we can discuss together.

cvvz: Pipe can probably solve the panic problem, since no more temp file creation and deletion.

go.mod
```diff
@@ -29,3 +28,5 @@ require (
 )
 
 replace github.com/spf13/cobra => github.com/gapra-msft/cobra v1.4.1-0.20220411185530-5b83e8ba06dd
+
+replace github.com/sevlyar/go-daemon => github.com/cvvz/go-daemon v0.1.5
```
vibhansa-msft, line 32: Why we are doing this ? Is there something we have modified in the go-daemon itself ?

vibhansa-msft: This is not safe if we are modifying the third-party tool. Maintenance is a big question here and on top of that we cannot merge back with any critical fixes that comes in the main repo. Lot of manual effort will be there to keep these in sync, and we do not want to take such dependencies.

cvvz: I've filed a [pr ](https://github.com/sevlyar/go-daemon/pull/90)to go-daemon

vibhansa-msft: Let me take up this PR from here. I get what you are trying to achieve, let me fix it as per our regular blobfuse stnadards.

vibhansa-msft: Can you check the changes now

common/util.go
```diff
@@ -265,3 +266,17 @@ func ExpandPath(path string) string {
...
+
+// NotifyMountToParent : Send a signal to parent process about successful mount
+func NotifyMountToParent() error {
+	ppid := syscall.Getppid()
+	if ppid > 1 {
+		if err := syscall.Kill(ppid, syscall.SIGUSR2); err != nil {
+			return err
+		}
+	} else {
+		return fmt.Errorf("failed to get parent pid, received : %v", ppid)
```
cvvz, line 278: If parent exit before child, this will cause child exit, is that ok?

go.mod
cvvz: I've left some comments, the other looks good to me.

cmd/mount.go
```diff
@@ -431,28 +424,49 @@ var mountCmd = &cobra.Command{
...
-					return Destroy(fmt.Sprintf("libfuse mount failed: %s", buf))
+					// Get error string from the child
+					log.Info("mount: Child [%v] terminated from %s", child.Pid, options.MountPath)
+
+					fname := fmt.Sprintf("/tmp/blobfuse2.%v", child.Pid)
+
+					buff, err := ioutil.ReadFile(fname)
+					if err != nil {
+						log.Err("mount: failed to read child [%v] failure logs [%s]", child.Pid, err.Error())
+						return Destroy(fmt.Sprintf("failed to mount, please check logs [%s]", err.Error()))
```
cvvz, line 459: I think print out buff is better, what if the temp log file is cleaned up by system? We can keep the log file for more details.

vibhansa-msft: this is only when we fail to read the temp file, otherwise we are printing the contents of the temp file

common/util.go
vibhansa-msft: No, it will not error returned by this method is only logged and code moves on.

vibhansa-msft: oh ok you are saying signal send failure shall be ignored, ok will correct that.

vibhansa-msft: if err := common.NotifyMountToParent(); err != nil {
		log.Err("Libfuse::NotifyMountToParent : Failed to notify parent, error: [%v]", err)
	}

cvvz: That looks good to me.

cmd/mount.go
```diff
@@ -401,24 +404,70 @@ var mountCmd = &cobra.Command{
...
+			if child == nil { // execute in child only
 				defer dmnCtx.Release() // nolint
 				setGOConfig()
 				go startDynamicProfiler()
 
 				err = runPipeline(pipeline, ctx)
 				if err != nil {
+					pid := os.Getpid()
+					fname := fmt.Sprintf("/tmp/blobfuse2.%v", pid)
+					if err = ioutil.WriteFile(fname, []byte(err.Error()), 0777); err != nil {
```
cvvz, line 439: libfuse print out error message by stderr, you should use [SetLogFile](https://github.com/sevlyar/go-daemon/pull/90/files#diff-bf608eab570a8660c0f9a7c6712ee27da25ecb3246c8aaad0131f7005a8a675aR56-R59) to set stderr as PIPE or [set log file name](https://github.com/sevlyar/go-daemon/blob/master/daemon_unix.go#L24) as this file. However, it can only be done before creating child process.

cvvz:
```shell
git checkout 4c9ab71dadf7b4f79cb992655962ed20598d9ca6
```
or 
```shell
git checkout 1c304b820c7c02e0b68fc25cae34071cbef589ca
```
and build blobfuse2 you can see the error message returned by libfuse in terminal.

vibhansa-msft (approved):

cmd/mount.go
```diff
@@ -394,29 +397,69 @@ var mountCmd = &cobra.Command{
...
+				defer os.Remove(fname)
+
+				select {
+				case <-sigusr2:
+					log.Info("mount: Child [%v] mounted successfully at %s", child.Pid, options.MountPath)
+
+				case <-sigchild:
+					// Get error string from the child
+					log.Info("mount: Child [%v] terminated from %s", child.Pid, options.MountPath)
+					if dmnCtx.LogFileName != "" {
```
souravgupta-msft, line 451: Since the `LogFileName` is set while creating the `dmnCtx`, this check can be removed.

souravgupta-msft (approved):

PR #1099, opened 2023-03-28
Subdir mount fails to list blobs

vibhansa-msft:


component/azstorage/block_blob.go
```diff
@@ -468,6 +468,8 @@ func (bb *BlockBlob) getAttrUsingRest(name string) (attr *internal.ObjAttr, err
 func (bb *BlockBlob) getAttrUsingList(name string) (attr *internal.ObjAttr, err error) {
 	log.Trace("BlockBlob::getAttrUsingList : name %s", name)
 
+	listPath := filepath.Join(bb.Config.prefixPath, name)
```
gapra-msft, line 471: Can we add a test that validates this fix?

vibhansa-msft: As of today we do not have any test case which validates subdirectory mount.

vibhansa-msft: subdirectory mount can only be validated in E2E and for those test suits we create new container and mount it so subdirectory mount can not be tested as such.

vibhansa-msft: Done, have added UT to validate the function which is suppose to remove prefix path from the received path.
Also added subdirectory based getAttr method in existing UT.

component/azstorage/utils_test.go
```diff
@@ -75,6 +76,45 @@ type contentTypeVal struct {
...
+		result string
+	}
+
+	var inputs = []PrefixPath{
+		{prefix: "", path: "abc.txt", result: "abc.txt"},
+		{prefix: "", path: "ABC", result: "ABC"},
+		{prefix: "", path: "ABC/DEF.txt", result: "ABC/DEF.txt"},
+		{prefix: "", path: "ABC/DEF/1.txt", result: "ABC/DEF/1.txt"},
+
+		{prefix: "ABC", path: "ABC/DEF/1.txt", result: "DEF/1.txt"},
```
gapra-msft, line 94: can we ever expect a path with / at the end? is that something this method should support?

gapra-msft (approved):

Issue #1100, opened 2023-03-30
blobfuse2 on AKS 1.25.5 does not support files with `Content-Encoding: gzip`


eyedvabny:
### Which version of blobfuse was used?
`blobfuse2 version 2.0.1`

### Which OS distribution and version are you using?
AKS 1.25.5 running  `AKSUbuntu-2204gen2containerd-2023.02.15` node image

```
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=22.04
DISTRIB_CODENAME=jammy
DISTRIB_DESCRIPTION="Ubuntu 22.04.1 LTS"
```

### If relevant, please share your mount command.
blobfuse2 mount /var/lib/kubelet/plugins/kubernetes.io/csi/blob.csi.azure.com/40774c1cc34ceb8f8428e46f2de90f65dcd5f007fd997f8fc8e2f8c8ff924f60/globalmount --file-cache-timeout-in-seconds=120 -o negative_timeout=120 --use-attr-cache=true --virtual-directory=true --cache-size-mb=1000 -o allow_other --log-level=LOG_WARNING -o attr_timeout=120 -o entry_timeout=120 --pre-mount-validate=true --use-https=true --cancel-list-on-mount-seconds=10 --empty-dir-check=false --tmp-path=<redacted> --container-name=<redacted> --ignore-open-flags=true

### What was the issue encountered?
blobfuse2 seems to be confused by files which have a Content-Encoding of `gzip`.
<img width="194" alt="image" src="https://user-images.githubusercontent.com/1400408/228942324-dcabb2aa-89ab-4495-bc46-3ca50c8f4e1d.png">

The files show up in the directory just fine, but any attempt to copy or move them results in an IO error:
![image](https://user-images.githubusercontent.com/1400408/228943063-f674f4ba-b52a-4b5e-88de-ccb3a42aac93.png)

This only impacts blobfuse-mounted directories. The file was downloadable from Azure using Storage Explorer and perfectly valid.

### Have you found a mitigation/solution?
Removing or changing Content-Encoding seems to fix the behavior. No value other than `gzip` seems to trigger this issue - we tried `deflate` and `utf-8` and `base64` and even `asdf`. We ended up setting the Content-Type to `application/x-gzip` which is what Azure Storage Explorer sets it to when uploading a gzip file, and not setting Content-Encoding at all.

### Please share logs if available.
[...]

PR #1104, opened 2023-04-04
Disable compression to be a configurable feature

vibhansa-msft:
Fix for #1100

### DIFF
--- a/component/azstorage/config.go
+++ b/component/azstorage/config.go
@@ -465,7 +465,7 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 		log.Warn("unsupported v1 CLI parameter: debug-libcurl is not applicable in blobfuse2.")
 	}
 
-	log.Info("ParseAndValidateConfig : Account: %s, Container: %s, AccountType: %s, Auth: %s, Prefix: %s, Endpoint: %s, ListBlock: %d, MD5 : %v %v, Virtual Directory: %v, Compression: %v",
+	log.Info("ParseAndValidateConfig : Account: %s, Container: %s, AccountType: %s, Auth: %s, Prefix: %s, Endpoint: %s, ListBlock: %d, MD5 : %v %v, Virtual Directory: %v, Disable Compression: %v",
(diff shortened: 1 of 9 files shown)


component/azstorage/config.go
```diff
@@ -464,9 +465,9 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 		log.Warn("unsupported v1 CLI parameter: debug-libcurl is not applicable in blobfuse2.")
 	}
 
-	log.Info("ParseAndValidateConfig : Account: %s, Container: %s, AccountType: %s, Auth: %s, Prefix: %s, Endpoint: %s, ListBlock: %d, MD5 : %v %v, Virtual Directory: %v",
+	log.Info("ParseAndValidateConfig : Account: %s, Container: %s, AccountType: %s, Auth: %s, Prefix: %s, Endpoint: %s, ListBlock: %d, MD5 : %v %v, Virtual Directory: %v, Compression: %v",
```
souravgupta-msft, line 468:
```suggestion
	log.Info("ParseAndValidateConfig : Account: %s, Container: %s, AccountType: %s, Auth: %s, Prefix: %s, Endpoint: %s, ListBlock: %d, MD5 : %v %v, Virtual Directory: %v, Disable Compression: %v",
```

souravgupta-msft (approved):

setup/baseConfig.yaml
```diff
@@ -144,7 +144,7 @@ azstorage:
   update-md5: true|false <set md5 sum on upload. Impacts performance. works only when file-cache component is part of the pipeline>
   validate-md5: true|false <validate md5 on download. Impacts performance. works only when file-cache component is part of the pipeline>
   virtual-directory: true|false <support virtual directories without existence of a special marker blob>
-
+  disable-compression: true|false <disable transport layer content encoding like gzip, set this flag to true if blobs have content-encoding set in container>
```
gapra-msft, line 147: Should we add a test for this feature?

vibhansa-msft: we can just add a cli validation kind of a test. Data validation kind of thing can not be done as this goes to http client options and we do not have any direct control on testing it.

vibhansa-msft: Added config test and moved block blob based proxy test in code coverage to use disable-compression

Issue #1118, opened 2023-04-22
flag for redownload check via LMT


YinTatMSR:
### Which version of blobfuse was used?
V2

### Which OS distribution and version are you using?
ubuntu 20.04 (WSL2)

### What was the issue encountered?
In my use cases, some files are seldom updated but will get access frequently. I noticed that after the file-cache-timeout, the file is redownloaded despite the file didn't get updated. I am trying to avoid such downloads. I do not care about the exact version of the file, it can be off a little bit, but I don't want to set file-cache-timeout to be 1 year to avoid redownloading those files.

### Have you found a mitigation/solution?
When blobfuse2 thinks "downloadRequired" (line 798 in file_cache/file_cache.go), check the LMT again (it is okay we use the cache version of LMT, we can control attr-timeout anyway).

If the LMT has not changed, do not download the file.

If I understand the program correctly, the whole process shouldn't involve any extra Azure call because if blobfuse2 indeed downloads the file, it needs to get the attr anyway to know the file size.

In this way, if I use the default setting, the file might be out of sync for 240 sec, but when it is not changed and if it is not evicted from the cache, I don't need to redownload the file which is very helpful for me.

I do not know what is the best exact logic to avoid redownloading, this is just a suggestion. But I think it should be possible to avoid redownloading with a small/no cost.

vibhansa-msft:
Were you able to validate using the private branch ?

PR #1120, opened 2023-04-24
Adding feature to download file based on LMT check

vibhansa-msft:
Feature addition as per request #1118


souravgupta-msft (approved):

nakulkar-msft (approved):

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+## 2.0.4 (WIP)
```
gapra-msft, line 1: We should probably try to add a scenario test for this.

vibhansa-msft: to validate this fix we need to create a simulation where files are being updated in parallel through other means on container. Any of the test infra that we have does not support such case of now. I am thinking over how to add case for this, but as of now no clue on this.

tasherif-msft (approved):

PR #1149, opened 2023-05-22
Performance Improvements for blobfuse attribute lookup

shubham808:
This PR adds new configs to blobfuse which impacts list and open operation time

### DIFF
--- a/component/attr_cache/attr_cache.go
+++ b/component/attr_cache/attr_cache.go
@@ -54,11 +54,12 @@ const defaultAttrCacheTimeout uint32 = (120)
 // Common structure for AttrCache Component
 type AttrCache struct {
 	internal.BaseComponent
-	cacheTimeout uint32
-	cacheOnList  bool
-	noSymlinks   bool
-	cacheMap     map[string]*attrCacheItem
-	cacheLock    sync.RWMutex
+	cacheTimeout  uint32
+	cacheOnList   bool
+	noSymlinks    bool
+	maxTotalFiles uint32
+	cacheMap      map[string]*attrCacheItem
+	cacheLock     sync.RWMutex
 }
 
 // Structure defining your config parameters
@@ -66,17 +67,19 @@ type AttrCacheOptions struct {
 	Timeout       uint32 `config:"timeout-sec" yaml:"timeout-sec,omitempty"`
 	NoCacheOnList bool   `config:"no-cache-on-list" yaml:"no-cache-on-list,omitempty"`
 	NoSymlinks    bool   `config:"no-symlinks" yaml:"no-symlinks,omitempty"`
-
+	
+	//maximum file attributes overall to be cached
+	MaxTotalFiles uint32 `config:"max-total-files" yaml:"max-total-files,omitempty"`
+	
 	// support v1
 	CacheOnList bool `config:"cache-on-list"`
 }
 
 const compName = "attr_cache"
 
-// for now caching only first 1 mil files in a directory
+// caching only first 5 mil files by default
 // caching more means increased memory usage of the process
-const maxFilesPerDir = 1000000 // 1 million max files to be cached per directory
-const maxTotalFiles = 10000000 // 10 million max files overall to be cached
+const defaultMaxTotalFiles = 5000000 // 5 million max files overall to be cached
 
 //  Verification to check satisfaction criteria with Component Interface
 var _ internal.Component = &AttrCache{}
@@ -138,6 +141,12 @@ func (ac *AttrCache) Configure(_ bool) error {
 		ac.cacheOnList = conf.CacheOnList
 	} else {
 		ac.cacheOnList = !conf.NoCacheOnList
+	}
+
+	if config.IsSet(compName + ".max-total-files") {
+		ac.maxTotalFiles = conf.MaxTotalFiles
+	} else {
+		ac.maxTotalFiles = defaultMaxTotalFiles
 	}
 
 	ac.noSymlinks = conf.NoSymlinks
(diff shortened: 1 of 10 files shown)


component/attr_cache/attr_cache.go
```diff
@@ -54,29 +54,32 @@ const defaultAttrCacheTimeout uint32 = (120)
...
 }
 
 // Structure defining your config parameters
 type AttrCacheOptions struct {
 	Timeout       uint32 `config:"timeout-sec" yaml:"timeout-sec,omitempty"`
 	NoCacheOnList bool   `config:"no-cache-on-list" yaml:"no-cache-on-list,omitempty"`
 	NoSymlinks    bool   `config:"no-symlinks" yaml:"no-symlinks,omitempty"`
-
+	
+	//maximum file attributes overall to be cached
```
vibhansa-msft, line 71: we can rename this to "max-files"

component/attr_cache/attr_cache.go
```diff
@@ -262,9 +271,7 @@ func (ac *AttrCache) StreamDir(options internal.StreamDirOptions) ([]*internal.O
 	log.Trace("AttrCache::ReadDir : %s", options.Name)
 
 	pathList, token, err := ac.NextComponent().StreamDir(options)
-	if err == nil && options.Offset < maxFilesPerDir {
-		ac.cacheAttributes(pathList)
-	}
+	ac.cacheAttributes(pathList)
```
vibhansa-msft, line 274: shouldn't we put the max files check here ?

component/attr_cache/attr_cache.go
```diff
@@ -485,8 +493,10 @@ func (ac *AttrCache) GetAttr(options internal.GetAttrOptions) (*internal.ObjAttr
 
 	if err == nil {
 		// Retrieved attributes so cache them
-		if len(ac.cacheMap) < maxTotalFiles {
+		if len(ac.cacheMap) < int(ac.maxTotalFiles) {
```
vibhansa-msft, line 496: If we declare the variable as int itself, we dont need to do this conversion for every call

component/attr_cache/attr_cache.go
```diff
@@ -569,12 +579,17 @@ func NewAttrCacheComponent() internal.Component {
...
+	
 	noSymlinks := config.AddBoolFlag("no-symlinks", false, "whether or not symlinks should be supported")
 	config.BindPFlag(compName+".no-symlinks", noSymlinks)
 
 	cacheOnList := config.AddBoolFlag("cache-on-list", true, "Cache attributes on listing.")
 	config.BindPFlag(compName+".cache-on-list", cacheOnList)
 	cacheOnList.Hidden = true
+	
+	maxTotalFiles := config.AddUint32Flag("max-total-files", defaultMaxTotalFiles, "Max total file attributes in cache.")
+	config.BindPFlag(compName+".max-total-files", maxTotalFiles)
```
vibhansa-msft, line 594: This part we can skip, this basically creates a command line parameter for this flag. I guess not many customers will go for this and based on that we can skip creating cli parameter for this field, just in the config it shall be fine.

component/libfuse/libfuse_handler.go
```diff
@@ -364,7 +364,7 @@ func (lf *Libfuse) fillStat(attr *internal.ObjAttr, stbuf *C.stat_t) {
 func libfuse_getattr(path *C.char, stbuf *C.stat_t, fi *C.fuse_file_info_t) C.int {
 	name := trimFusePath(path)
 	name = common.NormalizeObjectName(name)
-	// log.Trace("Libfuse::libfuse_getattr : %s", name)
+	log.Trace("Libfuse::libfuse_getattr : %s", name)
```
vibhansa-msft, line 367: comment this back, getattr is called for a very large number of times in the system and doing any logging here will hurt the performance.

component/attr_cache/attr_cache.go
shubham808: the was there in the cacheAttributes method so i thought we could skip her, but we can avoid iterating over path list if do the check here so added it back

souravgupta-msft:
Please add an entry in the changelog for this.

component/attr_cache/attr_cache.go
vibhansa-msft: I revisited the code, we can remove this check from here and just keep err=nil check as the ac.cacheAttributes() method is anyway checking on the length of list against the new configured parameter.

component/libfuse/libfuse_handler.go
vibhansa-msft: We need to comment this back.

vibhansa-msft:
There are go formatting errors in this PR and CI is not going through:
https://dev.azure.com/azstorage/BlobFuse/_build/results?buildId=16775&view=logs&j=1fbcc5aa-e08c-514c-179d-87ec73246166&t=122e66bd-a102-5d27-e872-5aaed2361566

component/azstorage/config.go
```diff
@@ -503,7 +504,11 @@ func ParseAndReadDynamicConfig(az *AzStorage, opt AzStorageOptions, reload bool)
 	} else {
 		az.stConfig.virtualDirectory = true
 	}
-
+	
+	if config.IsSet(compName + ".max-results-for-list") {
```
vibhansa-msft, line 508: set it to default value if config is not set.

component/azstorage/block_blob.go
```diff
@@ -475,8 +476,13 @@ func (bb *BlockBlob) getAttrUsingList(name string) (attr *internal.ObjAttr, err
 	var marker *string = nil
 	blobsRead := 0
 
+	maxResultsForList := bb.Config.maxResultsForList
```
vibhansa-msft, line 479: We can set the config value once during configure time and then use it in getAttrUsingList instead of checking it in everycall.

component/azstorage/block_blob.go
```diff
@@ -475,8 +475,10 @@ func (bb *BlockBlob) getAttrUsingList(name string) (attr *internal.ObjAttr, err
...
```
vibhansa-msft, line 478: there is no need to create this variable also, we can pass down bb.config.maxResults parameter as function argument below.

souravgupta-msft (approved):

PR #1162, opened 2023-06-01
Adding FUSE option direct_io

souravgupta-msft:
direct_io can be used to bypass the kernel cache. It disables the use of page cache in the kernel.

TODO: add this option for fuse2.

### DIFF
--- a/component/libfuse/libfuse_handler_test.go
+++ b/component/libfuse/libfuse_handler_test.go
@@ -58,12 +58,13 @@ func (suite *libfuseTestSuite) TestDefault() {
 	suite.assert.Equal(suite.libfuse.negativeTimeout, uint32(120))
 	suite.assert.False(suite.libfuse.disableWritebackCache)
 	suite.assert.True(suite.libfuse.ignoreOpenFlags)
+	suite.assert.False(suite.libfuse.directIO)
 }
 
 func (suite *libfuseTestSuite) TestConfig() {
 	defer suite.cleanupTest()
 	suite.cleanupTest() // clean up the default libfuse generated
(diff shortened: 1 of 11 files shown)


component/libfuse/libfuse_handler_test.go
```diff
@@ -58,12 +58,13 @@ func (suite *libfuseTestSuite) TestDefault() {
...
 	suite.assert.False(suite.libfuse.disableWritebackCache)
 	suite.assert.True(suite.libfuse.ignoreOpenFlags)
+	suite.assert.False(suite.libfuse.directIO)
 }
 
 func (suite *libfuseTestSuite) TestConfig() {
 	defer suite.cleanupTest()
 	suite.cleanupTest() // clean up the default libfuse generated
-	config := "allow-other: true\nread-only: true\nlibfuse:\n  attribute-expiration-sec: 60\n  entry-expiration-sec: 60\n  negative-entry-expiration-sec: 60\n  fuse-trace: true\n  disable-writeback-cache: true\n  ignore-open-flags: false\n"
+	config := "allow-other: true\nread-only: true\nlibfuse:\n  attribute-expiration-sec: 60\n  entry-expiration-sec: 60\n  negative-entry-expiration-sec: 60\n  fuse-trace: true\n  disable-writeback-cache: true\n  ignore-open-flags: false\n  direct-io: true\n"
```
gapra-msft, line 67: how safe is this? Should we add an e2e tests with this enabled?

souravgupta-msft: direct_io is disabled by default. If a user wants to disable kernel cache, then they can enable it.
I will add some e2e tests for it.

souravgupta-msft: Also this option is supported in v1. So, in my opinion we should support it in v2 as well.

souravgupta-msft: Yes

vibhansa-msft (approved):

component/libfuse/libfuse_handler_test.go
souravgupta-msft: Added it for fuse2

PR #1180, opened 2023-06-20
Fix function names in logs and log more

jfantinhardesty:
This PR fixes some of the logging in blobfuse2. Namely, that some functions did not provide trace logs, or they referred to the wrong function name when logging. I also tried to standardize some of the logging in files that were using different formats.

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -337,7 +337,7 @@ func (c *FileCache) StatFs() (*syscall.Statfs_t, bool, error) {
 	statfs := &syscall.Statfs_t{}
 	err := syscall.Statfs("/", statfs)
 	if err != nil {
-		log.Debug("FileCache: statfs err [%s].", err.Error())
+		log.Debug("FileCache::StatFs : statfs err [%s].", err.Error())
 		return nil, false, err
 	}
 	statfs.Blocks = uint64(maxCacheSize) / uint64(statfs.Frsize)
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -522,7 +522,6 @@ func (fc *FileCache) ReadDir(options internal.ReadDirOptions) ([]*internal.ObjAt
 
 // StreamDir : Add local files to the list retrieved from storage container
 func (fc *FileCache) StreamDir(options internal.StreamDirOptions) ([]*internal.ObjAttr, string, error) {
-	log.Trace("FileCache::StreamDir : name=%d, offset=%s", options.Name, options.Offset)
 	attrs, token, err := fc.NextComponent().StreamDir(options)
 
 	if token == "" {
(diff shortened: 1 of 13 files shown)


jfantinhardesty:
@microsoft-github-policy-service agree company="Seagate"

vibhansa-msft:
Some of these logs were left intentionally as the regular workflow will hit these functions quite often and any sort of logging here will degrade the performance. getAttr/Read/Write/StreamDir are hit very heavily and any logging shall be avoided here.
Is there any particular reason you wish to add these logs?

jfantinhardesty:
It was mainly to standardize the logging. We were doing some debugging and reading through the log files and it was sometimes difficult to follow when we were expecting a function to log something, but it didn't. For GetAttr every component except libfuse and stream do a trace log. For Read no component does a trace log. For Write only the Azstorage component does a trace log. And for StreamDir every component except file cache does a trace log.

vibhansa-msft:
As I mentioned some of the things are done intentionally. For e.g. getAttr generally do not get pass through the attr-cache (as it has higher timeouts) so putting a log in azstorage allows us to navigate when what the getAttr actually called on storage. On the flows where we hit the most, we try to avoid loggging.

jfantinhardesty:
Ok. I remove those added logs for read/write/getattr/streamdir.

component/file_cache/file_cache.go
```diff
@@ -337,7 +337,7 @@ func (c *FileCache) StatFs() (*syscall.Statfs_t, bool, error) {
 	statfs := &syscall.Statfs_t{}
 	err := syscall.Statfs("/", statfs)
 	if err != nil {
-		log.Debug("FileCache::StatFs : statfs err [%s].", err.Error())
+		log.Debug("FileCache: statfs err [%s].", err.Error())
```
vibhansa-msft, line 340: For any component level logs its always "<component>::<function> : " as prefix so existing log is correct as per our practices.

component/libfuse/libfuse2_handler.go
```diff
@@ -292,7 +292,7 @@ func libfuse2_init(conn *C.fuse_conn_info_t) (res unsafe.Pointer) {
 
 //export libfuse_destroy
 func libfuse_destroy(data unsafe.Pointer) {
```
vibhansa-msft, line 294: "<component>::<function> : "

tools/health-monitor/internal/stats_export.go
```diff
@@ -105,23 +105,23 @@ func (se *StatsExporter) Destroy() {
 	for i, op := range se.outputList {
 		jsonData, err := json.MarshalIndent(op, "", "\t")
 		if err != nil {
-			log.Err("stats_exporter::addToOutputFile : unable to marshal [%v]", err)
```
vibhansa-msft, line 108: I guess component name here is stats_exporter only

vibhansa-msft:
Have provided some comments kindly fix those and it shall be good to go.

jfantinhardesty:
Thanks. I think I addressed all your comments.

vibhansa-msft (approved):

PR #1208, opened 2023-07-19
Respect ACLs based on object id instead of using permissions

vibhansa-msft:


tasherif-msft (approved):
I do think the flags are getting overwhelming 😥

component/azstorage/azstorage.go
```diff
@@ -636,6 +636,9 @@ func init() {
 	disableCompression := config.AddBoolFlag("disable-compression", false, "Disable transport layer compression.")
 	config.BindPFlag(compName+".disable-compression", disableCompression)
 
+	resepectACL := config.AddBoolFlag("respect-acl", false, "Match ObjectID in ACL against the one used for authentication.")
```
souravgupta-msft, line 639: Add this option in the base config file.

souravgupta-msft (approved):

PR #1216, opened 2023-08-04
Rename leaving destination file in temp-cache for ever

vibhansa-msft:
- As file is renamed the destination file remains in file-cache. This file is not added to LRU because it was never opened or closed hence policy will never delete this file.
- As a Fix add the destination file to policy after successful rename,


CHANGELOG.md
```diff
@@ -1,4 +1,5 @@
 ## 2.0.6 (WIP)
+- Fix for rename leaving destination file in local file-cache.
```
souravgupta-msft, line 2: Can we make this little more descriptive?

souravgupta-msft (approved):

PR #1236, opened 2023-09-01
Move away from third party docker container base image

vibhansa-msft:


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 ## 2.1.1 (WIP)
 **Features**
+- Alpine based docker imgae publish to Azure Container Registry.
```
souravgupta-msft, line 3:
```suggestion
- Alpine based docker image publish to Azure Container Registry.
```

blobfuse2-release.yaml
```diff
@@ -429,6 +429,7 @@ stages:
               chmod 777 *.sh
               ./dockerinstall.sh
               ./buildcontainer.sh
+              ./buildcontainer_alpine.sh
```
souravgupta-msft, line 432: With this change, we will be building for both ubuntu and alpine based docker images, right?

docker/Dockerfile_Alpine
```diff
@@ -0,0 +1,51 @@
+# Create container based on Ubuntu-22.04 Jammy Jellyfish image
```
souravgupta-msft, line 1: This needs to be updated

souravgupta-msft (approved):

PR #1242, opened 2023-09-06
Ashruti/fixes

ashruti-msft:
Fixed the bug where content type was case sensitive for file extensions and added tests to check it as well.
Optimized the truncate process by eliminating download when truncate flag is received as it is redundant.
Added information about the above in CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,12 @@
 **Features**
 
 **Bug Fixes**
-- Fixed the case sensitivity of content type for file extensions.
+- [#1237](https://github.com/Azure/azure-storage-fuse/issues/1237) Fixed the case sensitivity of content type for file extensions.
 
 **Optimizations**
-- File is not downloaded in file_cache when truncate flag is received.
+- Optimized file-cache to skip download when O_TRUNC flag is provided in open call.
 
-## 2.1.0 (WIP)
+## 2.1.0 (2023-08-31)
 **Features**
 - Added support for ARM64 architecture.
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 2.1.1 (WIP)
+**Features**
+
+**Bug Fixes**
+- Fixed the case sensitivity of content type for file extensions.
+
+**Optimizations**
+- File is not downloaded in file_cache when truncate flag is received.
+
 ## 2.1.0 (WIP)
```
vibhansa-msft, line 10: 2.1.0 release date can be updated here instead of WIP

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
+## 2.1.1 (WIP)
+**Features**
+
+**Bug Fixes**
+- Fixed the case sensitivity of content type for file extensions.
```
vibhansa-msft, line 5: We need to mention the bug here against which we are fixing this.

CHANGELOG.md
```diff
@@ -1,3 +1,12 @@
...
+
+**Optimizations**
+- File is not downloaded in file_cache when truncate flag is received.
```
vibhansa-msft, line 8: - Optimized file-cache to skip download when O_TRUNC flag is provided in open call.

vibhansa-msft (approved):

souravgupta-msft (approved):

PR #1248, opened 2023-09-22
Truncate function changed for streaming 

ashruti-msft:
Fixed existing issues in the truncate function implementation for streaming.


vibhansa-msft (approved):

component/azstorage/block_blob.go
```diff
@@ -1021,8 +1031,10 @@ func (bb *BlockBlob) removeBlocks(blockList *common.BlockOffsetList, size int64,
 		}
 
 	}
-
+	blk := blockList.BlockList[index]
```
souravgupta-msft, line 1034: In the above section (lines [1023-1028](https://github.com/Azure/azure-storage-fuse/blob/ashruti/fixes/component/azstorage/block_blob.go#L1023-L1028)), we can skip the `ReadInBuffer` call.
`blk.Data` already has the block's data in it. So, we can just do something like,
`blk.Data = blk.Data[:blk.EndIndex-blk.StartIndex]`
Please confirm this part.

ashruti-msft: `blk.Data` does not contain any data before calling ReadInBuffer so this method will not work.

component/stream/read_write.go
```diff
@@ -47,6 +48,8 @@ import (
 	"github.com/pbnjay/memory"
 )
 
+var file_created = false
```
vibhansa-msft, line 51: this is global variable, if we execute multiple file-open/file-create in parallel then this will not work.

component/azstorage/block_blob_test.go
```diff
@@ -3140,6 +3141,85 @@ func (s *blockBlobTestSuite) TestInvalidMD5OnReadNoVaildate() {
...
+func (suite *blockBlobTestSuite) TestTruncateBlockFileToSmaller() {
+	suite.UtilityFunctionTestTruncateFileToSmaller(300*MB, 290*MB)
+}
+
+func (suite *blockBlobTestSuite) TestTruncateBlockFileToLarger() {
+	suite.UtilityFunctionTruncateFileToLarger(290*MB, 300*MB)
+}
+
+func (suite *blockBlobTestSuite) TestTruncateNoBlockFileToLarger() {
+	suite.UtilityFunctionTruncateFileToLarger(200*MB, 300*MB)
```
vibhansa-msft, line 3161: Why we are not checking any failure or success in these cases, how do we know it works.

vibhansa-msft (approved):

souravgupta-msft (approved):

PR #1255, opened 2023-10-03
Make file cache limit a hard limit

bhansalivikas:


component/file_cache/file_cache.go
```diff
@@ -1340,6 +1371,16 @@ func (fc *FileCache) TruncateFile(options internal.TruncateFileOptions) error {
 	// If you call truncate CLI command from shell it always sends an open call first followed by truncate
 	// But if you call the truncate method from a C/C++ code then open is not hit and only truncate comes
 
+	if fc.diskHighWaterMark != 0 {
+		currSize, err := common.GetUsage(fc.tmpPath)
+		if err == nil {
```
ashruti-msft, line 1376: If err is not nil log the error

ashruti-msft (approved):

PR #1258, opened 2023-10-06
Write support using Block-Cache

vibhansa-msft:
Write support with block-cache (new streaming write model)
- Nighly pipeline has new data validation test added for block-cache where both upload and download are tested for data corruption
- Perf pipeline has a added step now to compare file-cache vs block-cache read/write performance.
- UT and code-coverage has been added for block-cache as well.


component/loopback/loopback_fs.go
```diff
@@ -208,6 +208,11 @@ func (lfs *LoopbackFS) RenameDir(options internal.RenameDirOptions) error {
 
 func (lfs *LoopbackFS) CreateFile(options internal.CreateFileOptions) (*handlemap.Handle, error) {
 	log.Trace("LoopbackFS::CreateFile : name=%s", options.Name)
+
+	if options.Name == "FailThis" {
```
souravgupta-msft, line 212: Any reason for this check?

vibhansa-msft: This is to run some negative test cases.

component/azstorage/block_blob.go
```diff
@@ -781,7 +781,11 @@ func (bb *BlockBlob) ReadInBuffer(name string, offset int64, len int64, data []b
 	blobURL := bb.Container.NewBlobURL(filepath.Join(bb.Config.prefixPath, name))
 	opt := bb.downloadOptions
 	opt.BlockSize = len
-	err := azblob.DownloadBlobToBuffer(context.Background(), blobURL, offset, len, data, opt)
+
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Minute)
```
souravgupta-msft, line 785: Any reason to use this. SDK should add this timeout value, right?
If we do go with using the context with timeout, should we update the other calls where `context.Background()` is used.

component/azstorage/connection.go
```diff
@@ -136,6 +136,10 @@ type AzConnection interface {
 	TruncateFile(string, int64) error
 	StageAndCommit(name string, bol *common.BlockOffsetList) error
 
+	GetCommittedBlockList(string) (*internal.CommittedBlockList, error)
+	StageBlock(string, []byte, string) error
+	CommitBlocks(string, []string) error
```
souravgupta-msft, line 141: We should add tests for these 3 methods in azstorage component. Not in this PR, maybe after the track2 migration.

component/block_cache/block_cache.go
```diff
@@ -80,24 +84,26 @@ type BlockCache struct {
...
 type BlockCacheOptions struct {
-	BlockSize      uint64 `config:"block-size-mb" yaml:"block-size-mb,omitempty"`
-	MemSize        uint64 `config:"mem-size-mb" yaml:"mem-size-mb,omitempty"`
-	TmpPath        string `config:"path" yaml:"path,omitempty"`
-	DiskSize       uint64 `config:"disk-size-mb" yaml:"disk-size-mb,omitempty"`
-	DiskTimeout    uint32 `config:"disk-timeout-sec" yaml:"timeout-sec,omitempty"`
-	PrefetchCount  uint32 `config:"prefetch" yaml:"prefetch,omitempty"`
-	Workers        uint32 `config:"parallelism" yaml:"parallelism,omitempty"`
-	PrefetchOnOpen bool   `config:"prefetch-on-open" yaml:"prefetch-on-open,omitempty"`
+	BlockSize      float64 `config:"block-size-mb" yaml:"block-size-mb,omitempty"`
```
souravgupta-msft, line 87: Why are we using floating values for BlockSize?

vibhansa-msft: We take input in terms of MBs. In initial reading block-cache there were couple of requests to make it in bytes instead as customer wanted to read only few bytes from a very large file and there no need to even download 1 mb data. Making this float gives us an oppurtunity to specify it in bytes. For e.g. a value "0.5" will mean 512KB bufer and "0.05" or values like these can take it further down.

component/azstorage/connection.go
vibhansa-msft: With dummy storage we can not do these tests, Will check if its possible to do this as part of this PR itself.

souravgupta-msft (approved):

component/azstorage/block_blob.go
souravgupta-msft: In case of throttling or network disruption, shouldn't we wait for the retries to complete?

component/azstorage/connection.go
souravgupta-msft: I would suggest doing this in some other PR or maybe after the track2 migration is done, I will add these tests.

component/azstorage/block_blob.go
souravgupta-msft: With the current SDK, I see the value of timeout as 901 seconds in the request URL.
`GET https://storageAccount.blob.core.windows.net/container/a.txt?timeout=901`

In my opinion using 3 mins might not be sufficient for retries (with exponential backoff).

README.md
```diff
@@ -7,7 +7,7 @@ Blobfuse2 is stable, and is ***supported by Microsoft*** provided that it is use
 
 ##  NOTICE
 - We have seen some customer issues around files getting corrupted when `streaming` is used in write mode. Kindly avoid using this feature for write while we investigate and resolve it.
-
+- You can now use block-cache instead of streaming for both read and write workflows, which offers much better performance compared to streaming. To enable `block-cache` instead of `streaming`, use `--block-cache` in CI param or `block-cache` as component in config file instead of `streaming`.
```
souravgupta-msft, line 10:
```suggestion
- You can now use block-cache instead of streaming for both read and write workflows, which offers much better performance compared to streaming. To enable `block-cache` instead of `streaming`, use `--block-cache` in CLI param or `block-cache` as component in config file instead of `streaming`.

```

souravgupta-msft (approved):

README.md
souravgupta-msft: Looks good

PR #1261, opened 2023-10-16
Added content-type for .usdz file

### LINKED ISSUE #1260: Blobfuse2 issue with .usdz and .glb files
2.0.5

When uploading .usdz and .glb files blobfuse2 is corrupting the file content. I compared the Md5 hash of original file it's different from the content of the on azure storage. Also it's updates the content-type as "application/octet-stream". On directly copying file to Azure Storage using Storage Explorer which uses AZCopy content-type stays as "application/zip" and Md5Hash remains same.

As per code  I don't see support for usdz and glb - https://github.com/Azure/azure-storage-fuse/blob/559d864ddb937e4a59fddd4b2361f592991a64c9/cpplite/src/blob/blob_client.cpp#L189

Is there a way to ignore content type and keep the content same during fuse upload?

No

ashruti-msft:
Fixes #1260 

For the .glb file uploading from AzCopy also gives content-type as "application/octet-stream" so changes have been made only for .usdz file


souravgupta-msft (approved):

souravgupta-msft:
Should we have a changelog for this?
@vibhansa-msft, @ashruti-msft

vibhansa-msft:
> Should we have a changelog for this? @vibhansa-msft, @ashruti-msft

No need as this does not impact any functionality or not a new feature either.

PR #1267, opened 2023-10-19
Randomise refresh token flow 

vibhansa-msft:
In case of AKS and HPC environment there are multiple instances of blobfuse running. If 200 instances were spawned together all will go for refreshing a token at the same time. In some customer cases we have seen these results into failure from AAD as too many requests are posted at the same time. With this changeset we are trying to randomize the refresh of token and in case of failure again take a backoff randomly and retry to refresh the token.


component/azstorage/azauthmsi.go
```diff
@@ -198,35 +203,53 @@ func (azmsi *azAuthBlobMSI) getCredential() interface{} {
...
-			newToken, err := azmsi.fetchTokenFromCLI()
-			if err != nil {
-				log.Err("azAuthBlobMSI::getCredential : Failed to refresh token [%s]", err.Error())
-				return 0
-			}
+			for failCount := 0; failCount < 5; failCount++ {
+				newToken, err := azmsi.fetchTokenFromCLI()
+				if err != nil {
+					log.Err("azAuthBlobMSI::getCredential : Failed to refresh token attempt %d [%s]", failCount, err.Error())
+					failCount++
```
souravgupta-msft, line 210: Why is `failCount` being incremented here? It is already done by the for loop.

vibhansa-msft: fixed

souravgupta-msft (approved):

PR #1275, opened 2023-10-26
Added decision tree for configuration and fixed config samples

tasherif-msft:


README.md
```diff
@@ -176,6 +176,11 @@ To learn about a specific command, just include the name of the command (For exa
 
 ***Please note: do not use quotations `""` for any of the config parameters***
 
+## Choosing Between File Cache and Stream Modes
```
vibhansa-msft, line 179: Not able to put comments in diagram so putting it here.
- Lets rename the file "Blank diagram.png" to have some meaningful name.
- I see readme links to this diagram is it possible to directly show it there in little shorter form
- We need more sample files with different configs. Lets create those in a seperate directory and not sure if possible but if respective flow in diagram can give a hyperlink to respective config file it would be great
- In case large files random reads we advocate file-cache in diagram, if its read only a part of large file then streaming shall be advocated and for entire file read may be file-cache.
- Also, we shall have some decision making based on available disk space. If disk space is not enough to accomodate 5 gb file then we need to go for streaming instead of file-cache. 5GB is just example here but any size for that matter.

sampleFileCacheConfig.yaml
```diff
@@ -1,7 +1,5 @@
 # Refer ./setup/baseConfig.yaml for full set of config parameters
 
-allow-other: true
```
souravgupta-msft, line 3: I think we should keep this in the sample configs because in most cases customers mount blobfuse2 as root(fstab for example) and try to access as regular user.

README.md
tasherif-msft: - renamed
- there is just one readme link, its name length is shorter than our sample config file names. I dont think this is a problem
- We can add more samples in a different pr, I think for a new user the three samples and this flow diagram are more than enough. Giving the user many files to choose from defeats the purpose of helping new customers onboard and get Blobfuse2 running. This is sufficient to get started. It's not possible unfortunately since the flow diagram is a png :/ I did add two lines under it linking to the two config files for file cache and stream.
- made those changes in the flow diagram

vibhansa-msft (approved):

souravgupta-msft (approved):

Issue #1060, opened 2023-02-15
Blobfuse2 unable to save changes in some text editors


jfantinhardesty:
### Which version of blobfuse was used?

2.0.1

### Which OS distribution and version are you using?

Ubuntu 20.04.5

### If relevant, please share your mount command.

blobfuse2 mount ./mount --config-file=./config.yaml

### What was the issue encountered?

I am unable to save changes to files in certain text editors.

I opened an existing text file using gedit in my mounted container. When I save the edited file I see the following error: "Unexpected error: Error renaming temporary file: Input/output error" and so I am unable to save the changes to the file. It seems this issue can be replicated with any editor that saves to a temporary file and does a rename like gedit does.

### Have you found a mitigation/solution?

No.

### Please share logs if available.

Log file
[blobfuse2.txt](https://github.com/Azure/azure-storage-fuse/files/10746020/blobfuse2.txt)

Config file as a text file
[config.txt](https://github.com/Azure/azure-storage-fuse/files/10746042/config.txt)

vibhansa-msft:
Add "--ignore-open-flags=true" to your mount command and retry.

jfantinhardesty:
I tried that in the mount command and got the same error. I also had that set to true originally in my config file.

vibhansa-msft:
As per logs, editor is creating a temperory file and doing chmod on it and later renaming it to original file name. As the temporary file is not yet uploaded to the container the rename command fails. Blobfuse uploads a file only on close, and here we do not see any close call before the rename. One possible workaround to get out of this situation is to instruct blobfuse to create a file on storage whenever application creates a file on mount path.
Add "create-empty-file: true" under "file-cache" section in your config file and that shall unblock you. I will try to reproduce this locally and investigate if we can do something better here.

vibhansa-msft:
Were you able to edit the file using above suggested config changes?

vibhansa-msft:
Closing this as this is resolved.

PR #1277, opened 2023-10-27
Flush dirty writes before renaming a file

### LINKED ISSUE #1259: Blobfuse2 uploading empty file on save using gedit
2.1.0

Ubuntu 20.04.6

blobfuse2 mount ./mount --config-file=./config.yaml

When saving changes to files in certain text editors a blank file is updated with no data in the file. Based on #1060 we set "create-empty-file: true" for the "file-cache" component. However, this results in files upload to Azure storage to be an empty file. The contents are saved in the file_cache so when reopening the file it appears to have the valid dat. But once the file_cache expires and we redownload the file, the file is empty.

It seems this issue can be replicated with any editor that saves to a temporary file and does a rename like gedit does.

No.

Log file
[blobfuse2.txt](https://github.com/Azure/azure-storage-fuse/files/12897017/blobfuse2.txt)

Config file as a text file
[config.txt](https://github.com/Azure/azure-storage-fuse/files/12897021/config.txt)

foodprocessor:
On rename, check source file for dirty writes - and flush them - before proceeding.

Related issues:
https://github.com/Azure/azure-storage-fuse/issues/1259
https://github.com/Azure/azure-storage-fuse/issues/1060

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -1317,6 +1317,29 @@ func (fc *FileCache) RenameFile(options internal.RenameFileOptions) error {
 	dflock.Lock()
 	defer dflock.Unlock()
 
+	// check if the source file has a dirty handle
+	var dirtySrcHandle *handlemap.Handle
+	handlemap.GetHandles().Range(
+		func(key any, value any) bool {
+			handle := value.(*handlemap.Handle)
+			if options.Src == handle.Path && handle.Dirty() {
+				dirtySrcHandle = value.(*handlemap.Handle)
+				return false
+			} else {
+				return true
+			}
+		},
+	)
+	if dirtySrcHandle != nil {
+		log.Warn("FileCache::RenameFile : src=%s has a dirty file handle. Flushing...", options.Src)
+		flushErr := fc.FlushFile(internal.FlushFileOptions{Handle: dirtySrcHandle})
+		if flushErr != nil {
+			log.Err("FileCache::RenameFile : Flushing dirty src=%s failed. Here's why: %v", options.Src, flushErr)
+			// abort rename to avoid losing data
+			return flushErr
+		}
+	}
+
 	err := fc.NextComponent().RenameFile(options)
 	err = fc.validateStorageError(options.Src, err, "RenameFile", false)
 	if err != nil {


foodprocessor:
> @foodprocessor please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement
> ### Contribution License Agreement
> 
> This Contribution License Agreement (**“Agreement”**) is agreed to by the party signing below (**“You”**), and conveys certain license rights to Microsoft Corporation and its affiliates (“Microsoft”) for Your contributions to Microsoft open source projects. This Agreement is effective as of the latest signature date below.
> 
>     1. **Definitions**.
>        **“Code”** means the computer software code, whether in human-readable or machine-executable form,
>        that is delivered by You to Microsoft under this Agreement.
>        **“Project”** means any of the projects owned or managed by Microsoft and offered under a license
>        approved by the Open Source Initiative ([www.opensource.org](http://www.opensource.org)).
>        **“Submit”** is the act of uploading, submitting, transmitting, or distributing code or other content to any
>        Project, including but not limited to communication on electronic mailing lists, source code control
>        systems, and issue tracking systems that are managed by, or on behalf of, the Project for the purpose of
>        discussing and improving that Project, but excluding communication that is conspicuously marked or
>        otherwise designated in writing by You as “Not a Submission.”
>        **“Submission”** means the Code and any other copyrightable material Submitted by You, including any
>        associated comments and documentation.
> 
>     2. **Your Submission**. You must agree to the terms of this Agreement before making a Submission to any
>        Project. This Agreement covers any and all Submissions that You, now or in the future (except as
>        described in Section 4 below), Submit to any Project.
> 
>     3. **Originality of Work**. You represent that each of Your Submissions is entirely Your original work.
>        Should You wish to Submit materials that are not Your original work, You may Submit them separately
>        to the Project if You (a) retain all copyright and license information that was in the materials as You
>        received them, (b) in the description accompanying Your Submission, include the phrase “Submission
>        containing materials of a third party:” followed by the names of the third party and any licenses or other
>        restrictions of which You are aware, and (c) follow any other instructions in the Project’s written
>        guidelines concerning Submissions.
> 
>     4. **Your Employer**. References to “employer” in this Agreement include Your employer or anyone else
>        for whom You are acting in making Your Submission, e.g. as a contractor, vendor, or agent. If Your
>        Submission is made in the course of Your work for an employer or Your employer has intellectual
>        property rights in Your Submission by contract or applicable law, You must secure permission from Your
>        employer to make the Submission before signing this Agreement. In that case, the term “You” in this
>        Agreement will refer to You and the employer collectively. If You change employers in the future and
>        desire to Submit additional Submissions for the new employer, then You agree to sign a new Agreement
>        and secure permission from the new employer before Submitting those Submissions.
> 
>     5. **Licenses**.
> 
> 
>     * **Copyright License**. You grant Microsoft, and those who receive the Submission directly or
>       indirectly from Microsoft, a perpetual, worldwide, non-exclusive, royalty-free, irrevocable license in the
>       Submission to reproduce, prepare derivative works of, publicly display, publicly perform, and distribute
>       the Submission and such derivative works, and to sublicense any or all of the foregoing rights to third
>       parties.
> 
>     * **Patent License**. You grant Microsoft, and those who receive the Submission directly or
>       indirectly from Microsoft, a perpetual, worldwide, non-exclusive, royalty-free, irrevocable license under
>       Your patent claims that are necessarily infringed by the Submission or the combination of the
>       Submission with the Project to which it was Submitted to make, have made, use, offer to sell, sell and
>       import or otherwise dispose of the Submission alone or with the Project.
> 
>     * **Other Rights Reserved**. Each party reserves all rights not expressly granted in this Agreement.
>       No additional licenses or rights whatsoever (including, without limitation, any implied licenses) are
>       granted by implication, exhaustion, estoppel or otherwise.
> 
> 
>     6. **Representations and Warranties**. You represent that You are legally entitled to grant the above
>        licenses. You represent that each of Your Submissions is entirely Your original work (except as You may
>        have disclosed under Section 3). You represent that You have secured permission from Your employer to
>        make the Submission in cases where Your Submission is made in the course of Your work for Your
>        employer or Your employer has intellectual property rights in Your Submission by contract or applicable
>        law. If You are signing this Agreement on behalf of Your employer, You represent and warrant that You
>        have the necessary authority to bind the listed employer to the obligations contained in this Agreement.
>        You are not expected to provide support for Your Submission, unless You choose to do so. UNLESS
>        REQUIRED BY APPLICABLE LAW OR AGREED TO IN WRITING, AND EXCEPT FOR THE WARRANTIES
>        EXPRESSLY STATED IN SECTIONS 3, 4, AND 6, THE SUBMISSION PROVIDED UNDER THIS AGREEMENT IS
>        PROVIDED WITHOUT WARRANTY OF ANY KIND, INCLUDING, BUT NOT LIMITED TO, ANY WARRANTY OF
>        NONINFRINGEMENT, MERCHANTABILITY, OR FITNESS FOR A PARTICULAR PURPOSE.
> 
>     7. **Notice to Microsoft**. You agree to notify Microsoft in writing of any facts or circumstances of which
>        You later become aware that would make Your representations in this Agreement inaccurate in any
>        respect.
> 
>     8. **Information about Submissions**. You agree that contributions to Projects and information about
>        contributions may be maintained indefinitely and disclosed publicly, including Your name and other
>        information that You submit with Your Submission.
> 
>     9. **Governing Law/Jurisdiction**. This Agreement is governed by the laws of the State of Washington, and
>        the parties consent to exclusive jurisdiction and venue in the federal courts sitting in King County,
>        Washington, unless no federal subject matter jurisdiction exists, in which case the parties consent to
>        exclusive jurisdiction and venue in the Superior Court of King County, Washington. The parties waive all
>        defenses of lack of personal jurisdiction and forum non-conveniens.
> 
>     10. **Entire Agreement/Assignment**. This Agreement is the entire agreement between the parties, and
>         supersedes any and all prior agreements, understandings or communications, written or oral, between
>         the parties relating to the subject matter hereof. This Agreement may be assigned by Microsoft.

@microsoft-github-policy-service agree [company="Seagate"]

foodprocessor:
@microsoft-github-policy-service agree company="Seagate"

component/file_cache/file_cache.go
```diff
@@ -1317,6 +1317,29 @@ func (fc *FileCache) RenameFile(options internal.RenameFileOptions) error {
 	dflock.Lock()
 	defer dflock.Unlock()
 
+	// check if the source file has a dirty handle
+	var dirtySrcHandle *handlemap.Handle
+	handlemap.GetHandles().Range(
+		func(key any, value any) bool {
```
vibhansa-msft, line 1323: I have two problems with such iterations:
- This will result in a crash because we are iterating it here and there can be other flows which are adding/removing items from the same map
- We do have customer cases where thousands of files are worked upon in parallel and for every rename if we go iterating the thread, it will consume too much of CPU
- Also, we are not supposed to flush the contents to storage container untill user flushes the contents manually or closes the file handle. As a result of this loop we will end up syncing data to container even for files which user has not closed yet.

Overall, this is not a safe handling.

foodprocessor: Thanks for taking a look at this so quickly!
I have been thinking about your concerns... Although I think [Range](https://pkg.go.dev/sync#Map.Range) is safe from crashes, I agree that my solution here is expensive and flushes a file that should not be flushed.
With those problems in mind, I came up with a new approach that should avoid both problems you pointed out. But this approach may raise new concerns, so please let me know what you think:

Instead of flushing the source file, we could call `CopyFromFile`, with `CopyFromFileOptions.Name` set to `RenameFileOptions.Dst` and `CopyFromFileOptions.File` set to the source file.
This would prevent a flush to the source file, which was not requested, but you could say it would instead "flush" the local source file to the destination file in cloud storage.

To avoid the concern you raised about the CPU cost of searching the `handlemap`, we would need to choose an approach that works without knowing whether the source file has a dirty open handle or not. To do this, I think we could check whether the source file is in the cache, using `isDownloadRequired` (is this a correct use of this function?).
- If `isDownloadRequired` returns true and the source file is not local, I think we can assume it does not have a dirty open handle, and the existing in-cloud rename process could be used in that case.
- If `isDownloadRequired` returns false and the source file is local, we could use the `CopyFromFile` process I outlined above. 

So, in a nutshell, my alternate proposal would be something like this:
```
localSrcPath := filepath.Join(fc.tmpPath, options.Src)
srcNotLocal, _, _, err := fc.isDownloadRequired(localSrcPath, options.Src, sflock)
if err != nil { ... }
if srcNotLocal {
	// rename file in cloud (previously the only code path)
	err := fc.NextComponent().RenameFile(options)
	...
} else {
	fc.policy.CacheValid(localSrcPath)
	uploadHandle, err := os.Open(localSrcPath)
	if err != nil { ... }
	err = fc.NextComponent().CopyFromFile(
		internal.CopyFromFileOptions{
			Name: options.Dst,
			File: uploadHandle,
		})
	uploadHandle.Close()
	if err != nil { ... }
}

// existing code resumes here untouched:
localDstPath := filepath.Join(fc.tmpPath, options.Dst)
...
```
What do you think of this alternate approach?

vibhansa-msft: I see what you are proposing here. In case of Rename if file does not exists locally then directly call rename on storage container. If the file exists locally then instead of doing rename on container and then locally, we just reupload the source file to container. It still has this challange that not all of local data might have flushed to local file and when we upload it migt still be stale contents as we will flush the local file when flush/close call comes. To avoid all this complexity only code was written to locally and remotely just rename the files and avoid any inconsistent uploads. Also, when we do rename on container we save a lot of cost to customer as data is copied by server from source to destination. In this approach we will end up uploading the file twice (when we rename and later if source is closed). Generally in case of editors a temp file is created and edited and at the end when you save the file its renamed and actual source gets deleted. Most editors have a config using which you can control where to keep these temp files. By default most editor tend to keep it in same directory (even in case of vim its the same). I do understand we are trying to solve the rename problem here but I see a much larger issue. When temp files are created in a mounted path we end up creating and uploading these temp files to Azure Container while these files are either going to be renamed or deleted at the end. We can save a lot of cost by keeping these temp files to some other directory instead of mounted path and that will solve this rename issue as well.

vibhansa-msft: I am not against solving the rename issue but if editor is causing this issue by creating temp files then keeping those temp files in mounted path is not a good approach as this is not a regular directory and rather a network storage. Any temp files creation will add cost to storage and network. For any network attached storage its always better to avoid such things. You can not only solve this problem but also save some $s if temp files are redirected to some other directory.

foodprocessor: I agree that users should not take the mounted folder to be a normal file system folder. And I agree that users should try to keep temp files outside of mounted folders. The temp file problem is definitely a headache.

I understand your objection: my proposal from October 29th would make it so that if a file is closed (and uploaded) before a rename, then it would be uploaded twice.
So the solution needs to be more specific to this situation. We are trying to detect when a file with uncommitted local writes is being renamed. But so far, I have only detected when a **_local_ file** is being renamed, which I agree is too broad.
Digging a bit deeper, I see an easy improvement to this: we could use `sflock.Count() > 0` (or maybe `sflock.Count() > 0 && !isDownloadRequired(...)`), which would detect when an **_open_ local file** is being renamed. This still isn't perfect... We don't know if the file has uncommitted writes, and we don't know if it is even open for writing. But I think this improvement would remove most of the cases where using only `isDownloadRequired` for this check would cause the software to upload the same file twice.
When we detect that an open local file is being renamed, maybe we could further reduce the risk of a redundant upload by comparing the modify date of the local file to the container's version. This adds the cost of sending a `GetAttr` to the container, but only in this already narrow case. So if that `GetAttr` call prevents a double-upload, maybe it's worth it.

What do you think of this latest proposal?

Thank you for your generous guidance as I search for a reasonable solution!

vibhansa-msft: Get-Attr calls are short-circuted. They go and fetch the data from attr-cache component and not directly from storage unless attr-cache says cache invalid. So, relying on getAttr from storage will not be helpful in all cases. This is not a simple problem to solve by just uploading as there are lot of complex scenarios involved here in case when application has multiple handles open for writing and reading. That is one of the reason to not jump on this solution quickly as we need to think through on what all can possibly go wrong with this change.

vibhansa-msft: Will try to go through the logs again and see if we are missing something which can help us sync such locally written files rename case. For now I suggest best bet is to change the default directory for temp files. Committing something quickly here is not a solution I am looking for. Need to think more on what shall be the correct approach to this problem and that will take some time and investigation. If you feel this change is good enough for you, you can create a private branch or fork out of the this repo and work with this change locally meanwhile.

foodprocessor: That sounds reasonable.
I will implement this approach on my fork for now, and I will look forward to your solution.
Thank you!

PR #1281, opened 2023-11-02
Correcting sync to flush behavior

vibhansa-msft:
sync-to-flush flag is suppose to fore upload a file to container. In previous fix it was just setting the dirty flag, which will upload the file only on close. Corrected the code to force upload the file on sync itself.

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -987,13 +987,10 @@ func (fc *FileCache) CloseFile(options internal.CloseFileOptions) error {
 
 	localPath := filepath.Join(fc.tmpPath, options.Handle.Path)
 
-	if options.Handle.Dirty() {
-		log.Info("FileCache::CloseFile : name=%s, handle=%d dirty. Flushing the file.", options.Handle.Path, options.Handle.ID)
-		err := fc.FlushFile(internal.FlushFileOptions{Handle: options.Handle}) //nolint
-		if err != nil {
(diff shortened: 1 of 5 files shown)


component/file_cache/file_cache.go
```diff
@@ -987,13 +987,10 @@ func (fc *FileCache) CloseFile(options internal.CloseFileOptions) error {
...
 	localPath := filepath.Join(fc.tmpPath, options.Handle.Path)
 
-	if options.Handle.Dirty() {
-		log.Info("FileCache::CloseFile : name=%s, handle=%d dirty. Flushing the file.", options.Handle.Path, options.Handle.ID)
-		err := fc.FlushFile(internal.FlushFileOptions{Handle: options.Handle}) //nolint
-		if err != nil {
-			log.Err("FileCache::CloseFile : failed to flush file %s", options.Handle.Path)
-			return err
-		}
+	err := fc.FlushFile(internal.FlushFileOptions{Handle: options.Handle}) //nolint
```
souravgupta-msft, line 990: In FileCache.CloseFile() shouldn't we call FlushFile() only when the handle is dirty?

vibhansa-msft: flush file internally checks the dirty flag, because flush can be called from outside also. As its already checked inside the method i have removed this.

souravgupta-msft (approved):

PR #1283, opened 2023-11-06
Symlink fix for ADLS accounts

### LINKED ISSUE #1243: Symlinks not working when mounting HNS-enabled blob storage usind ADLS
2.1.1-preview.1

Ubuntu 22.04

`sudo blobfuse2 mount ${mount_dir} --config-file="/etc/blobfuse_${storage_container}_config.yml --use-adls=true"`

Config:
```# Common configurations
allow-other: true

# Logger configuration
logging:
  type: base
  level: log_debug
  file-path: '/.blobfuse2.log'
  max-file-size-mb: 512
  file-count: 10
  track-time: true

# Pipeline configuration. Choose components to be engaged. The order below is the priority order that needs to be followed.
components:
  - libfuse
  - stream
  - attr_cache
  - azstorage

# Libfuse configuration
libfuse:
  default-permission: '0666'
  attribute-expiration-sec: 120
  entry-expiration-sec: 120
  negative-entry-expiration-sec: 240

# Streaming configuration
stream:
  # If block-size-mb, blocks-per-file or cache-size-mb are 0, the stream component will not cache blocks.
  block-size-mb: 8
  max-buffers: 3
  buffer-size-mb: 512

# Attribute cache related configuration
attr_cache:
  timeout-sec: 7200

# Azure storage configuration
azstorage:
# Required
  type: adls
  account-name: ${storage_account}
  container: ${storage_container}
  endpoint: https://${storage_account}.blob.core.windows.net/
  # Optional
  block-size-mb: 16
  max-concurrency: 32
  mode: spn
  tenantid: ${tenantid}
  clientid: ${clientid}
  clientsecret: ${clientsecret}

# Health Monitor configuration
health_monitor:
  output-path: /var/log/
  enable-monitoring: true
  stats-poll-interval-sec: 10
  process-monitor-interval-sec: 30
  # list of monitors to be disabled
  monitor-disable-list:
    - cpu_profiler
    - memory_profiler
```

The symlinks are not mounted. The name of the files appear but do no redirect to anything, `readlink` is empty, and `cd` returns `Error: Not a directory`.
Creating a symlink in the mounted directory after mounting also does not create a symlink, only a filename that appears when listing the directory.
Mounting the same storage using NFS instead of blobfuse does mount the symlinks. 
Mounting another storage with HNS-disabled storage account using `type: block` does mount symlinks

Not using blobfuse.

The mount is successful so there is no error log, but the symlinks don't work.

souravgupta-msft:
Fixes #1243

- In ADLS accounts, read symlink fails after dir list operation. This is because the ADLS List API does not return metadata and other properties in its response (unlike blob API which does).
- To get the metadata, GetProperties() call is explicitly made after the List() operation for each blob.
- To enable symlink support in ADLS accounts, either pass `--no-symlinks=false` in the mount command or set `no-symlinks: false` in the config under attr_cache section.

### DIFF
--- a/component/azstorage/datalake.go
+++ b/component/azstorage/datalake.go
@@ -550,7 +550,7 @@ func (dl *Datalake) List(prefix string, marker *string, count int32) ([]*interna
 		// to fetch metadata properties.
 		// Any method that populates the metadata should set the attribute flag.
 		// Alternatively, if you want Datalake list paths to return metadata/properties as well.
-		// pass CLI paramater --no-symlinks=false in the mount command.
+		// pass CLI parameter --no-symlinks=false in the mount command.
 		pathList = append(pathList, attr)
 	}
 
--- a/component/azstorage/datalake.go
+++ b/component/azstorage/datalake.go
@@ -430,7 +430,7 @@ func (dl *Datalake) GetAttr(name string) (attr *internal.ObjAttr, err error) {
 	}
 	attr.Flags.Set(internal.PropFlagMetadataRetrieved)
 
-	if dl.Config.HonourACL && dl.Config.authConfig.ObjectID != "" {
+	if dl.Config.honourACL && dl.Config.authConfig.ObjectID != "" {
 		acl, err := pathURL.GetAccessControl(context.Background())
 		if err != nil {
 			// Just ignore the error here as rest of the attributes have been retrieved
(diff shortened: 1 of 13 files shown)


component/azstorage/datalake.go
```diff
@@ -495,41 +495,52 @@ func (dl *Datalake) List(prefix string, marker *string, count int32) ([]*interna
 
 	// Process the paths returned in this result segment (if the segment is empty, the loop body won't execute)
 	for _, pathInfo := range listPath.Paths {
-		var mode fs.FileMode
-		if pathInfo.Permissions != nil {
-			mode, err = getFileMode(*pathInfo.Permissions)
+		var attr *internal.ObjAttr
+
+		if dl.Config.disableSymlink {
+			var mode fs.FileMode
```
vibhansa-msft, line 501: This peice shall be common. If disableSymlink is not there then we make an extra call to getAttr

vibhansa-msft (approved):
LGTM

PR #1296, opened 2023-12-05
Default working directory when $HOME is not present

souravgupta-msft:
If $HOME is not present, use the current working directory.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -305,3 +305,13 @@ func GetIdLength(id string) int64 {
 	existingBlockId, _ := base64.StdEncoding.DecodeString(id)
 	return int64(len(existingBlockId))
 }
+
+func init() {
+	val, present := os.LookupEnv("HOME")
+	if !present {
+		val = "./"
+	}
+	DefaultWorkDir = filepath.Join(val, ".blobfuse2")
+	DefaultLogFilePath = filepath.Join(DefaultWorkDir, "blobfuse2.log")
+	StatsConfigFilePath = filepath.Join(DefaultWorkDir, "stats_monitor.cfg")
+}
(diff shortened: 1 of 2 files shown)


common/types.go
```diff
@@ -305,3 +305,13 @@ func GetIdLength(id string) int64 {
...
 	return int64(len(existingBlockId))
 }
+
+func init() {
+	val, present := os.LookupEnv("HOME")
+	if !present {
+		val = "./"
+	}
+	DefaultWorkDir = filepath.Join(val, ".blobfuse2")
+	DefaultLogFilePath = filepath.Join(DefaultWorkDir, "blobfuse2.log")
```
vibhansa-msft, line 315: These variables shall already be populated somewhere in cmd processing. Even if we set them here, they might get overridden. I do not see any deletion of code which was setting these earlier.

souravgupta-msft: In the other code we are expanding the environment variables, $HOME in this case. But are not checking if the $HOME environment variable is present or not.

```
f, err := os.Stat(common.ExpandPath(common.DefaultWorkDir))
if err == nil && !f.IsDir() {
	return fmt.Errorf("default work dir '%s' is not a directory", common.DefaultWorkDir)
}

if err != nil && os.IsNotExist(err) {
	// create the default work dir
	if err = os.MkdirAll(common.ExpandPath(common.DefaultWorkDir), 0777); err != nil {
		return fmt.Errorf("failed to create default work dir [%s]", err.Error())
	}
}
```

If $HOME is not present, it tries to create the directory, `/.blobfuse2` which results in operation not permitted error.

vibhansa-msft (approved):

PR #1311, opened 2024-01-12
Mount on non-empty path

souravgupta-msft:
From [Libfuse 3.0.0 release](https://github.com/libfuse/libfuse/releases/tag/fuse-3.0.0),

> The -o nonempty mount point has been removed, mounting over non-empty directories is now always allowed. This brings the behavior of FUSE file systems in-line with the behavior of the
> regular mount command.
> 
> File systems that do not want to allow mounting to non-empty directories should perform this check themselves before handing control to libfuse.

This fixes mount on non-empty paths for fuse3.
For fuse2, this option is supported.

### DIFF
--- a/component/libfuse/libfuse2_handler.go
+++ b/component/libfuse/libfuse2_handler.go
@@ -110,7 +110,6 @@ func (lf *Libfuse) convertConfig() *C.fuse_options_t {
 	fuse_opts.allow_other = C.bool(lf.allowOther)
 	fuse_opts.allow_root = C.bool(lf.allowRoot)
 	fuse_opts.trace_enable = C.bool(lf.traceEnable)
-	fuse_opts.non_empty = C.bool(lf.nonEmptyMount)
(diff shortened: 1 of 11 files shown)


component/libfuse/libfuse2_handler.go
```diff
@@ -209,10 +208,6 @@ func populateFuseArgs(opts *C.fuse_options_t, args *C.fuse_args_t) (*C.fuse_opti
 		options += ",allow_root"
 	}
 
-	if opts.non_empty {
```
vibhansa-msft, line 212: we need to pass this down to fuse also, why we are removing this here?

souravgupta-msft: Libfuse 3.0.0 release has removed the `nonempty` option. Please refer the PR description.

vibhansa-msft (approved):

vibhansa-msft (approved):

PR #1334, opened 2024-02-01
Track2 Migration: Block Blob migration

souravgupta-msft:
This PR contains code for migrating block blob to use track2 azblob SDK.

NOTE: The tests migration will be done in some other PR.

### DIFF
--- a/component/azstorage/azstorage.go
+++ b/component/azstorage/azstorage.go
@@ -573,7 +573,7 @@ func NewazstorageComponent() internal.Component {
 		stConfig: AzStorageConfig{
 			blockSize:      0,
 			maxConcurrency: 32,
-			defaultTier:    getAccessTierType("none"),
+			defaultTier:    getAccessTierType(""),
 			authConfig: azAuthConfig{
 				AuthMode: EAuthType.KEY(),
 				UseHTTP:  false,
--- a/component/azstorage/block_blob.go
+++ b/component/azstorage/block_blob.go
@@ -120,15 +120,15 @@ func (bb *BlockBlob) UpdateConfig(cfg AzStorageConfig) error {
 	return nil
 }
 
-// NewServiceClient : Update the SAS specified by the user and create new service client
-func (bb *BlockBlob) NewServiceClient(key, value string) (err error) {
+// UpdateServiceClient : Update the SAS specified by the user and create new service client
+func (bb *BlockBlob) UpdateServiceClient(key, value string) (err error) {
 	if key == "saskey" {
 		bb.Auth.setOption(key, value)
 
 		// get the service client with updated SAS
 		svcClient, err := bb.Auth.getServiceClient(&bb.Config)
 		if err != nil {
-			log.Err("BlockBlob::NewServiceClient : Failed to get service client [%s]", err.Error())
+			log.Err("BlockBlob::UpdateServiceClient : Failed to get service client [%s]", err.Error())
 			return err
 		}
 
@@ -141,19 +141,19 @@ func (bb *BlockBlob) NewServiceClient(key, value string) (err error) {
 	return nil
 }
 
-// getServiceClient : Create the service client
-func (bb *BlockBlob) getServiceClient() (*service.Client, error) {
-	log.Trace("BlockBlob::getServiceClient : Getting service client")
+// createServiceClient : Create the service client
+func (bb *BlockBlob) createServiceClient() (*service.Client, error) {
+	log.Trace("BlockBlob::createServiceClient : Getting service client")
 
 	bb.Auth = getAzAuthT2(bb.Config.authConfig)
 	if bb.Auth == nil {
-		log.Err("BlockBlob::getServiceClient : Failed to retrieve auth object")
+		log.Err("BlockBlob::createServiceClient : Failed to retrieve auth object")
 		return nil, fmt.Errorf("failed to retrieve auth object")
 	}
 
 	svcClient, err := bb.Auth.getServiceClient(&bb.Config)
 	if err != nil {
-		log.Err("BlockBlob::getServiceClient : Failed to get service client [%s]", err.Error())
+		log.Err("BlockBlob::createServiceClient : Failed to get service client [%s]", err.Error())
 		return nil, err
 	}
 
@@ -166,7 +166,7 @@ func (bb *BlockBlob) SetupPipeline() error {
 	var err error
 
 	// create the service client
-	bb.Service, err = bb.getServiceClient()
+	bb.Service, err = bb.createServiceClient()
 	if err != nil {
 		log.Err("BlockBlob::SetupPipeline : Failed to get service client [%s]", err.Error())
 		return err
--- a/component/azstorage/block_blob.go
+++ b/component/azstorage/block_blob.go
@@ -217,11 +217,7 @@ func (bb *BlockBlob) ListContainers() ([]string, error) {
 			return cntList, err
 		}
 		for _, v := range resp.ContainerItems {
-			if v.Name != nil {
-				cntList = append(cntList, *v.Name)
-			} else {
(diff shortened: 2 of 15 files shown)


component/azstorage/azstorage.go
```diff
@@ -573,7 +573,7 @@ func NewazstorageComponent() internal.Component {
 		stConfig: AzStorageConfig{
 			blockSize:      0,
 			maxConcurrency: 32,
-			defaultTier:    getAccessTierType("none"),
+			defaultTier:    getAccessTierType(""),
```
vibhansa-msft, line 576: should it be 'nil' or empty string here?

souravgupta-msft: `getAccessTierType()` method returns nil when empty string is passed.

component/azstorage/block_blob.go
```diff
@@ -120,97 +120,60 @@ func (bb *BlockBlob) UpdateConfig(cfg AzStorageConfig) error {
...
 	}
 
-	// Create the service url
-	bb.Service = azblob.NewServiceURL(*bb.Endpoint, bb.Pipeline)
-
-	// Create the container url
-	bb.Container = bb.Service.NewContainerURL(bb.Config.container)
-
+	// create the container client
+	bb.Container = bb.Service.NewContainerClient(bb.Config.container)
```
vibhansa-msft, line 176: Do we need all three methods one creating service client, one creating container client, and this one which also creates container client. or we can create some common method and use it as needed.

component/azstorage/block_blob.go
```diff
@@ -65,44 +69,40 @@ const (
...
 	}
 
-	bb.downloadOptions = azblob.DownloadFromBlobOptions{
-		BlockSize:                bb.Config.blockSize,
-		Parallelism:              bb.Config.maxConcurrency,
-		ClientProvidedKeyOptions: bb.blobCPKOpt,
+	bb.downloadOptions = &blob.DownloadFileOptions{
+		BlockSize:   bb.Config.blockSize,
+		Concurrency: bb.Config.maxConcurrency,
+		CPKInfo:     bb.blobCPKOpt,
```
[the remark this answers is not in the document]
souravgupta-msft: The type of `bb.blobCPKOpt` is `*blob.CPKInfo`. Its value is nil by default unless has enabled CPK.

component/azstorage/block_blob.go
souravgupta-msft: We are only creating service client here. Container client is created from service client so that it uses the same pipeline options.
`bb.Service.NewContainerClient(bb.Config.container)` call goes to SDK.

component/azstorage/block_blob.go
```diff
@@ -496,6 +488,7 @@ func (bb *BlockBlob) getAttrUsingList(name string) (attr *internal.ObjAttr, err
 	var marker *string = nil
 	blobsRead := 0
 
+	// TODO:: track2 : review : why is the list call retried 20 times?
```
vibhansa-msft, line 491: Some time the list comes back empty though the blob exists and calling it again may fetch the requested blob. Better way would be to convert this function into a loop like dir listing where based on the token we decide whether to conitnue or not.

component/azstorage/block_blob.go
```diff
@@ -590,32 +588,39 @@ func (bb *BlockBlob) List(prefix string, marker *string, count int32) ([]*intern
...
-			Flags:  internal.NewFileBitMap(),
-			MD5:    blobInfo.Properties.ContentMD5,
+			Path:  split(bb.Config.prefixPath, *blobInfo.Name),
+			Name:  filepath.Base(*blobInfo.Name),
+			Mode:  0,
+			Flags: internal.NewFileBitMap(),
+			MD5:   blobInfo.Properties.ContentMD5,
+		}
+
+		if blobInfo.Properties.ContentLength != nil {
```
vibhansa-msft, line 610: These if conditions will slow down the system when we are listing large number of objects as they will be hit for each item. If we do not hit a situation where these are nil, may be we can ignore these checks.

component/azstorage/block_blob.go
```diff
@@ -854,15 +882,14 @@ func trackUpload(name string, bytesTransferred int64, count int64, uploadPtr *in
...
 	log.Trace("BlockBlob::WriteFromFile : name %s", name)
 	//defer exectime.StatTimeCurrentBlock("WriteFromFile::WriteFromFile")()
 
-	blobURL := bb.Container.NewBlockBlobURL(filepath.Join(bb.Config.prefixPath, name))
+	blobClient := bb.Container.NewBlockBlobClient(filepath.Join(bb.Config.prefixPath, name))
 	defer log.TimeTrack(time.Now(), "BlockBlob::WriteFromFile", name)
 
-	var uploadPtr *int64 = new(int64)
-	*uploadPtr = 1
+	uploadPtr := to.Ptr(int64(1))
```
vibhansa-msft, line 892: where is this used? Do we reallly need this variable.

component/azstorage/block_blob.go
souravgupta-msft: Yes, instead of retrying 20 times, it will be better if we depend on the marker returned by the List() call.

component/azstorage/block_blob.go
```diff
@@ -590,32 +588,39 @@ func (bb *BlockBlob) List(prefix string, marker *string, count int32) ([]*intern
...
-	}
-
 	// Process the blobs returned in this result segment (if the segment is empty, the loop body won't execute)
 	// Since block blob does not support acls, we set mode to 0 and FlagModeDefault to true so the fuse layer can return the default permission.
 
 	// For some directories 0 byte meta file may not exists so just create a map to figure out such directories
 	var dirList = make(map[string]bool)
 
 	for _, blobInfo := range listBlob.Segment.BlobItems {
+		if blobInfo.Name == nil {
```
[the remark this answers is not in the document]
souravgupta-msft: This check is to prevent panic situation because `blobInfo.Name` is a pointer.

component/azstorage/block_blob.go
souravgupta-msft: This is used for health monitor.

component/azstorage/block_blob.go
vibhansa-msft: There are three methods in a row which are creating clients so just see if we can extract few things out into a seperate method.

component/azstorage/block_blob.go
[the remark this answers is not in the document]
vibhansa-msft: We shall never get a blob item from backend where details are there but blob name is not there. Even if that situation is hit, better to crash as we do not even know what to do with such blobs.

component/azstorage/block_blob.go
souravgupta-msft: The three methods are,
`NewServiceClient`: this method is for dynamically updating the service client when SAS in the config is updated.
`getServiceClient`: this is the method where we set up the `Auth` and create the service client
`SetupPipeline`: this method calls the getServiceClient() to create the service client and then the container client from it.

souravgupta-msft: I can rename the methods if that makes sense. Like,
`NewServiceClient` to `UpdateServiceClient`
`getServiceClient` to `createServiceClient`

component/azstorage/block_blob.go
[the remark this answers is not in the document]
souravgupta-msft: Removed nil check

component/azstorage/block_blob.go
souravgupta-msft: Removed the nil checks

component/azstorage/block_blob.go
souravgupta-msft: Updated the logic. Please review

vibhansa-msft (approved):

PR #1342, opened 2024-02-12
Track2 Migration: Block blob tests migration

souravgupta-msft:


component/azstorage/azauth.go
```diff
@@ -58,46 +58,12 @@ func getAzAuth(config azAuthConfig) azAuth {
 		}(config.UseHTTP),
 		config.Endpoint)
 
-	if EAccountType.BLOCK() == config.AccountType {
-		return getAzAuthBlob(config)
```
vibhansa-msft, line 62: For blob endpoint we are not returning the config here?

component/azstorage/block_blob.go
```diff
@@ -806,24 +804,24 @@ func (bb *BlockBlob) calculateBlockSize(name string, fileSize int64) (blockSize
...
+		blockSize = blockblob.MaxUploadBlobBytes
 	} else {
 		// buffer / max blocks = block size to use all 50,000 blocks
-		blockSize = int64(math.Ceil(float64(fileSize) / azblob.BlockBlobMaxBlocks))
+		blockSize = int64(math.Ceil(float64(fileSize) / blockblob.MaxBlocks))
 
-		if blockSize < azblob.BlobDefaultDownloadBlockSize {
-			// Block size is smaller then 16MB then consider 16MB as default
-			blockSize = azblob.BlobDefaultDownloadBlockSize
+		if blockSize < blob.DefaultDownloadBlockSize {
```
vibhansa-msft, line 814: This is chaning our check from 16MB to 4MB which means now default block size is smaller.

component/azstorage/block_blob.go
```diff
@@ -806,24 +804,24 @@ func (bb *BlockBlob) calculateBlockSize(name string, fileSize int64) (blockSize
...
 	// If bufferSize <= BlockBlobMaxUploadBlobBytes, then Upload should be used with just 1 I/O request
-	if fileSize <= azblob.BlockBlobMaxUploadBlobBytes {
+	if fileSize <= blockblob.MaxUploadBlobBytes {
 		// Files up to 256MB can be uploaded as a single block
-		blockSize = azblob.BlockBlobMaxUploadBlobBytes
+		blockSize = blockblob.MaxUploadBlobBytes
 	} else {
 		// buffer / max blocks = block size to use all 50,000 blocks
-		blockSize = int64(math.Ceil(float64(fileSize) / azblob.BlockBlobMaxBlocks))
+		blockSize = int64(math.Ceil(float64(fileSize) / blockblob.MaxBlocks))
```
vibhansa-msft, line 812: We shall validate any of these values have changed or not. Existing code: < 256MB single shot upload, > 256MB divide into blocks. If user gives block size use that otherwise calculate in code assuming 16MB default.

component/azstorage/azauth.go
souravgupta-msft: Old unused code. So, removed it.
The blob endpoint calls the getAzAuthBlob() method from the track2 code.

component/azstorage/block_blob.go
souravgupta-msft: The value of the constants are same before and after track2 migration.

component/azstorage/block_blob.go
souravgupta-msft: The value of the constant `azblob.BlobDefaultDownloadBlockSize` was always 4MB. The comment was wrong.

vibhansa-msft (approved):

PR #1355, opened 2024-03-04
Track2 Migration: Adding headers and query params to be logged

souravgupta-msft:


component/azstorage/azstorage_constants.go
```diff
@@ -60,3 +60,14 @@ const (
 	size        = "Size"
 	target      = "Target"
 )
+
+// headers which should be logged and not redacted
+var allowedHeaders []string = []string{
+	"x-ms-version", "x-ms-date", "x-ms-range", "x-ms-delete-snapshots", "x-ms-delete-type-permanent", "x-ms-blob-content-type",
```
vibhansa-msft, line 66: request ids ? they will be important for backend tracing.

component/azstorage/utils.go
```diff
@@ -131,8 +131,8 @@ func getSDKLogOptions() policy.LogOptions {
 	} else {
 		// TODO:: track2 : check which headers and query params should not be redacted
 		return policy.LogOptions{
-			AllowedHeaders:     []string{"x-ms-version"},
-			AllowedQueryParams: []string{},
+			AllowedHeaders:     allowedHeaders,
```
vibhansa-msft, line 134: by default does it log something or everything is redacted ?

component/azstorage/azstorage_constants.go
souravgupta-msft: Request ids are logged by default.

component/azstorage/utils.go
souravgupta-msft: By default, these headers are logged, https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/policy_logging.go#L38-L64

And these query params are logged, https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azcore/runtime/policy_logging.go#L81-L83

souravgupta-msft: The list we provided here are the additional headers and query params that we want to be logged and not redacted.

vibhansa-msft (approved):

ashruti-msft (approved):

PR #1373, opened 2024-03-26
Add lazy-write feature to make flush and close async in nature.

vibhansa-msft:
- Close and flush will return immediately when lazy-write is enabled and upload of dirty file will be scheduled
- File will be locked till upload is in progress and no new handle for the same file can be created during this time
- Before unmount ensure all downloads are complete

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -83,8 +83,8 @@ type FileCache struct {
 	hardLimit         bool
 	diskHighWaterMark float64
 
-	lazyWrite  bool
-	asyncClose sync.WaitGroup
+	lazyWrite    bool
+	fileCloseOpt sync.WaitGroup
 }
 
 // Structure defining your config parameters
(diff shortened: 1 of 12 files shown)


component/file_cache/file_cache.go
```diff
@@ -983,42 +996,59 @@ func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Hand
...
+
+	if !fc.lazyWrite {
+		// Sync close is called so wait till the upload completes
+		return fc.closeFileInternal(options, flock)
+	}
+
+	// Async close is called so schedule the upload and return here
+	fc.asyncClose.Add(1)
+	go fc.closeFileInternal(options, flock) //nolint
+	return nil
```
souravgupta-msft, line 1011: How to know if the CloseFile returned error or not in case of lazy write?

vibhansa-msft: For user application is will always pass and only logs will tell whether file upload failed or not.
We need to think on if anything fails can we re-push the upload request and try again.

souravgupta-msft (approved):

PR #1381, opened 2024-04-05
Fix fuse minor version check

vibhansa-msft:


component/libfuse/libfuse_handler.go
```diff
@@ -322,7 +322,7 @@ func libfuse_init(conn *C.fuse_conn_info_t, cfg *C.fuse_config_t) (res unsafe.Po
 
 	// RHEL still has 3.3 fuse version and it does not allow max_write beyond 128K
 	// Setting this value to 1 MB will fail the mount.
-	fuse_minor := C.get_fuse_minor_version()
+	fuse_minor := common.GetFuseMinorVersion()
```
souravgupta-msft, line 325: If fusermount3 command fails, then should we revert back to 128 KB as max_write?

vibhansa-msft: yes if any error it will return back 0 as result and our check will fail resulting into 128KB max_write set.

souravgupta-msft: In my opinion, we should go with 1MB as default, since this error was only reported in RHEL. Will changing to 128KB have impact other than perf?

vibhansa-msft: Setting to 1MB will not work on any distro where older fuse version is installed. When we are not able to detect the version its safe to assume its older rather then new. When we do the wrong setting the mount will fail. So only when we are certain that its new fuse version we will set 1MB as write buffer otherwise revert back to 128KB default buffer.
Other then write operations this will not have any other impact on performance.

souravgupta-msft (approved):

PR #1382, opened 2024-04-08
TLRU large number of threads causing panic

vibhansa-msft:
runtime: program exceeds 10000-thread limit
fatal error: thread exhaustion

$ grep ".evictAsync in goroutine 416" 10000+Threads.txt  | wc -l
10122


go.mod
```diff
@@ -19,7 +19,7 @@ require (
 	github.com/spf13/pflag v1.0.5
 	github.com/spf13/viper v1.18.2
 	github.com/stretchr/testify v1.9.0
-	github.com/vibhansa-msft/tlru v0.0.0-20230621165448-dbd42234ad22
+	github.com/vibhansa-msft/tlru v0.0.0-20240409045834-9d4436645038
```
souravgupta-msft, line 22: Is the change in this repo?

vibhansa-msft: TLRU is another repo that I created and we take dependency on that. When it expires a node, it sends the details in a go routine, so in case where lru has 10K nodes and all are expired we will end up starting 10K go routines sending the callback. This will result in Blobfuse hitting its thread limit and crash.

CHANGELOG.md
```diff
@@ -1,4 +1,9 @@
-## 2.3.0~preview.1 (Unreleased)
+## 2.3.1 (Unreleased)
```
souravgupta-msft, line 1:
```suggestion
## 2.3.0 (Unreleased)
```

souravgupta-msft (approved):

souravgupta-msft (approved):

PR #1384, opened 2024-04-09
azcli auth mode

souravgupta-msft:
Adding authentication using Azure CLI


vibhansa-msft (approved):

vibhansa-msft:
Can we try running this mode on a system where cli is not installed and then see what it reports back?
Also, on a system where cli is not installed and mode is not set to cli it shall not fail due to any dependency or anything.

souravgupta-msft:
> Can we try running this mode on a system where cli is not installed and then see what it reports back? Also, on a system where cli is not installed and mode is not set to cli it shall not fail due to any dependency or anything.

If az cli is not installed, it will return back the error saying that "az is not found". I will add negative tests for this mode.

What do you mean by this,
"Also, on a system where cli is not installed and mode is not set to cli it shall not fail due to any dependency or anything."

vibhansa-msft (approved):

PR #1387, opened 2024-04-15
Get attribute fix for directory with cpk enabled

abhiguptacse:
- Get atrribute for directory file is failing when CPK is enabled as ListBlob  API does not return the metadata when the blob is encrypted [https://learn.microsoft.com/en-us/rest/api/storageservices/list-blobs](url)
- So fetching the metadata for encrypted blobs explicitly using get properties rest api [https://learn.microsoft.com/en-us/rest/api/storageservices/get-blob-properties](url)

### DIFF
--- a/component/azstorage/block_blob.go
+++ b/component/azstorage/block_blob.go
@@ -574,22 +574,30 @@ func (bb *BlockBlob) List(prefix string, marker *string, count int32) ([]*intern
 	var dirList = make(map[string]bool)
 
 	for _, blobInfo := range listBlob.Segment.BlobItems {
-		attr := &internal.ObjAttr{
-			Path:   split(bb.Config.prefixPath, *blobInfo.Name),
-			Name:   filepath.Base(*blobInfo.Name),
-			Size:   *blobInfo.Properties.ContentLength,
-			Mode:   0,
-			Mtime:  *blobInfo.Properties.LastModified,
-			Atime:  dereferenceTime(blobInfo.Properties.LastAccessedOn, *blobInfo.Properties.LastModified),
-			Ctime:  *blobInfo.Properties.LastModified,
-			Crtime: dereferenceTime(blobInfo.Properties.CreationTime, *blobInfo.Properties.LastModified),
-			Flags:  internal.NewFileBitMap(),
-			MD5:    blobInfo.Properties.ContentMD5,
+		attr := &internal.ObjAttr{}
+		if blobInfo.Properties.CustomerProvidedKeySHA256 != nil && *blobInfo.Properties.CustomerProvidedKeySHA256 != "" {
+			attr, err = bb.getAttrUsingRest(*blobInfo.Name)
+			if err != nil {
+				log.Err("BlockBlob::List : Failed to get properties of blob %s", blobInfo.Name)
+				return blobList, nil, err
+			}
+		} else {
+			attr = &internal.ObjAttr{
+				Path:   split(bb.Config.prefixPath, *blobInfo.Name),
+				Name:   filepath.Base(*blobInfo.Name),
+				Size:   *blobInfo.Properties.ContentLength,
+				Mode:   0,
+				Mtime:  *blobInfo.Properties.LastModified,
+				Atime:  dereferenceTime(blobInfo.Properties.LastAccessedOn, *blobInfo.Properties.LastModified),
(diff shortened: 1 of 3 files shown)


component/azstorage/block_blob.go
```diff
@@ -572,24 +572,38 @@ func (bb *BlockBlob) List(prefix string, marker *string, count int32) ([]*intern
...
+			log.Trace("BlockBlob::List : blob is encrypted with customer provided key so fetching metadata explicitly using REST")
+			attr, err = bb.getAttrUsingRest(*blobInfo.Name)
+			if err != nil {
+				log.Err("BlockBlob::List : Failed to get properties of blob %s", *blobInfo.Name)
+				if errors == nil {
+					errors = err
+				} else {
+					errors = fmt.Errorf("error fetching metadata for blob: %s, %w; %w", *blobInfo.Name, errors, err)
+				}
+				continue
```
souravgupta-msft, line 588: `continue` should not be present here.
If the current blob is a directory (having hdi_isfolder=true), then it needs to be appended to the dirList below. Also, attr needs to be appended to the blobList.

component/azstorage/block_blob.go
```diff
@@ -572,24 +572,38 @@ func (bb *BlockBlob) List(prefix string, marker *string, count int32) ([]*intern
...
+		if blobInfo.Properties.CustomerProvidedKeySHA256 != nil && *blobInfo.Properties.CustomerProvidedKeySHA256 != "" {
+			log.Trace("BlockBlob::List : blob is encrypted with customer provided key so fetching metadata explicitly using REST")
+			attr, err = bb.getAttrUsingRest(*blobInfo.Name)
+			if err != nil {
+				log.Err("BlockBlob::List : Failed to get properties of blob %s", *blobInfo.Name)
+				if errors == nil {
+					errors = err
+				} else {
+					errors = fmt.Errorf("error fetching metadata for blob: %s, %w; %w", *blobInfo.Name, errors, err)
+				}
```
souravgupta-msft, line 587: What is the use of this? We can directly return the error after this loop.

souravgupta-msft: Aren't you updating errors variable with the last returned err?
My suggestion would be to log the error here (if not nil) and return err after this loop.

souravgupta-msft: Maybe we can just log and return the first error we got here. If GetProperties fails for a blob, we should not continue.
In this case, you can remove the errors variable and lines 615-617.

component/azstorage/block_blob.go
```diff
@@ -572,24 +572,32 @@ func (bb *BlockBlob) List(prefix string, marker *string, count int32) ([]*intern
...
-			Size:   *blobInfo.Properties.ContentLength,
-			Mode:   0,
-			Mtime:  *blobInfo.Properties.LastModified,
-			Atime:  dereferenceTime(blobInfo.Properties.LastAccessedOn, *blobInfo.Properties.LastModified),
-			Ctime:  *blobInfo.Properties.LastModified,
-			Crtime: dereferenceTime(blobInfo.Properties.CreationTime, *blobInfo.Properties.LastModified),
-			Flags:  internal.NewFileBitMap(),
-			MD5:    blobInfo.Properties.ContentMD5,
+		attr := &internal.ObjAttr{}
+		if blobInfo.Properties.CustomerProvidedKeySHA256 != nil && *blobInfo.Properties.CustomerProvidedKeySHA256 != "" {
```
souravgupta-msft, line 577: Can you please add changelog and tests for this change.

abhiguptacse: For the tests i think the ideal way to go about this would be running all the tests in block_blob_test.go with CPK enabled in config, i think the change might go in pipeline for this. 
Me adding test for only get attribute could lead to cases being missed just like this bug was missed. 
Let me know what you think ?

souravgupta-msft: Each test in block_blob_test.go runs with a separate config. So, it might not be straight forward to enable CPK and run those tests. Can you instead add the tests for the use case which was failing.

abhiguptacse: updated.

souravgupta-msft (approved):

vibhansa-msft (approved):

PR #1400, opened 2024-05-07
Display mount command

ashruti-msft:
Added code to display the entire mount command and fixed a spelling error.

### DIFF
--- a/component/libfuse/libfuse.go
+++ b/component/libfuse/libfuse.go
@@ -36,7 +36,6 @@ package libfuse
 import (
 	"context"
 	"fmt"
-	"os"
 
 	"github.com/Azure/azure-storage-fuse/v2/common"
 	"github.com/Azure/azure-storage-fuse/v2/common/config"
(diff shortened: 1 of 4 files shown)


component/libfuse/libfuse.go
```diff
@@ -251,7 +252,7 @@ func (lf *Libfuse) Validate(opt *LibfuseOptions) error {
 //	Return failure if any config is not valid to exit the process
 func (lf *Libfuse) Configure(_ bool) error {
 	log.Trace("Libfuse::Configure : %s", lf.Name())
-
+	log.Info("Mount Command: %s", os.Args)
```
vibhansa-msft, line 255: This will print only once, lets add it to every log otherwise customer has to write scripts to figure out pid from here and then grep on logs to filter other logs from same mount.

vibhansa-msft (approved):

souravgupta-msft (approved):

PR #1407, opened 2024-05-09
Added mount path in the syslog 

ashruti-msft:
Example of syslog :
```
May  9 15:26:03 anubhuti-Virtual-Machine blobfuse2[2332884]: [/home/anubhuti/mntdir] LOG_DEBUG [azstorage.go (306)]: AzStorage::StreamDir : Retrieved 6 objects with  marker for Path
May  9 15:26:03 anubhuti-Virtual-Machine blobfuse2[2332884]: [/home/anubhuti/mntdir] LOG_TRACE [libfuse_handler.go (493)]: Libfuse::libfuse_releasedir : , handle: 19

```
Example of base log:

`Fri May 10 15:27:07 IST 2024 : blobfuse2[2352962] : [/home/anubhuti/mntdir] LOG_DEBUG [utils.go (151)]: SDK(Retry) : response 200
Fri May 10 15:27:07 IST 2024 : blobfuse2[2352962] : [/home/anubhuti/mntdir] LOG_DEBUG [utils.go (151)]: SDK(Retry) : exit due to non-retriable status code`

### DIFF
--- a/common/log/sys_logger.go
+++ b/common/log/sys_logger.go
@@ -38,6 +38,7 @@ import (
 	"fmt"
 	"log"
 	"log/syslog"
+	"os"
 	"path/filepath"
 	"runtime"
 
@@ -120,7 +121,11 @@ func getSyslogLevel(lvl common.LogLevel) syslog.Priority {
 func (l *SysLogger) write(lvl string, format string, args ...interface{}) {
 	_, fn, ln, _ := runtime.Caller(3)
 	msg := fmt.Sprintf(format, args...)
-	l.logger.Print(lvl, " [", filepath.Base(fn), " (", ln, ")]: ", msg)
+	mountPath := os.Args[1]
+	if mountPath == "mount" {
+		mountPath = os.Args[2]
+	}
+	l.logger.Print("[", mountPath, "] ", lvl, " [", filepath.Base(fn), " (", ln, ")]: ", msg)
 }
 
 func (l *SysLogger) Debug(format string, args ...interface{}) {
--- a/common/log/sys_logger.go
+++ b/common/log/sys_logger.go
@@ -38,7 +38,6 @@ import (
 	"fmt"
 	"log"
 	"log/syslog"
-	"os"
 	"path/filepath"
 	"runtime"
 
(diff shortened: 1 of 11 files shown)


common/log/sys_logger.go
```diff
@@ -120,7 +121,11 @@ func getSyslogLevel(lvl common.LogLevel) syslog.Priority {
 func (l *SysLogger) write(lvl string, format string, args ...interface{}) {
 	_, fn, ln, _ := runtime.Caller(3)
 	msg := fmt.Sprintf(format, args...)
-	l.logger.Print(lvl, " [", filepath.Base(fn), " (", ln, ")]: ", msg)
+	mountPath := os.Args[1]
+	if mountPath == "mount" {
+		mountPath = os.Args[2]
+	}
+	l.logger.Print("[", mountPath, "] ", lvl, " [", filepath.Base(fn), " (", ln, ")]: ", msg)
```
souravgupta-msft, line 128: Also add this in base level logging

common/util.go
```diff
@@ -66,6 +67,7 @@ func IsDirectoryMounted(path string) bool {
 
 	// removing trailing / from the path
 	path = strings.TrimRight(path, "/")
+	MountPath = path
```
vibhansa-msft, line 70: Instead of doing this in this method, we shall do it in the command processing
there is a step where all validations are done and we start setting the logging infra.
"common.EnableMonitoring = options.MonitorOpt.EnableMon" at this line we can save the mount path as well.

common/log/sys_logger.go
```diff
@@ -120,7 +120,7 @@ func getSyslogLevel(lvl common.LogLevel) syslog.Priority {
 func (l *SysLogger) write(lvl string, format string, args ...interface{}) {
 	_, fn, ln, _ := runtime.Caller(3)
 	msg := fmt.Sprintf(format, args...)
-	l.logger.Print(lvl, " [", filepath.Base(fn), " (", ln, ")]: ", msg)
+	l.logger.Print("[", common.MountPath, "] ", lvl, " [", filepath.Base(fn), " (", ln, ")]: ", msg)
```
vibhansa-msft, line 123: Instead of dumping path after the pid and all, can we do something like [pid:path] and then rest of the log remains the same. This help us identify that pid is associated with the mount path.

ashruti-msft: Syslog computes the time and pid internally I think because I couldn't find any code written by us for it. I can make this change in base log, but I just tried to keep the format consistent.

souravgupta-msft (approved):

vibhansa-msft (approved):

PR #1416, opened 2024-05-16
Autoconfig

ashruti-msft:
Added calcultions for parameters for prefetch, parallelism, mem-size-mb, disk-size-mb in block_cache. 
 Updated test cases according to new calculations.
 
 P.S: While printing blockCache configuations some values were in bytes and some in MB so I made the output uniform for all in bytes.


component/block_cache/block_cache.go
```diff
@@ -41,16 +41,19 @@ import (
...
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/Azure/azure-storage-fuse/v2/common"
 	"github.com/Azure/azure-storage-fuse/v2/common/config"
 	"github.com/Azure/azure-storage-fuse/v2/common/log"
 	"github.com/Azure/azure-storage-fuse/v2/internal"
 	"github.com/Azure/azure-storage-fuse/v2/internal/handlemap"
+	"github.com/pbnjay/memory"
```
vibhansa-msft, line 56: We shall use some standard library instead of relying on some project here.

component/block_cache/block_cache.go
```diff
@@ -262,6 +261,13 @@ func (bc *BlockCache) Configure(_ bool) error {
 			}
 		}
 	}
+	var stat syscall.Statfs_t
```
vibhansa-msft, line 264: This shall be done only if tmp-path is configured.

component/block_cache/block_cache.go
```diff
@@ -206,22 +208,23 @@ func (bc *BlockCache) Configure(_ bool) error {
 		bc.blockSize = uint64(conf.BlockSize * float64(_1MB))
 	}
 
-	bc.memSize = uint64(4192) * _1MB
+	var sysinfo syscall.Sysinfo_t
+	err = syscall.Sysinfo(&sysinfo)
+	if err != nil {
+		panic(err)
```
vibhansa-msft, line 214: We shall not panic here, instead return back error as we do in all configure methods.

component/block_cache/block_cache.go
```diff
@@ -261,14 +264,21 @@ func (bc *BlockCache) Configure(_ bool) error {
 				return fmt.Errorf("config error in %s [%s]", bc.Name(), err.Error())
 			}
 		}
+		var stat syscall.Statfs_t
+		_ = syscall.Statfs(bc.tmpPath, &stat)
+		bc.diskSize = uint64(0.6 * float64(stat.Bavail) * float64(stat.Bsize))
```
vibhansa-msft, line 269: lets make this also as 80%

component/block_cache/block_cache.go
souravgupta-msft: We can log the error here and assign memSize with some default value.

component/block_cache/block_cache.go
```diff
@@ -1258,7 +1268,7 @@ func (bc *BlockCache) diskEvict(node *list.Element) {
 // checkDiskUsage : Callback to check usage of disk and decide whether eviction is needed
 func (bc *BlockCache) checkDiskUsage() bool {
 	data, _ := common.GetUsage(bc.tmpPath)
-	usage := uint32((data * 100) / float64(bc.diskSize))
+	usage := uint32((data * 100) / float64(bc.diskSize/_1MB))
```
vibhansa-msft, line 1271: why are we changing this method here ? disk size is in MB and data is also in MB right?

component/block_cache/block_cache.go
```diff
@@ -261,14 +264,21 @@ func (bc *BlockCache) Configure(_ bool) error {
 				return fmt.Errorf("config error in %s [%s]", bc.Name(), err.Error())
 			}
 		}
+		var stat syscall.Statfs_t
+		_ = syscall.Statfs(bc.tmpPath, &stat)
```
vibhansa-msft, line 268: we shall check for error here if statfs fails

component/block_cache/block_cache_test.go
```diff
@@ -159,11 +160,11 @@ func (suite *blockCacheTestSuite) TestEmpty() {
...
 	suite.assert.EqualValues(tobj.blockCache.blockSize, 16*_1MB)
-	suite.assert.EqualValues(tobj.blockCache.memSize, 4192*_1MB)
-	suite.assert.EqualValues(tobj.blockCache.diskSize, 4192)
+	//Removed memory check since memory calculated during testing may differ
+	suite.assert.EqualValues(tobj.blockCache.diskSize, 0)
 	suite.assert.EqualValues(tobj.blockCache.diskTimeout, defaultTimeout)
-	suite.assert.EqualValues(tobj.blockCache.workers, 128)
-	suite.assert.EqualValues(tobj.blockCache.prefetch, MIN_PREFETCH)
+	suite.assert.EqualValues(tobj.blockCache.workers, uint32(3*runtime.NumCPU()))
+	suite.assert.EqualValues(tobj.blockCache.prefetch, uint32(2*runtime.NumCPU()))
```
vibhansa-msft, line 167: we shall check for disk and memory size also here for autoconfiguration

component/block_cache/block_cache.go
souravgupta-msft, line 268: Same as previous comment. Log the error here and assign diskSize with some default value.

component/block_cache/block_cache.go
ashruti-msft: For mem-size and block-size the values are saved in bytes but for disk-size the value was saved in MB so I changed it to bytes to make the output uniform. Hence here I need to convert bytes to MB for the test case

component/block_cache/block_cache_test.go
ashruti-msft: I added a memory check but 2/3 times it showed the test as failed because it wasn't able to capture the exact same value for free memory which the code had captured before while calculating.

souravgupta-msft:
Can you also write tests for validating these options that are auto computed. For example, you can compute free memory using the `free` command in Linux and compare the output against the one computed by the code.

component/block_cache/block_cache_test.go
ashruti-msft: Another thing for this test case the config is empty with contents read-only: true. Since tmp-path is not configured and disk-size-mb is not given the disk-size is not set and value is 0. So, I am validating that here.

component/block_cache/block_cache_test.go
```diff
@@ -159,16 +163,76 @@ func (suite *blockCacheTestSuite) TestEmpty() {
...
+	free, err := strconv.Atoi(strings.TrimSpace(out.String()))
+	suite.assert.Nil(err)
+	expected := uint64(0.8 * float64(free))
+	actual := tobj.blockCache.memSize
+	var difference float64
+	if actual > expected {
+		difference = float64(actual - expected)
+	} else {
+		difference = float64(expected - actual)
+	}
```
souravgupta-msft, line 203: You can use `math.Abs()` method here

component/block_cache/block_cache_test.go
```diff
@@ -159,16 +163,76 @@ func (suite *blockCacheTestSuite) TestEmpty() {
...
+	freeDisk, err := strconv.Atoi(strings.TrimSpace(out.String()))
+	suite.assert.Nil(err)
+	expected := uint64(0.8 * float64(freeDisk))
+	actual := tobj.blockCache.diskSize
+	var difference float64
+	if actual > expected {
+		difference = float64(actual - expected)
+	} else {
+		difference = float64(expected - actual)
+	}
```
souravgupta-msft, line 231: Same as previous.

souravgupta-msft (approved):

souravgupta-msft (approved):

component/block_cache/block_cache.go
```diff
@@ -206,22 +208,26 @@ func (bc *BlockCache) Configure(_ bool) error {
...
 	}
 
-	bc.memSize = uint64(4192) * _1MB
 	if config.IsSet(compName + ".mem-size-mb") {
 		bc.memSize = conf.MemSize * _1MB
+	} else {
+		var sysinfo syscall.Sysinfo_t
+		err = syscall.Sysinfo(&sysinfo)
+		if err != nil {
+			log.Err("BlockCache::Configure : config error %s [%s]. Assigning a pre-defined value of 4GB.", bc.Name(), err.Error())
```
vibhansa-msft, line 217: Let's add some section in Readme which describes our logic of auto-config like 2xCPU or 80% of free disk etc. Also, you shall update changelog with the same inputs.

vibhansa-msft (approved):

README.md
```diff
@@ -145,6 +145,12 @@ To learn about a specific command, just include the name of the command (For exa
     * `--block-cache-prefetch=<Number of blocks>`: Number of blocks to prefetch at max when sequential reads are in progress.
     * `--block-cache-parallelism=<count>`: Number of parallel threads doing upload/download operation.
     * `--block-cache-prefetch-on-open=true`: Start prefetching on open system call instead of waiting for first read. Enhances perf if file is read sequentially from offset 0.
+    Default values, if not assigned in config, for the following parameters in block-cache are calculated as follows:
+        * Memory preallocated for Block-Cache is 80% of free memory available
+        * Disk Cache size is 80% of free disk space
+        * Prefetch is 2 times number of CPU cores
+        * Parallelism is 3 times the number of CPU cores
```
souravgupta-msft, line 152: Nit, you can mention this with the above options, like
`--block-cache-disk-size=<SIZE IN MB>`: Disk space to be used for caching. Default - 80% of free disk space.

PR #1435, opened 2024-06-13
Flexible File Filtering Feature for Blobfuse Mounting

t-yashmahale:
This pull request introduces advanced filtering capabilities to Blobfuse, empowering users to selectively mount cloud files based on size, regex patterns, modification time, and formats like PDF, JPG, TXT, and more. Efficient thread pool management at both file and filter levels ensures swift processing, enhancing user flexibility and optimizing resource utilization. This enhancement streamlines the file mounting process, enabling users to tailor their cloud storage access with precision and performance.


.gitignore
```diff
@@ -21,3 +21,4 @@ lint.log
 azure-storage-fuse
 bfusemon
 test/scripts/dirIterate.go
+filterbin
```
vibhansa-msft, line 24: There is this binary file __debug_* which can be removed from the PR.

component/azstorage/config.go
```diff
@@ -182,10 +183,12 @@ type AzStorageOptions struct {
 	MaxResultsForList       int32  `config:"max-results-for-list" yaml:"max-results-for-list"`
 	DisableCompression      bool   `config:"disable-compression" yaml:"disable-compression"`
 	Telemetry               string `config:"telemetry" yaml:"telemetry"`
-	HonourACL               bool   `config:"honour-acl" yaml:"honour-acl"`
-	CPKEnabled              bool   `config:"cpk-enabled" yaml:"cpk-enabled"`
-	CPKEncryptionKey        string `config:"cpk-encryption-key" yaml:"cpk-encryption-key"`
-	CPKEncryptionKeySha256  string `config:"cpk-encryption-key-sha256" yaml:"cpk-encryption-key-sha256"`
+	BlobFilter              string `config:"blobFilter" yaml:"blobFilter"`
```
vibhansa-msft, line 186: Move this new config towards the end of the structure so that diff does not show existing items.

component/azstorage/config.go
```diff
@@ -385,7 +388,18 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 	az.stConfig.cancelListForSeconds = opt.CancelListForSeconds
 
 	az.stConfig.telemetry = opt.Telemetry
-
+	az.stConfig.blobFilter = opt.BlobFilter
+	if len(opt.BlobFilter) > 0 {
+		fmt.Println(opt.BlobFilter)
```
vibhansa-msft, line 393: You can move fmt.Println to log.Debug as blobfuse will have no console attached to print,

component/azstorage/config.go
```diff
@@ -385,7 +388,18 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
...
+		erro := filter.ParseInp(&opt.BlobFilter)
+		fmt.Println(len(filter.GlbFilterArr))
+		fmt.Println(len(opt.BlobFilter))
+		if erro != nil {
+			log.Err("no filters applied, mount failed")
```
vibhansa-msft, line 398: For log.Debug/Err we are following a format which is consistent across all our logs. Just check other existing logs in this file and modify accordingly.

component/azstorage/config.go
```diff
@@ -385,7 +388,18 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 	az.stConfig.cancelListForSeconds = opt.CancelListForSeconds
 
 	az.stConfig.telemetry = opt.Telemetry
-
+	az.stConfig.blobFilter = opt.BlobFilter
+	if len(opt.BlobFilter) > 0 {
+		fmt.Println(opt.BlobFilter)
+		erro := filter.ParseInp(&opt.BlobFilter)
+		fmt.Println(len(filter.GlbFilterArr))
```
vibhansa-msft, line 395: These shall be converted to log.Debug or log.Info

component/azstorage/azstorage.go
```diff
@@ -331,6 +333,10 @@ func (az *AzStorage) StreamDir(options internal.StreamDirOptions) ([]*internal.O
 	// increment streamdir call count
 	azStatsCollector.UpdateStats(stats_manager.Increment, streamDir, (int64)(1))
 
+	//check for filters provided
+	if len(az.stConfig.blobFilter) > 0 { //only apply filter if user has given
```
vibhansa-msft, line 337: After parsing we might have created a filter object so instead of checking the string length here rather we can check if that object is nil or not, will be a faster check.

component/azstorage/config.go
```diff
@@ -385,7 +388,18 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
...
+		fmt.Println(len(opt.BlobFilter))
+		if erro != nil {
+			log.Err("no filters applied, mount failed")
+			return erro
+		}
+	}
+	//blobfilter
```
vibhansa-msft, line 402: Deadcode can be removed.

component/azstorage/azstorage.go
```diff
@@ -521,7 +527,23 @@ func (az *AzStorage) ReadLink(options internal.ReadLinkOptions) (string, error)
 // Attribute operations
 func (az *AzStorage) GetAttr(options internal.GetAttrOptions) (attr *internal.ObjAttr, err error) {
 	//log.Trace("AzStorage::GetAttr : Get attributes of file %s", name)
-	return az.storage.GetAttr(options.Name)
+	resp, err := az.storage.GetAttr(options.Name)
+	if err != nil {
+		return resp, err
+	}
+
+	if len(az.stConfig.blobFilter) > 0 {
```
vibhansa-msft, line 535: Instead of checking length try to check object is nil or not.

component/azstorage/azstorage.go
```diff
@@ -521,7 +527,23 @@ func (az *AzStorage) ReadLink(options internal.ReadLinkOptions) (string, error)
...
+		fv1 := &filter.FileValidator{
+			FilterArr: filter.GlbFilterArr,
```
vibhansa-msft, line 537: Here we can name the variables better than fv1 and GlbFilterArr.

component/azstorage/azstorage.go
```diff
@@ -331,6 +333,10 @@ func (az *AzStorage) StreamDir(options internal.StreamDirOptions) ([]*internal.O
 	// increment streamdir call count
 	azStatsCollector.UpdateStats(stats_manager.Increment, streamDir, (int64)(1))
 
+	//check for filters provided
+	if len(az.stConfig.blobFilter) > 0 { //only apply filter if user has given
+		new_list = filter.ApplyFilterOnBlobs(new_list)
```
vibhansa-msft, line 338: Shall we keep the input and output variables same ? just think it might not be a safer way to do conversion.

internal/filter/filter.go
```diff
@@ -0,0 +1,41 @@
+package filter
+
+import (
+	"sync/atomic"
+
+	"github.com/Azure/azure-storage-fuse/v2/internal"
+)
+
+var GlbFilterArr [][]Filter //it will store the fliters, outer array splitted by ||, inner array splitted by &&
```
vibhansa-msft, line 9: As part of design we shall move these filters to be a class so that there can be multiple instances of such filters across the blobfuse code. Current design of having it as a global variable restricts that only one such filter can exists in scope of blobfuse instance.

internal/filter/filter.go
```diff
@@ -0,0 +1,41 @@
...
+
+func ApplyFilterOnBlobs(fileInfos []*internal.ObjAttr) []*internal.ObjAttr { //function called from azstorage.go streamDir func
```
vibhansa-msft, line 11: From outside it shall be an object based call.
- read config as a string from cli
- parse it and create a filter object
- filtering code shall be executed on this object
- if filters are not applied then object shall be nil
- if not nil then apply the filters and it will return back a new slice.
- this model will allow us to have multiple instances of filter across the code if required,

tools/filter/formatfilter.go
```diff
@@ -0,0 +1,34 @@
+package main
```
souravgupta-msft, line 1: tools/filter can be removed

component/azstorage/azstorage.go
```diff
@@ -331,6 +333,12 @@ func (az *AzStorage) StreamDir(options internal.StreamDirOptions) ([]*internal.O
 	// increment streamdir call count
 	azStatsCollector.UpdateStats(stats_manager.Increment, streamDir, (int64)(1))
 
+	//check for filters provided
+	if az.stConfig.filters != nil { //only apply if user has given filter
+		filtered_list := az.stConfig.filters.ApplyFilterOnBlobs(new_list)
+		// return filtered_list, *new_marker, nil
```
vibhansa-msft, line 339: Better to return here instead of copying over the slice again.

internal/filter/accesstier.go
```diff
@@ -0,0 +1,52 @@
...
+
+type AccessTierFilter struct {
+	opr  string
+	tier string
+}
+
+func (filter AccessTierFilter) Apply(fileInfo *internal.ObjAttr) bool {
+	// fmt.Println("AccessTier filter ", filter, " file name ", (*fileInfo).Name)  DEBUG PRINT
+	fmt.Println("inside filter tier ", filter, " with given tier ", filter.tier, " and file tier ", fileInfo.Tier)
+	if (filter.opr == "=") && (filter.tier == strings.ToLower(fileInfo.Tier)) {
```
vibhansa-msft, line 19: Multiple string checks can be avoided.

internal/filter/accesstier.go
```diff
@@ -0,0 +1,52 @@
...
+func newAccessTierFilter(args ...interface{}) Filter {
+	return AccessTierFilter{
+		opr:  args[0].(string),
+		tier: args[1].(string),
+	}
+}
+
+func giveAccessTierFilterObj(singleFilter *string) (Filter, error) {
+	(*singleFilter) = strings.Map(StringConv, (*singleFilter)) //remove all spaces and make all upperCase to lowerCase
+	sinChk := (*singleFilter)[4:5]                             //single char after tier (ex- tier=hot , here sinChk will be "=")
```
vibhansa-msft, line 37: Instead of magic numbers use len(<keyword>)

PR #1445, opened 2024-07-03
Redirecting stream config to BlockCache

ashruti-msft:


component/block_cache/stream.go
```diff
@@ -41,14 +41,12 @@ import (
 	"github.com/Azure/azure-storage-fuse/v2/common/config"
 	"github.com/Azure/azure-storage-fuse/v2/common/log"
 	"github.com/Azure/azure-storage-fuse/v2/internal"
-	"github.com/Azure/azure-storage-fuse/v2/internal/handlemap"
 
 	"github.com/pbnjay/memory"
 )
 
 type Stream struct {
 	internal.BaseComponent
```
souravgupta-msft, line 49: We can remove this and other methods like SetNextComponent(), Priority(), etc. if not needed.

component/block_cache/stream.go
```diff
@@ -127,93 +125,33 @@ func (st *Stream) Configure(_ bool) error {
...
-func NewStreamComponent() internal.Component {
-	comp := &Stream{}
-	comp.SetName(compName)
-	return comp
-}
-
 // On init register this component to pipeline and supply your constructor
 func init() {
-	internal.AddComponent(compName, NewStreamComponent)
+	internal.AddComponent(compName, NewBlockCacheComponent)
```
souravgupta-msft, line 146: This can be removed as it is already done in the block_cache init() method

internal/pipeline_test.go
```diff
@@ -88,6 +108,8 @@ func (suite *pipelineTestSuite) SetupTest() {
 	AddComponent("ComponentA", NewComponentA)
 	AddComponent("ComponentB", NewComponentB)
 	AddComponent("ComponentC", NewComponentC)
+	AddComponent("stream", NewComponentStream)
```
vibhansa-msft, line 111: In nightly there is a datavalidation using block-cache state. Let's duplicate that step and run it using stream config to verify things are working fine in case of auto injested block-cache to replace streaming options.

component/block_cache/block_cache.go
```diff
@@ -175,7 +175,13 @@ func (bc *BlockCache) Stop() error {
 //	Return failure if any config is not valid to exit the process
 func (bc *BlockCache) Configure(_ bool) error {
 	log.Trace("BlockCache::Configure : %s", bc.Name())
-
+	if common.IsStream {
```
jainakanksha-msft, line 178: Please add a unit test case for this.

vibhansa-msft:
Can we also add streaming config steps in pipeline to validate that things are working fine with streaming config.

vibhansa-msft (approved):

PR #1446, opened 2024-07-03
Added min prefetch check

ashruti-msft:


vibhansa-msft (approved):

component/block_cache/block_cache.go
```diff
@@ -227,7 +228,7 @@ func (bc *BlockCache) Configure(_ bool) error {
 	}
 
 	bc.prefetchOnOpen = conf.PrefetchOnOpen
-	bc.prefetch = uint32(2 * runtime.NumCPU())
+	bc.prefetch = uint32(math.Max((MIN_PREFETCH*2)+1, (float64)(2*runtime.NumCPU())))
```
souravgupta-msft, line 231: In case of auto-tuning the prefetch value, should we consider the case where the below check might fail resulting in mount failure.
`if (uint64(bc.prefetch) * uint64(bc.blockSize)) > bc.memSize`

ashruti-msft: We can set the value of prefetch to the min i.e., 11 in case mounting fails due to the default value calculated by us. Would that be better to ensure least amount of mount failures due to autoconfig?

vibhansa-msft: We shall not fail when values are calculated by us. If some parameter was given by user and we were not able to set other values then we can but everything was calculated by us then it does not make sense to fail the mount.

vibhansa-msft: Anubhuti, can you try this case out where allocated memory is lower then the number of prefetch blocks set. Just want to validate that nothing goes for a toss in that case may be system will slow down but it shall not crash or anything.

ashruti-msft: I already tried this case we have a check in place to ensure that memory should be higher than prefetch and it fails if prefetch is more than memory allocated.

ashruti-msft: > We shall not fail when values are calculated by us. If some parameter was given by user and we were not able to set other values then we can but everything was calculated by us then it does not make sense to fail the mount.

Ok I'll make the required change to make sure we don't fail when values are calculated by us.

souravgupta-msft (approved):

PR #1452, opened 2024-07-10
ObjectID info updated and simplified base config

ashruti-msft:
I have removed most options which i thought users may not use commonly. Let me know if I need to include/delete any config parameters. Can we remove any options under azstorage? it looks a bit cluttered right now.


README.md
```diff
@@ -10,6 +10,7 @@ Blobfuse2 is stable, and is ***supported by Microsoft*** provided that it is use
 ##  NOTICE
 - We have seen some customer issues around files getting corrupted when `streaming` is used in write mode. Kindly avoid using this feature for write while we investigate and resolve it.
 - You can now use block-cache instead of streaming for both read and write workflows, which offers much better performance compared to streaming. To enable `block-cache` instead of `streaming`, use `--block-cache` in CLI param or `block-cache` as component in config file instead of `streaming`.
+- Following the release of version 2.3.0, blobfuse no longer supports object-ID based OAuth for Managed Identity. Users requiring this feature must utilize CLI-based login and ensure the Azure CLI is installed on their system.
```
vibhansa-msft, line 13: Or migrate to using app-id based authenication.

README.md
```diff
@@ -167,7 +168,6 @@ To learn about a specific command, just include the name of the command (For exa
     * `AZURE_STORAGE_SAS_TOKEN`: Specifies the SAS token to use for authentication.
 - Managed Identity auth:
     * `AZURE_STORAGE_IDENTITY_CLIENT_ID`: Only one of these three parameters are needed if multiple identities are present on the system.
-    * `AZURE_STORAGE_IDENTITY_OBJECT_ID`: Only one of these three parameters are needed if multiple identities are present on the system.
```
vibhansa-msft, line 170: In AzStorage can we add an error message and fail the mount saying object-id is no longer supported kindly use azcli or use app-id based authentication. Otherwise there will be issues coming in where customers will say their mount failed and we need to debug why it failed. Better to give upfront error message stating the same.

README.md
souravgupta-msft: In my opinion, we should not say that we don't support ObjectID with MI. We can say that it is supported but with the dependency of Azure CLI being installed in their system. If they can't install Azure CLI, they can migrate to app-id based authentication.

README.md
souravgupta-msft: Same as previous, objectID is not deprecated, but has dependency on Azure CLI being installed.

vibhansa-msft: We shall also mark this PR and store it somewhere, when and if AzIdentity adds the support we will have to revert this exact change.

vibhansa-msft (approved):

cmd/mount.go
```diff
@@ -410,8 +410,13 @@ var mountCmd = &cobra.Command{
 		log.Debug("Mount allowed on nonempty path : %v", options.NonEmpty)
 		pipeline, err = internal.NewPipeline(options.Components, !daemon.WasReborn())
 		if err != nil {
-			log.Err("mount : failed to initialize new pipeline [%v]", err)
-			return Destroy(fmt.Sprintf("failed to initialize new pipeline [%s]", err.Error()))
+			if err.Error() == "Azure CLI not found on path" {
+				log.Err("mount : failed to initialize new pipeline :: To authenticate using MSI with object-ID, ensure Azure CLI is installed. Alternatively, use app/client ID for authentication. [%v]", err)
+				return Destroy(fmt.Sprintf("failed to initialize new pipeline :: To authenticate using MSI with object-ID, ensure Azure CLI is installed. Alternatively, use app/client ID for authentication. [%s]", err.Error()))
```
souravgupta-msft, line 415: Same change in other places where this notice has been added. Resource ID can also be used.
```suggestion
				return Destroy(fmt.Sprintf("failed to initialize new pipeline :: To authenticate using MSI with object-ID, ensure Azure CLI is installed. Alternatively, use app/client ID or resource ID for authentication. [%s]", err.Error()))
```

setup/baseConfig.yaml
```diff
@@ -27,10 +27,6 @@
 #  10. 'sdk-trace' has been removed with v2.3.0 release and setting log level to log_debug will auto enable these logs.
```
souravgupta-msft, line 27: Should we also rename baseConfig to simplifiedConfig or something else?

README.md
```diff
@@ -10,6 +10,7 @@ Blobfuse2 is stable, and is ***supported by Microsoft*** provided that it is use
 ##  NOTICE
 - We have seen some customer issues around files getting corrupted when `streaming` is used in write mode. Kindly avoid using this feature for write while we investigate and resolve it.
 - You can now use block-cache instead of streaming for both read and write workflows, which offers much better performance compared to streaming. To enable `block-cache` instead of `streaming`, use `--block-cache` in CLI param or `block-cache` as component in config file instead of `streaming`.
+- As of version 2.3.0, blobfuse has updated its authentication methods. For Managed Identity, Object-ID based OAuth is solely accessible via CLI-based login, requiring Azure CLI on the system. For a dependency-free option, users may utilize Application/Client-ID based authentication.
```
souravgupta-msft, line 13: Mention ResourceID also here

souravgupta-msft (approved):

setup/baseConfig.yaml
ashruti-msft: I wanted to change it to basicConfig since its similar to baseConfig? I'm not sure because most customers might be accustomed to looking for baseConfig. What do you think about this?

souravgupta-msft: baseConfig name is fine. Let' remove the comments above from line 1-28, since these are already present in the advanced config

ashruti-msft: Cool

PR #1453, opened 2024-07-10
Cleanup stale mount in remount case

vibhansa-msft:
Auto cleanup stale mounts when remounting.


ashruti-msft (approved):

common/util.go
```diff
@@ -85,21 +85,86 @@ func IsDirectoryMounted(path string) bool {
...
+			continue
+		}
+
+		cmd = exec.Command("ps", "-o", "args=", "-p", pid)
+		out.Reset()
+		cmd.Stdout = &out
+
+		err := cmd.Run()
+		if err != nil {
+			return true, fmt.Errorf("failed to get command line arguments for pid %s", pid)
```
souravgupta-msft, line 121: Same as previous, log or return the error

cmd/mount.go
```diff
@@ -106,7 +106,32 @@ func (opt *mountOptions) validate(skipNonEmptyMount bool) error {
...
+		log.Info("Mount::validate : Mount directory is already mounted, trying to cleanup")
+		active, err := common.IsMountActive(opt.MountPath)
+		if active || err != nil {
+			// Previous mount is still active so we need to fail this mount
+			return fmt.Errorf("directory is already mounted")
+		} else {
+			// Previous mount is in stale state so lets cleanup the state
+			log.Info("Mount::validate : Cleaning up stale mount")
+			if err = unmountBlobfuse2(opt.MountPath); err != nil {
+				return fmt.Errorf("directory is already mounted, unmount manually before remount.")
```
souravgupta-msft, line 119:
```suggestion
				return fmt.Errorf("directory is already mounted, unmount manually before remount [%s]", err.Error())
```

vibhansa-msft: For any reason if we failed to unmount, we do not want user to have more info on the console and rather ask them to try to unmount manually. Hence I skipped adding the err.Error() here intentionally. If user gets any sort of error like permission denided or anything, it will not make any sense to it.

common/util.go
```diff
@@ -85,21 +85,86 @@ func IsDirectoryMounted(path string) bool {
...
+	var out bytes.Buffer
+	cmd := exec.Command("pidof", "blobfuse2")
+	cmd.Stdout = &out
+
+	err := cmd.Run()
+	if err != nil {
+		if err.Error() == "exit status 1" {
+			return false, nil
+		} else {
+			return true, fmt.Errorf("failed to get pid of blobfuse2")
```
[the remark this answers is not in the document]
vibhansa-msft: Logging is initialized after the validate() method so anything we log here will not be found in our log files.

common/util.go
vibhansa-msft: Logging is intialised after validate() method hence can not log here.

common/util.go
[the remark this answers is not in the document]
vibhansa-msft: Due to cyclic dependency we can not add logs to utils.

souravgupta-msft (approved):

PR #1458, opened 2024-07-13
Directory listing caching component

vibhansa-msft:
Add component to cache the directory entries. Helps in reducing REST calls when multiple threads are trying to list the same directory in parallel.


component/attr_cache/attr_cache_test.go
```diff
@@ -210,8 +207,7 @@ func (suite *attrCacheTestSuite) TestDefault() {
 	defer suite.cleanupTest()
 	suite.assert.Equal(suite.attrCache.Name(), "attr_cache")
 	suite.assert.EqualValues(suite.attrCache.cacheTimeout, 120)
-	suite.assert.Equal(suite.attrCache.cacheOnList, true)
-	suite.assert.Equal(suite.attrCache.noSymlinks, false)
+	// suite.assert.Equal(suite.attrCache.noSymlinks, false)
```
ashruti-msft, line 210: Is there a reason we have commented this?

vibhansa-msft: after this change this flag is just exported from attr-cache but is not used in attr-cache to make any decision so removed all refrences of it. Kept the flag as is because azstorage component refers this as well.

PR #1461, opened 2024-07-18
Making pipeline consistent with the units

syeleti-msft:


perf_testing/scripts/highspeed_create.py
```diff
@@ -29,7 +29,7 @@ def main(folder, num_files):
     total_data_written = num_files * 20  # in GB
     speed_gbps = (total_data_written * 8) / total_time  # converting GB to Gb and then calculating Gbps
 
-    print(json.dumps({"name": "create_10_20GB_file", "total_time": total_time, "speed": speed_gbps, "unit": "GiB/s"}))
+    print(json.dumps({"name": "create_10_20GB_file", "total_time": total_time, "speed": speed_gbps / 8, "unit": "GiB/s"}))
```
vibhansa-msft, line 32: This will invalidate the current values in the graph and we might have to cleanup all old data. How do you want to clean those up.

syeleti-msft: Thanks for the heads up, will find a way to do that

vibhansa-msft (approved):

vibhansa-msft (approved):

perf_testing/scripts/highspeed_create.py
vibhansa-msft: We might need to delete all of old data before making this change. Let's clean it up just before merging this.

PR #1462, opened 2024-07-19
Block-cache read Pipeline 

ashruti-msft:
We copied the entire block regardless of whether it was fully used, leading to copying over garbage data.

PR #1467, opened 2024-07-22
Reset block when released

souravgupta-msft:
Reset the block data to null when it is released back to the block pool. As a result of this, a block retrieved from the block pool will have null data instead of garbage data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 - Fixed the case where file creation using SAS on HNS accounts was returning back wrong error code.
 - [#1402](https://github.com/Azure/azure-storage-fuse/issues/1402) Fixed proxy URL parsing.
 - If earlier instance of Blobfuse2 crashed and mount is unstable then next mount to same path will automatically cleanup the system.
+- Reset block data to null before reuse to avoid corruption
 
 **Features**
 
--- a/component/block_cache/block_cache.go
+++ b/component/block_cache/block_cache.go
@@ -337,6 +337,7 @@ func (bc *BlockCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Han
 
 	if options.Flags&os.O_TRUNC != 0 || options.Flags&os.O_WRONLY != 0 {
 		// If file is opened in truncate or wronly mode then we need to wipe out the data consider current file size as 0
+		log.Debug("BlockCache::OpenFile : Truncate %v to 0", options.Name)
 		handle.Size = 0
 		handle.Flags.Set(handlemap.HandleFlagDirty)
 	} else if options.Flags&os.O_RDWR != 0 && handle.Size != 0 {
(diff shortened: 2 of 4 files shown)


vibhansa-msft (approved):

syeleti-msft (commented):
Suggestion:
1. I think it would be better to first merge this pr [https://github.com/Azure/azure-storage-fuse/pull/1462](https://github.com/Azure/azure-storage-fuse/pull/1462) before merging this pr, as reading the block is causing us the issue, if it is fixed we can verify all other scenarios here as well.

Queries:
1. I believe, blobfuse is not POSIX complaint, It doesn't implement lseek(atleast for block cache). If we open a file, we write it sequentially and data integrity is preserved. 
but, if we seek to some offset and write at random offsets in the file handle we doesn't preserve data Integrity. It seems to me that the current fix is regarding this scenario.
Ex: In blobfuse if my block size is 32MB, and I open a file and write some data at start say we created first block and now when we seek to 100MB offset and write at that position, I believe we corrupt the data even if we patch it with the current fix. 
2. I suppose that the current fix only works when user seeks in the allocated blocks, **I think it is better to mention that we don't support seek at all, hence user is aware of that.**
3. are there any sequential write problems that we are fixing using this pr?


Please correct me if I am wrong.

souravgupta-msft:
> I think it would be better to first merge this pr [Block-cache read block correction #1462](https://github.com/Azure/azure-storage-fuse/pull/1462) before merging this pr, as reading the block is causing us the issue, if it is fixed we can verify all other scenarios here as well.

This fix is for write and not related to read scenario. So, it is independent of the above PR.

> Queries:
>
> 1. I believe, blobfuse is not POSIX complaint, It doesn't implement lseek(atleast for block cache). If we open a file, we write it sequentially and data integrity is preserved.
>    but, if we seek to some offset and write at random offsets in the file handle we doesn't preserve data Integrity. It seems to me that the current fix is regarding this scenario.
>    Ex: In blobfuse if my block size is 32MB, and I open a file and write some data at start say we created first block and now when we seek to 100MB offset and write at that position, I believe we corrupt the data even if we patch it with the current fix.

We have decided to fix the issues with random write in different PRs, where each PR will be fixing one sub-issue. This makes the code review easy. This PR just fixes the part where say first few bytes in a block are filled and the rest data is null. So, during releasing a block we clear the data in block to prevent this. The write issue in the example you mentioned is related to sparse files which will be fixed in the future PRs.


> 2. I suppose that the current fix only works when user seeks in the allocated blocks, **I think it is better to mention that we don't support seek at all, hence user is aware of that.**

We will support seek in block cache. The fix for it will be coming in future PRs.

> 3. are there any sequential write problems that we are fixing using this pr?

Mentioned in the answer for the first query.

syeleti-msft (approved):

ashruti-msft (approved):

vibhansa-msft (approved):

PR #1470, opened 2024-07-24
Added statfs for block-cache

ashruti-msft:
Added statfs for block_cache which will show the memory available if disk is not configured otherwise the disk space. Added corresponding tests for this and fixed a few unit changes in file_cache.

Used a VM with an attached disk of size 5G for the follwoing outputs:

When file cache/block cache uses the attached disk for disk caching
Screenshot 2024-11-22 153142

When config has block_cache component without disk cache it will show the memory available:
Screenshot 2024-11-22 153029

Output when file cache/block cache uses the same VM as disk cache
Screenshot 2024-11-22 152615


component/block_cache/block_cache.go
```diff
@@ -298,6 +298,40 @@ func (bc *BlockCache) Configure(_ bool) error {
...
+	usage = usage * float64(_1MB)
+
+	available := (float64)(maxCacheSize) - usage
+	statfs := &syscall.Statfs_t{}
+	err := syscall.Statfs("/", statfs)
+	if err != nil {
+		log.Debug("BlockCache::StatFs : statfs err [%s].", err.Error())
+		return nil, false, err
+	}
+	statfs.Blocks = uint64(maxCacheSize) / uint64(statfs.Frsize)
```
vibhansa-msft, line 328: Need to correct the computation here as statfs.frsize will point to disk blocks and free size

jainakanksha-msft:
@vibhansa-msft 
This PR changes will represent temp disk details with the notation of root Filesystem.
I think that can confuse the users.
Should we change the FileSystem representation also for this somehow if that is feasible?

ashruti-msft:
> @ashruti-msft Please add the output in the PR description and details of the change.

@jainakanksha-msft I have added the output please have a look

vibhansa-msft (approved):

component/block_cache/block_cache.go
```diff
@@ -1758,6 +1758,41 @@ func (bc *BlockCache) SyncFile(options internal.SyncFileOptions) error {
 	return nil
 }
 
+func (bc *BlockCache) StatFs() (*syscall.Statfs_t, bool, error) {
+	// cache_size = f_blocks * f_frsize/1024
```
jainakanksha-msft, line 1762: Please remove the comment, which is not needed in this calculation

syeleti-msft (approved):

vibhansa-msft (approved):

jainakanksha-msft (approved):

PR #1472, opened 2024-07-26
Block cache openFile if condition bug

syeleti-msft:
Correct the Condition check that allows last block to be size greater than block size.

### DIFF
--- a/component/block_cache/block_cache.go
+++ b/component/block_cache/block_cache.go
@@ -380,6 +380,7 @@ func (bc *BlockCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Han
 
 // validateBlockList: Validates the blockList.
 // Each Block's size must equal to blockSize set in config and last block size <= config's blockSize
+// returns true, if blockList is valid
 func (bc *BlockCache) validateBlockList(fileName string, blockSize uint64, blockList *internal.CommittedBlockList) bool {
 	listLen := len(*blockList)
 	for idx, block := range *blockList {
--- a/component/block_cache/block_cache.go
+++ b/component/block_cache/block_cache.go
@@ -350,18 +350,17 @@ func (bc *BlockCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Han
 			return nil, fmt.Errorf("failed to retrieve block list for %s", options.Name)
 		}
 
+		valid := bc.validateBlockList(options.Name, bc.blockSize, blockList)
+		if !valid {
+			return nil, fmt.Errorf("block size mismatch for %s", options.Name)
+		}
+
+		// blockList is valid, Copy the blockList to the handle
 		lst, _ := handle.GetValue("blockList")
 		listMap := lst.(map[int64]string)
 
-		listLen := len(*blockList)
 		for idx, block := range *blockList {
 			listMap[int64(idx)] = block.Id
-			// All blocks shall of same size otherwise fail the open call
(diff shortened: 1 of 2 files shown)


component/block_cache/block_cache.go
```diff
@@ -358,7 +358,7 @@ func (bc *BlockCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Han
 			listMap[int64(idx)] = block.Id
 			// All blocks shall of same size otherwise fail the open call
 			// Last block is allowed to be of smaller size as it can be partial block
-			if block.Size != bc.blockSize && idx != (listLen-1) {
+			if (idx < (listLen-1) && block.Size != bc.blockSize) || (idx == (listLen-1) && block.Size > bc.blockSize) {
```
vibhansa-msft, line 361: Is it possible to add some test case around this

syeleti-msft: Added Unit tests for the code, please review it again

vibhansa-msft (approved):

component/block_cache/block_cache.go
```diff
@@ -350,18 +350,17 @@ func (bc *BlockCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Han
 			return nil, fmt.Errorf("failed to retrieve block list for %s", options.Name)
 		}
 
+		valid := bc.validateBlockList(options.Name, bc.blockSize, blockList)
+		if !valid {
+			return nil, fmt.Errorf("block size mismatch for %s", options.Name)
+		}
+
+		// blockList is valid, Copy the blockList to the handle
 		lst, _ := handle.GetValue("blockList")
```
vibhansa-msft, line 359: from here till the end of the loop can be moved to another function instead of iterating the blocks twice.

component/block_cache/block_cache.go
```diff
@@ -350,18 +350,17 @@ func (bc *BlockCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Han
 			return nil, fmt.Errorf("failed to retrieve block list for %s", options.Name)
 		}
 
+		valid := bc.validateBlockList(options.Name, bc.blockSize, blockList)
```
vibhansa-msft, line 353: blockSize is a class member only no need to pass it down to another method.
Ideally we shall just send the handle and the blockList and blockList map updation shall be done in the new method.

component/block_cache/block_cache.go
```diff
@@ -379,6 +378,20 @@ func (bc *BlockCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Han
...
 }
 
+// validateBlockList: Validates the blockList.
+// Each Block's size must equal to blockSize set in config and last block size <= config's blockSize
+// returns true, if blockList is valid
+func (bc *BlockCache) validateBlockList(fileName string, blockSize uint64, blockList *internal.CommittedBlockList) bool {
+	listLen := len(*blockList)
+	for idx, block := range *blockList {
+		if (idx < (listLen-1) && block.Size != blockSize) || (idx == (listLen-1) && block.Size > blockSize) {
+			log.Err("BlockCache::OpenFile : Block size mismatch for %s [block: %v, size: %v]", fileName, block.Id, block.Size)
```
vibhansa-msft, line 388: In log we print the class name and function name. here function name shall be used instead of "OpenFile"

souravgupta-msft (approved):

PR #1473, opened 2024-07-26
Blocker and Vulnerability update

vibhansa-msft:


cmd/root.go
```diff
@@ -151,21 +151,32 @@ func beginDetectNewVersion() chan interface{} {
 			return
 		}
 
+		blockedVersions := common.Blobfuse2ListContainerURL + "/blockedversions/" + common.Blobfuse2Version
+		isBlocked := checkVersionExists(blockedVersions)
```
souravgupta-msft, line 155: Can we add some tests for this

cmd/root.go
```diff
@@ -151,21 +151,32 @@ func beginDetectNewVersion() chan interface{} {
...
+
+		warningsUrl := common.Blobfuse2ListContainerURL + "/securitywarnings/" + common.Blobfuse2Version
+		hasWarnings := checkVersionExists(warningsUrl)
+
+		if hasWarnings {
+			warningsPage := common.BlobFuse2WarningsURL + "#" + strings.ReplaceAll(strings.ReplaceAll(common.Blobfuse2Version, ".", ""), "~", "")
+			fmt.Fprintf(stderr, "Visit %s to see the list of known issues associated with your current version [%s]\n", warningsPage, common.Blobfuse2Version)
+			log.Warn("Visit %s to see the list of known issues associated with your current version [%s]\n", warningsPage, common.Blobfuse2Version)
+		}
+
```
ashruti-msft, line 172: It's better to check if the version has a security warning first then check if the version is blocked.

vibhansa-msft: updated. Kindly review

ashruti-msft (approved):

souravgupta-msft (approved):

PR #1475, opened 2024-07-26
Block cache random write in sparse files

souravgupta-msft:
- For blocks where kernel doesn't send write call, stage a block having null data and use its block id for these blocks
- When first few bytes of a block (not last block) are written, and it is sent for staging, then append null at the end so that the size of the block being staged is same as the block size defined in config.

### DIFF
--- a/component/block_cache/block_cache.go
+++ b/component/block_cache/block_cache.go
@@ -943,7 +943,10 @@ func (bc *BlockCache) WriteFile(options internal.WriteFileOptions) (int, error) 
 
 		if block.endIndex < uint64(options.Offset) {
 			block.endIndex = uint64(options.Offset)
-			options.Handle.Size = int64(block.endIndex)
+		}
+
+		if options.Handle.Size < options.Offset {
+			options.Handle.Size = options.Offset
 		}
 	}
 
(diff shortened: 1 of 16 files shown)


component/block_cache/block_cache.go
```diff
@@ -1119,12 +1136,22 @@ func (bc *BlockCache) upload(item *workItem) {
 	flock.Lock()
 	defer flock.Unlock()
 
+	endIndex := item.block.endIndex
+	// if a block has data less than block size and is not the last block,
+	// add null at the end and upload the full block
+	if item.block.endIndex-item.block.offset < bc.blockSize && item.block.endIndex < uint64(item.handle.Size) {
```
vibhansa-msft, line 1142: (item.block.endIndex-item.block.offset < bc.blockSize) : this will never be true

component/block_cache/block_cache.go
```diff
@@ -1119,12 +1136,22 @@ func (bc *BlockCache) upload(item *workItem) {
...
+		endIndex = item.block.offset + bc.blockSize
+		log.Debug("BlockCache::upload : Appending null for block %v", item.block.id)
```
vibhansa-msft, line 1144: This code can be moved to lineupupload

component/block_cache/block_cache.go
```diff
@@ -1119,12 +1136,22 @@ func (bc *BlockCache) upload(item *workItem) {
 	flock.Lock()
 	defer flock.Unlock()
 
+	endIndex := item.block.endIndex
+	// if a block has data less than block size and is not the last block,
+	// add null at the end and upload the full block
+	if item.block.endIndex-item.block.offset < bc.blockSize && item.block.endIndex < uint64(item.handle.Size) {
+		endIndex = item.block.offset + bc.blockSize
```
vibhansa-msft, line 1143: We shall update the actual block.endindex as well as now the block is full for us.

component/block_cache/block_cache.go
souravgupta-msft: Updated

component/block_cache/block_cache.go
souravgupta-msft: Done

component/block_cache/block_cache.go
souravgupta-msft: Done

vibhansa-msft (approved):

ashruti-msft (approved):

syeleti-msft (approved):

PR #1476, opened 2024-07-26
Add support for no_proxy

vibhansa-msft:


component/azstorage/utils.go
```diff
@@ -162,7 +162,7 @@ func setSDKLogListener() {
 func newBlobfuse2HttpClient(conf *AzStorageConfig) (*http.Client, error) {
 	var ProxyURL func(req *http.Request) (*url.URL, error)
 	if conf.proxyAddress == "" {
-		ProxyURL = nil
+		ProxyURL = http.ProxyFromEnvironment
```
souravgupta-msft, line 165: Please confirm if the existing proxy test setup works with environment variable or config parameters. If it is configured for config parameters, can we also add tests for environment variables

PR #1481, opened 2024-07-29
Race condition in random write between uploading and writing to a same block

souravgupta-msft:
In random write scenarios where a block is being uploaded (or staged via PutBlock) and being written to in parallel, wait for the block to be uploaded first and then write to the same block.


component/block_cache/block_cache.go
```diff
@@ -1006,6 +1006,19 @@ func (bc *BlockCache) getOrCreateBlock(handle *handlemap.Handle, offset uint64)
 			<-block.state
 			block.flags.Clear(BlockFlagDownloading)
 			block.Unblock()
+		} else if block.flags.IsSet(BlockFlagUploading) {
+			// If the block is being staged, then wait till it is uploaded,
+			// and then write to the same block and move it back to cooking queue
+			log.Debug("BlockCache::getOrCreateBlock : Waiting for the block %v to upload for %v=>%s", block.id, handle.ID, handle.Path)
+			<-block.state
+			block.flags.Clear(BlockFlagUploading)
```
vibhansa-msft, line 1014: Do we need to set the dirty flag here for the block, or it will be taken care by the caller.

souravgupta-msft: When we return the block to the WriteFile() method which requested for the block to write to it, it marks the block as dirty.

ashruti-msft (approved):

vibhansa-msft (approved):

syeleti-msft (approved):

vibhansa-msft (approved):

PR #1484, opened 2024-07-30
Block cache disable random write

souravgupta-msft:
Random write operations in block cache will be disabled by default. It can be enabled by passing the CLI flag `--block-cache-enable-random-write` in the mount command.

This PR also checks and returns error in the following cases:

1. Race condition between uploading and writing to a same block.
2. Random write scenario where a block has been staged and cleared from the in-memory cache and write call comes for that block.
3. A staged block (not last block) has data less than the block size.

The fixes for the above will be done in future PRs.

### DIFF
--- a/component/block_cache/block_cache.go
+++ b/component/block_cache/block_cache.go
@@ -937,7 +937,8 @@ func (bc *BlockCache) WriteFile(options internal.WriteFileOptions) (int, error) 
 
 	// check for possible random write scenario and block the write request
 	blockIndex := bc.getBlockIndex(uint64(options.Offset))
-	if int64(blockIndex*bc.blockSize) < options.Handle.Size-int64(MIN_WRITE_BLOCK*bc.blockSize) {
+	lastBlockIndex := bc.getBlockIndex(uint64(options.Offset))
+	if lastBlockIndex-blockIndex > MIN_WRITE_BLOCK {
 		log.Debug("BlockCache::WriteFile : Random write detection for write offset %v and block %v, where handle size is %v", options.Offset, blockIndex, options.Handle.Size)
 		if !bc.enableRandomWrite {
 			log.Err("BlockCache::WriteFile : Blocking random write")
--- a/component/block_cache/block_cache.go
+++ b/component/block_cache/block_cache.go
@@ -938,7 +938,7 @@ func (bc *BlockCache) WriteFile(options internal.WriteFileOptions) (int, error) 
 	// check for possible random write scenario and block the write request
 	blockIndex := bc.getBlockIndex(uint64(options.Offset))
 	lastBlockIndex := bc.getBlockIndex(uint64(options.Offset))
-	if lastBlockIndex-blockIndex > MIN_WRITE_BLOCK {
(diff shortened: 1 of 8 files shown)


vibhansa-msft (approved):

component/block_cache/block_cache.go
```diff
@@ -932,6 +935,16 @@ func (bc *BlockCache) WriteFile(options internal.WriteFileOptions) (int, error)
 
 	// log.Debug("BlockCache::WriteFile : Writing handle %v=>%v: offset %v, %v bytes", options.Handle.ID, options.Handle.Path, options.Offset, len(options.Data))
 
+	// check for possible random write scenario and block the write request
+	blockIndex := bc.getBlockIndex(uint64(options.Offset))
+	if int64(blockIndex*bc.blockSize) < options.Handle.Size-int64(MIN_WRITE_BLOCK*bc.blockSize) {
```
syeleti-msft, line 940: Lets say we have n blocks, this condition will be true for first n-3 blocks, does that mean we don't allow write on those blocks even if they are sequential?

component/block_cache/block_cache.go
```diff
@@ -68,36 +68,38 @@ import (
...
-	PrefetchOnOpen bool    `config:"prefetch-on-open" yaml:"prefetch-on-open,omitempty"`
+	BlockSize         float64 `config:"block-size-mb" yaml:"block-size-mb,omitempty"`
+	MemSize           uint64  `config:"mem-size-mb" yaml:"mem-size-mb,omitempty"`
+	TmpPath           string  `config:"path" yaml:"path,omitempty"`
+	DiskSize          uint64  `config:"disk-size-mb" yaml:"disk-size-mb,omitempty"`
+	DiskTimeout       uint32  `config:"disk-timeout-sec" yaml:"timeout-sec,omitempty"`
+	PrefetchCount     uint32  `config:"prefetch" yaml:"prefetch,omitempty"`
+	Workers           uint32  `config:"parallelism" yaml:"parallelism,omitempty"`
+	PrefetchOnOpen    bool    `config:"prefetch-on-open" yaml:"prefetch-on-open,omitempty"`
+	EnableRandomWrite bool    `config:"enable-random-write" yaml:"enable-random-write,omitempty"`
```
syeleti-msft, line 102: What does actually --block-cache-enable-random-write mean for user?

if I make the above flag as false, what happens to the following conditions allowed/not allowed?
1. user can write sequentially from 0th offset to the end of the file, without allowing seeking in the file.
2. user can start seek to some offset and start writing sequentially from there.

component/block_cache/block_cache.go
souravgupta-msft: Updated the logic

component/block_cache/block_cache.go
souravgupta-msft: If the flag is false, the random write will be blocked in the following scenarios
1. The block where the write call comes is not in the buffer list (both cooking and cooked)
2. Write call comes for the block which is not the last block
3. The block is present in the blockIdList map indicating it was staged earlier and deleted from the buffer list

For your case,
1. User can seek back if the block has not been deleted from the buffer list
2. This is allowed as per the above conditions

Issue #1426, opened 2024-06-06
Problems with multithreaded reading/writing via block_cache


olegshv:
### Which version of blobfuse was used?
blobfuse2 version 2.3.0

### Which OS distribution and version are you using?
Debian GNU/Linux 11 (bullseye)

### If relevant, please share your mount command.
blobfuse2 mount /opt/blobfuse/fuse_container/ --config-file=/opt/blobfuse/fuse_connection.yaml

### My fuse_connection.yaml file
![image](https://github.com/Azure/azure-storage-fuse/assets/9905160/3f43c198-b8bb-4435-9111-885c0ccc5a5d)

### What was the issue encountered?
A brief description of what we do with blogfusion2:
- We use it to make DB backups directly to the Storage Account and restore from it. Usually, we run backups of 10-15 threads in parallel (we would like to have more in the future, but we need to understand how to scale). Database sizes range from 200mb to 4TB.
- Now we use Azure instance with size Standard L8as v3 (8 vcpus, 64 GiB memory). 
- Previously, we used file_cache, but now we have a critical problem with the disk space for the cache, as the total amount of backups running in one thread can take up to 12-15TB, and it becomes very unreliable.

We are currently testing the transition on stream or block_cache. After a few days of testing, the stream was abandoned altogether, as we never managed to successfully back up. block_cache gives us hope in our hearts :)

For example, here is a 3.6TB database backup running in 5 threads in parallel. It's ok. 
![image](https://github.com/Azure/azure-storage-fuse/assets/9905160/babaf33c-51b2-4799-948e-2342809bc1ca)
![image](https://github.com/Azure/azure-storage-fuse/assets/9905160/cf9808da-2378-4320-9329-186744657ddf)

But here's an example of 15-thread jobs, where there are small databases and one large one, the load is not high, and it failed.
![image](https://github.com/Azure/azure-storage-fuse/assets/9905160/2b8d7df4-72fa-4249-918d-e47688c121d0)
There is no load, but we get an error.

1. From Windows
![image](https://github.com/Azure/azure-storage-fuse/assets/9905160/ef4e0230-7997-4fec-a188-8e66a1598e3b)
2. An error from SQL Server Management Studio:  1117(The request could not be performed because of an I/O device error.)

Unfortunately, after reading a lot of information and documentation, I did not find an answer on scaling and configurations for large loads. We ran tests even with large VMs (32cpu, 256ram), but it didn't work.
[...]

PR #1491, opened 2024-08-05
panic on calling close on close channel

syeleti-msft:
Issue: https://github.com/Azure/azure-storage-fuse/issues/1426

Closing the channel twice causing the panic.
This can happen in 3 scenarios:
1. Closing the file.
2. when requesting a block for writing.
3. Evicting the blocks when the file is in rand-read mode.

Both call the same method waitAndFreeUploadedBlocks, which unblocks the requested no of blocks inside the cooked list. but cooked list might contain the blocks which were unblocked already (i.e., when blocks are downloaded they were moved to cooked list when there is a reader for them and unblocked), hence calling unblock on them again causes panic.

### DIFF
--- a/component/block_cache/block_cache.go
+++ b/component/block_cache/block_cache.go
@@ -670,15 +670,18 @@ func (bc *BlockCache) startPrefetch(handle *handlemap.Handle, index uint64, pref
 				block := handle.Buffers.Cooking.Remove(node).(*Block)
 				block.node = nil
 				i++
-
+				//This list may contain dirty blocks which are yet to be committed.
 				select {
-				case <-block.state:
+				case _, ok := <-block.state:
 					// As we are first reader of this block here its important to unblock any future readers on this block
-					block.flags.Clear(BlockFlagDownloading)
-					block.Unblock()
-
-					// Block is downloaded so it's safe to ready it for reuse
-					block.node = handle.Buffers.Cooked.PushBack(block)
+					if ok {
+						block.flags.Clear(BlockFlagDownloading)
+						block.Unblock()
+						// Block is downloaded so it's safe to ready it for reuse
+						block.node = handle.Buffers.Cooked.PushBack(block)
+					} else {
+						block.node = handle.Buffers.Cooking.PushBack(block)
+					}
 
 				default:
 					// Block is still under download so can not reuse this
@@ -1218,6 +1221,8 @@ func (bc *BlockCache) waitAndFreeUploadedBlocks(handle *handlemap.Handle, cnt in
 				block.Unblock()
 			}
 			block.flags.Clear(BlockFlagUploading)
+		} else {
+			block.Unblock()
 		}
 
 		if block.IsFailed() {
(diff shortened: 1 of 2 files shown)


component/block_cache/block_cache.go
```diff
@@ -1213,12 +1213,13 @@ func (bc *BlockCache) waitAndFreeUploadedBlocks(handle *handlemap.Handle, cnt in
...
 			// Wait for upload of this block to complete
-			<-block.state
+			_, ok := <-block.state
+			if ok {
+				block.Unblock()
+			}
 			block.flags.Clear(BlockFlagUploading)
 		}
 
-		block.Unblock()
```
vibhansa-msft, line 1220: block-id can be -1 and for those we need to close the channel as well but if id is  != -1 then we shall not wait to receive anything. Hence the code was written like this. block-id == -1 is a valid case where due to prefetch we allocated more blocks then required and we never used them to fill data as there offsets were beyond the file length. So may be better fix is to release those blocks when you try to do lineupDownload() and function returns EOF. That will fix this problem too.

component/block_cache/block_cache.go
```diff
@@ -1098,17 +1101,22 @@ func (bc *BlockCache) getOrCreateBlock(handle *handlemap.Handle, offset uint64)
 			block.flags.Clear(BlockFlagSynced)
 		} else if block.flags.IsSet(BlockFlagDownloading) {
 			log.Debug("BlockCache::getOrCreateBlock : Waiting for download to finish for committed block %v for %v=>%s", block.id, handle.ID, handle.Path)
-			<-block.state
+			_, ok := <-block.state
+			if ok {
+				block.Unblock()
+			}
 			block.flags.Clear(BlockFlagDownloading)
```
vibhansa-msft, line 1108: This line we can move up, generally we are doing unblock after all flags are reset.

component/block_cache/block_cache.go
```diff
@@ -1098,17 +1101,22 @@ func (bc *BlockCache) getOrCreateBlock(handle *handlemap.Handle, offset uint64)
...
 		} else if block.flags.IsSet(BlockFlagUploading) {
 			// If the block is being staged, then wait till it is uploaded,
 			// and then write to the same block and move it back to cooking queue
 			log.Debug("BlockCache::getOrCreateBlock : Waiting for the block %v to upload for %v=>%s", block.id, handle.ID, handle.Path)
-			<-block.state
+			_, ok := <-block.state
+			if ok {
+				block.Unblock()
+			}
 			block.flags.Clear(BlockFlagUploading)
```
vibhansa-msft, line 1118: same as above: clear flags before the unblock. if ok{ check can be done after the reset

component/block_cache/block_cache.go
```diff
@@ -1234,10 +1242,10 @@ func (bc *BlockCache) waitAndFreeUploadedBlocks(handle *handlemap.Handle, cnt in
 		if block.id != -1 {
 			// Wait for upload of this block to complete
 			_, ok := <-block.state
-			block.flags.Clear(BlockFlagUploading)
```
vibhansa-msft, line 1237: This one is the correct way to do it.

vibhansa-msft (approved):

souravgupta-msft (approved):

PR #1504, opened 2024-08-19
Added kernel tar tests

ashruti-msft:


testdata/config/azure_block_perf.yaml
```diff
@@ -1,5 +1,5 @@
 logging:
-  level: log_err
+  level: log_debug
```
vibhansa-msft, line 2: Why do we need log_debug here. Perf tests shall always run with log_err only.

blobfuse2-data-validation.yaml
```diff
@@ -63,12 +70,120 @@ stages:
...
             parameters:
-              working_dir: $(WORK_DIR)
               mount_dir: $(MOUNT_DIR)
               temp_dir: $(TEMP_DIR)
-              prefix: 'ubn-22'
               kversion: "6.10.2"
+              idstring: 'Block-Cache'
+              config_file: $(BLOBFUSE2_CFG)
+
+  - stage: VScodeBuild
```
vibhansa-msft, line 79: Is this working locally now?

vibhansa-msft (approved):

PR #1507, opened 2024-08-20
File Cache failed to upload if Read perm is not given 

ashruti-msft:
Fixed File Cache failed to upload if Read perm is not given by changing file perm to read and resetting it after upload.

### DIFF
--- a/component/file_cache/file_cache.go
+++ b/component/file_cache/file_cache.go
@@ -1234,7 +1234,8 @@ func (fc *FileCache) FlushFile(options internal.FlushFileOptions) error {
 			if os.IsPermission(err) {
 				info, _ := os.Stat(localPath)
 				orgMode = info.Mode()
-				err = os.Chmod(localPath, os.FileMode(0666))
+				newMode := orgMode | 0444
+				err = os.Chmod(localPath, newMode)
 				if err == nil {
 					modeChanged = true
 					uploadHandle, err = os.Open(localPath)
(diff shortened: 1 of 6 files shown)


component/file_cache/file_cache.go
```diff
@@ -1226,12 +1226,27 @@ func (fc *FileCache) FlushFile(options internal.FlushFileOptions) error {
...
 		// The local handle can still be used for read and write.
+		var orgMode fs.FileMode
+		modeChanged := false
+
 		uploadHandle, err := os.Open(localPath)
 		if err != nil {
-			log.Err("FileCache::FlushFile : error [unable to open upload handle] %s [%s]", options.Handle.Path, err.Error())
-			return nil
-		}
+			if os.IsPermission(err) {
```
vibhansa-msft, line 1234: Add a changelog entry for this change.

vibhansa-msft (approved):

component/file_cache/file_cache.go
```diff
@@ -1226,12 +1226,27 @@ func (fc *FileCache) FlushFile(options internal.FlushFileOptions) error {
...
+				info, _ := os.Stat(localPath)
+				orgMode = info.Mode()
+				err = os.Chmod(localPath, os.FileMode(0666))
```
syeleti-msft, line 1237: Changing the file mode to 0666 will make the in progress read and write requests to success without respecting their original filemode, we are not maintaining any lock in flush, read, write.

vibhansa-msft: We do not want to lock because earlier also parallel operations on file was allowd so adding a new lock here may break some existing scenarios. Adding mode to read the file is required otherwise SDK will not be able to read the file and upload the data. There is a small window where permissions are granted even when user did not give it. Is there any other better solution here ?

vibhansa-msft: Instead of setting it to 0666 blindly, we shall just add read bits to existing mode here.

vibhansa-msft (approved):

PR #1511, opened 2024-08-30
Update block cache limitations in README

souravgupta-msft:


vibhansa-msft (approved):

syeleti-msft (approved):

ashruti-msft (approved):

README.md
```diff
@@ -8,9 +8,16 @@ Blobfuse2 is stable, and is ***supported by Microsoft*** provided that it is use
 [This](https://github.com/Azure/azure-storage-fuse/tree/main?tab=readme-ov-file#config-guide) section will help you choose the correct config for Blobfuse2.
 
 ##  NOTICE
-- If you are using versions 2.2.0, 2.2.1 and 2.3.0, refrain from using Block-cache mode and switch to `file-cache` mode till [known issues](https://github.com/Azure/azure-storage-fuse/wiki/Blobfuse2-Known-issues) are fixed.
+- If you are using versions 2.2.0, 2.2.1 and 2.3.0, refrain from using Block-cache mode and switch to `file-cache` mode. [Known issues](https://github.com/Azure/azure-storage-fuse/wiki/Blobfuse2-Known-issues) in these versions are fixed in version **`2.3.2`**.
```
tjvishnu, line 11: We should ask customers to upgrade as the primary action. Reworded statement below

To avoid encountering [known issues](https://github.com/Azure/azure-storage-fuse/wiki/Blobfuse2-Known-issues) when using the 'Block-cache' mode, users of Blobfuse versions 2.2.0, 2.2.1, and 2.3.0 are advised to promptly upgrade to version 2.3.2.

README.md
```diff
@@ -8,9 +8,16 @@ Blobfuse2 is stable, and is ***supported by Microsoft*** provided that it is use
 [This](https://github.com/Azure/azure-storage-fuse/tree/main?tab=readme-ov-file#config-guide) section will help you choose the correct config for Blobfuse2.
 
 ##  NOTICE
-- If you are using versions 2.2.0, 2.2.1 and 2.3.0, refrain from using Block-cache mode and switch to `file-cache` mode till [known issues](https://github.com/Azure/azure-storage-fuse/wiki/Blobfuse2-Known-issues) are fixed.
+- If you are using versions 2.2.0, 2.2.1 and 2.3.0, refrain from using Block-cache mode and switch to `file-cache` mode. [Known issues](https://github.com/Azure/azure-storage-fuse/wiki/Blobfuse2-Known-issues) in these versions are fixed in version **`2.3.2`**.
 - As of version 2.3.0, blobfuse has updated its authentication methods. For Managed Identity, Object-ID based OAuth is solely accessible via CLI-based login, requiring Azure CLI on the system. For a dependency-free option, users may utilize Application/Client-ID or Resource ID based authentication.
 - `streaming` mode is being deprecated.
+
+## Limitations in Block Cache
```
tjvishnu, line 15: We will have to rewrite this to make sure we communicate to customers in better detail. My Comment for each point below

1. Concurrent Write operations on the same file using multiple handles is not checked for data consistency and may lead to incorrect data being written.
2. 

> [Comment] I am not sure if I have understood this problem clearly. For example, if we have a file with 1000 blocks and the customer is reading this using block-cache mode sequentially from 0 to 999. During this process, if the file is updated from another client, and if this client is at Block 50 when this happens, will the next set of blocks to be returned be the updated blocks or old blocks? Also, this point needs to go to the bottom. 

3.

>  [Comment] I believe this cp thing applicable to only Block-cache mode? Also, if they do not use the sparse never, what would happen? And what if they wanted to create a sparse file? (I'm not sure that's a real use case, but asking for clarification)

4. 

> [Comment] Unclear what this means. What does "persistence" mean here and what is the customer supposed to do when reading a file using Blobfuse. As an ML user who use Blobfuse as a mount to get my training dataset and uses PyTorch, I do not use any of these operations. How do I interpret these?

5. 

>[Comment] Again, unclear what the customer is expected to check. What is "code" here? And how are they supposed to "check"?

souravgupta-msft: Please find the reply below.

2. This is for the case when there are two handles for a same file where one is writing and the other is reading parallelly. Let's say the first handle updated the block 1. Now when the second handle tries to read block 1, it won't get the data written by the first handle.

3. In block cache when sparse files are copied using cp, then the sparse=never flag should be used. If they don't use it, then the data may not be uploaded completely. This is for the case when there are trailing null data in sparse file and kernel does not send call to Blobfuse for this data. This flag ensures that kernel will send call whether the data is null or not.

4. This case is related to write operation only. Persistence here means that the staged blocks are committed to the blob via PutBlockList only in close, sync or flush calls.

5. Code here means the error code returned by the filesystem, if it is success or failure.

I have created a new PR #1513 addressing these changes. Let's continue our discussion there

PR #1513, opened 2024-09-02
README update

souravgupta-msft:


README.md
```diff
@@ -8,16 +8,16 @@ Blobfuse2 is stable, and is ***supported by Microsoft*** provided that it is use
...
 - `streaming` mode is being deprecated.
 
 ## Limitations in Block Cache
-- Parallel write operations using multiple handles on a same file is not supported and might lead to data inconsistency.
-- Read operation on a file which is being written via another handle will not return updated data.
-- When using `cp` utility on mounted path, always use `--sparse=never` parameter. For example, `cp --sparse=never src dest`
-- In write operations data is persisted in storage only on close, sync or flush calls.
-- User applications must check the returned code for write, close and flush operations.
+- Concurrent write operations on the same file using multiple handles is not checked for data consistency and may lead to incorrect data being written.
+- Read operation on a file which is being written parallely via another handle will not return updated data.
```
tjvishnu: 2. A read operation on a file that is being written to simultaneously by another process or handle will not return the most up-to-date data.

3. 

> [Comment] It looks like only null bytes in the end will be an issue and not those in between. How does a customer even specify null bytes in the end?

4. 

> [Comment] What this means is the data being streamed in `Block-cache` mode, is not being sent to the Storage account while write is happening. Does this mean that if a customer writes a file bigger than available memory defined in config or is bigger than the RAM, file write will fail?

5. 

> [Comment] I'm not able to figure out where and when to use this. Also, assuming the customer does not get to see these file system call failures, what would happen? It is not clear if they should write a separate app to do this or is it just a disclaimer for a specific scenario.

vibhansa-msft: # Sample data to write to the file
data = "Hello, this is a sample text."

# Create and open a file in write mode
with open("sample_file.txt", "w") as file:
    # Write data to the file
    bytes_written = file.write(data)
    
    # Check if the write was successful
    if bytes_written == len(data):
        print("Write successful. All data written.")
    else:
        print("Write failed. Not all data written.")

tjvishnu: Thanks. For #3, here's the statement: When copying files with trailing null bytes using using `cp` utility to a Blobfuse2 mounted path, use `--sparse=never` parameter to avoid data being trimmed. For example, `cp --sparse=never src dest`.

For #4, If i understand your concern, you want to say that the file will not be saved if the customer fails to do a flush/close/sync call, but the blocks are already in storage account in an uncommitted state. Is that correct? I just wanted to make sure we explain to the users the impact of these calls for the file's persistence in Storage account vs how we stream the data to storage account using block-cache mode

For #5,  I believe you want the customers to check every time, but it is unclear why you ask for it and when there could be an issue. If its not a limitation, rather a recommendation, we should call it accordingly and not include it as part of "Limitations of Block-cache mode"

vibhansa-msft: for #4, customer is not aware of what is staging and when we stage a block. For that reason, I just wanted to write in simple terms saying data is not saved/persisted. For them data not being updated is what matters and not the staging part.
For #5, you are right. It's more of a recommendation and not a mandate. Most application developers choose not to rely on these error codes but this means some of the errors may go unnoticed. It's always good practice to check the return codes of any API you call and validate that it succeeded.

souravgupta-msft: @tjvishnu, I have updated the points as suggested and added a new point(5th) in the limitations section. Also added a recommendation section. Please review.

PR #1515, opened 2024-09-06
Entry cache component

vibhansa-msft:
Cache the directory listing


component/attr_cache/attr_cache_test.go
```diff
@@ -210,8 +207,7 @@ func (suite *attrCacheTestSuite) TestDefault() {
 	defer suite.cleanupTest()
 	suite.assert.Equal(suite.attrCache.Name(), "attr_cache")
 	suite.assert.EqualValues(suite.attrCache.cacheTimeout, 120)
-	suite.assert.Equal(suite.attrCache.cacheOnList, true)
-	suite.assert.Equal(suite.attrCache.noSymlinks, false)
+	// suite.assert.Equal(suite.attrCache.noSymlinks, false)
```
jainakanksha-msft, line 210: I think this is not needed. Please remove it.

component/attr_cache/attr_cache.go
```diff
@@ -139,22 +138,18 @@ func (ac *AttrCache) Configure(_ bool) error {
 		ac.cacheTimeout = defaultAttrCacheTimeout
 	}
 
-	if config.IsSet(compName + ".cache-on-list") {
-		ac.cacheOnList = conf.CacheOnList
-	} else {
```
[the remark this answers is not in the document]
vibhansa-msft: So it's more like only if user has given the value we want to honor that otherwise use the default one. When we use structure we have no way to know whether user set the value or its coming from default one so we always check whether the flag was set explicitly by user or not.

component/attr_cache/attr_cache_test.go
vibhansa-msft: this is to remove an older feature which is not implemented correctly. We have symlink related flags being handled at both attr-cache and azstorage and for some reason we had different defualt behaviour like azstorage assume symlink is on by default while attr-cache assume its off. So as part of this PR I am removing that duplicate flag from attr-cache.

component/entry_cache/entry_cache.go
```diff
@@ -0,0 +1,202 @@
...
+	}
+
+	return nil
+}
+
+// Configure : Pipeline will call this method after constructor so that you can read config and initialize yourself
+//
+//	Return failure if any config is not valid to exit the process
+func (c *EntryCache) Configure(_ bool) error {
+	log.Trace("EntryCache::Configure : %s", c.Name())
```
vibhansa-msft, line 119: Fail if Read-only mode is not used to mount.

cmd/mount.go
```diff
@@ -94,6 +94,7 @@ type mountOptions struct {
 	AttrCache      bool     `config:"use-attr-cache"`
 	LibfuseOptions []string `config:"libfuse-options"`
 	BlockCache     bool     `config:"block-cache"`
+	EntryCache     int      `config:"list-cache-timeout"`
```
souravgupta-msft, line 97: Rename this to something more meaningful, `EntryCacheTimeout`

cmd/mount.go
```diff
@@ -313,6 +314,10 @@ var mountCmd = &cobra.Command{
 			options.Components = pipeline
 		}
 
+		if config.IsSet("entry_cache.timeout-sec") || options.EntryCache > 0 {
+			options.Components = append(options.Components[:1], append([]string{"entry_cache"}, options.Components[1:]...)...)
+		}
```
souravgupta-msft, line 319: Why are we doing this explicitly here? config.Unmarshal should do this

souravgupta-msft: Is this for the case when entry cache timeout is passed as CLI parameter?

souravgupta-msft (approved):

cmd/mount.go
vibhansa-msft: Corrected.

cmd/mount.go
vibhansa-msft: If user pass the entry-cache timeout on cli but did not mention the component in the pipeline I am injecting the compoent explicitly in the pipeline to reduce the user configuration. Ideally we should have done this for other components as well.

PR #1517, opened 2024-09-06
Vibhansa/truncate update

vibhansa-msft:


component/azstorage/block_blob.go
```diff
@@ -1068,119 +1044,246 @@ func (bb *BlockBlob) TruncateFile(name string, size int64) error {
...
+			log.Err("BlockBlob::updateBlocks : Failed to stage block for %s [%s]", name, err.Error())
+			return err
 		}
+
+		blkList = append(blkList, id)
+		newSize = 0
 	}
-	return data, err
+
+	return bb.createSparseFile(name, newSize, blockSize, blkList, idLen)
```
syeleti-msft, line 1233: If file size is not aligning with the block size and truncate is used to expand the size, this will lead to irregular blocksizes in the block list.
Eg: lets say blocksize = 4M
file size = 5M (i.e., 2blocks) block1 = 4M, block2 = 1M

Call for truncate(10M)

I suppose the logic in this PR will make the block list look like this after truncate operation: 
block1 = 4M, block2 = 1M, block3 = 4M, block4 = 1M. 
Although this looks correct but the subsequent operations in block cache cannot utilize this sort of blocklist.
Please correct me if I am wrong

component/azstorage/block_blob.go
```diff
@@ -1068,119 +1044,246 @@ func (bb *BlockBlob) TruncateFile(name string, size int64) error {
...
+
+	err := bb.ReadInBuffer(name, 0, size, data)
+	if err != nil {
+		log.Err("BlockBlob::readFileData : Failed to read small file %s", name, err.Error())
+	}
+
+	return data, err
+}
+
+func (bb *BlockBlob) createSparseFile(name string, size int64, blockSize int64, blkList []string, idLen int64) error {
```
syeleti-msft, line 1125: It is better to have 2 functions originating from truncate function which are 
1. Expand
2. Shrink
This will make the code more readable

component/azstorage/block_blob_trucate_test.go
```diff
@@ -0,0 +1,212 @@
...
+
+		h, err := os.Open(filename)
+		suite.assert.Nil(err)
+		suite.assert.NotNil(h)
+
+		suite.az.CopyFromFile(internal.CopyFromFileOptions{Name: filename, File: h, Metadata: nil})
+		_ = h.Close()
+		os.Remove(filename)
+
+		err = suite.az.TruncateFile(internal.TruncateFileOptions{Name: filename, Size: fs.TruncatedSize, BlockSize: 16 * MB})
```
jainakanksha-msft, line 197: Should we also verify the block size for 1,2 use case after truncate.

test/scripts/flush_test.c
```diff
@@ -0,0 +1,55 @@
+
```
jainakanksha-msft, line 1: Is this file needed?

PR #1521, opened 2024-09-13
Upgrading go version to 1.23.1

vibhansa-msft:
Upgrade to latest go version


go_installer.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 work_dir=$(echo $1 | sed 's:/*$::')
-version="1.22.4"
+version="1.23.1"
```
souravgupta-msft, line 3: You can also update the dependencies in go.mod file.

ashruti-msft (approved):

go_installer.sh
vibhansa-msft: Done

souravgupta-msft (approved):

syeleti-msft (approved):

PR #1524, opened 2024-09-23
Delete empty directories from local cache

souravgupta-msft:
Renaming a directory does not delete it in the local cache. This causes rmdir operation failure of parent directory, as local cache is not empty. Steps to replicate this,
```
mkdir test && cd test

mkdir -p one/two \
&& touch one/two/a.txt \
&& mv one/two one/three \
&& touch one/three/a.txt

rm -rf *
```


internal/base_component.go
```diff
@@ -111,6 +112,14 @@ func (base *BaseComponent) IsDirEmpty(options IsDirEmptyOptions) bool {
 	return false
 }
 
+func (base *BaseComponent) DeleteEmptyDirs(options DeleteDirOptions) error {
+	if base.next != nil {
+		return base.next.DeleteEmptyDirs(options)
+	}
+	// this method is only implemented in file cache
+	return fmt.Errorf("%s component does not implement DeleteEmptyDirs", base.Name())
```
vibhansa-msft, line 120: we shall return nil here, not all components will implement this method.

component/file_cache/file_cache.go
```diff
@@ -614,6 +614,43 @@ func (fc *FileCache) IsDirEmpty(options internal.IsDirEmptyOptions) bool {
 	return fc.NextComponent().IsDirEmpty(options)
 }
 
+// DeleteEmptyDirs: delete empty directories in local cache, return error if directory is not empty
+func (fc *FileCache) DeleteEmptyDirs(options internal.DeleteDirOptions) error {
```
souravgupta-msft, line 618: Add tests for parallel rename dir and create operation in the dir

internal/base_component.go
souravgupta-msft: `DeleteEmptyDirs` method is only implemented in file cache and is only called in `libfuse_rmdir` method when directory is not empty.
```
empty := fuseFS.NextComponent().IsDirEmpty(internal.IsDirEmptyOptions{Name: name})
if !empty {
// delete empty directories from local cache directory
err := fuseFS.NextComponent().DeleteEmptyDirs(internal.DeleteDirOptions{Name: name})
if err != nil {
	log.Err("Libfuse::libfuse_rmdir : Failed to delete %s [%s]", name, err.Error())
	return -C.ENOTEMPTY
    }
}
```

If file cache is not present in the config, this should ideally be pass through method. So, if nil is returned in base component, it won't return `C.ENOTEMPTY` error.

vibhansa-msft: If there are all components in the pipeline and tommorow for some reason we decide to forward the call, it will reach azstorage and it will return error if it has not implemented this method. For all other methods as well we just return back nil if they  have not implemented it. Just to be consistent with other methods also we shall return nil and not error.

souravgupta-msft: Updated the return type of this method to return boolean value and error. For components which do not implement this, the default value returned is false and nil error.

vibhansa-msft (approved):

component/file_cache/file_cache.go
```diff
@@ -651,6 +651,46 @@ func (fc *FileCache) IsDirEmpty(options internal.IsDirEmptyOptions) bool {
 	return fc.NextComponent().IsDirEmpty(options)
 }
 
+// DeleteEmptyDirs: delete empty directories in local cache, return error if directory is not empty
+func (fc *FileCache) DeleteEmptyDirs(options internal.DeleteDirOptions) (bool, error) {
```
syeleti-msft, line 655: Why do we have 2 return parameters here if they both signify the same thing!

souravgupta-msft: You can refer the previous comment. It's related to why I updated this method to return two values.

PR #1529, opened 2024-09-28
Custom component support

abhiguptacse:
- Custom component feature support via go plugin.
- Added an E2E test to verify read/write with two sample custom components.
POC doc: [custom-component](https://microsoftapc-my.sharepoint.com/:w:/g/personal/abhinavgupta_microsoft_com/EWRc-EzxE2xPkQOn4Rsvuv0BjjY_SX0jxaxjZrqPWG5-og?e=bdvcBD)


sample_custom_component2/main.go
```diff
@@ -0,0 +1,222 @@
+package main
```
vibhansa-msft, line 1: Let's move these sample components under the Test directory as they are not part of the main line code.

sample_custom_component2/main.go
```diff
@@ -0,0 +1,222 @@
...
+	"syscall"
+	"time"
+
+	"github.com/Azure/azure-storage-fuse/v2/common/config"
+	"github.com/Azure/azure-storage-fuse/v2/common/log"
+	"github.com/Azure/azure-storage-fuse/v2/external"
+	"github.com/Azure/azure-storage-fuse/v2/internal"
+)
+
+// SAMPLE EXTERNAL COMPONENT IMPLEMENTATION
```
vibhansa-msft, line 16: Add a comment explaining how to build this component inform of a dll/so file which can be later linked to main code.

vibhansa-msft: Sample code shall be clear enough to tell a user on how to compile this code.

internal/component_options.go
```diff
@@ -191,9 +191,10 @@ type ChownOptions struct {
 }
 
 type StageDataOptions struct {
-	Name string
-	Id   string
-	Data []byte
+	Name   string
+	Id     string
+	Data   []byte
+	Offset uint64
```
vibhansa-msft, line 197: Any particular reason to use offset here?

component/custom/custom_loader.go
```diff
@@ -0,0 +1,67 @@
+package custom
```
vibhansa-msft, line 1: As a common coding practice we use the same name in both file and directory so lets rename this file to custom.go

component/custom/custom_loader.go
```diff
@@ -0,0 +1,67 @@
...
+
+import (
```
vibhansa-msft, line 3: As a good coding practice we have always created a test file for each component in the same directory. I do not see a UT file here for custom. This will be flagged in our code coverage pipelines and it will fail.

component/custom/custom_loader.go
```diff
@@ -0,0 +1,67 @@
...
+	"strings"
+
+	"github.com/Azure/azure-storage-fuse/v2/common/log"
+	"github.com/Azure/azure-storage-fuse/v2/external"
+	"github.com/Azure/azure-storage-fuse/v2/internal"
+)
+
+func init() {
+	// Environment variable for directory containing the plugin .so files.
+	pluginDirPath := os.Getenv("BLOBFUSE_PLUGIN_DIRECTORY")
```
vibhansa-msft, line 16: Are we keeping this only in Env variables or in the mount command this can be supplied as CLI parameter as well?

component/custom/custom_loader.go
```diff
@@ -0,0 +1,67 @@
...
+	if pluginDirPath == "" {
+		return
+	}
+
+	// Environment variable which expects file names as a colon-separated list of `.so` files located in
+	// the specified plugin directory. For example: BLOBFUSE_PLUGIN_NAMES=plugin1.so:plugin2.so:plugin3.so
+	pluginFilesPath := os.Getenv("BLOBFUSE_PLUGIN_NAMES")
```
vibhansa-msft, line 23: Do we need two env variables for this. Instead of path we could have asked user to provide fully qualified path in this variable itself seperated by ":" or something

component/custom/custom_loader.go
```diff
@@ -0,0 +1,67 @@
...
+				pluginFiles = append(pluginFiles, filepath.Join(pluginDirPath, file.Name()))
+			}
+		}
+	}
+
+	for _, file := range pluginFiles {
+		log.Info("custom_loader::Opening plugin: %s", file)
+		p, err := plugin.Open(file)
+		if err != nil {
+			log.Err("custom_loader::Error opening plugin: %s", err.Error())
```
vibhansa-msft, line 49: As this is init phase, our logging might not be initialed yet. We shall discuss on whether to throw this message on console or not.

component/custom/custom_loader.go
```diff
@@ -0,0 +1,67 @@
...
+			log.Err("custom_loader::Error looking up GetExternalComponent function: %s", err.Error())
+			os.Exit(1)
+		}
+		getExternalComponent, ok := getExternalComponentFunc.(func() (string, func() external.Component))
+		if !ok {
+			log.Err("custom_loader::Error casting GetExternalComponent function")
+			os.Exit(1)
+		}
+		compName, initExternalComponent := getExternalComponent()
+		internal.AddComponent(compName, initExternalComponent)
```
vibhansa-msft, line 65: Idea wise this looks interesting that external component can define its name and here we register its name just like a regular inbuilt component. Good design :)

component/block_cache/block_cache.go
```diff
@@ -1535,9 +1536,10 @@ func (bc *BlockCache) stageZeroBlock(handle *handlemap.Handle, tryCnt int) (stri
 
 	log.Debug("BlockCache::stageZeroBlock : Staging zero block for %v=>%v, try = %v", handle.ID, handle.Path, tryCnt)
 	err := bc.NextComponent().StageData(internal.StageDataOptions{
-		Name: handle.Path,
-		Data: bc.blockPool.zeroBlock.data[:],
-		Id:   id,
+		Name:   handle.Path,
+		Data:   bc.blockPool.zeroBlock.data[:],
+		Offset: 0,
```
souravgupta-msft, line 1541: This can be removed as the default value is 0 and for staging a block offset should not be required.

component/custom/custom_loader.go
souravgupta-msft: Can you also add some documentation on it in README mentioning the env variables and other parameters if any.

external/external.go
```diff
@@ -0,0 +1,107 @@
+package external
```
souravgupta-msft, line 1: Can you rename this package to `exported` and the file to `exported.go`. Reason for saying this is to follow the same naming convention across Blobfuse and GO SDK.

sample_custom_component2/main.go
abhiguptacse: Will take renaming and movement changes at last iteration to keep it easy for reviewers.

internal/component_options.go
abhiguptacse: Ideally if a component(block_cache) is staging data(write operation) it should also send the offset from where we need to start writing. Next Component of block_cache doesn't necessarily have to be AzStorage.

abhiguptacse: We need offset in encryptor component for our use case because for encryptor it is just file operation to another mount point.

component/custom/custom_loader.go
abhiguptacse: I gave a thought about adding a unit test but couldn't come up something that made sense for UT. Any ideas around how to test an init() that is just registering new component without having a blobfuse mount. ?
I have added an E2E which expects you two have a blobfuse mount and validates read/write on that mount path.

component/custom/custom_loader.go
abhiguptacse: Ideally in these plugin models there should be just one specific folder for .so files. As a good practice used two variables for one folder and another to specify the file names. Open for suggestions on this.

component/custom/custom_loader.go
abhiguptacse: Logging on console was looking odd. I verified that logger is getting initialised and the logs are coming to syslog.

component/custom/custom_loader.go
abhiguptacse: This is only in Env variables right now.

component/custom/custom.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/Azure/azure-storage-fuse/v2/common/log"
+	"github.com/Azure/azure-storage-fuse/v2/exported"
+	"github.com/Azure/azure-storage-fuse/v2/internal"
+)
+
+func init() {
+	// Environment variable for directory containing the plugin .so files.
+	pluginDirPath := os.Getenv("BLOBFUSE_PLUGIN_DIRECTORY")
+	if pluginDirPath == "" {
```
vibhansa-msft, line 17: I still feel instead of giving a directory path and a files list is bit complicated for the user.
If I just have the plugin list then I can set for e.g. "BLOBFUSE_PLUGIN_NAMES=/home/user1/testcomponent1.so;/home/user2/testcomponent2.so;/mnt/temp/testcomponent3.so". Directory to load all .so files from it is not safe and bug prone.

component/custom/custom.go
```diff
@@ -0,0 +1,67 @@
...
+	"path/filepath"
+	"plugin"
+	"strings"
+
+	"github.com/Azure/azure-storage-fuse/v2/common/log"
+	"github.com/Azure/azure-storage-fuse/v2/exported"
+	"github.com/Azure/azure-storage-fuse/v2/internal"
+)
+
+func init() {
```
vibhansa-msft, line 14: UT can be written to explicitly call the init method of this module and give a directory with some dummy .so files and validate one shall load successfully, one shall fail and all those combinations like .so does not exists, path is invalid etc.

component/custom/custom.go
abhiguptacse: updated.

component/custom/custom.go
abhiguptacse: As init() cannot be invoked explicitly used a helper method instead to test. Also dummy logic won't work with .so files. so had to validate via sample custom components.

component/custom/custom.go
```diff
@@ -0,0 +1,57 @@
+package custom
```
vibhansa-msft, line 1: Copyright statements are missing. Run "copyright_fix.sh" from workspace root to fix all required files.

vibhansa-msft: "./copyright_fix.sh " is the command to use.

PR #1533, opened 2024-10-07
Fix ReadPanic

syeleti-msft:
https://github.com/Azure/azure-storage-fuse/issues/1426

Read is getting panicked due to block.endIndex being incorrectly modified.

**Fix:**
Removed block.endIndex field and added a simple way to find the block size.

### DIFF
--- a/component/block_cache/block.go
+++ b/component/block_cache/block.go
@@ -61,13 +61,12 @@ const (
 
 // Block is a memory mapped buffer with its state to hold data
 type Block struct {
-	offset   uint64          // Start offset of the data this block holds
-	id       int64           // Id of the block i.e. (offset / block size)
-	endIndex uint64          // Length of the data this block holds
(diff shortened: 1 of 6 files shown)


component/block_cache/block.go
```diff
@@ -61,13 +61,12 @@ const (
 
 // Block is a memory mapped buffer with its state to hold data
 type Block struct {
-	offset   uint64          // Start offset of the data this block holds
-	id       int64           // Id of the block i.e. (offset / block size)
-	endIndex uint64          // Length of the data this block holds
```
jainakanksha-msft, line 66: What's the reason to remove endIndex ?

vibhansa-msft (approved):

jainakanksha-msft (approved):

souravgupta-msft (approved):

PR #1535, opened 2024-10-08
Generate default config 

ashruti-msft:
## ✅ What

Added a new command to generate optimal default configurations.

## 👩‍🔬 How to validate

```
./blobfuse2 gen-config --component block_cache --o console
```
## 🔖 Related links

image


cmd/default-config.go
```diff
@@ -0,0 +1,136 @@
...
+
+		_, err = internal.NewPipeline(pipeline, true)
+		if err != nil {
+			return fmt.Errorf("defaultConfig:: error creating pipeline [%s]", err.Error())
+		}
+
+		yamlContent = "\n#Required\n#azstorage:\n  #  type: block|adls \n  #  account-name: <name of the storage account>\n  #  container: <name of the storage container to be mounted>\n  #  endpoint: <example - https://account-name.blob.core.windows.net>\n  #  mode: key|sas|spn|msi|azcli \n  #  account-key: <storage account key>\n  # OR\n  #  sas: <storage account sas>\n  # OR\n  #  appid: <storage account app id / client id for MSI>\n  # OR\n  #  tenantid: <storage account tenant id for SPN"
+
+		// Open the file in append mode, create it if it doesn't exist
+		file, err = os.OpenFile("defaultConfig.yaml", os.O_APPEND|os.O_CREATE|os.O_WRONLY, 0644)
```
vibhansa-msft, line 116: Why do we need to open the config file again here?

component/block_cache/block_cache.go
```diff
@@ -295,6 +299,25 @@ func (bc *BlockCache) Configure(_ bool) error {
 		}
 	}
 
+	if common.GenConfig {
+		log.Info("BlockCache::Configure : config generation started")
```
vibhansa-msft, line 303: We shall discuss on this model, each component opening the config file and dumping its own config part?
If we can come up with some better approach to do this will be great.

cmd/default-config.go
```diff
@@ -0,0 +1,136 @@
...
+
+type defaultConfigOptions struct {
+	configComp     string
+	configTmp      string
+	configDirectIO bool
+}
+
+var opts2 defaultConfigOptions
+
+var defaultConfig = &cobra.Command{
```
souravgupta-msft, line 53: Add UT for this command

cmd/default-config.go
```diff
@@ -0,0 +1,136 @@
...
+	Use:               "gen-config",
+	Short:             "Generate default config file.",
+	Long:              "Generate default config file with the values pre-caculated by blobfuse2.",
+	SuggestFor:        []string{"default config", "generate default config"},
+	Hidden:            true,
```
souravgupta-msft, line 58: Do we have to keep this command as hidden?

component/block_cache/block_cache.go
ashruti-msft: Yes maybe we can store the yaml content from each component in a common string and finally in the default-config we can open and dump it. Or some other way maybe we can discuss.

jainakanksha-msft:
Thinking about our genConfig functionality should be treated as an independent feature. Much like how our mount and unmount commands operate as standalone functions, it makes sense for genConfig to independently generate and dump all default configurations into a file. This way, it won't rely on the existing implementation, ensuring a cleaner, more modular approach.

Flow - 
image

So rather than just extending the current functionality make it an independent feature as describe in sequence diagram.

jainakanksha-msft:
also rename the command file to genConfig to align with product design.

cmd/default-config.go
vibhansa-msft: I guess this shall be fine to have as a hidden command as not everyone will use this.

cmd/default-config.go
ashruti-msft: Added

ashruti-msft:
@jainakanksha-msft made the reqd changes according to your suggestions lmk if this looks good.

component/libfuse/libfuse.go
```diff
@@ -246,6 +246,19 @@ func (lf *Libfuse) Validate(opt *LibfuseOptions) error {
 	return nil
 }
 
+func (lf *Libfuse) GenConfig() error {
```
jainakanksha-msft, line 249: Rather than using string, let's use string builder.
Other way to write the same
func (lf *Libfuse) GenConfig() error {
	// Initialize the libfuse options based on current values
	options := LibfuseOptions{
		AttributeExpiration:     lf.attributeExpiration,
		EntryExpiration:         lf.entryExpiration,
		NegativeEntryExpiration: lf.negativeTimeout,
	}

	// If DirectIO is enabled, override expiration values
	if common.DirectIO {
		options.AttributeExpiration = 0
		options.EntryExpiration = 0
		options.NegativeEntryExpiration = 0
	}

	// Use a string builder to construct the YAML content
	var sb strings.Builder
	sb.WriteString("\nlibfuse:\n")

	// DirectIO check merged within value setting
	sb.WriteString(fmt.Sprintf(
		"  attribute-expiration-sec: %v\n  entry-expiration-sec: %v\n  negative-entry-expiration-sec: %v\n",
		options.AttributeExpiration, options.EntryExpiration, options.NegativeEntryExpiration))

	if common.DirectIO {
		sb.WriteString("  direct-io: true\n")
	}

	// Append generated YAML to the common configuration
	common.ConfigYaml += sb.String()

	return nil
}

Stringbuilder always perform better than string when we concatenate and build the result.

component/libfuse/libfuse.go
```diff
@@ -304,6 +317,10 @@ func (lf *Libfuse) Configure(_ bool) error {
 	log.Info("Libfuse::Configure : read-only %t, allow-other %t, allow-root %t, default-perm %d, entry-timeout %d, attr-time %d, negative-timeout %d, ignore-open-flags %t, nonempty %t, direct_io %t, max-fuse-threads %d, fuse-trace %t, extension %s, disable-writeback-cache %t, dirPermission %v, mountPath %v, umask %v",
 		lf.readOnly, lf.allowOther, lf.allowRoot, lf.filePermission, lf.entryExpiration, lf.attributeExpiration, lf.negativeTimeout, lf.ignoreOpenFlags, lf.nonEmptyMount, lf.directIO, lf.maxFuseThreads, lf.traceEnable, lf.extensionPath, lf.disableWritebackCache, lf.dirPermission, lf.mountPath, lf.umask)
 
+	if common.GenConfig {
```
jainakanksha-msft, line 320: What's the significance of doing this?

cmd/gen-config.go
```diff
@@ -0,0 +1,127 @@
...
+		_, err := internal.NewPipeline(pipeline, true)
+		if err != nil {
+			return fmt.Errorf("generatedConfig:: error creating pipeline [%s]", err.Error())
+		}
+
+		yamlContent += common.ConfigYaml
+
+		yamlContent += "\n#Required\n#azstorage:\n  #  type: block|adls \n  #  account-name: <name of the storage account>\n  #  container: <name of the storage container to be mounted>\n  #  endpoint: <example - https://account-name.blob.core.windows.net>\n  #  mode: key|sas|spn|msi|azcli \n  #  account-key: <storage account key>\n  # OR\n  #  sas: <storage account sas>\n  # OR\n  #  appid: <storage account app id / client id for MSI>\n  # OR\n  #  tenantid: <storage account tenant id for SPN"
+
+		// Open the file in append mode, create it if it doesn't exist
```
jainakanksha-msft, line 106: For the tasks of opening and adding content to a file, it would be great if we could use an existing utility or common function. If such a function doesn't already exist, let's create one and place it in a common location.

cmd/gen-config.go
```diff
@@ -0,0 +1,127 @@
...
+			pipeline = append(pipeline, "file_cache")
+			common.TmpPath = opts2.configTmp
+		}
+		if opts2.configDirectIO {
+			common.DirectIO = true
+		} else {
+			pipeline = append(pipeline, "attr_cache")
+		}
+
+		yamlContent := "# Logger configuration\n#logging:\n  #  type: syslog|silent|base\n  #  level: log_off|log_crit|log_err|log_warning|log_info|log_trace|log_debug\n  #  file-path: <path where log files shall be stored. Default - '$HOME/.blobfuse2/blobfuse2.log'>\n"
```
jainakanksha-msft, line 88: Let's replace yamlContent from string to stringbuilder.
Let's also make it in multi-line for easy understanding.
Sample -

var sb strings.Builder
		sb.WriteString("# Logger configuration\n#logging:\n")
		sb.WriteString("  #  type: syslog|silent|base\n")
		sb.WriteString("  #  level: log_off|log_crit|log_err|log_warning|log_info|log_trace|log_debug\n")
		sb.WriteString("  #  file-path: <path where log files shall be stored. Default - '$HOME/.blobfuse2/blobfuse2.log'>\n")

cmd/gen-config.go
```diff
@@ -0,0 +1,127 @@
...
+		}
+
+		pipeline := []string{"libfuse"}
+		if opts2.configComp == "block_cache" {
+			pipeline = append(pipeline, "block_cache")
+			common.TmpPath = opts2.configTmp
+		} else if opts2.configComp == "file_cache" {
+			pipeline = append(pipeline, "file_cache")
+			common.TmpPath = opts2.configTmp
+		}
```
jainakanksha-msft, line 81: Can we replace it like this
pipeline := []string{"libfuse", opts2.configComp}
common.TmpPath = opts2.configTmp

or are there any other possibilities in opts2.configComp?

component/attr_cache/attr_cache.go
```diff
@@ -119,6 +120,16 @@ func (ac *AttrCache) Stop() error {
 	return nil
 }
 
+// GenConfig : Generate the default config for the component
+func (ac *AttrCache) GenConfig() error {
+	log.Info("AttrCache::Configure : config generation started")
```
jainakanksha-msft, line 125: I think command level (means in gen-Config command file) logging is enough.
Adding info level into every component, doesn't have any significance.

ashruti-msft: Yes Ik anyways we won't get the logs as we aren't giving user to add the log path. I just added this log in case there is some error which leads to this method getting called in future without gen-config? Then we would know something's wrong. Idk I felt it's just a way to track incase of some issue in our code later on.

cmd/gen-config.go
ashruti-msft: yes this is redundant i guess i added this check previously for some reason but now its not needed , will change it

component/libfuse/libfuse.go
ashruti-msft: so that we don't call genconfig func in normal flow of blobfuse but only while using gen-config command. I guess it wouldn't matter much since we are only writing to the config file in gen-config but yeah that's why its there

cmd/gen-config.go
```diff
@@ -0,0 +1,127 @@
...
+		yamlContent := "# Logger configuration\n#logging:\n  #  type: syslog|silent|base\n  #  level: log_off|log_crit|log_err|log_warning|log_info|log_trace|log_debug\n  #  file-path: <path where log files shall be stored. Default - '$HOME/.blobfuse2/blobfuse2.log'>\n"
+		yamlContent += "\ncomponents:\n"
+
+		// Iterate through the pipeline and add each component to the YAML content
+		for _, component := range pipeline {
+			yamlContent += fmt.Sprintf("  - %s\n", component)
+		}
+		yamlContent += "  - azstorage\n"
+
+		_, err := internal.NewPipeline(pipeline, true)
```
jainakanksha-msft, line 97: Do we require to call the whole NewPipeline?
NewPipeline > register all the components.
we are already excluding priority part.
Registration is needed for genConfig or not.
If not can we direct the call genConfig call without calling NewPiepline itself.?

ashruti-msft: I'm just trying to reuse the code in newpipeline otherwise I'll write the same logic to call the configure of each comp. All the components get registered in the beginning and then we extract from the map and call their constructor method then configure.

component/libfuse/libfuse.go
ashruti-msft: done

cmd/gen-config.go
ashruti-msft: added

jainakanksha-msft (approved):
Looks good.

internal/pipeline.go
```diff
@@ -70,15 +71,17 @@ func NewPipeline(components []string, isParent bool) (*Pipeline, error) {
 				return nil, err
 			}
 
-			if !(comp.Priority() <= lastPriority) {
-				log.Err("Pipeline::NewPipeline : Invalid Component order [priority of %s higher than above components]", comp.Name())
-				return nil, fmt.Errorf("config error in Pipeline [component %s is out of order]", name)
-			} else {
-				lastPriority = comp.Priority()
-			}
+			if !common.GenConfig {
```
jainakanksha-msft, line 74: Do we have to add some condition exclusion in unit test file for this.?

ashruti-msft: Added test for this

cmd/gen-config.go
```diff
@@ -0,0 +1,115 @@
...
+	SuggestFor:        []string{"generate default config", "generate config"},
+	Hidden:            true,
+	Args:              cobra.ExactArgs(0),
+	FlagErrorHandling: cobra.ExitOnError,
+	RunE: func(cmd *cobra.Command, args []string) error {
+
+		common.GenConfig = true
+
+		if opts2.configComp != "block_cache" && opts2.configComp != "file_cache" {
+			return fmt.Errorf("component is required and should be either block_cache or file_cache")
```
souravgupta-msft, line 67: You can specify in this error how to provide component name using the `--component` flag

cmd/gen-config.go
```diff
@@ -0,0 +1,115 @@
...
+		err = common.WriteToFile("generatedConfig.yaml", sb.String(), common.WriteToFileOptions{Flags: os.O_TRUNC, Permission: 0644})
+		return err
+	},
+}
+
+func init() {
+	rootCmd.AddCommand(generatedConfig)
+	generatedConfig.Flags().StringVar(&opts2.configComp, "component", "", "Input block_cache or file_cache")
+	generatedConfig.Flags().StringVar(&opts2.configTmp, "tmp-path", "", "Input path for caching")
+	generatedConfig.Flags().BoolVar(&opts2.configDirectIO, "direct-io", false, "Choose direct-io mode")
```
souravgupta-msft, line 114: Should we use `direct_io` here so that it is same as the fuse CLI option.

cmd/gen-config.go
```diff
@@ -0,0 +1,115 @@
...
+		common.TmpPath = opts2.configTmp
+
+		if opts2.configDirectIO {
+			common.DirectIO = true
+		} else {
+			pipeline = append(pipeline, "attr_cache")
+		}
+
+		var sb strings.Builder
+		sb.WriteString("# Logger configuration\n#logging:\n  #  type: syslog|silent|base\n  #  level: log_off|log_crit|log_err|log_warning|log_info|log_trace|log_debug\n")
```
souravgupta-msft, line 85: Why are the logging and azstorage parts commented in the config generated?

cmd/gen-config.go
```diff
@@ -0,0 +1,115 @@
...
+		if err != nil {
+			return fmt.Errorf("generatedConfig:: error creating pipeline [%s]", err.Error())
+		}
+
+		sb.WriteString(common.ConfigYaml)
+
+		sb.WriteString("\n#Required\n#azstorage:\n  #  type: block|adls \n  #  account-name: <name of the storage account>\n  #  container: <name of the storage container to be mounted>\n  #  endpoint: <example - https://account-name.blob.core.windows.net>\n  ")
+		sb.WriteString("#  mode: key|sas|spn|msi|azcli \n  #  account-key: <storage account key>\n  # OR\n  #  sas: <storage account sas>\n  # OR\n  #  appid: <storage account app id / client id for MSI>\n  # OR\n  #  tenantid: <storage account tenant id for SPN")
+
+		err = common.WriteToFile("generatedConfig.yaml", sb.String(), common.WriteToFileOptions{Flags: os.O_TRUNC, Permission: 0644})
```
souravgupta-msft, line 105: We should write this config file to some default path or take it as input from the user

cmd/gen-config_test.go
```diff
@@ -0,0 +1,187 @@
...
+	//check if the generated file is not empty
+	file, err := os.ReadFile("generatedConfig.yaml")
+	suite.assert.Nil(err)
+	suite.assert.NotEmpty(file)
+
+	//check if the generated file has the correct direct io flag
+	suite.assert.Contains(string(file), "direct-io: true")
+}
+
+func (suite *genConfig) TestInvalidComponent() {
```
jainakanksha-msft, line 141: if it is not file cache or block cache but certain custom component,
should we fail it or generate the remaining configurations?

common/util.go
```diff
@@ -90,7 +90,7 @@ func IsMountActive(path string) (bool, error) {
 	var out bytes.Buffer
 	cmd := exec.Command("pidof", "blobfuse2")
 	cmd.Stdout = &out
-
+	print(out.String() + "Pids printed")
```
jainakanksha-msft, line 93: We can remove this

component/file_cache/file_cache.go
```diff
@@ -316,6 +331,11 @@ func (c *FileCache) Configure(_ bool) error {
 	log.Info("FileCache::Configure : create-empty %t, cache-timeout %d, tmp-path %s, max-size-mb %d, high-mark %d, low-mark %d, refresh-sec %v, max-eviction %v, hard-limit %v, policy %s, allow-non-empty-temp %t, cleanup-on-start %t, policy-trace %t, offload-io %t, sync-to-flush %t, ignore-sync %t, defaultPermission %v, diskHighWaterMark %v, maxCacheSize %v, mountPath %v",
 		c.createEmptyFile, int(c.cacheTimeout), c.tmpPath, int(cacheConfig.maxSizeMB), int(cacheConfig.highThreshold), int(cacheConfig.lowThreshold), c.refreshSec, cacheConfig.maxEviction, c.hardLimit, conf.Policy, c.allowNonEmpty, c.cleanupOnStart, c.policyTrace, c.offloadIO, c.syncToFlush, c.syncToDelete, c.defaultPermission, c.diskHighWaterMark, c.maxCacheSize, c.mountPath)
 
+	if common.GenConfig {
+		err = c.GenConfig()
```
jainakanksha-msft, line 335: direct return, rather than creating a new variable

component/block_cache/block_cache.go
```diff
@@ -295,6 +312,10 @@ func (bc *BlockCache) Configure(_ bool) error {
 		}
 	}
 
+	if common.GenConfig {
+		err = bc.GenConfig()
+		return err
```
jainakanksha-msft, line 317: direct return, rather than creating a new variable

component/attr_cache/attr_cache.go
```diff
@@ -156,6 +167,11 @@ func (ac *AttrCache) Configure(_ bool) error {
 	log.Info("AttrCache::Configure : cache-timeout %d, symlink %t, cache-on-list %t, max-files %d",
 		ac.cacheTimeout, ac.noSymlinks, ac.cacheOnList, ac.maxFiles)
 
+	if common.GenConfig {
+		err = ac.GenConfig()
+		return err
```
jainakanksha-msft, line 172: direct return, rather than creating a new variable

component/libfuse/libfuse.go
```diff
@@ -304,6 +334,11 @@ func (lf *Libfuse) Configure(_ bool) error {
...
+		err = lf.GenConfig()
+		return err
```
jainakanksha-msft, line 339: direct return, rather than creating a new variable

vibhansa-msft (approved):

cmd/gen-config.go
ashruti-msft: Logging is commented because its optional and azstorage is commented because we are not populating storage accounts values through the gen-config command. So I have kind of given a blueprint of azstorage component and commented it for the cx to edit as reqd.

cmd/gen-config_test.go
ashruti-msft: Right now we only support fc, bc if the user gives an unsupported component its better if we fail and inform them during config generation to avoid future complications.

cmd/gen-config.go
ashruti-msft: added default location

jainakanksha-msft (approved):
LGTM

PR #1542, opened 2024-10-15
Custom component feature

abhiguptacse:
Custom component feature support via go plugin.
Added an E2E test to verify read/write with two sample custom components.
POC doc: [custom-component](https://microsoftapc-my.sharepoint.com/:w:/g/personal/abhinavgupta_microsoft_com/EWRc-EzxE2xPkQOn4Rsvuv0BjjY_SX0jxaxjZrqPWG5-og?e=bdvcBD)


vibhansa-msft (approved):

component/custom/custom.go
```diff
@@ -0,0 +1,90 @@
...
+		return nil
+	}
+
+	pluginFiles := strings.Split(pluginFilesPath, ":")
+
+	for _, file := range pluginFiles {
+		if strings.HasSuffix(file, ".so") {
+			p, err := plugin.Open(file)
+			if err != nil {
+				return fmt.Errorf("error opening plugin %s: %s", file, err.Error())
```
souravgupta-msft, line 62: Can you also log the errors returned

component/custom/custom.go
```diff
@@ -0,0 +1,90 @@
...
+
+			getExternalComponent, ok := getExternalComponentFunc.(func() (string, func() exported.Component))
+			if !ok {
+				return fmt.Errorf("error casting GetExternalComponent function in %s", file)
+			}
+
+			compName, initExternalComponent := getExternalComponent()
+			internal.AddComponent(compName, initExternalComponent)
+		} else {
+			return fmt.Errorf("invalid plugin file extension: %s", file)
```
souravgupta-msft, line 78: Same as previous, log the error

souravgupta-msft (approved):

component/custom/custom.go
```diff
@@ -0,0 +1,90 @@
...
+		}
+	}
+	return nil
+}
+
+func init() {
+	err := initializePlugins()
+	if err != nil {
+		log.Err("custom::Error initializing plugins: %s", err.Error())
```
vibhansa-msft, line 87: When load is failing in the pipelines its silent on the console. 
Add "fmt.Printf("failed to initialize plugin [%s]", err.Error())" after log and before exit(1) so that user gets an error on console as well.

component/custom/custom.go
```diff
@@ -0,0 +1,90 @@
...
+	// Example BLOBFUSE_PLUGIN_PATH="/path/to/plugin1.so:/path/to/plugin2.so"
+	pluginFilesPath := os.Getenv("BLOBFUSE_PLUGIN_PATH")
+	if pluginFilesPath == "" {
+		return nil
+	}
+
+	pluginFiles := strings.Split(pluginFilesPath, ":")
+
+	for _, file := range pluginFiles {
+		if strings.HasSuffix(file, ".so") {
```
vibhansa-msft, line 59: Let's print each file picked up and tried in the log as well at info level.

component/custom/custom.go
```diff
@@ -0,0 +1,90 @@
...
+				return fmt.Errorf("error looking up GetExternalComponent function in %s: %s", file, err.Error())
+			}
+
+			getExternalComponent, ok := getExternalComponentFunc.(func() (string, func() exported.Component))
+			if !ok {
+				return fmt.Errorf("error casting GetExternalComponent function in %s", file)
+			}
+
+			compName, initExternalComponent := getExternalComponent()
+			internal.AddComponent(compName, initExternalComponent)
```
vibhansa-msft, line 76: Once component is added successfully, lets add a info log on which component was loaded successfully.

component/custom/custom.go
abhiguptacse: Errors are already logged at line 86. Here we are just returning errors to the caller method.

exported/exported.go
```diff
@@ -0,0 +1,140 @@
...
+	PropFlagNotExists
+	PropFlagIsDir
+	PropFlagEmptyDir
+	PropFlagSymlink
+	PropFlagMetadataRetrieved
+	PropFlagModeDefault
+)
+
+// Type aliases for base component
+type BaseComponent = internal.BaseComponent
```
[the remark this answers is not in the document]
abhiguptacse: Already there at line 41. This is just how we are exposing internal attributes for outside user so that they can write the custom components as internal package cannot be imported.

PR #1545, opened 2024-10-17
Add Checks for temp path in Block Cache while mouting

syeleti-msft:
Why - We are facing issues when tempPath is not Empty.  Mount is successful even if tempPath is not empty in case of block Cache scenario. Ideally temp path sould be empty before proceeding ahead after unmount or over a clean start.
what - Check temppath is not same as mount path, Check temppath is empty to get the succeess from the mount


component/block_cache/block_cache.go
```diff
@@ -242,17 +243,30 @@ func (bc *BlockCache) Configure(_ bool) error {
 	bc.tmpPath = ""
 	if conf.TmpPath != "" {
 		bc.tmpPath = common.ExpandPath(conf.TmpPath)
+		//check mnt path is not same as temp path
+
+		err = config.UnmarshalKey("mount-path", &bc.mntPath)
+		if err == nil && bc.mntPath == bc.tmpPath {
```
jainakanksha-msft, line 249: Test case for error scenarios are missing

syeleti-msft: done

component/file_cache/file_cache.go
```diff
@@ -241,7 +241,11 @@ func (c *FileCache) Configure(_ bool) error {
 	}
 
 	err = config.UnmarshalKey("mount-path", &c.mountPath)
-	if err == nil && c.mountPath == c.tmpPath {
+	if err != nil {
```
jainakanksha-msft, line 244: Please add this into UT.

component/block_cache/block_cache.go
```diff
@@ -242,17 +243,34 @@ func (bc *BlockCache) Configure(_ bool) error {
...
+		//check mnt path is not same as temp path
+
+		err = config.UnmarshalKey("mount-path", &bc.mntPath)
+		if err != nil {
+			log.Err("BlockCache: config error [unable to obtain Mount Path]")
+			return fmt.Errorf("config error in %s [%s]", bc.Name(), err.Error())
+		}
+		if bc.mntPath == bc.tmpPath {
+			log.Err("BlockCache: config error [tmp-path is same as mount path]")
+			return fmt.Errorf("config error in %s error [tmp-path is same as mount path]", bc.Name())
```
jainakanksha-msft, line 255: Can't see test case of this

component/block_cache/block_cache.go
```diff
@@ -242,17 +243,34 @@ func (bc *BlockCache) Configure(_ bool) error {
 	bc.tmpPath = ""
 	if conf.TmpPath != "" {
 		bc.tmpPath = common.ExpandPath(conf.TmpPath)
+		//check mnt path is not same as temp path
+
+		err = config.UnmarshalKey("mount-path", &bc.mntPath)
+		if err != nil {
+			log.Err("BlockCache: config error [unable to obtain Mount Path]")
```
jainakanksha-msft, line 250: Can't see test case for this.

component/file_cache/file_cache.go
syeleti-msft: mount-path's value is set in mount.go, which we are not mocking this file currently from the individual components. Currently in tests we only mock individual components. Any suggestion on how to mock this value is appreciated.

component/block_cache/block_cache.go
syeleti-msft: Added!

vibhansa-msft (approved):

component/block_cache/block_cache.go
```diff
@@ -70,6 +70,7 @@ type BlockCache struct {
 
 	blockSize       uint64          // Size of each block to be cached
 	memSize         uint64          // Mem size to be used for caching at the startup
+	mntPath         string          // Mountpoint path
```
souravgupta-msft, line 73: Nit
```suggestion
	mntPath         string          // Mount path
```

souravgupta-msft (approved):

ashruti-msft (approved):

PR #1550, opened 2024-10-29
Move version checks from public container to static website

vibhansa-msft:
## ✅ What
 
Version check is done on a public container with anonymous access, but due to security issues the public access will be terminated soon.
Now onwards the version check with work through a static website hosted on same storage account,
 
## 🤔 Why
 
Security vulnerability restricting anonymous access.
 
Check the URI used to retrieve the latest version
 
## 🔖 Related links
NA


vibhansa-msft:
Current changeset is getting an index.xml file from the website and expects it will have latest version written under \<latest\> tag in that xml file.

cmd/root.go
```diff
@@ -104,18 +103,23 @@ func getRemoteVersion(req string) (string, error) {
 		return "", err
 	}
 
+	if len(body) > 50 {
```
souravgupta-msft, line 106: Should we do something like pattern based here instead of length?

vibhansa-msft: After the check we are doing unmarshall of the xml contents so pattern matching will fail if the expected tags are not present. Hence did not do that explicitly.

common/version.go
```diff
@@ -39,7 +39,7 @@ import (
 	"strings"
 )
 
-const Blobfuse2ListContainerURL = "https://blobfuse2.blob.core.windows.net/release"
+const Blobfuse2ListContainerURL = "https://blobfuse2.z13.web.core.windows.net/release"
```
souravgupta-msft, line 42: Can we make a short URL for this. Also please add a comment what this short URL will have for reference.

blobfuse2-release.yaml
```diff
@@ -1782,7 +1797,39 @@ stages:
                 workingDirectory: $(root_dir)
 
               - script: |
-                  python3 releaseVersionUpdate.py "$(VERSION_CNT_SAS_URL)" "`blobfuse2 version`"
```
souravgupta-msft, line 1785: This python script can be removed as it is not being used anywhere else

souravgupta-msft (approved):

common/version.go
vibhansa-msft: does not work as the aka.ms link needs exact url destination and we suffix the url with the version numbers at times.

ashruti-msft (approved):

souravgupta-msft (approved):

PR #1556, opened 2024-11-05
feat: support workload identity token

cvvz:
## ✅ What
 
Support provide workload identity token when authenticating rather than use the token file.
 
## 🤔 Why
 
Kubelet generate the token when mounting blob volume and we should pass the token to blobfuse2 rather than create a token file. 
 
I've already tested with blob-csi-driver
 
## 🔖 Related links
 
- [Issues](<link>)
- [Team thread](<link>)
- [Documents](<link>)
- [Email Subject]


andyzhangx:
@vibhansa-msft this PR exposes `WORKLOAD_IDENTITY_TOKEN` as env variable, could you take a look? thanks.

component/azstorage/config.go
```diff
@@ -124,24 +124,25 @@ const DefaultMaxResultsForList int32 = 2
...
+	EnvAzStorageAccessKey                = "AZURE_STORAGE_ACCESS_KEY"
+	EnvAzStorageSasToken                 = "AZURE_STORAGE_SAS_TOKEN"
+	EnvAzStorageIdentityClientId         = "AZURE_STORAGE_IDENTITY_CLIENT_ID"
+	EnvAzStorageIdentityResourceId       = "AZURE_STORAGE_IDENTITY_RESOURCE_ID"
+	EnvAzStorageIdentityObjectId         = "AZURE_STORAGE_IDENTITY_OBJECT_ID"
+	EnvAzStorageSpnTenantId              = "AZURE_STORAGE_SPN_TENANT_ID"
+	EnvAzStorageSpnClientId              = "AZURE_STORAGE_SPN_CLIENT_ID"
+	EnvAzStorageSpnClientSecret          = "AZURE_STORAGE_SPN_CLIENT_SECRET"
+	EnvAzStorageSpnOAuthTokenFilePath    = "AZURE_OAUTH_TOKEN_FILE"
+	EnvAzStorageSpnWorkloadIdentityToken = "WORKLOAD_IDENTITY_TOKEN"
```
andyzhangx, line 138: if we use `WORKLOAD_IDENTITY_TOKEN` env var, who is responsible for token refresh?

component/azstorage/azauthspn.go
```diff
@@ -69,6 +71,20 @@ func (azspn *azAuthSPN) getTokenCredential() (azcore.TokenCredential, error) {
...
 			return nil, err
 		}
+	} else if azspn.config.WorkloadIdentityToken != "" {
+		log.Trace("AzAuthSPN::getTokenCredential : Going for fedrated token flow ")
+
+		cred, err = azidentity.NewClientAssertionCredential(
+			azspn.config.TenantID,
+			azspn.config.ClientID,
+			func(ctx context.Context) (string, error) {
+				return azspn.config.WorkloadIdentityToken, nil
```
vibhansa-msft, line 81: Is this option something like where user is providing the token directly to blobfuse?
If so, what is the process to refresh the token or validation of token ?

component/azstorage/config.go
```diff
@@ -450,14 +453,15 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 		az.stConfig.authConfig.ResourceID = opt.ResourceID
 	case EAuthType.SPN():
 		az.stConfig.authConfig.AuthMode = EAuthType.SPN()
-		if opt.ClientID == "" || (opt.ClientSecret == "" && opt.OAuthTokenFilePath == "") || opt.TenantID == "" {
+		if opt.ClientID == "" || (opt.ClientSecret == "" && opt.OAuthTokenFilePath == "" && opt.WorkloadIdentityToken == "") || opt.TenantID == "" {
 			//lint:ignore ST1005 ignore
-			return errors.New("Client ID, Tenant ID or Client Secret not provided")
+			return errors.New(fmt.Sprintf("Client ID %s, Tenant ID %s or Client Secret %s, OAuthTokenFilePath %s, WorkloadIdentityToken %s not provided", opt.ClientID, opt.TenantID, opt.ClientSecret, opt.OAuthTokenFilePath, opt.WorkloadIdentityToken))
```
souravgupta-msft, line 458: We should not log the secret or token here

component/azstorage/azauthspn.go
andyzhangx: why do we need to save the token in a file? that's not safe

jainakanksha-msft:
@cvvz Please update PR description

component/azstorage/azauthspn.go
vibhansa-msft: My question was are we expecting a file name here in this env variable or directly the token. If directly token is what we expect then who will be responsible for the refresh. If its a file then re-reding the file (assuming the file now has the latest token) can work but giving directly the token will not work.

component/azstorage/config.go
cvvz: The token was generated by kubelet when it try to mount volume, and it will be passed to csi driver and csi driver set it as WORKLOAD_IDENTITY_TOKEN env. We should only use this token for mounting.

component/azstorage/azauthspn.go
cvvz: The token was generated by kubelet when try to mount the blob volume and pass to blobfuse for authentication. I think we only need the token for mounting.

component/azstorage/config.go
cvvz: Ok, fixed

component/azstorage/azauthspn.go
andyzhangx: I think here the problem is when token is expired, whether blobfuse mount is still valid or not

andyzhangx: in the before, we use workload identity token to get account key for blobfuse mount, since account key is usually not rotated, then blobfuse mount is always valid, but now the situation is different.

even we store the token in file, but when token context in file changes, the blobfuse mount is always valid? does it need to remount after token refreshed? I think we have never tested on that scenario.

component/azstorage/config.go
andyzhangx: `AZURE_OAUTH_TOKEN_FILE` is used here, it's for blobfuse mount, I think you also need to change the code here when WORKLOAD_IDENTITY_TOKEN env is provided explicitly by the user:

https://github.com/Azure/azure-storage-fuse/blob/42b3b19b42dbe36e5b37e7b4d81115c6a319b277/component/azstorage/azauthspn.go#L58-L66

component/azstorage/azauthspn.go
andyzhangx: I think it's clear now, originally blobfuse2 driver only supports OAuthTokenFilePath, and with this PR, it also support env var. this should work in the first mount, but when workload identity token is expired, whether the blobfuse2 mount would be still valid, we need to verify on that. That may requires remount using OAuthTokenFilePath or env var again

souravgupta-msft: In case of OAuthTokenFilePath, Kubernetes is responsible for updating the token in the given file and the code reads the updated token before its expiry.
https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/azidentity/workload_identity.go#L117

Here in case of `ClientAssertionCredential`, you can have the token refresh logic in the `getAssertion()` method similar to the above link where before expiry, it reads the updated token from the environment variable. The responsibility of updating the env variable should be of the user or AKS in this case.

cvvz: I have just confirmed that token expiration will not affect Blobfuse I/O. Therefore, we don’t need to refresh the token.

vibhansa-msft: How did you confirm this part ? If the token is expired from backend we will start receiving 4xx Auth failure errors. Storage token needs to be refreshed on each expiry. If the token that you use to generate storage token expires, it will not have any impact on Blobfuse IO

cvvz: I created a blobfuse pod and the io was not impacted when the token expires. 
If the token expired, kubelet will generate a new token when we try to mount again.

vibhansa-msft: If you remount then its fine as its a fresh start. Workflow we are discussing here is when Blobfuse remains mounted for a longer duration and in between the token expires. There will not be any remount in this case and newly generated token will not be forwarded to Blobfuse as well. In such case all storage REST call shall fail with 4xx errors. Can you validate this flow.

vibhansa-msft: Blobfuse interacts with Azure Storage via REST call only. Lets assume you are trying to upload a 10GB file. Job starts and file is divided into block and each block is uploaded via a seperate REST call (few uploads going on in parallel). Now while the upload is in progress the token expires. As Blobfuse is not updated with the new token, it will continue using the expired token to upload the remaining chunks but the storage backend will fail these calls with Auth error as the token is expired. This is the part where you need to somehow convey Blobfuse to start using a new token. Generally MSI/SPN case Blobfuse acquires a token and 5 minute before its expiry it refreshes the token internally. Here as the token is provided as an input, its responsibility of the token provider to take care of the refresh and provide the updated token to Blobfuse while the job is going on. This is what Sourav pointed earlier in his comment that the function you have implemented shall take care of token update as well.

vibhansa-msft: If the token that you are sending it across to Blobfuse is not going to expire then its fine and I can approve the PR

andyzhangx: as @cvvz has tested, the workload identity token is only used during blobfuse mount, so this PR is ok to merge

vibhansa-msft: My understanding might be wrong here so trying to clarify again. "workload Identity" is used only during the mount time but you get a storage token using this identity correct, in exchange of this  token provided at mount time ?
Now, there are two possibilities:
1. Token provided by user at mount time is the one which is used directly to access the storage account. In this case every REST call that we send across to storage will have the token. Hence saying that token is only used once is not correct and token may potentially expire during lifetime of blobfuse2 process.
2. If the token provided at mount time is used to get a storage token, in exchange, then the storage token will have  a life time and once it expires we need to renew the storage token. To do so we will need a fresh workload identity token and if user can only provide it once at mount then this will fail.

In either case if token is used to access storage account it will expire and blobfuse will need some way to get a fresh token, which is not considered as part of this PR.  If pod is short lived and we are very well aware that token will not expire before the pod goes off, then it's a fair assumption to not refresh the token.

vibhansa-msft (approved):

souravgupta-msft (approved):

PR #1557, opened 2024-11-05
Support back for object id using azidentity

vibhansa-msft:
## ✅ What
 
- MSI auth with object-id was relying on azcli because object-id support was removed from azidentity. As it's supported now we are removing our dependency on azcli for this.
 
## 🤔 Why
 
- Re-addition of the feature
 
- MSI auth using object-id based login
 
## 🔖 Related links
- NA


souravgupta-msft (approved):

go.mod
```diff
@@ -5,12 +5,12 @@ go 1.22.0
...
 
 require (
-	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.14.0
-	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.7.0
-	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.4.0
-	github.com/Azure/azure-sdk-for-go/sdk/storage/azdatalake v1.2.0
+	github.com/Azure/azure-sdk-for-go/sdk/azcore v1.16.0
+	github.com/Azure/azure-sdk-for-go/sdk/azidentity v1.8.0
+	github.com/Azure/azure-sdk-for-go/sdk/storage/azblob v1.4.1
+	github.com/Azure/azure-sdk-for-go/sdk/storage/azdatalake v1.2.1
```
souravgupta-msft, line 11: CI is failing because these SDK releases have the shared key bug, which has been fixed in the latest beta release. So, we can move these dependencies back to `v1.4.0` and `v1.2.0` and upgrade once these beta releases are GA.

souravgupta-msft (approved):

PR #1558, opened 2024-11-05
Adding custom component feature

abhiguptacse:
## ✅ What
 
Custom component feature support via go plugin.
 
## 🤔 Why

 This allows customers to create their own components and add it in blobfuse pipeline dynamically using go plugin.
 
## 👩‍🔬 How to validate

while running blobfuse we can use the custom plugin and plug it with blobfuse.
Sample command to validate - 
set env variable
```
BLOBFUSE_PLUGIN_PATH="<path>/sample_plugin.so"
./blobfuse2 mount ~/mnt
```
 
## 🔖 Related links

- [Documents](https://microsoftapc-my.sharepoint.com/:w:/g/personal/abhinavgupta_microsoft_com/EWRc-EzxE2xPkQOn4Rsvuv0BjjY_SX0jxaxjZrqPWG5-og?e=bdvcBD)


component/custom/custom.go
```diff
@@ -0,0 +1,96 @@
...
+	"plugin"
+	"strings"
+	"time"
+
+	"github.com/Azure/azure-storage-fuse/v2/common/log"
+	"github.com/Azure/azure-storage-fuse/v2/exported"
+	"github.com/Azure/azure-storage-fuse/v2/internal"
+)
+
+func initializePlugins() error {
```
vibhansa-msft, line 48: We need to ignore this file in blobfuse2-code-coverage.yml file. Kindly add the path to the exclude filter list:
"                cat ./blobfuse2_coverage_raw.rpt | grep -v mock_component | grep -v base_component | grep -v loopback | grep -v tools | grep -v "common/log" | grep -v "common/exectime" | grep -v "common/types.go" | grep -v "internal/stats_manager" | grep -v "main.go" | grep -v "component/azstorage/azauthmsi.go" | grep -v "component/azstorage/azauthspn.go" | grep -v "component/stream" | grep -v "component/azstorage/azauthcli.go" > ./blobfuse2_coverage.rpt 
"
In above line you can just add "grep -v "component/custom" after "component/stream" option

vibhansa-msft (approved):

component/custom/custom.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+	return nil
+}
+
+func init() {
+	err := initializePlugins()
+	if err != nil {
+		log.Err("custom::Error initializing plugins: %s", err.Error())
+		fmt.Printf("failed to initialize plugin: %s\n", err.Error())
+		os.Exit(1)
```
jainakanksha-msft, line 94: @vibhansa-msft Is this correct to exit from the blobfuse here?

vibhansa-msft: Yes if the customer has chosen to load a component and we fail to do that then we shall exit with throwing an error on console. Looks correct to me. Also, if we do not fail here and allow system to run anyway during init of the pipeline its going to fail as it will not be able to find the component. Better to fail early here.

jainakanksha-msft (approved):

PR #1571, opened 2024-11-18
Preserve ACL/Permissions while uploading file over datalake

vibhansa-msft:
## ✅ What
 
When files are uploaded to datalake storage, Permissions and ACL set on the file previously are lost.
 
## 🤔 Why
 
Upload APIs do not set ACLs back. We need to make additional REST call to get the access control list and set it back after the upload.
 
Create a file on datalake, set the permissions from portal and then edit the file from blobfuse. ACLs/Permissions will be lost once file is saved from blobfuse end.
 
## 🔖 Related links
 
NA


component/azstorage/datalake.go
```diff
@@ -553,7 +553,34 @@ func (dl *Datalake) ReadInBuffer(name string, offset int64, len int64, data []by
 
 // WriteFromFile : Upload local file to file
 func (dl *Datalake) WriteFromFile(name string, metadata map[string]*string, fi *os.File) (err error) {
-	return dl.BlockBlob.WriteFromFile(name, metadata, fi)
+	// File in DataLake may have permissions and ACL set. Just uploading the file will override them.
+	// So, we need to get the existing permissions and ACL and set them back after uploading the file.
+
+	fileClient := dl.Filesystem.NewFileClient(filepath.Join(dl.Config.prefixPath, name))
```
souravgupta-msft, line 559: Should we do this behind a hidden flag so that there is no impact to existing flow?

vibhansa-msft: Current implementation is incorrect so better to do this always. If someone comes back with complain on the perf part we may consider that. Other way is to just validate if its default or not, if the permissions are default and no ACLs are applied then we can skip the setCall but even in that case Get call will be done additionally

vibhansa-msft: We already have an "HonourACL" flag, we can reuse this as well. This flag basically means whatever ACL is set show permissions in local file according to that and not by the permissions.

souravgupta-msft: If we don't have this behind a flag, it will always fail for users using OAuth with `Blob Data Contributor` role, because `SetAccessControl` will only work with `Blob Data Owner` role.

component/azstorage/azstorage.go
```diff
@@ -662,6 +662,9 @@ func init() {
 	cpkEnabled := config.AddBoolFlag("cpk-enabled", false, "Enable client provided key.")
 	config.BindPFlag(compName+".cpk-enabled", cpkEnabled)
 
+	preserveACL := config.AddBoolFlag("preserve-acl", false, "Preserve ACL in case of datalake storage")
```
souravgupta-msft, line 665: Add this in the advancedConfig.yaml

souravgupta-msft: Update the description also

component/azstorage/datalake.go
```diff
@@ -578,6 +611,18 @@ func (dl *Datalake) TruncateFile(name string, size int64) error {
 	return dl.BlockBlob.TruncateFile(name, size)
 }
 
+func (dl *Datalake) getAccessControl(name string) (string, error) {
```
souravgupta-msft, line 614: Can this be renamed to getACL as it sounds confusing with the SDK's GetAccessControl method.

component/azstorage/datalake.go
```diff
@@ -553,7 +553,40 @@ func (dl *Datalake) ReadInBuffer(name string, offset int64, len int64, data []by
...
+	var fileClient *file.Client = nil
+
+	if dl.Config.preserveACL {
+		fileClient = dl.Filesystem.NewFileClient(filepath.Join(dl.Config.prefixPath, name))
+		resp, err := fileClient.GetAccessControl(context.Background(), nil)
+		if err != nil {
+			// Earlier code was ignoring this so it might break customer cases where they do not have auth to update ACL
+			log.Err("Datalake::WriteFromFile : Failed to get ACL for %s [%s]", name, err.Error())
+		}
+		acl = *resp.ACL
```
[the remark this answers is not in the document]
vibhansa-msft: updated

component/azstorage/datalake.go
vibhansa-msft: updated

component/azstorage/datalake.go
```diff
@@ -553,7 +553,34 @@ func (dl *Datalake) ReadInBuffer(name string, offset int64, len int64, data []by
...
+	}
+
+	// Upload the file, which will override the permissions and ACL
+	retCode := dl.BlockBlob.WriteFromFile(name, metadata, fi)
+
+	if err == nil && acl != "" {
+		// Cannot set both permissions and ACL in one call. ACL includes permission as well so just setting those back
+		// Just setting up the permissions will delete existing ACLs applied on the blob so do not convert this code to
+		// just set the permissions.
+		_, err := fileClient.SetAccessControl(context.Background(), &file.SetAccessControlOptions{
```
souravgupta-msft, line 573: How is this working as I don't see fileClient initialized?

component/azstorage/datalake.go
```diff
@@ -578,6 +605,18 @@ func (dl *Datalake) TruncateFile(name string, size int64) error {
...
+func (dl *Datalake) getACL(name string) (string, error) {
+	log.Trace("Datalake::getACL : Get ACLs for file %s", name)
+	fileClient := dl.Filesystem.NewFileClient(filepath.Join(dl.Config.prefixPath, name))
+	acl, err := fileClient.GetAccessControl(context.Background(), nil)
+	if err != nil {
+		log.Err("Datalake::getACL : Failed to get ACLs for file %s [%s]", name, err.Error())
+		return "", err
+	}
+
+	return *acl.ACL, nil
```
souravgupta-msft, line 617: This can result in panic in these cases,
1. Error is not nil and acl is nil
2. Error is nil, acl is not nil, but acl.ACL is nil

You can have a check something like,
`if acl != nil && acl.ACL != nil`

vibhansa-msft: This method can be removed as its needed only by test and not used in datalake.go code anymore. Will try to refactor the code here.

component/azstorage/datalake.go
```diff
@@ -553,7 +553,40 @@ func (dl *Datalake) ReadInBuffer(name string, offset int64, len int64, data []by
...
-	return dl.BlockBlob.WriteFromFile(name, metadata, fi)
+	// File in DataLake may have permissions and ACL set. Just uploading the file will override them.
+	// So, we need to get the existing permissions and ACL and set them back after uploading the file.
+
+	var acl string = ""
+	var fileClient *file.Client = nil
+
+	if dl.Config.preserveACL {
+		fileClient = dl.Filesystem.NewFileClient(filepath.Join(dl.Config.prefixPath, name))
+		resp, err := fileClient.GetAccessControl(context.Background(), nil)
```
jainakanksha-msft, line 564: File ACL should be present locally before upload, I am guessing.
Do we require to fetch it from storage(backend datalake)?

component/azstorage/datalake.go
```diff
@@ -553,7 +553,40 @@ func (dl *Datalake) ReadInBuffer(name string, offset int64, len int64, data []by
 
 // WriteFromFile : Upload local file to file
 func (dl *Datalake) WriteFromFile(name string, metadata map[string]*string, fi *os.File) (err error) {
-	return dl.BlockBlob.WriteFromFile(name, metadata, fi)
+	// File in DataLake may have permissions and ACL set. Just uploading the file will override them.
+	// So, we need to get the existing permissions and ACL and set them back after uploading the file.
+
+	var acl string = ""
```
jainakanksha-msft, line 559: Refactor the function 
and move setACL getACL in proper file


func (dl *Datalake) WriteFromFile(name string, metadata map[string]*string, fi *os.File) (err error) {
	var acl string
	var fileClient *file.Client

	// Only retrieve ACL if we need to preserve it
	if dl.Config.preserveACL {
		// Prepare the file client for the specific file
		fileClient = dl.Filesystem.NewFileClient(filepath.Join(dl.Config.prefixPath, name))

		// Get the current ACL
		if acl, err = getACL(fileClient); err != nil {
			log.Err("Datalake::getACL : Failed to get ACLs for file %s [%s]", name, err.Error())
		}
	}

	// Upload the file; this can overwrite any existing ACLs/permissions
	if err = dl.BlockBlob.WriteFromFile(name, metadata, fi); err != nil {
		return fmt.Errorf("failed to upload file: %v", err)
	}

	// If ACL was retrieved, restore it after the upload
	if acl != "" {
		if err = setACL(fileClient, acl); err != nil {
			log.Err("Datalake::WriteFromFile : Failed to set ACL for %s [%s]", name, err.Error())
		}
	}

	return nil
}

// Helper function to retrieve ACL
func getACL(fileClient *file.Client) (string, error) {
	resp, err := fileClient.GetAccessControl(context.Background(), nil)
	return *resp.ACL, err
}

// Helper function to set ACL
func setACL(fileClient *file.Client, acl string) error {
	_, err := fileClient.SetAccessControl(context.Background(), &file.SetAccessControlOptions{
		ACL: &acl,
	})
	return err
}

component/azstorage/datalake_test.go
```diff
@@ -2568,6 +2570,138 @@ func (s *datalakeTestSuite) TestUploadWithCPKEnabled() {
...
+	s.assert.Contains(acl, "user::rw-")
+	s.assert.Contains(acl, "other::---")
+
+	os.Remove(name + "_local")
+}
+
+func (s *datalakeTestSuite) TestPermissionPreservationWithFlag() {
+	defer s.cleanupTest()
+	// Setup
+	conf := fmt.Sprintf("azstorage:\n  preserve-acl: true\n  account-name: %s\n  endpoint: https://%s.dfs.core.windows.net/\n  type: adls\n  account-key: %s\n  mode: key\n  container: %s\n  fail-unsupported-op: true",
```
jainakanksha-msft, line 2621: All 3 tests having common sets of step, can we just refactor that part also.

setup/advancedConfig.yaml
```diff
@@ -150,6 +150,7 @@ azstorage:
   cpk-enabled: true|false <enable client provided key encryption>
   cpk-encryption-key: <customer provided base64-encoded AES-256 encryption key value>
   cpk-encryption-key-sha256:  <customer provided base64-encoded sha256 of the encryption key>
+  preserve-acl: true|false <preserve ACLs and Permissions set on file during updates>
```
jainakanksha-msft, line 153: Shouldn't we make it default behaviour, rather than rely over a flag.
We can discuss with Vishnu once, but In my opinion, it should be default behaviour, rather than relying over a flag.

component/azstorage/datalake.go
vibhansa-msft: Permissions are present in local like the "rwx" triplet but on blob you can have ACLs as well like giving permissions to a specific MSI/SPN/Identity. These are not replicated on local system and we do not cache them anywhere in blobfuse as well. This is the reason we need to get the latest ACLs and permissions from storage before we upload the file. Uploading a file using PutBlob is just resetting all the applied permissions and ACLs.
That being said, this has an extra cost of 2 additional REST calls. By default we were not doing this and customers who are ok with that may not want to take this additional cost. Hence we have given a flag to configure.

component/azstorage/datalake.go
vibhansa-msft: getACL is not used anywhere else hence did not create it a seperate method. Can do if that improves code readabiltiy.

component/azstorage/datalake_test.go
vibhansa-msft: all 3 tests have different config or differnet api set being called. In one we do upload file without the config flag, in second we do it with config flag and third one is using putblocklist  api. Let me know how you suggest to refactor those.

setup/advancedConfig.yaml
vibhansa-msft: By default we are not doing this in existing builds. If we enable by default then it means 2 additional REST calls for all customers hitting this flow which will degrade performance and cost mroe to customers. This shall hit only if customers really want to preserve the permissions. Ideally it shall always do but just to avoid extra cost we need the flag.

syeleti-msft: Correctness of the tool should be the priority imo.

souravgupta-msft: If we make this the default behavior, it will break many customers workflows who use OAuth with `Storage Blob Data Contributor` role. This is the RBAC role generally being used and it fails on SetAccessControl API. This API only works with `Storage Blob Data Owner` role.
Also, if some customers only care about perf and not ACLs, then making this the default behavior will also impact them.

syeleti-msft: SetAccessControl is also used while creating the file. 

now setting them when creating the file and droping them again while updating the file make sense?

jainakanksha-msft (approved):

component/azstorage/datalake_test.go
```diff
@@ -2568,6 +2571,127 @@ func (s *datalakeTestSuite) TestUploadWithCPKEnabled() {
...
+	f, err := os.OpenFile(name+"_local", os.O_RDWR, mode)
+	s.assert.Nil(err)
+
+	err = s.az.CopyFromFile(internal.CopyFromFileOptions{Name: name, File: f, Metadata: nil})
+	s.assert.Nil(err)
+
+	attr, err := s.az.GetAttr(internal.GetAttrOptions{Name: name})
+	s.assert.Nil(err)
+	s.assert.NotNil(attr)
+	s.assert.NotEqual(0764, attr.Mode)
```
jainakanksha-msft, line 2646: This check is incorrect

component/azstorage/datalake_test.go
```diff
@@ -2568,6 +2571,127 @@ func (s *datalakeTestSuite) TestUploadWithCPKEnabled() {
...
+	err = s.az.CopyFromFile(internal.CopyFromFileOptions{Name: name, File: f, Metadata: nil})
+	s.assert.Nil(err)
+	attr, err := s.az.GetAttr(internal.GetAttrOptions{Name: name})
+	s.assert.Nil(err)
+	s.assert.NotNil(attr)
+	s.assert.NotEqual(0764, attr.Mode)
+
+	acl, err := getACL(s.az.storage.(*Datalake), name)
+	s.assert.Nil(err)
+	s.assert.Contains(acl, "user::rw-")
```
ashruti-msft, line 2618: We can also assert group permissions to ensure they are correct

syeleti-msft (approved):

setup/advancedConfig.yaml
vibhansa-msft: Creation time we do it because user sends the permissions while creating a file and we need to save those to file. This path is for any upload/update. So for creation of a new file I do see that there is an overhead.

vibhansa-msft: Not able to come up with any way to skip this on upload as we never know during upload whether the file was created newly or its updating of an existing file.

PR #1580, opened 2024-11-28
bug in block cache open call

syeleti-msft:
## ✅ What
 
Current implementation of open file when opened in O_WRONLY mode truncates the file to zero.
While this mode - means we should have opened the file in read write mode, not truncate it.
So, This is incorrect behaviour.

## 👩‍🔬 How to validate

We don't see it in the normal scenario as write-back cache is on by default. All the open calls with O_WRONLY will be redirected to O_RDWR mode.
So, To simulate this issue, turn off the write-back cache while mounting with --disable-writeback-cache=true and then open file in O_WRONLY mode, you see that the file is truncated to size zero.


component/block_cache/block_cache.go
```diff
@@ -397,7 +397,7 @@ func (bc *BlockCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Han
 	log.Debug("BlockCache::OpenFile : Size of file handle.Size %v", handle.Size)
 	bc.prepareHandleForBlockCache(handle)
 
-	if options.Flags&os.O_TRUNC != 0 || (options.Flags&os.O_WRONLY != 0 && options.Flags&os.O_APPEND == 0) {
+	if options.Flags&os.O_TRUNC != 0 {
```
vibhansa-msft, line 400: Earlier we were not changing the open flags in libfuse later due to some other issues we encountered we started doing that (opening all files in RDWR mode). Hence this check is independent and not relying on what libfuse does. Tomorow if we change that logic we need to remember to correct this flag here. Also, this is just in open and one time call, so it does not have any impact on the system. I suggest we leave this here as is as some day we might correct the libfuse behavior of overriding the flags.

syeleti-msft: I agree while write back cache is enabled then kernel also make read requests for WRONLY file mode, as we made writeback caching default we are redirecting the writeonly mode to RDWR. but the case here is for the writeonly mode when writeback cache is disabled, The default behaviour is not to change the filesize to zero.

vibhansa-msft, line 400: We shall add a UT case around this validating that truncate clears the file contents while wronly does not.

jainakanksha-msft:
@syeleti-msft Can you please add the steps to validate this flow?

jainakanksha-msft:
@syeleti-msft Also tag the related issue with this PR.

component/block_cache/block_cache.go
syeleti-msft: Added, Please review it

syeleti-msft:
> @syeleti-msft Can you please add the steps to validate this flow?

Added in the description part of the PR

jainakanksha-msft (approved):

vibhansa-msft (approved):

vibhansa-msft (approved):

PR #1585, opened 2024-12-04
Xload prefetch initial code

souravgupta-msft:
## ✅ What
 
Initial code for prefetch feature in Xload component
 
## 🤔 Why
 
Download all the blobs on mount
 
## NOTE

Please note that this the first PR for prefetch. Below items are in TODO list and will be added in future PRs
- Stats collection
- Open, close, getAttr and read dir calls in xload component
- UT and E2E
- Retry mechanism

## Related Links
https://microsoft.sharepoint.com/:w:/t/AzureStorage/EYpVETudnXJNiYSiYDwRiR0BycWuL10UOK3g32eiHubEWQ?e=TqEFWP


component/azstorage/block_blob.go
```diff
@@ -1080,14 +1079,14 @@ func (bb *BlockBlob) TruncateFile(name string, size int64) error {
 			blobClient := bb.Container.NewBlockBlobClient(blobName)
 
 			blkList := make([]string, 0)
-			id := base64.StdEncoding.EncodeToString(common.NewUUIDWithLength(16))
+			id := common.GetBlockID(16)
```
dphulkar-msft, line 1082: Should we consider declaring `16` as a constant. This approach would make it easier to update the value in the future, as it would only need to be changed in one place. Additionally, assigning it to a well-named constant would enhance readability for developers.

component/xload/block.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("mmap error: %v", err)
+	}
+
+	block := &Block{
+		index:  0,
+		data:   addr,
+		offset: 0,
+		length: 0,
+		id:     "",
```
dphulkar-msft, line 68: Should we omit explicit assignments for default values (`index: 0`, `offset: 0`, etc.) as they are implicit in Go. This could make the code cleaner unless clarity is the goal.

component/xload/lister.go
```diff
@@ -0,0 +1,199 @@
...
+
+	marker := ""
+	var cnt, iteration int
+	for {
+		entries, new_marker, err := rl.getRemote().StreamDir(internal.StreamDirOptions{
+			Name:  absPath,
+			Token: marker,
+		})
+		if err != nil {
+			log.Err("remoteLister::process : Remote listing failed for %s [%s]", absPath, err.Error())
```
dphulkar-msft, line 146: If remote listing fails, should we continue from here?

component/xload/xload.go
```diff
@@ -0,0 +1,300 @@
...
+func (xl *Xload) Stop() error {
+	log.Trace("Xload::Stop : Stopping component %s", xl.Name())
+
+	xl.comps[0].stop()
+	xl.blockPool.Terminate()
+
+	// TODO:: xload : should we delete the files from local path
+	err := common.TempCacheCleanup(xl.path)
+	if err != nil {
+		log.Err("unable to clean xload local path [%s]", err.Error())
```
dphulkar-msft, line 225: Should we return an error from this function? If not, can we remove the return type entirely since the function currently only returns `nil`.

component/xload/xload.go
```diff
@@ -0,0 +1,300 @@
...
+	var err error
+
+	// Xload : start code goes here
+	switch xl.mode {
+	case xcommon.EMode.CHECKPOINT():
+		// Start checkpoint thread here
+		return fmt.Errorf("checkpoint is currently unsupported")
+	case xcommon.EMode.DOWNLOAD():
+		// Start downloader here
+		err = xl.startDownloader()
```
vibhansa-msft, line 196: We are doing "startDownloader" here and then at the end of the function we call "startComponents". As the later is actually starting the download part, this method shall be called something else as its preparing the pipeline for download and not actually tirggering any download at this point.

component/xload/xload.go
```diff
@@ -0,0 +1,300 @@
...
+	if xl.blockPool == nil {
+		log.Err("Xload::Start : Failed to create block pool")
+		return fmt.Errorf("failed to create block pool")
+	}
+
+	var err error
+
+	// Xload : start code goes here
+	switch xl.mode {
+	case xcommon.EMode.CHECKPOINT():
```
vibhansa-msft, line 191: Let's rename this to UPLOAD instead of CHECKPOINT

component/xload/xload.go
```diff
@@ -0,0 +1,300 @@
...
+	log.Trace("Xload::startDownloader : Starting downloader")
+
+	// Create remote lister pool to list remote files
+	rl, err := comp.NewRemoteLister(xl.path, xl.NextComponent())
+	if err != nil {
+		log.Err("Xload::startDownloader : Unable to create remote lister [%s]", err.Error())
+		return err
+	}
+
+	ds, err := comp.NewDownloadSplitter(xl.blockSize, xl.blockPool, xl.path, xl.NextComponent())
```
vibhansa-msft, line 241: If we provide blockpool as an object to the splitter, does it matter whether it knows the size of the block or not.

component/azstorage/block_blob.go
vibhansa-msft: Instead of passing down 16 here, we can put this as hard coded value int the GetBlockID method itself as we will use 16 for now for all components.

component/xload/xload.go
vibhansa-msft: I see there is a seperate UPLOAD as well, not sure what thoughts we had to put CHECKPOINT as a seperate mode.

component/xload/comp/lister.go
```diff
@@ -0,0 +1,198 @@
...
+	if rl.GetThreadPool() != nil {
+		rl.GetThreadPool().Stop()
+	}
+	rl.GetNext().Stop()
+}
+
+// wait for the configured block-list-on-mount-sec to make the list call
+func waitForListTimeout() error {
+	var blockListSeconds uint16 = 0
+	err := config.UnmarshalKey("azstorage.block-list-on-mount-sec", &blockListSeconds)
```
vibhansa-msft, line 112: Why do we want to wait here for this timeout. Ideally customer going for prefetch shall not have any wait timeout given,

vibhansa-msft: If the 10 second timeout is default, then on enabling this prefetch feature that shall get disabled as whole idea is to download all the contents as fast as possible so waiting for 10 seconds to unblock does not make sense here.

component/xload/comp/lister.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (rl *remoteLister) Process(item *common.WorkItem) (int, error) {
+	absPath := item.Path // TODO:: xload : check this for subdirectory mounting
+
+	log.Debug("remoteLister::Process : Reading remote dir %s", absPath)
+
+	// this block will be executed only in the first list call for the remote directory
+	// so haven't made the listBlocked variable atomic
+	if !rl.listBlocked {
```
vibhansa-msft, line 127: Same as above comment, we shall not wait here for anything.

component/xload/comp/lister.go
```diff
@@ -0,0 +1,198 @@
...
+			break
+		}
+	}
+
+	return cnt, nil
+}
+
+func (rl *remoteLister) mkdir(name string) error {
+	log.Debug("remoteLister::mkdir : Creating local path: %s", name)
+	return os.MkdirAll(name, 0777)
```
vibhansa-msft, line 197: If we are doing only mkdir in this function than better to remove the function call overhead and do it in the process method only.

component/xload/comp/lister.go
```diff
@@ -0,0 +1,198 @@
...
+
+					// push the directory to input pool for its listing
+					rl.GetThreadPool().Schedule(&common.WorkItem{
+						CompName: rl.GetName(),
+						Path:     name,
+					})
+				}(entry.Path)
+			} else {
+				// send file to the output channel for chunking
+				rl.GetNext().GetThreadPool().Schedule(&common.WorkItem{
```
vibhansa-msft, line 178: Instead of getting the threadpool object and then call schedule on that ideally the sub-component shall have a schedule method which inserts the item in its own threadpool so that external sub module is not aware whether the next component has multiple threads or works in single threaded model. Here we are mandating that next component shall always have a threadpool.

component/xload/comp/lister.go
```diff
@@ -0,0 +1,198 @@
...
+					localPath := filepath.Join(rl.path, name)
+					err = rl.mkdir(localPath)
+					// TODO:: xload : handle error
+					if err != nil {
+						log.Err("remoteLister::Process : Failed to create directory [%s]", err.Error())
+						return
+					}
+
+					// push the directory to input pool for its listing
+					rl.GetThreadPool().Schedule(&common.WorkItem{
```
vibhansa-msft, line 171: Component pushing back into self queue shall have a check on the channel size. If channel is full its going to hand up and then we may end up having many threads in this state. Better to have a log if the channel is already full so that if we hit any such issue on customer setup its easier to debug.

component/xload/comp/splitter.go
```diff
@@ -0,0 +1,201 @@
...
+			}
+			log.Debug("downloadSplitter::Process : Scheduling download for %s offset %v", item.Path, offset)
+			d.GetNext().GetThreadPool().Schedule(splitItem)
+		}
+
+		offset += int64(d.blockSize)
+	}
+
+	wg.Wait()
+	err = item.FileHandle.Truncate(int64(item.DataLen))
```
vibhansa-msft, line 180: We wrote only what we downloaded so do we really need to truncate the file to size after download is complete?

component/xload/comp/splitter.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
+		log.Err("downloadSplitter::Process : Failed to truncate file %s [%s]", item.Path, err.Error())
+		operationSuccess = false
+	}
+
+	if !operationSuccess {
```
vibhansa-msft, line 186: Shall we update the last modified time of the local file to the one mentioned in container? Not sure but applications might take decisions based on that.

component/xload/comp/data_manager.go
```diff
@@ -0,0 +1,120 @@
...
+	if item.Download {
+		return rdm.ReadData(item)
+	} else {
+		return rdm.WriteData(item)
+	}
+}
+
+// ReadData reads data from the data manager
+func (rdm *remoteDataManager) ReadData(item *common.WorkItem) (int, error) {
+	log.Debug("remoteDataManager::ReadData : Scheduling download for %s offset %v", item.Path, item.Block.Offset)
```
vibhansa-msft, line 100: Where do we need to read data from handle, as we are just downloading and writing data to local disk,

component/xload/comp/data_manager.go
```diff
@@ -0,0 +1,120 @@
...
+
+	h := handlemap.NewHandle(item.Path)
+	h.Size = int64(item.DataLen)
+	return rdm.GetRemote().ReadInBuffer(internal.ReadInBufferOptions{
```
vibhansa-msft, line 104: Where do we put the responce back to channel from where splitter is reading.

component/xload/xload.go
```diff
@@ -0,0 +1,300 @@
...
+// Pipeline will call this method to create your object, initialize your variables here
+func NewXloadComponent() internal.Component {
+	comp := &Xload{}
+	comp.SetName(compName)
+	return comp
+}
+
+// On init register this component to pipeline and supply your constructor
+func init() {
+	internal.AddComponent(compName, NewXloadComponent)
```
vibhansa-msft, line 299: We shall map the path parameter to tmp-path cli param that file-cache uses so that we dont need to introduce any new cli parameter.

cmd/imports.go
```diff
@@ -42,4 +42,5 @@ import (
 	_ "github.com/Azure/azure-storage-fuse/v2/component/file_cache"
 	_ "github.com/Azure/azure-storage-fuse/v2/component/libfuse"
 	_ "github.com/Azure/azure-storage-fuse/v2/component/loopback"
+	_ "github.com/Azure/azure-storage-fuse/v2/component/xload"
```
vibhansa-msft, line 45: Based on a cli, the component shall be auto inserted in the pipeline. Check mount.go for block-cache addition. If user give "--preload" parameter then this mode shall get enabled and we shall auto create pipeline for prefetching.

component/xload/xload.go
```diff
@@ -0,0 +1,300 @@
...
+		}
+
+		if !common.IsDirectoryEmpty(xl.path) {
+			log.Err("Xload::Configure : config error %s directory is not empty", xl.path)
+			return fmt.Errorf("config error in %s [temp directory not empty]", xl.Name())
+		}
+	}
+
+	var mode xcommon.Mode
+	err = mode.Parse(conf.Mode)
```
vibhansa-msft, line 160: Mode is a mandatory parameter as of now. As we are only having download feature for now the default value shall be preload related mode.

component/xload/xload.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	var err error
+
+	// Xload : start code goes here
+	switch xl.mode {
+	case xcommon.EMode.CHECKPOINT():
+		// Start checkpoint thread here
+		return fmt.Errorf("checkpoint is currently unsupported")
+	case xcommon.EMode.DOWNLOAD():
```
vibhansa-msft, line 194: We shall call this mode as 'preload', as that's what we are projecting as a feature.

component/xload/comp/splitter.go
```diff
@@ -0,0 +1,201 @@
...
+// download data in chunks and then write to the local file
+func (d *downloadSplitter) Process(item *common.WorkItem) (int, error) {
+	var err error
+
+	log.Debug("downloadSplitter::Process : Splitting data for %s", item.Path)
+	if len(item.Path) == 0 {
+		return 0, nil
+	}
+
+	numBlocks := ((item.DataLen - 1) / d.blockSize) + 1
```
vibhansa-msft, line 112: When data len is 0, this is overrunning the value into a very large number of numBlocks.

component/xload/comp/splitter.go
```diff
@@ -0,0 +1,201 @@
...
+	offset := int64(0)
+
+	// TODO:: xload : should we delete the file if it already exists
+	// TODO:: xload : what should be the flags and mode and should we allocate the full size to the file
+	item.FileHandle, err = os.OpenFile(filepath.Join(d.path, item.Path), os.O_WRONLY|os.O_CREATE, 0644)
```
vibhansa-msft, line 117: For HNS case premission of blob shall be preserved in local file as well.

component/xload/comp/splitter.go
```diff
@@ -0,0 +1,201 @@
...
+	if err != nil {
```
vibhansa-msft, line 118: If blob is a symlink we might need to handle it differently here. Let's add a task for future about this.

component/xload/common/block.go
```diff
@@ -0,0 +1,96 @@
...
+type Block struct {
+	Index  int    // Index of the block in the pool
+	Offset int64  // Start offset of the data this block holds
+	Length int64  // Length of data that this block holds
+	Id     string // ID to represent this block in the blob
+	Data   []byte // Data this block holds
+}
+
+// AllocateBlock creates a new memory mapped buffer for the given size
+func AllocateBlock(size uint64) (*Block, error) {
```
vibhansa-msft, line 51: Block, Blockpool and Threadpool are used by block-cache as well, can we generalize the code so that we have only one copy of these common utils,

component/xload/comp/data_manager.go
```diff
@@ -0,0 +1,120 @@
...
+	} else {
+		return rdm.WriteData(item)
+	}
+}
+
+// ReadData reads data from the data manager
+func (rdm *remoteDataManager) ReadData(item *common.WorkItem) (int, error) {
+	log.Debug("remoteDataManager::ReadData : Scheduling download for %s offset %v", item.Path, item.Block.Offset)
+
+	h := handlemap.NewHandle(item.Path)
```
vibhansa-msft, line 102: We shall not create dummy handle here, instead pass the path, file-size as input parameter in option bag

component/azstorage/block_blob.go
souravgupta-msft: Create a constant with value as 16 instead of hard coding it in the method. The block id length may not always be 16. For example, in the truncate flow if we are increasing the size of the blob, we use the existing block id lengths in the blob.

component/xload/block.go
souravgupta-msft: Removed, relying on the default values in GO.

component/xload/xload.go
souravgupta-msft: Returning error here

component/xload/xload.go
souravgupta-msft: Renamed this method to `createDownloader`.

component/xload/comp/splitter.go
souravgupta-msft, line 117: Truncate the file to its actual size post creation so that if the disk is full, it fails here

component/xload/xload.go
souravgupta-msft: Removed checkpoint mode

component/xload/xload.go
souravgupta-msft: Removed the block size parameter and reused the one set in block pool

component/xload/comp/lister.go
souravgupta-msft: The wait logic is for cases when user has set `block-list-on-mount-sec` parameter in the config. The default value of this parameter is 0. So, customers won't be impacted if they don't set this parameter.
Another reason for adding this is because we call the List() method in azstorage component, which waits till this timeout before making the REST call.

component/xload/comp/lister.go
souravgupta-msft: Just trying to separate the flow functionality wise.

component/xload/comp/lister.go
souravgupta-msft: Removed dependency on GetThreadPool method. Added schedule method in subcomponent level so that if it does not have a thread pool, it will directly call its process method.

component/xload/xload.go
souravgupta-msft: Done, renamed this mode to `preload`

component/xload/comp/splitter.go
souravgupta-msft: Added the check when dataLen is 0

component/xload/comp/lister.go
souravgupta-msft: Also having this as a separate function, so that when the support for upload is added, this method will have logic to create directory in the storage.

component/xload/xload.go
souravgupta-msft: Setting default mode to "preload"

component/xload/comp/splitter.go
souravgupta-msft: Removed the truncate operation post download completion. Truncate will be done at the starting when file is created.

component/xload/comp/data_manager.go
souravgupta-msft: Done, added `Path` and `Size` parameters in `ReadInBufferOptions` so that we don't need to create a dummy handle.

component/xload/comp/splitter.go
souravgupta-msft: Done, doing truncate after file creation and before the download starts. If the disk is full, truncate operation will fail and return back, and the download of the file will not start.

component/xload/comp/splitter.go
souravgupta-msft: Created ADO workitem and TODO item in the code

component/xload/comp/lister.go
vibhansa-msft: Ideally customers using prefetch shall not have this wait time configured so it shall be safe to assume its 0 and not wait here. We can discuss this with the team,

vibhansa-msft (approved):

component/xload/comp/lister.go
souravgupta-msft: Yes, the default value is 0. So, if not configured, users won't have to wait for this timeout

component/xload/xload.go
souravgupta-msft: Done, reusing the "tmp-path" and "block-size-mb" cli flags for the xload config.

cmd/imports.go
souravgupta-msft: Added "--preload" flag to add xload to pipeline

component/xload/comp/splitter.go
souravgupta-msft: Done

component/xload/comp/splitter.go
```diff
@@ -197,6 +197,13 @@ func (d *downloadSplitter) Process(item *common.WorkItem) (int, error) {
 
 	wg.Wait()
 
+	// update the last modified time
+	err = os.Chtimes(localPath, item.Atime, item.Mtime)
+	if err != nil {
+		log.Err("downloadSplitter::Process : Failed to change times of file %s [%s]", item.Path, err.Error())
+		operationSuccess = false
```
vibhansa-msft, line 206: Should we fail the preload if we fail to set the mod time, if user does not have permission it might fail so shall we just log and move on ?

souravgupta-msft: Done

component/xload/xload.go
```diff
@@ -0,0 +1,332 @@
...
+	return internal.EComponentPriority.LevelMid()
+}
+
+// Configure : Pipeline will call this method after constructor so that you can read config and initialize yourself
+func (xl *Xload) Configure(_ bool) error {
+	log.Trace("Xload::Configure : %s", xl.Name())
+
+	// xload component should be used only in readonly mode
+	var readonly bool
+	err := config.UnmarshalKey("read-only", &readonly)
```
jainakanksha-msft, line 102: type XloadConfig struct {
	ReadOnly   bool    `yaml:"read-only" validate:"required"`
	BlockSize  float64 `yaml:"block-size-mb" validate:"gt=0"`
	Path       string  `yaml:"path" validate:"required"`
	Mode       string  `yaml:"mode"`
	AllowOther bool    `yaml:"allow-other"`
	MountPath  string  `yaml:"mount-path"`
}

var conf XloadConfig
err := config.UnmarshalKey(xl.Name(), &conf)

Use single Unmarshalling to import all the config

jainakanksha-msft: Use 
"github.com/go-playground/validator/v10"
"gopkg.in/yaml.v2" 
for yaml transformation and validation purpose

jainakanksha-msft: with that you can do basic validation like this

`

func LoadConfig() error {
      var conf XloadConfig
	err := LoadConfig(filename , &conf)
	if err != nil {
		log.Fatalf("Error loading config: %v", err)
	}

	// If no error, proceed with configuration
	fmt.Println("Config loaded and validated:", conf)

	// Read the YAML file
	file, err := os.Open(filename)
	if err != nil {
		return fmt.Errorf("failed to open config file: %v", err)
	}
	defer file.Close()

	// Unmarshal the YAML into the struct
	decoder := yaml.NewDecoder(file)
	err = decoder.Decode(config)
	if err != nil {
		return fmt.Errorf("failed to unmarshal YAML: %v", err)
	}

	// Validate the struct using go-playground validator
	validate := validator.New()
	err = validate.Struct(config)
	if err != nil {
		return fmt.Errorf("validation error: %v", err)
	}

	return nil
// add remaining things

}

`

component/azstorage/azstorage.go
```diff
@@ -440,22 +441,36 @@ func (az *AzStorage) ReadFile(options internal.ReadFileOptions) (data []byte, er
...
 
-	if options.Offset > atomic.LoadInt64(&options.Handle.Size) {
+	var size int64
+	var path string
+	if options.Handle != nil {
+		size = atomic.LoadInt64(&options.Handle.Size)
+		path = options.Handle.Path
+	} else {
+		size = options.Size
+		path = strings.TrimSpace(options.Path)
```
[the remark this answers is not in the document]
souravgupta-msft: Done

component/azstorage/azstorage.go
```diff
@@ -440,22 +441,36 @@ func (az *AzStorage) ReadFile(options internal.ReadFileOptions) (data []byte, er
 func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (length int, err error) {
 	//log.Trace("AzStorage::ReadInBuffer : Read %s from %d offset", h.Path, offset)
 
-	if options.Offset > atomic.LoadInt64(&options.Handle.Size) {
+	var size int64
```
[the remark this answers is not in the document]
souravgupta-msft: Done

component/xload/common/xcomponent.go
```diff
@@ -0,0 +1,124 @@
...
+
+func (xb *XBase) Process(item *WorkItem) (int, error) {
+	return 0, nil
+}
+
+func (xb *XBase) GetNext() XComponent {
+	return xb.next
+}
+
+func (xb *XBase) SetNext(s XComponent) {
```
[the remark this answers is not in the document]
souravgupta-msft: Done

component/xload/common/utils.go
```diff
@@ -0,0 +1,96 @@
+/*
```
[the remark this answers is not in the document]
souravgupta-msft: Done

component/xload/comp/lister.go
```diff
@@ -0,0 +1,209 @@
...
+	var blockListSeconds uint16 = 0
+	err := config.UnmarshalKey("azstorage.block-list-on-mount-sec", &blockListSeconds)
+	if err != nil {
+		return err
+	}
+	time.Sleep(time.Duration(blockListSeconds) * time.Second)
+	return nil
+}
+
+func (rl *remoteLister) Process(item *common.WorkItem) (int, error) {
```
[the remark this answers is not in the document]
souravgupta-msft: Will do in future iteration

component/xload/xload.go
```diff
@@ -0,0 +1,332 @@
...
+
+// Structure defining your config parameters
+type XloadOptions struct {
+	BlockSize float64 `config:"block-size-mb" yaml:"block-size-mb,omitempty"`
+	Mode      string  `config:"mode" yaml:"mode,omitempty"`
+	Path      string  `config:"path" yaml:"path,omitempty"`
+	// TODO:: xload : add parallelism parameter
+}
+
+const (
```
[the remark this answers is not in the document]
souravgupta-msft: Done

component/xload/comp/splitter.go
```diff
@@ -0,0 +1,222 @@
...
+func (d *downloadSplitter) Stop() {
+	log.Debug("downloadSplitter::Stop : stop download splitter for %s", d.path)
+	if d.GetThreadPool() != nil {
+		d.GetThreadPool().Stop()
+	}
+	d.GetNext().Stop()
+}
+
+// download data in chunks and then write to the local file
+func (d *downloadSplitter) Process(item *common.WorkItem) (int, error) {
```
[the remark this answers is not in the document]
souravgupta-msft: Will do in future iteration

component/xload/xload.go
souravgupta-msft: Discussed offline, changeset will be very large. Not feasible for current iteration

PR #1595, opened 2024-12-19
Feature: Blob filter

vibhansa-msft:
## ✅ What
 
Blob filter integration to filter out blob based on various parameters.
Filter library can be found [here](https://github.com/vibhansa-msft/blobFilter/blob/main/README.md)
 
## 🤔 Why
 
When user mounts blobfuse, all blob are visible and user is free to work on any blob. There are scenarios where customer do not want to have visibility of all the blobs that exists in container and rather have a constrained view based on blob names, size, LMT etc. This feature will allow that limited scope to be created.
 
Mount with filter option and check all the blobs are visible or only the one matching the given filter

## 🔖 Related links
 
- [Filter Library code base](https://github.com/vibhansa-msft/blobFilter)
- [How to use filter lib](https://github.com/vibhansa-msft/blobFilter/blob/main/filter_test.go)


component/azstorage/utils.go
```diff
@@ -596,3 +596,21 @@ func removeLeadingSlashes(s string) string {
 	}
 	return s
 }
+
```
vibhansa-msft, line 599: Tag filter is not there right now, but later at some point we will add it hence keeping this code as commented for now.

component/azstorage/config.go
```diff
@@ -517,6 +525,25 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 	return nil
 }
 
+func configureBlobFilter(az *AzStorage, opt AzStorageOptions) error {
```
jainakanksha-msft, line 528: Kindly rename az and opt with proper names

component/azstorage/datalake.go
```diff
@@ -421,6 +422,17 @@ func (dl *Datalake) GetAttr(name string) (attr *internal.ObjAttr, err error) {
 		}
 	}
 
+	if dl.Config.filter != nil {
```
jainakanksha-msft, line 425: Shouldn't we filter out the file before even reaching to GetAttr for particular file?
I understand for filtering we will require the properties.
But filtering out inside GetAttribute looks incorrect.

Let's discuss quickly for this

component/azstorage/utils.go
```diff
@@ -596,3 +596,21 @@ func removeLeadingSlashes(s string) string {
...
+// func parseBlobTags(tags *container.BlobTags) map[string]string {
+
+// 	if tags == nil {
+// 		return nil
```
jainakanksha-msft, line 603: Should we have to return nil or empty map that is blobtags?

component/azstorage/block_blob.go
```diff
@@ -595,19 +609,32 @@ func (bb *BlockBlob) processBlobItems(blobItems []*container.BlobItem) ([]*inter
...
 			dirList[*blobInfo.Name+"/"] = true
 			attr.Size = 4096
 		}
+
+		if bb.Config.filter != nil && !attr.IsDir() {
+			filterAttr.Name = attr.Name
+			filterAttr.Mtime = attr.Mtime
+			filterAttr.Size = attr.Size
+			if bb.Config.filter.IsFileAcceptable(&filterAttr) {
+				blobList = append(blobList, attr)
```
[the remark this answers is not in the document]
vibhansa-msft: Library does not care whether attributes are coming from a file or a blob it just takes some parameters as input and filters on basis of that. As the name, type, size, lmt are common file parameters API is created as "IsFileAcceptible"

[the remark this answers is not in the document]
vibhansa-msft: as API filters based on file attributes, the variable is name fileAttr while 'attr' just means the blob attributes as component is azStorage. If there is any confusion we can think of better naming.

jainakanksha-msft (approved):

jainakanksha-msft (approved):

ashruti-msft (approved):

vibhansa-msft:
> Should we support filtering for directories for HNS accounts as well? Right now, filtering is only applied on blob items and all directories are visible which doesn't entirely restrict view as per user's request.

We are not filtering directories because directories do not have tags, types, size and lmt populated so only name based filter can work on them. User might give name based filter only on the matching blob names hence I ignored the directories from the filtering logic.

souravgupta-msft (approved):

PR #1604, opened 2025-01-06
Add strong consistency check for data on disk

vibhansa-msft:
## ✅ What
Data being downloaded by block-cache is saved to local disk, if user choses to enable persistence. This feature is useful if application is reading the same files multiple times and we do not have enough in memory cache to hold all of them. When a block is downloaded it's saved on local disk. There are chances of accidental overwrite on the disk file which might corrupt the data. Later when same block is read by application Blobfuse will load the data from this disk file and end up serving corrupt data to user application.

Having a CRC check on the data when it was downloaded, protects us from those accidental corruptions.
 
## 🤔 Why
To provide strong data consistency checks and protect data persisted on local disk.
 
 If user choses to persist the downloaded content then this additional flag can provide and extra layer of security.
 
## 🔖 Related links
 NA


README.md
```diff
@@ -166,6 +166,8 @@ To learn about a specific command, just include the name of the command (For exa
     * `--block-cache-prefetch=<Number of blocks>`: Number of blocks to prefetch at max when sequential reads are in progress. Default - 2 times number of CPU cores.
     * `--block-cache-parallelism=<count>`: Number of parallel threads doing upload/download operation. Default - 3 times number of CPU cores.
     * `--block-cache-prefetch-on-open=true`: Start prefetching on open system call instead of waiting for first read. Enhances perf if file is read sequentially from offset 0.
+    * `--block-cache-strong-consistency=true`: Enable strong data consistency checks in block-cache. This will increase load on your CPU and may introduce some latency. 
+    This will need support of `xattr` on your system. Kindly install the feature manually before using this cli parameter.
```
syeleti-msft, line 170: 1. Why are we using xattr, why can't we put the computed hash in memory, What is the logic behind saving the hash again on disk using xattr, if we are checking the consistency of the disk.
2. Why are we using MD5sum, is CRC not enough to serve the job?

jainakanksha-msft:
@vibhansa-msft Please add the whole use case to validate the scenario

component/block_cache/block_cache.go
```diff
@@ -997,11 +1003,35 @@ func (bc *BlockCache) download(item *workItem) {
...
 
 				f.Close()
-				// We have read the data from disk so there is no need to go over network
-				// Just mark the block that download is complete
+
 				if successfulRead {
-					item.block.Ready(BlockStatusDownloaded)
-					return
+					// If user has enabled consistency check then compute the md5sum and match it in xattr
+					if bc.consistency {
```
jainakanksha-msft, line 1009: extract this md5Validation into separate method.
we will require to write the test case for that validation only. No extra steps needed.

README.md
vibhansa-msft: If there are lot of files being worked on in parallel, we might end up bloating up our memory usage. Disk files are created for each block and not per blob, so one blob can end up having 5000 blocks at max to keep them in memory will need map add/delete and searches on every disk read. Just wanted to save on the memory front by saving the data on disk itself.

component/block_cache/block_cache.go
vibhansa-msft: Did not put this in a separate function intentionally as we need to pass down the data, size and local file path to that function and only thing it does is to extract the sum and compare. Felt overhead of sending down all the items to another function will have more impact as this is a trivial task.

README.md
vibhansa-msft: CRC32 is faster but it can not detect all kind of errors. CRC64 and MD5Sum are similar in terms of performance. Does it make any difference on what algo we use to compute the hash?

component/block_cache/block_cache.go
jainakanksha-msft: This validation is standalone. can be called as separate validation altogether. 
kind of Part of a Validation Layer.
If the flag is not true. things will not go for validation.
That's why rather than adding it here suggestion is to separate it out.

README.md
vibhansa-msft: Migrated code to CRC64 instead of md5 sum.

ashruti-msft (approved):

jainakanksha-msft:
@vibhansa-msft , Do we need to update any public documentation for this new argument?
 Additionally, do we need to verify our default config generation script
 or any update in E2E test cases?

vibhansa-msft:
> @vibhansa-msft , Do we need to update any public documentation for this new argument? Additionally, do we need to verify our default config generation script or any update in E2E test cases?

ETAG validation can not be done in UT as we do not hit actual storage in UT. In E2E we can try but for that we need multipe handles to read-write which we do not support at the time with block-cache directly. I have manually tested this so far.

jainakanksha-msft (approved):

PR #1608, opened 2025-01-15
Enable ETAG based validation on every block download to provide higher consistency

vibhansa-msft:
## ✅ What
 
In block-cache mode, every download is not validated for higher consistency. If a file is open or half-read and contents are modified on the container, subsequent reads will start fetching new content and provide inconsistent data to application. Adding an ETAG based validation will guard applications against such data corruption. When an 'open' file call comes ETAG from blob will be preserved by block-cache. Any subsequent block download will revalidate that ETAG is still the same. If ETAG has changed, then it means blob was modified on container after `open` and in such case download and read operation will fail.
 
## 🤔 Why
 
This will safeguard application from data corruption where blob is modified on container through other means.
 
Open a file and then modify the file on container followed by a read operation on mounted path.
 
## 🔖 Related links
 
NA


component/azstorage/block_blob.go
```diff
@@ -918,6 +921,56 @@ func (bb *BlockBlob) ReadInBuffer(name string, offset int64, len int64, data []b
...
+
+	opt := &blob.DownloadStreamOptions{
+		Range: blob.HTTPRange{
+			Offset: offset,
+			Count:  len,
+		},
+		CPKInfo: bb.blobCPKOpt,
+	}
+
+	dr, err := blobClient.DownloadStream(ctx, opt)
```
souravgupta-msft, line 939: Use DownloadBuffer() method here. DownloadStream() is basically a single GetBlob call which will take a lot of time if the buffer is huge.
Also, with DownloadBuffer(), you can remove the below code of copying the data to the buffer.

souravgupta-msft: If the objective to get ETag, then we can make an extra GetProperties call at the end. Since, this feature will be enabled by a flag, we should have DownloadBuffer() + GetProperties() call, which will generally be more performant than DownloadStream().

component/azstorage/azstorage.go
```diff
@@ -453,9 +453,17 @@ func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (length
 		return 0, nil
 	}
 
-	err = az.storage.ReadInBuffer(options.Handle.Path, options.Offset, dataLen, options.Data)
-	if err != nil {
-		log.Err("AzStorage::ReadInBuffer : Failed to read %s [%s]", options.Handle.Path, err.Error())
+	if options.Etag != nil {
```
souravgupta-msft, line 456: Please add tests for this change.

component/azstorage/azstorage.go
```diff
@@ -453,9 +453,17 @@ func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (length
...
+		etag, err := az.storage.ReadInBufferWithETag(options.Handle.Path, options.Offset, dataLen, options.Data)
+		if err != nil {
+			log.Err("AzStorage::ReadInBuffer : Failed to read %s [%s]", options.Handle.Path, err.Error())
+		}
+		*(options.Etag) = etag
```
souravgupta-msft, line 461: You can also use this,
`options.Etag = to.Ptr(etag)`

component/azstorage/block_blob.go
vibhansa-msft: downloadBuffer that we are using is internally using download stream only and then copies the data over to our buffer. Reason to use the stream directly is to get the etag. In downloadstream we have the responce but in download buffer we just return back data len and error so that actual responce is trimmed. If we use downloadStream directly then there is no need to make an extra getProperties call.

component/azstorage/azstorage.go
vibhansa-msft: No I do not want the pointer to etag, rather I have sent a pointer and I want to copy the value of etag to my variable.

component/azstorage/azstorage.go
vibhansa-msft: done

ashruti-msft:
Later we should run some perf tests with data consistency on to inform users what % diff in performance, memory usage they can expect with this option on and off.

ashruti-msft (approved):

component/azstorage/azstorage.go
jainakanksha-msft: I think UT test is not pushed yet.
Can you please check once.
If it is pushed, could you please point us.?

component/azstorage/block_blob.go
jainakanksha-msft, line 939: Please rename dr

component/azstorage/azstorage.go
vibhansa-msft: done now

component/azstorage/azstorage.go
```diff
@@ -453,9 +453,17 @@ func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (length
...
+	} else {
+		err = az.storage.ReadInBuffer(options.Handle.Path, options.Offset, dataLen, options.Data)
+		if err != nil {
```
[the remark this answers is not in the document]
vibhansa-msft: done

component/azstorage/block_blob.go
vibhansa-msft: done

component/azstorage/block_blob.go
```diff
@@ -918,6 +921,56 @@ func (bb *BlockBlob) ReadInBuffer(name string, offset int64, len int64, data []b
 	return nil
 }
 
+// ReadInBufferWithETag : Download specific range from a file to a user provided buffer, return back the current etag along with data
+func (bb *BlockBlob) ReadInBufferWithETag(name string, offset int64, len int64, data []byte) (string, error) {
```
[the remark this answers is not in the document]
vibhansa-msft: Across Blobfuse object name is only initials of component/class only, so this is in line with our coding standards across the code.

component/azstorage/block_blob.go
```diff
@@ -918,6 +921,56 @@ func (bb *BlockBlob) ReadInBuffer(name string, offset int64, len int64, data []b
...
+		}
+
+		log.Err("BlockBlob::ReadInBuffer : Failed to download blob %s [%s]", name, err.Error())
+		return "", err
+	}
+
+	var streamBody io.ReadCloser = dr.NewRetryReader(ctx, nil)
+	dataRead, err := io.ReadFull(streamBody, data)
+
+	if err != nil && err != io.EOF && err != io.ErrUnexpectedEOF {
```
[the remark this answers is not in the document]
vibhansa-msft: EOF is not an error, it just says you have read everything that exists. So we dont treat EOF like an error.

component/block_cache/block_cache.go
```diff
@@ -952,31 +956,31 @@ func (bc *BlockCache) lineupDownload(handle *handlemap.Handle, block *Block, pre
 }
 
 // download : Method to download the given amount of data
-func (blockCache *BlockCache) download(item *workItem) {
+func (bc *BlockCache) download(item *workItem) {
```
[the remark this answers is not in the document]
vibhansa-msft: Its uniform across all functions and all files.

component/azstorage/block_blob.go
```diff
@@ -918,6 +921,56 @@ func (bb *BlockBlob) ReadInBuffer(name string, offset int64, len int64, data []b
...
+			Offset: offset,
+			Count:  len,
+		},
+		CPKInfo: bb.blobCPKOpt,
+	}
+
+	dr, err := blobClient.DownloadStream(ctx, opt)
+
+	if err != nil {
+		e := storeBlobErrToErr(err)
```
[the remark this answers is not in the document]
vibhansa-msft: done

jainakanksha-msft (approved):

component/azstorage/block_blob.go
```diff
@@ -918,6 +921,56 @@ func (bb *BlockBlob) ReadInBuffer(name string, offset int64, len int64, data []b
...
+
+	opt := &blob.DownloadStreamOptions{
+		Range: blob.HTTPRange{
+			Offset: offset,
+			Count:  len,
+		},
+		CPKInfo: bb.blobCPKOpt,
+	}
+
+	downloadResponse, err := blobClient.DownloadStream(ctx, opt)
```
syeleti-msft, line 939: why not use DownloadStream inside ReadinBuffer as anyways we are not using concurrency of DonwloadBuffer in go SDK. Doing this would make the Code much more readable and also don't need to implement these new functions.

CHANGELOG.md
```diff
@@ -10,7 +10,8 @@
 
 **Features**
 - Mount container or directory but restrict the view of blobs that you can see. This feature is available only in read-only mount.
-- To protect against accidental overwrites on data stored by block-cache on temp path, md5 sums will be validated on read. This feature can be enabled by using `--block-cache-strong-consistency` cli flag.
+- To protect against accidental overwrites on data stored by block-cache on temp path, crc64 hash will be validated on read. This feature can be enabled by using `--block-cache-strong-consistency` cli flag.
```
syeleti-msft, line 13: Why to put this behind a flag, if the feature is not imposing any extra REST calls. Why can't it be default

CHANGELOG.md
```diff
@@ -10,7 +10,8 @@
...
+- To provide strong consistency check, ETAG of the file will be preserved on open. For any subsequent block download, with block-cache, ETAG will be verified and if the blob has changed in container the download will be declare failure resulting into read failure. This feature will be enabled with above cli option. Disable `attr_cache` when you want this feature to work with latest contents updated on blob.
```
syeleti-msft, line 14: Why should this etag thing must be combined with block-cache-consistency? as this is not imposing any extra REST calls, this can be the default behavior imo

component/block_cache/block_cache.go
```diff
@@ -400,6 +400,10 @@ func (bc *BlockCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Han
 	handle.Mtime = attr.Mtime
 	handle.Size = attr.Size
 
+	if bc.consistency && attr.ETag != "" {
+		handle.SetValue("ETAG", attr.ETag)
```
syeleti-msft, line 404: It is ok that we are putting etag here. But one must also ensure to update the etag when we update the blob. for example if we do putblocklist when flush, etag would change which should be reflected here. To simulate this testcase you can clone the fd and then close one fd and read from another which would fail the read operation by the current implemenation of this feature but in general this shouldn't be the case.

component/azstorage/block_blob.go
vibhansa-msft: ReadInBuffer internally calls DownloadStream but it does not return back the responce that backend sends and instead sends only size and error. Here we need the responce back to parse some fields from it. Whatever DownloadBuffer was doing we are doing exactly the same. In face DownloadBuffer was trying to split the download in parallel routines which anyway block-cache does not need so there are overheads in DownloadBuffer that we can avoid using stream method directly.

CHANGELOG.md
vibhansa-msft: It's not about extra call, a flag saying strong-consistency is just provided so that customer can choose if they want all these features or they are fine who the current version is running. If we make this feature enable by default it may start failing some existing customer flows. Also, some of these enhanced integrity checks have their own cost so we want customer to choose if they want to incur this cost or not.

component/block_cache/block_cache.go
vibhansa-msft: Yes that workflow makes sense, let me see what we can do here to support that.

component/azstorage/block_blob.go
syeleti-msft: > why not use DownloadStream inside ReadinBuffer as anyways we are not using concurrency of DonwloadBuffer in go SDK. Doing this would make the Code much more readable and also don't need to implement these new functions which were introduced.

I am saying that why can't we use DownloadStream as default download methods for our default ReadinBuffer method. Doing this would make the code changes very minimal

CHANGELOG.md
syeleti-msft: > If we make this feature enable by default it may start failing some existing customer flows

This is good thing for us and them as well, as this check atleast tells something is going wrong in their workflow.

>Also, some of these enhanced integrity checks have their own cost so we want customer to choose if they want to incur this cost or not.

I didn't understand this

component/azstorage/block_blob.go
vibhansa-msft: You mean use downloadStream instead of downloadBuffer in our code base in general ?

CHANGELOG.md
vibhansa-msft: in the same check we are doing crc64 computation and storing in xattr and on read we will compute back and compare so that will have additional cpu cost. Wanted to keep all these enhanced features in one group so that customer can wisely choose whether they want to incur this cost or not.

component/azstorage/azstorage.go
```diff
@@ -453,7 +453,15 @@ func (az *AzStorage) ReadInBuffer(options internal.ReadInBufferOptions) (length
 		return 0, nil
 	}
 
-	err = az.storage.ReadInBuffer(options.Handle.Path, options.Offset, dataLen, options.Data)
+	var etag string = ""
+	if options.Etag != nil {
+		etag, err = az.storage.ReadInBufferWithETag(options.Handle.Path, options.Offset, dataLen, options.Data)
+		*(options.Etag) = etag
+	} else {
```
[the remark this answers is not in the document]
vibhansa-msft: With DownloadStream changes, this if-else is removed.

CHANGELOG.md
vibhansa-msft: If things start failing for existing customers, which was working earlier it might be good that they realise there was something wrong earlier but also means we getting a lot of icms and github issues around this.
We can discuss with the team whether we want this to be default option or not.

vibhansa-msft: Removed the flag check and make this default, logically it makes sense to have this as default option.

component/azstorage/block_blob.go
```diff
@@ -888,20 +891,26 @@ func (bb *BlockBlob) ReadBuffer(name string, offset int64, len int64) ([]byte, e
...
-func (bb *BlockBlob) ReadInBuffer(name string, offset int64, len int64, data []byte) error {
+func (bb *BlockBlob) ReadInBuffer(name string, offset int64, len int64, data []byte, etag *string) error {
 	// log.Trace("BlockBlob::ReadInBuffer : name %s", name)
-	blobClient := bb.Container.NewBlobClient(filepath.Join(bb.Config.prefixPath, name))
-	opt := (blob.DownloadBufferOptions)(*bb.downloadOptions)
-	opt.BlockSize = len
-	opt.Range = blob.HTTPRange{
-		Offset: offset,
-		Count:  len,
+	if etag != nil {
```
syeleti-msft, line 896: No need to check the nil value as we are passing the pointer to a zerovalue

vibhansa-msft: Only block-cache is sending it, file-cache and UTs dont so we need to check here otherwise it will PANIC.

syeleti-msft (approved):

ashruti-msft (approved):

PR #1614, opened 2025-01-20
Reduce the number of rest calls while doing Rename Operation

syeleti-msft:
## ✅ What
 Currently we invalidate the cache item of destination (to remove the older reference of detination in our cache) after the rename operation completes[, this can be identified in the final call of the following flow:
**Rename on FNS:**
API calls: There are 5 REST api calls assosiated with this operation.

1. GetAttr(src)---------list call (filesystem call)
2. GetAttr(dst)---------list call (filesystem call)
3. copyblob(src->dst)----------(rename function)
4. deleteblob(src)--------------(rename function)
5. GetAttr(dst)---------list call (filesystem call)

**Rename on HNS:**
API Calls: There are 4 REST api calls assosiated with this operation.

1. GetAttr(Dst)
2. GetAttr(Src)
3. RenamePathFile
4. GetAttr(dst)

 goal is to serve the last call(invalidatePath) for FNS&HNS in the above REST calls from the attribute cache.

## 🤔 Why
 The idea is to simply copy source Attributes in the destination file object when the rename is success. 
 The only thing that modified in the destination was the last modified time, it will be modified according to the response of REST call for rename in AzStorage Component.
 
 Do rename operation with "mv" and see that the Final GetAttr on Dst after rename is success will be served from the cache.


component/attr_cache/attr_cache.go
```diff
@@ -360,14 +360,27 @@ func (ac *AttrCache) DeleteFile(options internal.DeleteFileOptions) error {
...
+			dstCacheEntry.attr.Size = srcAttr.Size
+			dstCacheEntry.attr.Path = options.Dst
+			dstCacheEntry.attr.Mode = srcAttr.Mode
+			dstCacheEntry.attr.Flags = srcAttr.Flags
+			dstCacheEntry.attr.MD5 = srcAttr.MD5
+			dstCacheEntry.attr.Metadata = srcAttr.Metadata
+			// Dst blob may not exist before
+			dstCacheEntry.attrFlag.Set(AttrFlagExists)
+		} else {
+			ac.invalidatePath(options.Dst)
```
vibhansa-msft, line 381: If its' not found or not valid, then why shall be invalidate it ?

component/attr_cache/attr_cache.go
```diff
@@ -360,14 +360,27 @@ func (ac *AttrCache) DeleteFile(options internal.DeleteFileOptions) error {
...
 	err := ac.NextComponent().RenameFile(options)
 	if err == nil {
+		// Copy source attribute to destination.
+		// LMT of destination will be modified by next component if the copy is success.
 		ac.cacheLock.RLock()
 		defer ac.cacheLock.RUnlock()
-
+		dstCacheEntry, found := ac.cacheMap[options.Dst]
+		if found && dstCacheEntry.valid() {
+			dstCacheEntry.attr.Size = srcAttr.Size
```
vibhansa-msft, line 372: Instead of copying down each field seperatly can we not do dst = src and just change the path and name,

component/azstorage/block_blob.go
```diff
@@ -349,9 +354,14 @@ func (bb *BlockBlob) RenameFile(source string, target string) error {
 		if err != nil {
 			log.Err("BlockBlob::RenameFile : CopyStats : Failed to get blob properties for %s [%s]", source, err.Error())
 		}
+		dstLMT = prop.LastModified
```
vibhansa-msft, line 357: We need not copy lmt on each iteration, only after copy is complete we can do it once.

component/attr_cache/attr_cache.go
syeleti-msft: That would make the changes to the local copy of dstAttr but we should also make the changes to its original copy.
For Ex: https://go.dev/play/p/P0OjTdVSrGb

component/azstorage/block_blob.go
syeleti-msft: to keep code simple I assigned variable inside that code block. I think It wouldn't be a problem

component/attr_cache/attr_cache.go
syeleti-msft: refactored

component/attr_cache/attr_cache.go
syeleti-msft: refactored

component/azstorage/block_blob.go
vibhansa-msft: Each instruction will have its own cost, code simplicity shall not be an excuse for the performance override. If this is not required you can do it outside the loop only.

component/azstorage/block_blob.go
```diff
@@ -316,7 +316,11 @@ func (bb *BlockBlob) DeleteDirectory(name string) (err error) {
 
 // RenameFile : Rename the file
 // Source file must exist in storage account before calling this method.
-func (bb *BlockBlob) RenameFile(source string, target string) error {
+// When the rename is success, Data, metadata, of the blob will be copied to the destination.
+// Creation time and LMT is not preserved for copyBlob API.
+// Copy the LMT to the src attr if the copy is success.
```
jainakanksha-msft, line 321: I think this comment should be Copy the LMT to the destination object if copy is a success.

component/azstorage/block_blob.go
```diff
@@ -340,6 +344,8 @@ func (bb *BlockBlob) RenameFile(source string, target string) error {
 		return err
 	}
 
+	var dstLMT *time.Time = startCopy.LastModified
```
jainakanksha-msft, line 347: Why we are picking up LMT from startCopy.
It can be from srcAttr also if needed.
But bigger question is why did we need it at this point.
Only after copying the blob we would need to update the LMT over destination object right?

component/azstorage/block_blob.go
```diff
@@ -349,9 +355,14 @@ func (bb *BlockBlob) RenameFile(source string, target string) error {
 		if err != nil {
 			log.Err("BlockBlob::RenameFile : CopyStats : Failed to get blob properties for %s [%s]", source, err.Error())
 		}
+		dstLMT = prop.LastModified
 		copyStatus = prop.CopyStatus
 	}
 
+	if copyStatus != nil && *copyStatus == blob.CopyStatusTypeSuccess {
+		modifyLMT(srcAttr, dstLMT)
```
jainakanksha-msft, line 363: Actually, modifing srcAttribute LMT might be slightly misleading in this context.
Now, i understand that in the next step we are just copying the attribute from source to destination.
Therefore, just a suggestion instead of updating the attribute on the source, we can set the LMT to the current value in attr_cache itself?
This suggestion is because the file is already modified 
and if other attribute like access time, change time and modify time all these needs to be updated we can set them to current value.

This way change will not be required at block_blob and datalake end.
we will be able to handle it at parent level itself.

syeleti-msft: Yes, We can do that completely in attribute cache as you said, but for that we need to return the last modified time we got from the azstorage. This LMT is only used by attribute cache and not by any other component, hence I thought it would be better to modify the attribute itself.

component/azstorage/block_blob.go
syeleti-msft: > Why we are picking up LMT from startCopy.

we picking it up from the startCopy because many times the Copyblob API copies the file immediately(i.e., you get the x-ms-copystatus=success in the response). in this case we can use LMT that was returned by the copyBlob to update our attribute 

> Only after copying the blob we would need to update the LMT over destination object right?

I think the variable name is culprit here for misunderstanding, I will update it

component/azstorage/block_blob.go
syeleti-msft: as you said, this is the reason
> Now, i understand that in the next step we are just copying the attribute from source to destination.

component/azstorage/block_blob.go
syeleti-msft: True, will update it

component/attr_cache/attr_cache_test.go
```diff
@@ -676,15 +676,28 @@ func (suite *attrCacheTestSuite) TestRenameFile() {
...
+	// Src, Dst Entry Already Exists
 	addPathToCache(suite.assert, suite.attrCache, src, false)
 	addPathToCache(suite.assert, suite.attrCache, dst, false)
+	options.SrcAttr = suite.attrCache.cacheMap[src].attr
+	options.DstAttr = suite.attrCache.cacheMap[dst].attr
 	suite.mock.EXPECT().RenameFile(options).Return(nil)
+	err = suite.attrCache.RenameFile(options)
+	suite.assert.Nil(err)
+	assertDeleted(suite, src)
+	assertUntouched(suite, dst)
```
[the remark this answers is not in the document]
syeleti-msft: >Shouldn't destination attribute definition be updated by source attribute information?

We are updating the src attribute

component/attr_cache/attr_cache.go
```diff
@@ -360,14 +360,24 @@ func (ac *AttrCache) DeleteFile(options internal.DeleteFileOptions) error {
...
-
+		dstCacheEntry, found := ac.cacheMap[options.Dst]
+		if found {
+			// Copy the Src Attr to Dst
+			dstCacheEntry.attr = srcAttr
+			dstCacheEntry.attr.Path = options.Dst
+			// Dst blob may not exist before
+			dstCacheEntry.attrFlag.Set(AttrFlagExists)
+			// Refresh the cache
+			dstCacheEntry.cachedAt = time.Now()
```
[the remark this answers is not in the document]
syeleti-msft: Sure

component/attr_cache/attr_cache_test.go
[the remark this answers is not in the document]
jainakanksha-msft: In attr_cache, we make sure things are copied to destination, so destination attribute should be updated right?

component/azstorage/block_blob.go
syeleti-msft: Updated

component/attr_cache/attr_cache_test.go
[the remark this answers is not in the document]
syeleti-msft: Updated the Tests

component/attr_cache/attr_cache.go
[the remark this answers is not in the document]
syeleti-msft: Updated

vibhansa-msft (approved):

jainakanksha-msft (approved):

Issue #1612, opened 2025-01-17
File open error for storage mounted with subdirectory and block cache.


listcrawler:
When mounting a storage subdirectory with a block cache, opening of files may fail. Using a file cache or without a subdirectory shows no such problems.

```bash
$ sha256sum somefile.bin
LOG_TRACE [libfuse_handler.go (687)]: Libfuse::libfuse_open : somefile.bin
LOG_TRACE [block_cache.go (385)]: BlockCache::OpenFile : name=somefile.bin, flags=32768, mode=-rwxrwxrwx
LOG_TRACE [attr_cache.go (472)]: AttrCache::GetAttr : somefile.bin
LOG_DEBUG [attr_cache.go (482)]: AttrCache::GetAttr : somefile.bin served from cache
LOG_ERR [block_cache.go (389)]: BlockCache::OpenFile : Failed to get attr of somefile.bin [no such file or directory]
LOG_ERR [libfuse_handler.go (721)]: Libfuse::libfuse_open : Failed to open somefile.bin [no such file or directory]
sha256sum: somefile.bin: No such file or directory
```

Inside the directory, some files can be opened and others not, seemingly arbitrarily. The success and failures are consistent across remounts though. The storage has a flat namespace and block blobs.

From what I can discern, the failing code path is like this:
1. [attr_cache.go:GetAttr returns an ENOENT error](https://github.com/Azure/azure-storage-fuse/blob/32ba9ccff233795b83121f0bbab482b2a651ad14/component/attr_cache/attr_cache.go#L484).
2. [block_cache.go:OpenFile picks up the error and returns the same ENOENT error](https://github.com/Azure/azure-storage-fuse/blob/32ba9ccff233795b83121f0bbab482b2a651ad14/component/block_cache/block_cache.go#L390).
3. [libfuse_handler.go:libfuse_open picks up the error and fails](https://github.com/Azure/azure-storage-fuse/blob/32ba9ccff233795b83121f0bbab482b2a651ad14/component/libfuse/libfuse_handler.go#L723).

config.yaml:
```yaml
allow-other: true

logging:
  type: syslog
  level: log_debug

components:
  - libfuse
  - block_cache
  - attr_cache
  - azstorage

libfuse:
  attribute-expiration-sec: 120
  entry-expiration-sec: 120
  negative-entry-expiration-sec: 240
  fuse-trace: true

block_cache:
  path: /tmp/blobfuse/block_cache
  block-size-mb: 32
[...]

PR #1618, opened 2025-01-27
 BUG#31069208:  Fixed Prefix filtering from File Path

jainakanksha-msft:
## ✅ What

This update addresses a bug related to subdirectory separation in file paths.
 
## 🤔 Why
 
A customer reported an issue where mounting blobfuse with the subdirectory option caused failures in opening certain files. Upon investigation, we discovered that files named identically to directories failed to open.
 
Before change 

1. In your storage account create a folder abc
2.  Create File abc.txt
3. "./blobfuse2 mount --subdirectory=abc"
4. ls abc.txt
5. It is failing saying that "abc.txt: No such file or directory"

After Change

1. "./blobfuse2 mount --subdirectory=abc"
2. ls abc.txt
3. It will result with proper ls output.

 
## 🔖 Related links
 
- [Issues] #1612


component/azstorage/utils_test.go
```diff
@@ -533,6 +533,33 @@ func (s *utilsTestSuite) TestRemoveLeadingSlashes() {
...
+func (suite *utilsTestSuite) TestRemovePrefixPath() {
+	assert := assert.New(suite.T())
+
+	var inputs = []struct {
+		prefixPath string
+		path       string
+		result     string
+	}{
+		{prefixPath: "", path: "abc.txt", result: "abc.txt"},
+		{prefixPath: "ABC", path: "ABC/DEF/1.txt", result: "DEF/1.txt"},
```
vibhansa-msft, line 545: Shall we try prefixPath also starting with "/" as part of this test case.

component/azstorage/utils_test.go
```diff
@@ -533,6 +533,33 @@ func (s *utilsTestSuite) TestRemoveLeadingSlashes() {
...
+		{prefixPath: "ABC/", path: "ABC/DEF/1.txt", result: "DEF/1.txt"},
+		{prefixPath: "ABC/DEF", path: "ABC/DEF/1.txt", result: "1.txt"},
```
vibhansa-msft, line 547: Also, add negative test case where the file name does not match any prefix and the same path is retruned back

jainakanksha-msft: Done

component/azstorage/utils_test.go
jainakanksha-msft: Done

component/azstorage/utils.go
```diff
@@ -553,6 +553,14 @@ func split(prefixPath string, path string) string {
 	return filepath.Join(paths...)
 }
 
+// removePrefixPath removes the given prefixPath from the beginning of path,
+// if it exists, and returns the resulting string without leading slashes.
+func removePrefixPath(prefixPath, path string) string {
+	prefixPath = strings.TrimLeft(prefixPath, "/")
```
vibhansa-msft, line 559: This just removes the "/" from prefixPath. Do we really need to call TrimLeft or just checking the first character will be faster here?

component/azstorage/utils.go
```diff
@@ -553,6 +553,24 @@ func split(prefixPath string, path string) string {
 	return filepath.Join(paths...)
 }
 
+// removePrefixPath removes the given prefixPath from the beginning of path,
+// if it exists, and returns the resulting string without leading slashes.
+func removePrefixPath(prefixPath, path string) string {
```
souravgupta-msft, line 558: The `split()` method is called from other places also like DeleteDirectory and RenameDirectory. We should remove the `split()` method and replace it with this method.

component/azstorage/block_blob.go
```diff
@@ -653,7 +653,7 @@ func (bb *BlockBlob) getBlobAttr(blobInfo *container.BlobItem) (*internal.ObjAtt
 	}
 
 	attr := &internal.ObjAttr{
-		Path:   split(bb.Config.prefixPath, *blobInfo.Name),
+		Path:   removePrefixPath(bb.Config.prefixPath, *blobInfo.Name),
```
souravgupta-msft, line 656: Please add an entry in the changelog

component/azstorage/utils.go
```diff
@@ -553,6 +553,24 @@ func split(prefixPath string, path string) string {
...
+	if prefixPath == "" {
+		return path
+	}
+	prefixPath = trimSlash(prefixPath)
```
vibhansa-msft, line 562: prefixPath is take from config, why do we want to remove "/" from it for every single call. Rather it can be done only once when we mount and then reduce this additional cost. Also, instead of calling 3 different apis we can just take the prefix path length and remove that much of characters from the path by forming a new string for e.g. path = path[10:]. This will remove the additional text in one operation instead of doing 3 operations, as written in code here.

component/azstorage/utils.go
souravgupta-msft: You can do something like,

- While prefix path is read from config, remove the leading slash and add a trailing slash. For example, if user has given "/abc" in config, the prefix path will be set to "abc/" after parsing. This will be done only once in the Config() method.
- In this method `removePrefixPath` you can just remove the prefix path directly,
`return path[len(prefixPath):]`

jainakanksha-msft: During config read, prefixPath is properly build.
Therefore, no need of slash handing at this point.
Rest of the method is updated.

component/azstorage/utils.go
jainakanksha-msft: Done

component/azstorage/block_blob.go
```diff
@@ -725,7 +725,7 @@ func (bb *BlockBlob) createDirAttr(name string) *internal.ObjAttr {
 	// For these dirs we get only the name and no other properties so hardcoding time to current time
 	name = strings.TrimSuffix(name, "/")
 	attr := &internal.ObjAttr{
-		Path:  split(bb.Config.prefixPath, name),
+		Path:  removePrefixPath(bb.Config.prefixPath, name),
```
vibhansa-msft, line 728: Kindly double check if there are any more instances of similar parsing which needs correction.

vibhansa-msft: I see below methods also have similar issue
- block_blob.go::
    - DeleteDirectory
    - RenameDirectory
    - createDirAttrWithPermissions

jainakanksha-msft: yes I am doing testing for those calls

vibhansa-msft (approved):

component/azstorage/block_blob.go
```diff
@@ -299,7 +299,7 @@ func (bb *BlockBlob) DeleteDirectory(name string) (err error) {
 
 		// Process the blobs returned in this result segment (if the segment is empty, the loop body won't execute)
 		for _, blobInfo := range listBlobResp.Segment.BlobItems {
-			err = bb.DeleteFile(split(bb.Config.prefixPath, *blobInfo.Name))
+			err = bb.DeleteFile(removePrefixPath(bb.Config.prefixPath, *blobInfo.Name))
 			if err != nil {
 				log.Err("BlockBlob::DeleteDirectory : Failed to delete file %s [%s]", *blobInfo.Name, err.Error())
 			}
```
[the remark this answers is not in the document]
souravgupta-msft: Yes, sounds logical to me also to not list the directory again.

souravgupta-msft (approved):

syeleti-msft (approved):

PR #1627, opened 2025-02-06
File open call in preload

souravgupta-msft:
## ✅ What
 
Preload operation downloads the blobs in parallel. If open call comes for a file and it has not been downloaded yet, then it will be scheduled for download on priority. 
  
## 🤔 Why
 
The open call for a file downloads it on priority, if not yet scheduled for download. This is done so that the users will not have to wait till the scheduler picks it up for download.
 
### NOTE

The UT will come in the next PR


component/xload/blockpool.go
```diff
@@ -97,15 +99,44 @@ func (pool *BlockPool) GetBlockSize() uint64 {
...
+}
+
+func (pool *BlockPool) mustGet() *Block {
+	var b *Block = nil
+
+	select {
+	case b = <-pool.blocksCh:
+		break
+
+	default:
```
vibhansa-msft, line 127: If we allocate a new block here then release shall also check if queue is full or not, and if so it shall release the memory otherwise this one block will linger for the lifespan of blobfuse.

component/xload/utils.go
```diff
@@ -104,3 +108,13 @@ func RoundFloat(val float64, precision int) float64 {
 	ratio := math.Pow10(precision)
 	return math.Round(val*ratio) / ratio
 }
+
+func IsFilePresent(localPath string) (bool, int64) {
```
vibhansa-msft, line 112: IsFilePresent lets make this method private.

component/xload/blockpool.go
souravgupta-msft: Already doing that, https://github.com/Azure/azure-storage-fuse/blob/sourav/prefetchOpen/component/xload/blockpool.go#L144-L151
The release of block back to the pool is non-blocking and if the pool is full, the block will be deleted, and its memory will be released.

```
func (pool *BlockPool) Release(b *Block) {
	select {
	case pool.blocksCh <- b:
		break
	default:
		_ = b.Delete()
	}
}
```

component/xload/utils.go
souravgupta-msft: Done

vibhansa-msft (approved):

component/xload/blockpool.go
```diff
@@ -97,15 +99,44 @@ func (pool *BlockPool) GetBlockSize() uint64 {
...
+		return pool.mustGet()
+	} else {
+		return pool.tryGet()
+	}
+}
+
+// TryGet a block from the pool. If the pool is empty, wait till a block is released back to the pool
+func (pool *BlockPool) tryGet() *Block {
 	// getting a block from pool will be a blocking operation if the pool is empty
 	b := <-pool.blocksCh
```
jainakanksha-msft, line 113: Kindly rename `b`

component/xload/blockpool.go
```diff
@@ -97,15 +99,44 @@ func (pool *BlockPool) GetBlockSize() uint64 {
...
 
 	// Mark the buffer ready for reuse now
-	if b != nil {
-		b.ReUse()
+	b.ReUse()
+	return b
+}
+
+func (pool *BlockPool) mustGet() *Block {
+	var b *Block = nil
```
jainakanksha-msft, line 121: please rename b

component/xload/blockpool.go
```diff
@@ -97,15 +99,44 @@ func (pool *BlockPool) GetBlockSize() uint64 {
...
+		break
+
+	default:
+		// There are no free blocks so we must allocate one and return here
+		// As the consumer of the pool needs a block immediately
+		log.Info("BlockPool::MustGet : No free blocks, allocating a new one")
+		var err error
+		b, err = AllocateBlock(pool.blockSize)
+		if err != nil {
+			return nil
```
jainakanksha-msft, line 134: If there is no more memory available. should we fallback in waiting state for a thread to gets free from the pool?

jainakanksha-msft, line 134: If there is no more memory available. should we fallback in waiting state for a thread to gets free from the pool?

component/xload/xload.go
```diff
@@ -326,11 +329,111 @@ func (xl *Xload) startComponents() error {
...
+// OpenFile: Download the file if not already downloaded and return the file handle
+func (xl *Xload) OpenFile(options internal.OpenFileOptions) (*handlemap.Handle, error) {
+	log.Trace("Xload::OpenFile : name=%s, flags=%d, mode=%s", options.Name, options.Flags, options.Mode)
+	localPath := filepath.Join(xl.path, options.Name)
+
+	flock := xl.fileLocks.Get(options.Name)
+	flock.Lock()
+	defer flock.Unlock()
+
+	downloadRequired, attr, err := xl.isDownloadRequired(localPath, options.Name)
```
jainakanksha-msft, line 367: Don't bind getAttr call inside isDownloadRequired.
Let's place it after the downloadRequired.
There might be possibility that call is not required.

component/xload/xload.go
```diff
@@ -326,11 +329,111 @@ func (xl *Xload) startComponents() error {
...
+	filePresent, size := isFilePresent(localPath)
+	downloadRequired := !filePresent
+
+	var attr *internal.ObjAttr
+	var err error
+	attr, err = xl.NextComponent().GetAttr(internal.GetAttrOptions{Name: blobPath})
+	if err != nil {
+		log.Err("Xload::isDownloadRequired : Failed to get attr of %s [%s]", blobPath, err.Error())
+	} else if filePresent {
+		downloadRequired = attr.Size != size
```
jainakanksha-msft, line 342: Why do we need to check the size. If file is already download. this size will be returned from cached attribute right?

component/xload/xload.go
```diff
@@ -326,11 +329,111 @@ func (xl *Xload) startComponents() error {
...
+		splitter := xl.getSplitter()
+		if splitter == nil {
+			log.Err("Xload::OpenFile : failed to  get download splitter for %s", options.Name)
+			return nil, fmt.Errorf("failed to  get download splitter")
+		}
+
+		_, err = splitter.Process(&WorkItem{
+			CompName: splitter.GetName(),
+			Path:     options.Name,
+			DataLen:  uint64(attr.Size),
```
jainakanksha-msft, line 384: if DataLength is not given in splitter process, shouldn't it consider to download the whole file?

component/xload/xload.go
```diff
@@ -326,11 +329,111 @@ func (xl *Xload) startComponents() error {
...
+	downloadRequired, attr, err := xl.isDownloadRequired(localPath, options.Name)
+	if err != nil {
+		log.Err("Xload::OpenFile : failed to open file %s [%s]", options.Name, err.Error())
+		return nil, err
+	}
+
+	if downloadRequired {
+		// send to splitter to download on priority
+		splitter := xl.getSplitter()
+		if splitter == nil {
```
jainakanksha-msft, line 376: Is there a chance that the splitter could be null?

component/xload/xload.go
```diff
@@ -326,11 +329,111 @@ func (xl *Xload) startComponents() error {
...
+		handle.Size = inf.Size()
+	}
+
+	handle.UnixFD = uint64(fh.Fd())
+	handle.Flags.Set(handlemap.HandleFlagCached)
+
+	log.Info("Xload::OpenFile : file=%s, fd=%d", options.Name, fh.Fd())
+	handle.SetFileObject(fh)
+
+	return handle, nil
```
jainakanksha-msft, line 417: After Download, should xload only serve the file? or transfer the call to subsequent component, which will Open the file accordingly?

component/xload/xload.go
```diff
@@ -326,11 +329,111 @@ func (xl *Xload) startComponents() error {
...
+		if c.GetName() == SPLITTER {
+			return c
+		}
+	}
+
+	return nil
+}
+
+// OpenFile: Download the file if not already downloaded and return the file handle
+func (xl *Xload) OpenFile(options internal.OpenFileOptions) (*handlemap.Handle, error) {
```
jainakanksha-msft, line 359: breakdown the implementation into 2 parts.
1st part downloadtheFile
2nd Part Serving the FIle

component/xload/xload.go
```diff
@@ -326,11 +329,111 @@ func (xl *Xload) startComponents() error {
...
+	handle.UnixFD = uint64(fh.Fd())
+	handle.Flags.Set(handlemap.HandleFlagCached)
+
+	log.Info("Xload::OpenFile : file=%s, fd=%d", options.Name, fh.Fd())
+	handle.SetFileObject(fh)
+
+	return handle, nil
+}
+
+func (xl *Xload) CloseFile(options internal.CloseFileOptions) error {
```
jainakanksha-msft, line 420: Similarly close file will not be required in xload, if we delegate the open call working the according component.

component/xload/xload.go
```diff
@@ -326,11 +329,111 @@ func (xl *Xload) startComponents() error {
...
+	if err != nil {
+		log.Err("Xload::OpenFile : error opening cached file %s [%s]", options.Name, err.Error())
+		return nil, err
+	}
+
+	// Increment the handle count in this lock item as there is one handle open for this now
+	flock.Inc()
+
+	handle := handlemap.NewHandle(options.Name)
+	inf, err := fh.Stat()
```
jainakanksha-msft, line 406: please rename `inf` either to0 `fileInfo` or `info`

component/xload/xload.go
```diff
@@ -326,11 +329,111 @@ func (xl *Xload) startComponents() error {
 	return nil
 }
 
+func (xl *Xload) isDownloadRequired(localPath string, blobPath string) (bool, *internal.ObjAttr, error) {
+	filePresent, size := isFilePresent(localPath)
+	downloadRequired := !filePresent
+
+	var attr *internal.ObjAttr
+	var err error
+	attr, err = xl.NextComponent().GetAttr(internal.GetAttrOptions{Name: blobPath})
```
jainakanksha-msft, line 338: attr, err := xl.NextComponent().GetAttr(internal.GetAttrOptions{Name: blobPath})
I think single line is enough

component/xload/splitter.go
```diff
@@ -51,6 +52,7 @@ type splitter struct {
 	XBase
 	blockPool *BlockPool
 	path      string
+	fileLocks *common.LockMap
```
jainakanksha-msft, line 55: Do we require to use LockMap and LockMapItem here?
Normal sync map with mutex functionality will be fine right?
LockMapItem serving the feature of handleCount and exLocked.

component/xload/splitter.go
```diff
@@ -102,6 +105,20 @@ func (d *downloadSplitter) Process(item *WorkItem) (int, error) {
...
 	localPath := filepath.Join(d.path, item.Path)
 
+	// if priority is false, it means that it has been scheduled by the lister and not OpenFile call
+	// so get a lock and wait if file is already under download by the OpenFile thread
+	// OpenFile thread already has a lock on the file, so don't take it again
+	if !item.Priority {
+		flock := d.fileLocks.Get(item.Path)
+		flock.Lock()
+		defer flock.Unlock()
+	}
```
jainakanksha-msft, line 115: If it's a normal scheduled item,
Do we require a locking?
It will be processed only once right?

component/xload/utils.go
```diff
@@ -104,3 +108,13 @@ func RoundFloat(val float64, precision int) float64 {
 	ratio := math.Pow10(precision)
 	return math.Round(val*ratio) / ratio
 }
+
+func isFilePresent(localPath string) (bool, int64) {
+	fileInfo, err := os.Stat(localPath)
+	if err != nil {
+		log.Debug("utils::isDownloadRequired : %s is not present in local path [%v]", localPath, err.Error())
```
jainakanksha-msft, line 115: Update logging `isDownloadRequired` to `isFilePresent`

component/xload/splitter.go
```diff
@@ -102,6 +105,20 @@ func (d *downloadSplitter) Process(item *WorkItem) (int, error) {
 	var err error
 	localPath := filepath.Join(d.path, item.Path)
 
+	// if priority is false, it means that it has been scheduled by the lister and not OpenFile call
+	// so get a lock and wait if file is already under download by the OpenFile thread
+	// OpenFile thread already has a lock on the file, so don't take it again
```
jainakanksha-msft, line 110: Update the comment -
if priority is false, it means that it has been scheduled by the lister and not by the OpenFile call
So, get a lock. If the locking goes into wait state, it means the file is already under download by the OpenFile call.
Otherwise, if there are no other locks, acquire a lock to prevent any OpenFile call from adding a request again.

component/xload/threadpool.go
```diff
@@ -61,48 +65,92 @@ func NewThreadPool(count uint32, callback func(*WorkItem) (int, error)) *ThreadP
...
 
 	return &ThreadPool{
-		worker:    count,
-		callback:  callback,
-		workItems: make(chan *WorkItem, count*2),
+		worker:        count,
+		callback:      callback,
+		priorityItems: make(chan *WorkItem, count*2),
+		workItems:     make(chan *WorkItem, count*4),
+		done:          make(chan int, count),
```
jainakanksha-msft, line 72: Instead of using done, use Go’s built-in context.Context to stop workers gracefully:
ctx context.Context
cancel context.CancelFunc

func NewThreadPool(count uint32, callback func(*WorkItem) (int, error)) *ThreadPool {
    ctx, cancel := context.WithCancel(context.Background())
    
func (threadPool *ThreadPool) Do(priority bool) {
...
   case <-tp.ctx.Done():  // Stop signal from context
            return 
        }
   
func (tp *ThreadPool) Stop() {
    tp.cancel() // Sends stop signal to all workers

component/xload/blockpool.go
souravgupta-msft: Done

component/xload/xload.go
souravgupta-msft: Ideally no, but this is a safe check to prevent panic scenarios

component/xload/xload.go
souravgupta-msft: CloseFile does not has any specific logic. It is only decreasing the handle count for the lock we incremented in the OpenFile call. This is similar to what we do in file cache where we download and maintain the file in local path

component/xload/xload.go
souravgupta-msft: Done

component/xload/xload.go
souravgupta-msft: Done

component/xload/utils.go
souravgupta-msft: Done

component/xload/splitter.go
souravgupta-msft: Done

component/xload/blockpool.go
souravgupta-msft: This is a good point. I will add a TODO item here and let's discuss with the team as what should be the best approach in this case.

component/xload/splitter.go
souravgupta-msft: Reusing the `LockMap` code here which is already used by the other components.
We can have the sync map and mutex in xload, but then we will have to maintain its code in the xload component. So, reusing the already written and well tested code is better imo.
For the other point, we are using the handle count in Open and Close calls. And, as xload component will grow, using the LockMap will be a better option.

component/xload/blockpool.go
souravgupta-msft: Done

component/xload/xload.go
souravgupta-msft: Discussed offline

component/xload/threadpool.go
souravgupta-msft: Done, using context with cancel method instead of buffered done channel

component/xload/xload.go
souravgupta-msft: No, it will create a 0 byte sized file in the local path

component/xload/xload.go
souravgupta-msft: Removed getAttr from this method

component/xload/xload.go
souravgupta-msft: Removed this check

component/xload/xload.go
souravgupta-msft: Refactored

component/xload/threadpool.go
```diff
@@ -60,49 +66,90 @@ func NewThreadPool(count uint32, callback func(*WorkItem) (int, error)) *ThreadP
...
+
+			case item := <-threadPool.workItems:
+				threadPool.process(item)
+
+			case <-threadPool.ctx.Done(): // listen to cancellation signal
+				return
+			}
+		}
+	}
+}
```
jainakanksha-msft, line 140: optimised version
`func (threadPool *ThreadPool) Do(priority bool) {
	defer threadPool.waitGroup.Done()

	for {
		select {
		case item := <-threadPool.priorityItems:
			threadPool.process(item)

		case <-threadPool.ctx.Done(): // listen to cancellation signal
			return

		default:
			if !priority {
				select {
				case item := <-threadPool.workItems:
					threadPool.process(item)
				case <-threadPool.ctx.Done(): // listen to cancellation signal
					return
				}
			}
		}
	}
}`
Please check the correctness because you default behaviour is our workItem in my implementation if I understood your point correctly.

component/xload/xload.go
```diff
@@ -326,11 +329,109 @@ func (xl *Xload) startComponents() error {
...
+		handle.Size = info.Size()
+	}
+
+	handle.UnixFD = uint64(fh.Fd())
+	handle.Flags.Set(handlemap.HandleFlagCached)
+
+	log.Info("Xload::OpenFile : file=%s, fd=%d", options.Name, fh.Fd())
+	handle.SetFileObject(fh)
+
+	return handle, nil
```
jainakanksha-msft, line 415: `
return openFile(localPath, options, flock)
}

func openFile(localPath string, options internal.OpenFileOptions, flock *common.LockMapItem) (*handlemap.Handle, error) {
	fh, err := os.OpenFile(localPath, options.Flags, options.Mode)
	if err != nil {
		log.Err("Xload::OpenFile : error opening cached file %s [%s]", options.Name, err.Error())
		return nil, err
	}

	// Increment the handle count in this lock item as there is one handle open for this now
	flock.Inc()

	handle := handlemap.NewHandle(options.Name)
	info, err := fh.Stat()
	if err == nil {
		handle.Size = info.Size()
	}

	handle.UnixFD = uint64(fh.Fd())
	handle.Flags.Set(handlemap.HandleFlagCached)

	log.Info("Xload::OpenFile : file=%s, fd=%d", options.Name, fh.Fd())
	handle.SetFileObject(fh)

	return handle, nil
}
`

Extract separate method to segregate out validations and proper openFile implementation

jainakanksha-msft (approved):

component/xload/threadpool.go
souravgupta-msft: The above code is incorrect because priority threads should not receive items from workItems channel. Just adding if condition is not sufficient as the priority threads can get an item from the workItems channel and not do anything because of the if condition.
For the unit tests, it will be added in the next PR.

component/xload/xload.go
souravgupta-msft: Already refactored the download logic in a different method. Rest of the code is specific to open call and I don't think will be reused.

PR #1636, opened 2025-02-18
Cancel download of a blob if one block fails + Unit tests

souravgupta-msft:
## ✅ What
 
If one of the blocks of a blob fails to download, cancel the entire download operation for that blob.
This PR also adds unit tests
 
## 🤔 Why
 
If one block fails to download, there is no need to download the remaining blocks because at the end, the blob will be marked as failed. So, using context in GO cancel the other download operations if one fails.


component/xload/data_manager.go
```diff
@@ -55,6 +57,11 @@ type remoteDataManager struct {
 func NewRemoteDataManager(remote internal.Component, statsMgr *StatsManager) (*remoteDataManager, error) {
```
vibhansa-msft, line 57:
```suggestion
func newRemoteDataManager(remote internal.Component, statsMgr *StatsManager) (*remoteDataManager, error) {
```

vibhansa-msft (approved):

component/xload/data_manager.go
souravgupta-msft: Done

PR #1637, opened 2025-02-19
In case of mount and mount all, if the container name has '$' do not treat it as env variable

vibhansa-msft:
## Type of Change
- [X] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description
In case of mount all or while mounting Azure special container like "$web", expandPath assumes it's an enviornment variable and tries to expand it while it shall skip these special container names. As part of this change we will not expand any environment variable in any path or config file name if env variable name matches reserved Azure Container name.

## How Has This Been Tested?
UT has been added and related bug has been linked.

## Checklist
- [X] The purpose of this PR is explained in this or a referenced issue.

## Related Links
NA

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -334,7 +334,13 @@ func ExpandPath(path string) string {
 		path = filepath.Join(homeDir, path[2:])
 	}
 
-	path = os.ExpandEnv(path)
+	path = os.Expand(path, func(key string) string {
+		if azureSpecialContainers[key] {
+			return "$" + key // Keep it as is
+		}
+		return os.Getenv(key) // Expand normally
+	})
+
 	path, _ = filepath.Abs(path)
 	return path
 }
--- a/common/types.go
+++ b/common/types.go
@@ -322,3 +322,9 @@ func init() {
 	DefaultLogFilePath = filepath.Join(DefaultWorkDir, "blobfuse2.log")
 	StatsConfigFilePath = filepath.Join(DefaultWorkDir, "stats_monitor.cfg")
 }
+
+var azureSpecialContainers = map[string]bool{
+	"web":        true,
+	"logs":       true,
+	"changefeed": true,
+}
--- a/common/util_test.go
+++ b/common/util_test.go
@@ -257,6 +257,16 @@ func (suite *utilTestSuite) TestExpandPath() {
 	path = ""
 	expandedPath = ExpandPath(path)
 	suite.assert.Equal(expandedPath, path)
+
+	path = "$HOME/.blobfuse2/config_$web.yaml"
+	expandedPath = ExpandPath(path)
+	suite.assert.NotEqual(expandedPath, path)
+	suite.assert.Contains(path, "$web")
+
+	path = "$HOME/.blobfuse2/$web"
+	expandedPath = ExpandPath(path)
+	suite.assert.NotEqual(expandedPath, path)
+	suite.assert.Contains(path, "$web")
 }
 
 func (suite *utilTestSuite) TestGetUSage() {


common/util.go
```diff
@@ -334,7 +334,13 @@ func ExpandPath(path string) string {
 		path = filepath.Join(homeDir, path[2:])
 	}
 
-	path = os.ExpandEnv(path)
+	path = os.Expand(path, func(key string) string {
+		if azureSpecialContainers[key] {
```
syeleti-msft, line 338: This fix seems to be very specific to that customer having $web in their config file imo. In linux $ is valid character that can be used in file path. So one possible solution is if the env variable is not set then return the exact same string rather than predefining some strings using some assumption.

Something like this:

```
	path = os.Expand(path, func(key string) string {
		value, ok := os.LookupEnv(key)
		if ok {
			return value
		}
		return "$" + key
	})
```

syeleti-msft (changes requested):

syeleti-msft:
is this pr about container name having $ symbol or config file having $ symbol, seems ambiguous to me!

common/util.go
vibhansa-msft: This might also create an issue what if I have $web set in env variable and using $web as container name as well.

vibhansa-msft:
> is this pr about container name having $ symbol or config file path having $ symbol, seems ambiguous to me!

Issue is both with mount path and config file name having "$" in their names.

common/util.go
syeleti-msft: if someone sets the env variable it means the user want to use the value of it when expanding the path. In the issue as well, the user tries to set the $web env variable as a workaround.

In the issue user might have $web in their file path, It does not have to do anything with azure special containers. tomorrow I may path like this in my config, /this/is/my/path/$syeleti

syeleti-msft:
> > is this pr about container name having $ symbol or config file path having $ symbol, seems ambiguous to me!
> 
> Issue is both with mount path and config file name having "$" in their names.

Then it is basically about filepath having $  symbol in it. I don't understand the description stating that container name is having $ in it. If it is not, then please modify the description

common/util.go
vibhansa-msft: in case of mount all we make directory same as container so if the container list gave a container named  '$web' we will create a directory with that name and mount will be done on that directory only. In such case we do not want to expand the env variables. Hence in case of mount path and config file name we do not want to expand if the key matches a special container name. I guess this is a safe check as we are not expanding only the env variables if their name matches a Azure special container name not otherwise.

vibhansa-msft: web / logs / changefeed : other than these 3 we will expand all other env variables anyway.

vibhansa-msft:
> > > is this pr about container name having $ symbol or config file path having $ symbol, seems ambiguous to me!
> > 
> > 
> > Issue is both with mount path and config file name having "$" in their names.
> 
> Then it is basically about filepath having $ symbol in it. I don't understand the description stating that container name is having $ in it. If it is not, then please modify the description

Expanded description.

common/util.go
syeleti-msft: Why cant we first expand the base path and then add the container name to it! In that case you don't need to declare these web / logs / changefeed  in the code.

souravgupta-msft: In mount all, the config file is created first for each container where the file name includes the container name. For example, /home/path/{container_name}.yaml. Then mount command is called for each container passing its respective config file in its mount command. So, even if we expand the base path first, the mount command will again try to expand the config file path which will again result in the same issue.

souravgupta-msft (approved):

syeleti-msft (approved):

PR #1642, opened 2025-02-25
Xload: Unit tests

souravgupta-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description
Add unit tests for xload

## How Has This Been Tested?
In the CI and code coverage pipelines

## Checklist

### DIFF
--- a/component/xload/data_manager_test.go
+++ b/component/xload/data_manager_test.go
@@ -34,6 +34,7 @@
 package xload
 
 import (
+	"context"
 	"testing"
 
 	"github.com/Azure/azure-storage-fuse/v2/component/loopback"
@@ -66,8 +67,30 @@ func (suite *dataManagerTestSuite) TestNewRemoteDataManager() {
 	suite.assert.NotNil(rdm)
 }
 
+func (suite *dataManagerTestSuite) TestProcessErrors() {
+	rdm := &remoteDataManager{}
+
+	ctx, cancel := context.WithCancel(context.Background())
+	item := &WorkItem{
+		CompName: DATA_MANAGER,
+		Path:     "test",
+		Block:    &Block{},
+		Download: false,
+		Ctx:      ctx,
+	}
+
+	n, err := rdm.Process(item)
+	suite.assert.NotNil(err)
+	suite.assert.Equal(n, 0)
+
+	// cancel the context
+	cancel()
+
+	n, err = rdm.Process(item)
+	suite.assert.NotNil(err)
+	suite.assert.Equal(n, 0)
+}
+
 func TestDatamanagerSuite(t *testing.T) {
 	suite.Run(t, new(dataManagerTestSuite))
 }
-
-// TODO:: xload : add tests for context cancelled
--- a/component/xload/data_manager_test.go
+++ b/component/xload/data_manager_test.go
@@ -69,3 +69,5 @@ func (suite *dataManagerTestSuite) TestNewRemoteDataManager() {
 func TestDatamanagerSuite(t *testing.T) {
 	suite.Run(t, new(dataManagerTestSuite))
 }
+
+// TODO:: xload : add tests for context cancelled
--- a/component/xload/splitter.go
+++ b/component/xload/splitter.go
@@ -39,6 +39,7 @@ import (
 	"os"
 	"path/filepath"
 	"sync"
+	"time"
 
 	"github.com/Azure/azure-storage-fuse/v2/common"
 	"github.com/Azure/azure-storage-fuse/v2/common/log"
@@ -107,7 +108,9 @@ func (d *downloadSplitter) Stop() {
 
 // download data in chunks and then write to the local file
 func (d *downloadSplitter) Process(item *WorkItem) (int, error) {
-	log.Debug("downloadSplitter::Process : Splitting data for %s, mode %v", item.Path, item.Mode)
+	log.Debug("downloadSplitter::Process : Splitting data for %s, size %v, mode %v, priority %v, access time %v, modified time %v", item.Path, item.DataLen,
+		item.Mode, item.Priority, item.Atime.Format(time.DateTime), item.Mtime.Format(time.DateTime))
+
 	var err error
 	localPath := filepath.Join(d.path, item.Path)
 
@@ -123,6 +126,7 @@ func (d *downloadSplitter) Process(item *WorkItem) (int, error) {
 
 	filePresent, size := isFilePresent(localPath)
 	if filePresent && item.DataLen == uint64(size) {
+		log.Debug("downloadSplitter::Process : %s will be served from local path, priority %v", item.Path, item.Priority)
 		return int(size), nil
 	}
 
@@ -260,6 +264,6 @@ func (d *downloadSplitter) Process(item *WorkItem) (int, error) {
 		return -1, fmt.Errorf("failed to download data for file %s", item.Path)
 	}
 
-	log.Debug("downloadSplitter::Process : Download completed for file %s", item.Path)
+	log.Debug("downloadSplitter::Process : Download completed for file %s, priority %v", item.Path, item.Priority)
 	return 0, nil
 }
--- a/component/xload/splitter.go
+++ b/component/xload/splitter.go
@@ -124,14 +124,14 @@ func (d *downloadSplitter) Process(item *WorkItem) (int, error) {
 		defer flock.Unlock()
 	}
 
-	filePresent, size := isFilePresent(localPath)
+	filePresent, isDir, size := isFilePresent(localPath)
+	if isDir {
+		log.Err("downloadSplitter::Process : %s is a directory", item.Path)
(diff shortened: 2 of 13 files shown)


component/xload/data_manager_test.go
```diff
@@ -66,6 +67,30 @@ func (suite *dataManagerTestSuite) TestNewRemoteDataManager() {
...
+		Download: false,
+		Ctx:      ctx,
+	}
+
+	n, err := rdm.Process(item)
+	suite.assert.NotNil(err)
+	suite.assert.Equal(n, 0)
+
+	// cancel the context
+	cancel()
```
jainakanksha-msft, line 87: Before canceling the context, if the item processing is failing in the download false scenario, please add a message in the Assertion part to add a context of this assertion.
suite.Assert().Equal(n, 0, "Expected n to be 0, but download is not supported now")

component/xload/data_manager_test.go
```diff
@@ -66,6 +67,30 @@ func (suite *dataManagerTestSuite) TestNewRemoteDataManager() {
...
+
+	n, err = rdm.Process(item)
```
jainakanksha-msft, line 89: Is there any scenerio where after cancelling the context we could have recevied the item in rdm channel

component/xload/data_manager_test.go
```diff
@@ -66,6 +67,30 @@ func (suite *dataManagerTestSuite) TestNewRemoteDataManager() {
...
+	suite.assert.NotNil(err)
+	suite.assert.Equal(n, 0)
```
jainakanksha-msft, line 91: Kindly rename n

component/xload/splitter.go
```diff
@@ -121,15 +124,16 @@ func (d *downloadSplitter) Process(item *WorkItem) (int, error) {
 		defer flock.Unlock()
 	}
 
-	filePresent, size := isFilePresent(localPath)
+	filePresent, isDir, size := isFilePresent(localPath)
+	if isDir {
```
jainakanksha-msft, line 128: In which scenario, it can be a directory call?
Ideally isFilePresent function should be returning just false and true. not other information related to File

component/xload/xload.go
```diff
@@ -351,17 +351,33 @@ func (xl *Xload) downloadFile(fileName string) error {
 	attr, err := xl.NextComponent().GetAttr(internal.GetAttrOptions{Name: fileName})
 	if err != nil {
 		log.Err("Xload::downloadFile : Failed to get attr of %s [%s]", fileName, err.Error())
+		return err
```
jainakanksha-msft, line 354: At line no 351, when we received the attribute, we should validate if it's a directory or not at this place, not at the splitter level.
The splitter should get a call corresponding to the file always, I believe.

component/xload/xload_test.go
```diff
@@ -318,6 +319,254 @@ func (suite *xloadTestSuite) TestUnsupportedModes() {
...
+		} else {
+			relPath := strings.TrimPrefix(localFilePath, suite.local_path+"/")
+			fh, err := suite.xload.OpenFile(internal.OpenFileOptions{Name: relPath, Flags: os.O_RDONLY, Mode: common.DefaultFilePermissionBits})
+			suite.assert.Nil(err)
+			suite.assert.NotNil(fh)
+
+			l, err := computeMD5(localFilePath)
+			suite.assert.Nil(err)
+
+			r, err := computeMD5(remoteFilePath)
```
jainakanksha-msft, line 559: Please add proper naming `r`

component/xload/xload_test.go
```diff
@@ -318,6 +319,254 @@ func (suite *xloadTestSuite) TestUnsupportedModes() {
...
+
+	fh1, err := suite.xload.OpenFile(internal.OpenFileOptions{Name: "file_4", Flags: os.O_RDONLY, Mode: common.DefaultFilePermissionBits})
+	suite.assert.Nil(err)
+	suite.assert.NotNil(fh1)
+	suite.assert.Equal(fh1.Size, (int64)(36))
+
+	err = suite.xload.CloseFile(internal.CloseFileOptions{Handle: fh1})
+	suite.assert.Nil(err)
+
+	fh2, err := suite.xload.OpenFile(internal.OpenFileOptions{Name: "dir_0/file_3", Flags: os.O_RDWR, Mode: common.DefaultFilePermissionBits})
```
jainakanksha-msft, line 529: We have opened the file with Read Write mode.
Is this correct?
As xload is working only in `read only` mode

component/xload/xload_test.go
```diff
@@ -318,6 +319,254 @@ func (suite *xloadTestSuite) TestUnsupportedModes() {
...
+	xl := &Xload{}
+	err := xl.createDownloader()
+	suite.assert.NotNil(err)
+	suite.assert.Len(xl.comps, 0)
+
+	xl.path = suite.local_path
+	xl.SetNextComponent(xl)
+	xl.statsMgr = &StatsManager{}
+	err = xl.createDownloader()
+	suite.assert.NotNil(err)
```
jainakanksha-msft, line 369: Also validate what error message we expect here if xload startup fails

component/xload/data_manager_test.go
souravgupta-msft: This test is specific to mock the context cancel case. So, the above scenario will not happen

component/xload/data_manager_test.go
souravgupta-msft: Yes, it is possible when download requests are already in the data manager's channel. If one block fails to download, it will cancel the context, and this will cause the rest of the incoming requests to cancel.

component/xload/data_manager_test.go
souravgupta-msft: test code

component/xload/splitter.go
souravgupta-msft: Splitter object can be created, and its process method can be called explicitly. So, this is a safe check if the process method is called not by the regular flow

component/xload/xload.go
souravgupta-msft: Agree, but the splitter's process method can be called outside of the regular flow (not scheduler or OpenFile). So, the dir check can help prevent those failures

component/xload/xload_test.go
souravgupta-msft: Since the call here is not coming from libfuse layer, this test passed. Will fix this in the next PR

component/xload/xload_test.go
souravgupta-msft: Will do in the next PR

component/xload/data_manager_test.go
jainakanksha-msft: still we should always follow proper naming in our code.
Test can be extended more and more in future.

component/xload/splitter.go
jainakanksha-msft: We have implemented a wrong behaviour here, filePresent is the first and correct check. if file is not present we shouldn't assume this check to be to be false itself.
With the current architecture, splitter should be calling over directory.
It should be the responsibility of the parent that only file is passed to the splitter.

component/xload/xload.go
jainakanksha-msft: We don't have that use case right now.
I think we shouldn't assume it that anyone can explicitly just called splitter.

component/xload/xload_test.go
```diff
@@ -318,6 +319,254 @@ func (suite *xloadTestSuite) TestUnsupportedModes() {
...
+
+		if entry.IsDir() {
+			suite.validateMD5WithOpenFile(localFilePath, remoteFilePath)
+		} else {
+			relPath := strings.TrimPrefix(localFilePath, suite.local_path+"/")
+			fh, err := suite.xload.OpenFile(internal.OpenFileOptions{Name: relPath, Flags: os.O_RDONLY, Mode: common.DefaultFilePermissionBits})
+			suite.assert.Nil(err)
+			suite.assert.NotNil(fh)
+
+			l, err := computeMD5(localFilePath)
```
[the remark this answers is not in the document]
jainakanksha-msft: Still need proper naming.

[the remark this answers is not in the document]
jainakanksha-msft: We can discuss over this.

component/xload/data_manager_test.go
souravgupta-msft: Will do in next PR

component/xload/splitter.go
souravgupta-msft: Refactored

component/xload/xload_test.go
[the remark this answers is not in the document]
souravgupta-msft: Done

component/xload/xload_test.go
souravgupta-msft: Done

PR #1655, opened 2025-03-07
Handled chown Scenario

jainakanksha-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description
This pull request introduces changes to add chown handling handling for adls account, and includes corresponding tests. 

## How Has This Been Tested?
image
image

Tested locally and with unit testing.
Still have to cover file_cache handling

## Checklist

- [x] Documentation update required.


component/azstorage/block_blob.go
```diff
@@ -666,6 +666,8 @@ func (bb *BlockBlob) getBlobAttr(blobInfo *container.BlobItem) (*internal.ObjAtt
 		Flags:  internal.NewFileBitMap(),
 		MD5:    blobInfo.Properties.ContentMD5,
 		ETag:   sanitizeEtag(blobInfo.Properties.ETag),
+		GID:    blobInfo.Metadata["gid"],
```
vibhansa-msft, line 669: We shall not use random names, check what azcopy uses and we shall be using the same key name in metadata to cross tool compatibility.

internal/attribute.go
```diff
@@ -81,6 +81,9 @@ type ObjAttr struct {
 	MD5      []byte             // MD5 of the blob as per last GetAttr
 	ETag     string             // ETag of the blob as per last GetAttr
 	Metadata map[string]*string // extra information to preserve
+	// Even though GID and UID are part of Metadata Map, I am defining new attributes to solve case sensitivity issue.
+	GID *string // Group ID
```
vibhansa-msft, line 85: for linux we can keep these values in integers so that we do not have to translate them later while giving it back to kernel

component/libfuse/libfuse_handler.go
```diff
@@ -346,8 +346,17 @@ func libfuse_destroy(data unsafe.Pointer) {
 }
 
 func (lf *Libfuse) fillStat(attr *internal.ObjAttr, stbuf *C.stat_t) {
-	(*stbuf).st_uid = C.uint(lf.ownerUID)
-	(*stbuf).st_gid = C.uint(lf.ownerGID)
+	if attr.UID != nil {
```
vibhansa-msft, line 349: ok for this kind of check may be either *string or *int is better but we might be reading from storage less often and returning it back to kernel more often so storing as *int will be a better choice.

component/azstorage/block_blob.go
```diff
@@ -1629,3 +1631,67 @@ func (bb *BlockBlob) SetFilter(filter string) error {
...
+		return err
+	}
+
+	// Update metadata if needed
+	if prop.Metadata == nil {
+		prop.Metadata = make(map[string]*string)
+	}
+
+	update := false
+	for key, value := range metadata {
```
vibhansa-msft, line 1658: Why we are doing so much of processing. We can get the metadata and set uid/gid related keys irrespective of whether they are there or not and send the metadata back.

component/azstorage/block_blob.go
```diff
@@ -1629,3 +1631,67 @@ func (bb *BlockBlob) SetFilter(filter string) error {
 	bb.Config.filter = &blobfilter.BlobFilter{}
 	return bb.Config.filter.Configure(filter)
 }
+
+func (bb *BlockBlob) updateMetadata(name string, metadata map[string]*string) error {
```
vibhansa-msft, line 1635: changeowner method anyways gets the uid and gid so there is no need to implement another method as this could have been done in that method itself,

component/azstorage/datalake.go
```diff
@@ -561,28 +564,23 @@ func (dl *Datalake) ChangeMod(name string, mode os.FileMode) error {
...
 // ChangeOwner : Change owner of a path
-func (dl *Datalake) ChangeOwner(name string, _ int, _ int) error {
+func (dl *Datalake) ChangeOwner(name string, uid int, gid int) error {
 	log.Trace("Datalake::ChangeOwner : name %s", name)
 
-	if dl.Config.ignoreAccessModifiers {
-		// for operations like git clone where transaction fails if chown is not successful
-		// return success instead of ENOSYS
-		return nil
+	metadata := make(map[string]*string)
```
vibhansa-msft, line 570: just pass the uid/gid as is to blockblob changeOwner method and it can take care of the rest. There is no need to implement this here.

internal/attribute.go
jainakanksha-msft: For integer, there was a problem, as root value for GID and UID is zero, and default value is also zero.
We will not be able to distinguish that we have the value or not.

component/azstorage/datalake.go
jainakanksha-msft: Currently changeOwner wan't having any implementation.

common/types.go
```diff
@@ -72,6 +72,8 @@ const (
 	UserAgentHeader = "User-Agent"
 
 	BlockCacheRWErrMsg = "Notice: The random write flow using block cache is temporarily blocked due to potential data integrity issues. This is a precautionary measure. \nIf you see this message, contact blobfusedev@microsoft.com or create a GitHub issue. We're working on a fix. More details: https://aka.ms/blobfuse2warnings."
+	POSIXOwnerMeta     = "posix_owner"
```
vibhansa-msft, line 75: define permissions flags here as well for fns account we will persist the file mode in that.

component/azstorage/block_blob.go
```diff
@@ -78,6 +79,8 @@ type BlockBlob struct {
 	blockLocks      common.KeyedMutex
 }
 
+type Metadata map[string]*string
```
vibhansa-msft, line 82: We are using this map in ObjAttr, correct place to define this data type would be inside in the common where ObjAttr is defined.

component/azstorage/block_blob.go
```diff
@@ -666,6 +669,8 @@ func (bb *BlockBlob) getBlobAttr(blobInfo *container.BlobItem) (*internal.ObjAtt
 		Flags:  internal.NewFileBitMap(),
 		MD5:    blobInfo.Properties.ContentMD5,
 		ETag:   sanitizeEtag(blobInfo.Properties.ETag),
+		GID:    ReadMetadata(blobInfo.Metadata, common.POSIXGroupMeta),
```
vibhansa-msft, line 678: we already parse metadata somewhere to get directory and symlink stuff, ideally these shall be part of those or that function itself shall be optmised to get these fields.

component/azstorage/utils.go
```diff
@@ -309,6 +309,36 @@ func parseMetadata(attr *internal.ObjAttr, metadata map[string]*string) {
 	}
 }
 
+func AddMetadata(metadata Metadata, key, value string) bool {
+	if _, ok := metadata[key]; ok {
+		return false
+	}
+
+	if key != "" {
+		capitalizedKey := strings.ToUpper(string(key[0])) + key[1:]
```
vibhansa-msft, line 318: General convention to convert case is to always convert it to lower and not upper.

component/azstorage/block_blob.go
```diff
@@ -675,15 +676,25 @@ func (bb *BlockBlob) getBlobAttr(blobInfo *container.BlobItem) (*internal.ObjAtt
 	}
 
 	parseMetadata(attr, blobInfo.Metadata)
-	if !bb.listDetails.Permissions {
+	if !bb.listDetails.Permissions && !(mode != 0) {
```
vibhansa-msft, line 679: !(mode != 0) ?? why double negation in condition check ? does that mean mode == 0

component/azstorage/block_blob.go
```diff
@@ -1515,31 +1526,69 @@ func (bb *BlockBlob) StageAndCommit(name string, bol *common.BlockOffsetList) er
...
 
-	if bb.Config.ignoreAccessModifiers {
-		// for operations like git clone where transaction fails if chmod is not successful
-		// return success instead of ENOSYS
-		return nil
+	// Fetch existing properties
+	prop, err := blobClient.GetProperties(context.Background(), &blob.GetPropertiesOptions{
+		CPKInfo: bb.blobCPKOpt,
+	})
+	if err != nil {
```
vibhansa-msft, line 1537: storage error conversions on this and setmetadata calls are missing.
If the blob does not exists lets check the local file system call whether it goes with success or failure and based on enoent we might want to return back success.

component/azstorage/block_blob.go
```diff
@@ -1515,31 +1526,69 @@ func (bb *BlockBlob) StageAndCommit(name string, bol *common.BlockOffsetList) er
...
+	updatedGroup := AddMetadata(prop.Metadata, common.POSIXGroupMeta, strconv.FormatUint(uint64(gid), 10))
 
-	// This is not currently supported for a flat namespace account
-	return syscall.ENOTSUP
+	// Apply metadata update
+	if updatedOwner || updatedGroup {
+		_, err := blobClient.SetMetadata(context.Background(), prop.Metadata, &blob.SetMetadataOptions{
+			CPKInfo: bb.blobCPKOpt,
+		})
+		if err != nil {
```
vibhansa-msft, line 1585: storage error conversion missing here as well

component/azstorage/block_blob.go
jainakanksha-msft: I checked when the error conversion is happening,
In this case it most likely be authentication error only.

component/azstorage/utils.go
```diff
@@ -309,6 +309,27 @@ func parseMetadata(attr *internal.ObjAttr, metadata map[string]*string) {
...
+		if strings.ToLower(k) == lowerKey {
+			existingKey, found = k, true
+			break
+		}
+	}
+	if found {
+		if *metadata[existingKey] == value {
+			return false
+		}
+		metadata[existingKey] = &value
```
vibhansa-msft, line 326: we modify the value at two places we shall unify this saying if the key was a match and value is same then skip otherwise update the value

component/libfuse/libfuse_handler.go
```diff
@@ -1156,6 +1167,25 @@ func libfuse_chown(path *C.char, uid C.uid_t, gid C.gid_t, fi *C.fuse_file_info_
 	name = common.NormalizeObjectName(name)
 	log.Trace("Libfuse::libfuse_chown : %s", name)
 	// TODO: Implement
+	err := fuseFS.NextComponent().Chown(
```
vibhansa-msft, line 1170: This part is missing in libfuse2

component/attr_cache/attr_cache.go
```diff
@@ -580,8 +580,6 @@ func (ac *AttrCache) Chmod(options internal.ChmodOptions) error {
 
 // Chown : Update the file with its new owner and group (when datalake chown is implemented)
 func (ac *AttrCache) Chown(options internal.ChownOptions) error {
-	log.Trace("AttrCache::Chown : Change owner of file/directory %s", options.Name)
-
 	err := ac.NextComponent().Chown(options)
```
vibhansa-msft, line 583: We shall update the metadata to store user and group supplied as input for the file if remote call was successful

component/azstorage/datalake.go
```diff
@@ -567,28 +567,8 @@ func (dl *Datalake) ChangeMod(name string, mode os.FileMode) error {
 }
 
 // ChangeOwner : Change owner of a path
-func (dl *Datalake) ChangeOwner(name string, _ int, _ int) error {
```
vibhansa-msft, line 570: GetPriperties in datalake is not parsing uid and gid.

jainakanksha-msft: parseMetadata(blobAttr, prop.Metadata) is setting the metadata back to attr object, from where I am reading while returning the response.
So, no need of extra handling.

bhansalivikas (approved):

syeleti-msft (approved):

component/azstorage/block_blob.go
```diff
@@ -1515,31 +1526,86 @@ func (bb *BlockBlob) StageAndCommit(name string, bol *common.BlockOffsetList) er
 }
 
 // ChangeMod : Change mode of a blob
-func (bb *BlockBlob) ChangeMod(name string, _ os.FileMode) error {
+func (bb *BlockBlob) ChangeMod(name string, mode os.FileMode) error {
 	log.Trace("BlockBlob::ChangeMod : name %s", name)
+	blobClient := bb.Container.NewBlockBlobClient(filepath.Join(bb.Config.prefixPath, name))
 
-	if bb.Config.ignoreAccessModifiers {
-		// for operations like git clone where transaction fails if chmod is not successful
```
souravgupta-msft, line 1522: Can you test git clone operation of large repo like vscode and see how it goes.

component/azstorage/block_blob.go
souravgupta-msft: Not sure if this is resolved or not, I think we should convert the error using the `storeBlobErrToErr` method here.

component/azstorage/block_blob.go
```diff
@@ -1515,31 +1526,86 @@ func (bb *BlockBlob) StageAndCommit(name string, bol *common.BlockOffsetList) er
...
-	// This is not currently supported for a flat namespace account
-	return syscall.ENOTSUP
+	// Apply metadata update
+	if updatedMode {
+		_, err := blobClient.SetMetadata(context.Background(), prop.Metadata, &blob.SetMetadataOptions{
+			CPKInfo: bb.blobCPKOpt,
+		})
+		if err != nil {
+			log.Err("BlockBlob::ChangeMod : Failed to update Blob Metadata %s [%s]", name, err.Error())
+			return err
```
souravgupta-msft, line 1562: The error should be converted using the `storeBlobErrToErr` method here.

component/azstorage/block_blob.go
```diff
@@ -675,15 +676,25 @@ func (bb *BlockBlob) getBlobAttr(blobInfo *container.BlobItem) (*internal.ObjAtt
...
+	if !bb.listDetails.Permissions && mode == 0 {
 		// In case of HNS account do not set this flag
 		attr.Flags.Set(internal.PropFlagModeDefault)
 	}
 
 	return attr, nil
 }
 
-func (bb *BlockBlob) getFileMode(permissions *string) (os.FileMode, error) {
+func (bb *BlockBlob) getFileMode(permissions *string, metadata map[string]*string) (os.FileMode, error) {
```
souravgupta-msft, line 687: Can we move this method to utils.go file as `bb` is unused in this method.

component/azstorage/block_blob.go
```diff
@@ -654,7 +655,7 @@ func (bb *BlockBlob) getBlobAttr(blobInfo *container.BlobItem) (*internal.ObjAtt
 		log.Trace("BlockBlob::List : blob is encrypted with customer provided key so fetching metadata explicitly using REST")
 		return bb.getAttrUsingRest(*blobInfo.Name)
 	}
-	mode, err := bb.getFileMode(blobInfo.Properties.Permissions)
+	mode, err := bb.getFileMode(blobInfo.Properties.Permissions, blobInfo.Metadata)
```
souravgupta-msft, line 658: The `permissions` metadata parsing is missing in the `getAttrUsingRest` method. Please check

component/azstorage/datalake.go
```diff
@@ -567,28 +567,8 @@ func (dl *Datalake) ChangeMod(name string, mode os.FileMode) error {
...
-	// e := storeDatalakeErrToErr(err)
-	// if e == ErrFileNotFound {
-	// 	return syscall.ENOENT
-	// } else if err != nil {
-	// 	log.Err("Datalake::ChangeOwner : Failed to change ownership of file %s to %s [%s]", name, mode, err.Error())
-	// 	return err
-	// }
-	return syscall.ENOTSUP
+func (dl *Datalake) ChangeOwner(name string, uid int, gid int) error {
+	return dl.BlockBlob.ChangeOwner(name, uid, gid)
```
souravgupta-msft, line 571: I think you will have to call the HNS API `fileClient.SetAccessControl` here.

component/azstorage/block_blob.go
```diff
@@ -675,15 +676,25 @@ func (bb *BlockBlob) getBlobAttr(blobInfo *container.BlobItem) (*internal.ObjAtt
 	}
 
 	parseMetadata(attr, blobInfo.Metadata)
-	if !bb.listDetails.Permissions {
+	if !bb.listDetails.Permissions && mode == 0 {
```
souravgupta-msft, line 679: In the `ObjAttr` object we are only parsing and adding the mode field. Should we also parse and add the owner and group info?

component/azstorage/block_blob.go
jainakanksha-msft: storeBlobErrToErr is having this definition
case bloberror.BlobAlreadyExists:
			return ErrFileAlreadyExists
		case bloberror.BlobNotFound:
			return ErrFileNotFound
		case bloberror.InvalidRange:
			return InvalidRange
		case bloberror.LeaseIDMissing:
			return BlobIsUnderLease
		case bloberror.InsufficientAccountPermissions, bloberror.AuthorizationPermissionMismatch:
			return InvalidPermission
		default:
			return ErrUnknown

Which will not met in any situation after setMetadata call because we are already calling Getproperties after which we are doing the conversion.

component/azstorage/block_blob.go
```diff
@@ -675,15 +676,25 @@ func (bb *BlockBlob) getBlobAttr(blobInfo *container.BlobItem) (*internal.ObjAtt
...
 		// In case of HNS account do not set this flag
 		attr.Flags.Set(internal.PropFlagModeDefault)
 	}
 
 	return attr, nil
 }
 
-func (bb *BlockBlob) getFileMode(permissions *string) (os.FileMode, error) {
+func (bb *BlockBlob) getFileMode(permissions *string, metadata map[string]*string) (os.FileMode, error) {
+	if metadata != nil {
```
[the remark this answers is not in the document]
jainakanksha-msft: for datalake accounts metadata will not be set with permission, it's handled at datalake side while setting the mode.

component/azstorage/block_blob.go
jainakanksha-msft: Here also storeBlobErrToErr  mapping doesn't make any sense, bcz we are calling getproperties and converting the error there.

component/azstorage/datalake.go
jainakanksha-msft: for uid and gid, there is no option to set it over ACL.

component/azstorage/block_blob.go
jainakanksha-msft: That is already store in metadata field, So no separate field we have defined. And we are parsing when we are returning the result from libfuse_handler.

component/azstorage/block_blob.go
souravgupta-msft: Consider a scenario where the auth mode being used only has read permissions. In this case, GetProperties will work, however SetMetadata will return error. So, instead of returning the SDK error, we should map to the relevant filesystem error code.

component/azstorage/block_blob.go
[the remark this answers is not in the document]
souravgupta-msft: Suppose there is metadata defined for permissions in the blob in datalake account. In this case we are giving preference to the value in metadata rather than the one returned by the service.

component/azstorage/datalake.go
souravgupta-msft: `SetAccessControlOptions` does have owner and group options.
https://github.com/Azure/azure-sdk-for-go/blob/main/sdk/storage/azdatalake/internal/path/models.go#L110-L121

It should be similar to how the `ChangeMod` method is calling the `SetAccessControl` API.

component/azstorage/block_blob.go
bhansalivikas: git clone does not work right now due to block-cache issues, we can check it on file-cache though.

PR #1660, opened 2025-03-18
Xload: Config changes

souravgupta-msft:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [x] Other (describe):

## Description

- Config auto-tuning as to calculating the number of threads in different components in xload based on the number of cores in user's VM.
- Corrected the logic so that the priority channel is given preference over normal channel.
- Made read-only default mode when xload is enabled.
- Renamed the `consistency` config parameter to `validate-md5`.


vibhansa-msft (approved):

cmd/mount.go
```diff
@@ -321,6 +322,10 @@ var mountCmd = &cobra.Command{
 			options.Components = append(options.Components[:1], append([]string{"entry_cache"}, options.Components[1:]...)...)
 		}
 
+		if config.IsSet("xload") {
+			config.Set("read-only", "true") // preload is only supported in read-only mode
+		}
```
souravgupta-msft, line 327: Incorrect, even if xload is not present in the components section, but is present in the config file this block will execute making the mount read-only.

PR #1663, opened 2025-03-24
Adding Base Structure for Distributed Cache

jainakanksha-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description
This PR consist of the bootstrap part of distributed cache and Peer Manager.

## How Has This Been Tested?
This is being tested by unit testing, E2E manual testing and with 3 nodes local setup

## Checklist

## Related Links
- [Documents](<https://microsoftapc-my.sharepoint.com/:w:/g/personal/natomar_microsoft_com/ET1TW2MDQ9ZCvrmjJokMICMBYvpKle41zTw6OSQu7MA8Bg?e=UE8vgj>)
-->


component/azstorage/block_blob.go
```diff
@@ -241,18 +241,22 @@ func (bb *BlockBlob) SetPrefixPath(path string) error {
 func (bb *BlockBlob) CreateFile(name string, mode os.FileMode) error {
 	log.Trace("BlockBlob::CreateFile : name %s", name)
 	var data []byte
-	return bb.WriteFromBuffer(name, nil, data)
+	return bb.WriteFromBuffer(internal.WriteFromBufferOptions{Name: name,
```
syeleti-msft, line 244: I see most of the changes to Azstorage component are refactoring the code. Please Remove them from this PR, make those changes a seperate PR for main. we don't want the overall changes to be overwhelming.

component/distributed_cache/heartbeat_manager.go
```diff
@@ -0,0 +1,96 @@
...
+		log.Err("AddHeartBeat: Failed to evaluate VM storage: ", err)
+		return err
+	}
+	hostname, _ := common.GetHostName()
+	hbData := map[string]interface{}{
+		"ipaddr":             ipaddr,
+		"nodeid":             hm.nodeId,
+		"hostname":           hostname,
+		"last_heartbeat":     time.Now().Unix(),
+		"total_space_GB":     total / (1024 * 1024 * 1024),
```
syeleti-msft, line 72: leave the sizes as bytes, as it gives the more info on size.

component/distributed_cache/distributed_cache.go
```diff
@@ -0,0 +1,255 @@
...
+	CacheSize         uint64 `config:"cache-size" yaml:"cache-size,omitempty"`
+	Replicas          uint32 `config:"replicas" yaml:"replicas,omitempty"`
+	HeartbeatTimeout  uint32 `config:"heartbeat-timeout" yaml:"heartbeat-timeout,omitempty"`
+	HeartbeatDuration uint32 `config:"heartbeat-duration" yaml:"heartbeat-duration,omitempty"`
+	MissedHeartbeat   uint32 `config:"heartbeats-till-node-down" yaml:"heartbeats-till-node-down,omitempty"`
+}
+
+const (
+	compName          = "distributed_cache"
+	HeartBeatDuration = 30
```
syeleti-msft, line 81: why do we need this global variables, aren't this available from the DistributedCache struct?

component/distributed_cache/distributed_cache.go
```diff
@@ -0,0 +1,255 @@
...
+	dc.heartbeatManager = NewHeartbeatManager(dc.cachePath, dc.storage, dc.hbDuration, "__CACHE__"+dc.cacheID)
+	dc.heartbeatManager.Start()
+	return nil
+
+}
+
+// setupCacheStructure checks and creates necessary cache directories and metadata.
+// It's doing 4 rest api calls, 3 for directory and 1 for creator file.+1 call to check the creator file
+func (dc *DistributedCache) setupCacheStructure(cacheDir string) error {
+	_, err := dc.storage.GetAttr(cacheDir + "/creator.txt")
```
syeleti-msft, line 127: please justify the usecases for calling the Azstorage methods without the nextcomponent. If there are none, please fallback to the general approach(i.e., using the next component) we use, that will ease the testing and make the code follow the code pattern that we are using.

component/distributed_cache/distributed_cache.go
```diff
@@ -0,0 +1,255 @@
...
+				}
+			}
+		}
+
+		// Add metadata file with VM IP
+		ip, err := getVmIp()
+		if err != nil {
+			return logAndReturnError(fmt.Sprintf("DistributedCache::Start error [failed to get VM IP: %v]", err))
+		}
+		if err := dc.storage.WriteFromBuffer(internal.WriteFromBufferOptions{Name: cacheDir + "/creator.txt",
```
syeleti-msft, line 147: What is the point of creator.txt, if each directory will be created by multiple nodes? 
i.e., even if the cache directory is created by node1, it is not guarantee that creator.txt would be created by node1. lets discuss this on more detail.

component/distributed_cache/distributed_cache.go
```diff
@@ -0,0 +1,255 @@
...
+func init() {
+	internal.AddComponent(compName, NewDistributedCacheComponent)
+
+	cacheID := config.AddStringFlag("cache-id", "blobfuse", "Cache ID for the distributed cache")
+	config.BindPFlag(compName+".cache-id", cacheID)
+
+	cachePath := config.AddStringFlag("cache-dir", "/tmp", "Path to the cache")
+	config.BindPFlag(compName+".path", cachePath)
+
+	chunkSize := config.AddUint64Flag("chunk-size", 1024*1024, "Chunk size for the cache")
```
syeleti-msft, line 238: modify the default chunk size to 16M

component/distributed_cache/heartbeat_manager.go
jainakanksha-msft: It looks more readable, but that calculation we can do during the command execution I think while giving the node stats.

component/distributed_cache/distributed_cache.go
jainakanksha-msft: Few things are optional to configure, so even if user don't provide that info we assigned those to some values.

component/distributed_cache/distributed_cache.go
jainakanksha-msft: Not yet used.
Whenever needed we can modify that.

component/distributed_cache/distributed_cache.go
```diff
@@ -0,0 +1,255 @@
...
+	cacheSize := config.AddUint64Flag("cache-size", 1024*1024*1024, "Cache size for the cache")
+	config.BindPFlag(compName+".cache-size", cacheSize)
+
+	replicas := config.AddUint32Flag("replicas", 3, "Number of replicas for the cache")
+	config.BindPFlag(compName+".replicas", replicas)
+
+	heartbeatTimeout := config.AddUint32Flag("heartbeat-timeout", 30, "Heartbeat timeout for the cache")
+	config.BindPFlag(compName+".heartbeat-timeout", heartbeatTimeout)
+
+	heartbeatDuration := config.AddUint32Flag("heartbeat-duration", HeartBeatDuration, "Heartbeat duration for the cache")
```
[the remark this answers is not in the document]
jainakanksha-msft: HeartBeatDuration is a constant

component/distributed_cache/distributed_cache.go
syeleti-msft: Yes, but those values get default values from the init method, I think there is no need to define them again here

component/distributed_cache/distributed_cache.go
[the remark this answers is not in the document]
syeleti-msft: What if user configure them? how can you able change the global constant that you defined?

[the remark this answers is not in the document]
jainakanksha-msft: In configure method

component/distributed_cache/heartbeat_manager.go
```diff
@@ -0,0 +1,99 @@
...
+		return err
+	}
+	hostname, _ := common.GetHostName()
+	totalSpace = func() uint64 {
+		if hm.maxCacheSize != 0 {
+			return hm.maxCacheSize
+		}
+		return totalSpace
+	}()
+	hbData := map[string]interface{}{
```
syeleti-msft, line 65: Can we create a struct for hearbeat info as that would be used in many places in the code?

component/distributed_cache/peer_manager.go
```diff
@@ -0,0 +1,42 @@
...
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE
+*/
+package distributed_cache
+
+type Peer struct {
```
syeleti-msft, line 35: Why this struct? isn't it same as heartbeatData struct?

jainakanksha-msft: It can less or additional attributes related to node info e.g. connection or other details.

component/distributed_cache/heartbeat_manager.go
```diff
@@ -0,0 +1,199 @@
...
+	hbPath := hm.hbPath + "/Nodes/" + hm.nodeId + ".hb"
+	err := hm.comp.DeleteFile(internal.DeleteFileOptions{Name: hbPath})
+	if err != nil {
+		log.Err("HeartbeatManager::Stop Failed to delete heartbeat file: ", err)
+		return err
+	}
+	return nil
+}
+
+var PeersByNodeId map[string]*Peer = make(map[string]*Peer)
```
syeleti-msft, line 145: Why are we putting the data in the map here? Why not slice. This is because one can sort them using some comparator function. This data should also be lock protected. say if I am accessing the data of this variable and this function is updating it, that would be a race condition.

component/distributed_cache/peer_manager.go
syeleti-msft: Currently I see it same as the mentioned struct. if it works with heartbeatdata struct, I think it's better to use that only. Even having 1 or more additional field is just gives us more info. is intention behind creating new struct is to save some space?

component/distributed_cache/heartbeat_manager.go
jainakanksha-msft: Slice is not threadsafe. 
rather than that I will use sync.Map

component/distributed_cache/peer_manager.go
jainakanksha-msft: Do you see putting connection details in heartbeat structure make sense.
Peers gonna expand in future but heartbeat will remain same. That was the concept to have 2 different structure here.
Heartbeat structure work is done, when someone punched it. That's it.
It's helping in peer discovery but peer information should be segregated from heartbeat, that's what my thought of.

jainakanksha-msft: Yes, currently it is same.

component/distributed_cache/heartbeat_manager.go
syeleti-msft: you can use a lock variable with slice enclosing it in a struct.
I don't see any map operations we can do on node name. as the decision would be taken based on the size. Please correct me if I am wrong

component/distributed_cache/heartbeat_manager.go
```diff
@@ -0,0 +1,199 @@
...
+	attrs, err := hm.comp.ReadDir(internal.ReadDirOptions{Name: hm.hbPath + "/Nodes/"})
+	if err != nil {
+		log.Err("HeartbeatManager::StartDiscovery: Failed to read Cache Node directory: %v", err)
+		return
+	}
+	for _, attr := range attrs {
+		log.Info(attr.Name)
+		data, err := hm.comp.ReadFileWithName(internal.ReadFileWithNameOptions{
+			Path: attr.Path,
+		})
```
syeleti-msft, line 158: Remove the nodeid of the VM from this listing as it not a peer.

PR #1665, opened 2025-03-25
Refactored directory creation and write to file

jainakanksha-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description
Just refactored few functionalities to be used by distributed cache

## How Has This Been Tested?
Unit testing

## Checklist


component/azstorage/block_blob.go
```diff
@@ -241,18 +241,22 @@ func (bb *BlockBlob) SetPrefixPath(path string) error {
...
-func (bb *BlockBlob) CreateDirectory(name string) error {
+func (bb *BlockBlob) CreateDirectory(name string, etag bool) error {
 	log.Trace("BlockBlob::CreateDirectory : name %s", name)
 
 	var data []byte
 	metadata := make(map[string]*string)
 	metadata[folderKey] = to.Ptr("true")
 
-	return bb.WriteFromBuffer(name, metadata, data)
+	return bb.WriteFromBuffer(internal.WriteFromBufferOptions{Name: name,
```
vibhansa-msft, line 256: If ETAG is nil, where user does not want to compare then shall we even send the etag in REST call (we were not sending it earlier)

component/azstorage/azstorage.go
```diff
@@ -437,6 +437,10 @@ func (az *AzStorage) ReadFile(options internal.ReadFileOptions) (data []byte, er
 	return az.storage.ReadBuffer(options.Handle.Path, 0, 0)
 }
 
+func (az *AzStorage) ReadFileWithName(options internal.ReadFileWithNameOptions) (data []byte, err error) {
```
vibhansa-msft, line 440: API name does not sound correct, for ReadBuffer there shall already be a wrapper in azstorage.

component/azstorage/block_blob.go
```diff
@@ -1089,25 +1095,35 @@ func (bb *BlockBlob) WriteFromFile(name string, metadata map[string]*string, fi
...
 		AccessTier:  bb.Config.defaultTier,
 		HTTPHeaders: &blob.HTTPHeaders{
-			BlobContentType: to.Ptr(getContentType(name)),
+			BlobContentType: to.Ptr(getContentType(options.Name)),
 		},
 		CPKInfo: bb.blobCPKOpt,
-	})
+	}
+
+	if options.Etag {
```
vibhansa-msft, line 1115: should it just be a nil check or the value being not empty.

vibhansa-msft (approved):

cmd/imports.go
```diff
@@ -31,13 +33,12 @@
    SOFTWARE
 */
 
-package cmd
```
syeleti-msft, line 34: the text(package imports) got misaligned here, please revert

vibhansa-msft: This file is not supposed to be edited manually, there is a tool to create new component and I used that to create dcache. for renmaing it should bave been just string replacement.

vibhansa-msft (approved):

internal/pipeline.go
```diff
@@ -45,13 +45,38 @@ import (
...
 }
 
 // NewComponent : Function that all components have to register to allow their instantiation
 type NewComponent func() Component
 
 // Map holding all possible components along with their respective constructors
 var registeredComponents map[string]NewComponent
+var CurrentPipeline *Pipeline
+
+func GetCurrentPipeline() *Pipeline {
```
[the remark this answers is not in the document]
vibhansa-msft: these will be later used, not now.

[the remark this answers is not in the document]
vibhansa-msft: let me remove this as well.

vibhansa-msft (approved):

internal/pipeline.go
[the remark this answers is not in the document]
vibhansa-msft: cleandup, kindly review.

component/azstorage/block_blob.go
```diff
@@ -275,18 +275,22 @@ func (bb *BlockBlob) SetPrefixPath(path string) error {
...
 
 	var data []byte
 	metadata := make(map[string]*string)
 	metadata[folderKey] = to.Ptr("true")
 
-	return bb.WriteFromBuffer(name, metadata, data)
+	return bb.WriteFromBuffer(internal.WriteFromBufferOptions{Name: name,
+		Metadata: metadata,
+		Data:     data,
+		Etag:     etag})
```
souravgupta-msft, line 293: Rename this flag to `isEtagEnabled` or some other name. Just having this name as `Etag` sounds confusing as it means I am passing the etag value of the blob

component/azstorage/datalake.go
```diff
@@ -246,7 +246,7 @@ func (dl *Datalake) CreateFile(name string, mode os.FileMode) error {
 }
 
 // CreateDirectory : Create a new directory in the filesystem/directory
-func (dl *Datalake) CreateDirectory(name string) error {
+func (dl *Datalake) CreateDirectory(name string, _ bool) error {
```
souravgupta-msft, line 249: We should also look into the case when the account type is adls and we only want to create the `__CACHE__` directory once. In this case it is looking like every node will create the cache dir

jainakanksha-msft: This is already handled in datalake side.

component/azstorage/block_blob.go
```diff
@@ -1123,25 +1129,42 @@ func (bb *BlockBlob) WriteFromFile(name string, metadata map[string]*string, fi
...
+			BlobContentType: to.Ptr(getContentType(options.Name)),
 		},
 		CPKInfo: bb.blobCPKOpt,
-	})
+	}
+
+	if options.EtagNoneMatchConditions != "" {
+		uploadOptions.AccessConditions = &blob.AccessConditions{
+			ModifiedAccessConditions: &blob.ModifiedAccessConditions{
+				IfNoneMatch: to.Ptr(azcore.ETag(options.EtagNoneMatchConditions)),
```
souravgupta-msft, line 1152: Use the constant from SDK
```suggestion
				IfNoneMatch: to.Ptr(azcore.ETagAny),
```

component/azstorage/block_blob.go
```diff
@@ -1123,25 +1129,42 @@ func (bb *BlockBlob) WriteFromFile(name string, metadata map[string]*string, fi
...
+			},
+		}
+	}
+	if options.EtagMatchConditions != "" {
```
souravgupta-msft, line 1156: For which scenario are we using this option? I don't see this option being used anywhere in this PR

internal/component_options.go
```diff
@@ -114,10 +119,19 @@ type ReadInBufferOptions struct {
...
+	Data                    []byte
+	Metadata                map[string]*string
+	EtagNoneMatchConditions string
+}
+
+type WriteFromBufferOptions struct {
+	Name                    string
+	Metadata                map[string]*string
+	Data                    []byte
+	EtagNoneMatchConditions string
```
souravgupta-msft, line 133: The type of this should be bool instead of string. Do see any scenario where we need to pass some string value to IfNoneMatch header?

internal/component_options.go
```diff
@@ -114,10 +119,19 @@ type ReadInBufferOptions struct {
...
+	EtagMatchConditions     string
```
souravgupta-msft, line 134: Don't see this option used anywhere, please remove it if not needed in this PR

jainakanksha-msft: Hey, this is refactoring PR, which is used by dcache. How can we see the uses in this PR?

internal/component_options.go
jainakanksha-msft: It should be string as type ETag string in itself is a string.

component/azstorage/block_blob.go
jainakanksha-msft: It is used in later PR of dcache.

internal/component_options.go
souravgupta-msft: > If-None-Match: An ETag value, or the wildcard character (*). Specify this header to perform the operation only if the resource's ETag does not match the value specified.

Do you see any use case where we will perform the operation only if the ETag is not matching the value string value?

souravgupta-msft (approved):

ashruti-msft (approved):

syeleti-msft (approved):

component/azstorage/azstorage.go
```diff
@@ -207,7 +207,7 @@ func (az *AzStorage) ListContainers() ([]string, error) {
 func (az *AzStorage) CreateDir(options internal.CreateDirOptions) error {
 	log.Trace("AzStorage::CreateDir : %s", options.Name)
 
-	err := az.storage.CreateDirectory(internal.TruncateDirName(options.Name))
+	err := az.storage.CreateDirectory(internal.TruncateDirName(options.Name), options.IsNoneMatchEtagEnabled)
```
vibhansa-msft, line 210: If this flag is enabled, we always hardcode the filter to "*" which means in create directory we are only asking to check if the directory exists or not and the flag controls whether to create directory forcefully or fail if it exists then name of the flag is not depicting that correctly.

jainakanksha-msft:
This code is already present in feature/dcache branch.
So, closing this PR.

PR #1669, opened 2025-03-27
Detect Account Type

ashruti-msft:
Detect Account Type only if it's not specified in config file and --use-adls flag is not given.  
If the account is ADLS I'm changing the endpoint as well.
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description

## Checklist

- [x] Updates to module CHANGELOG.md are included.

## Related Links
- [Issues](<link>)


component/azstorage/config.go
```diff
@@ -309,15 +311,20 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
...
+		shouldDetectAccountType = true
 		opt.AccountType = "block"
 	}
 
 	if config.IsSet(compName + ".use-adls") {
 		if opt.UseAdls {
+			shouldDetectAccountType = false
 			az.stConfig.authConfig.AccountType = az.stConfig.authConfig.AccountType.ADLS()
 		} else {
+			shouldDetectAccountType = true
```
souravgupta-msft, line 327: `shouldDetectAccountType` should be set to false in both cases because `--use-adls` flag is already set. So, for both if-else cases, user has explicitly passed this parameter which means we should not check the account type.

ashruti-msft: done

component/azstorage/config.go
```diff
@@ -530,18 +537,53 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
...
 	log.Crit("ParseAndValidateConfig : Retry Config: retry-count %d, max-timeout %d, backoff-time %d, max-delay %d, preserve-acl: %v",
 		az.stConfig.maxRetries, az.stConfig.maxTimeout, az.stConfig.backoffTime, az.stConfig.maxRetryDelay, az.stConfig.preserveACL)
-
 	log.Crit("ParseAndValidateConfig : Telemetry : %s, honour-ACL %v", az.stConfig.telemetry, az.stConfig.honourACL)
 
 	return nil
 }
 
+// Auto-Detect Account Type is HNS or FNS
+func DetectAccountType(az *AzStorage, opt AzStorageOptions) {
```
jainakanksha-msft, line 556: Move this to azstroage, and we can call it over azstroage object

ashruti-msft: done

component/azstorage/azstorage.go
```diff
@@ -105,6 +106,38 @@ func (az *AzStorage) Configure(isParent bool) error {
...
+	containerClient := serviceClient.(*service.Client).NewContainerClient(az.stConfig.container)
+
+	accountInfo, err := containerClient.GetAccountInfo(context.Background(), nil)
+	if err != nil {
+		log.Err("DetectAccountType : Failed to get account info %s", err.Error())
+		return
+	}
+	if *accountInfo.IsHierarchicalNamespaceEnabled {
+		az.stConfig.authConfig.AccountType = EAccountType.ADLS()
+		opt.Endpoint = fmt.Sprintf("%s.dfs.core.windows.net", opt.AccountName)
```
souravgupta-msft, line 131: If some private endpoint is set in config, then this logic would override it to *.core.windows.net which is incorrect

component/azstorage/azstorage.go
```diff
@@ -105,6 +106,38 @@ func (az *AzStorage) Configure(isParent bool) error {
 	return nil
 }
 
+// Auto-Detect Account Type is HNS or FNS
+func (az *AzStorage) DetectAccountType(opt AzStorageOptions) {
```
souravgupta-msft, line 110: Add unit tests for this

component/azstorage/azstorage.go
```diff
@@ -105,6 +106,38 @@ func (az *AzStorage) Configure(isParent bool) error {
...
+	auth := getAzAuth(az.stConfig.authConfig)
+	if auth == nil {
+		log.Err("DetectAccountType : Failed to get auth type %s", az.stConfig.authConfig.AuthMode)
+		return
```
jainakanksha-msft, line 114: return the err and fail ParseAndValidateConfig if it fails to detect.
Also possible to write unit testing?

component/azstorage/azstorage.go
ashruti-msft: done

component/azstorage/config.go
```diff
@@ -530,13 +536,20 @@ func ParseAndValidateConfig(az *AzStorage, opt AzStorageOptions) error {
 		}
 	}
 
+	if shouldDetectAccountType {
```
jainakanksha-msft, line 539: I think we only need to call the method when shouldSetEndpoint is true.
we are changing the config only then.

PR #1674, opened 2025-04-04
Added base interface skeleton

jainakanksha-msft:


internal/dcache_lib/api/node_manager.go
```diff
@@ -0,0 +1,38 @@
+/*
+	_____           _____   _____   ____          ______  _____  ------
```
vibhansa-msft, line 2: Misaligned

vibhansa-msft (approved):

vibhansa-msft (approved):

PR #1688, opened 2025-04-15
Added Storage Interface and Dcache Model #3

jainakanksha-msft:


ashruti-msft (approved):

component/distributed_cache/storage_callback_impl.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// ReadDirFromStorage implements dcachelib.StorageCallbacks.
+func (sci *StorageCallbackImpl) ReadDirFromStorage(options internal.ReadDirOptions) ([]*internal.ObjAttr, error) {
+	return sci.storage.ReadDir(options)
+}
+
+// SetProperties implements dcachelib.StorageCallbacks.
+func (sci *StorageCallbackImpl) SetProperties(path string, properties map[string]string) error {
+	panic("unimplemented")
```
souravgupta-msft, line 87: Instead of panic, return error

souravgupta-msft (approved):

syeleti-msft (approved):

component/distributed_cache/storage_callback_impl.go
```diff
@@ -0,0 +1,110 @@
...
+}
+
+// SetPropertiesInStorage implements dcachelib.StorageCallbacks.
+func (sci *StorageCallbackImpl) SetPropertiesInStorage(path string, properties map[string]string) error {
+	panic("unimplemented")
```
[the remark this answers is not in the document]
ashruti-msft: For now its okay in my Pr I have made the change to call setmetadata 
with that this will be removed

component/distributed_cache/storage_callback_impl.go
ashruti-msft: For now its okay in my Pr I have made the change to call setmetadata

internal/dcache/storage_callbacks.go
```diff
@@ -0,0 +1,52 @@
...
+
+import "github.com/Azure/azure-storage-fuse/v2/internal"
+
+type StorageCallbacks interface {
+	DeleteBlobInStorage(opt internal.DeleteFileOptions) error
+	GetBlobFromStorage(opt internal.ReadFileWithNameOptions) ([]byte, error)
+	GetPropertiesFromStorage(opt internal.GetAttrOptions) (*internal.ObjAttr, error)
+	PutBlobInStorage(opt internal.WriteFromBufferOptions) error
+	ReadDirFromStorage(options internal.ReadDirOptions) ([]*internal.ObjAttr, error)
+	SetPropertiesInStorage(path string, properties map[string]string) error
```
ashruti-msft, line 44: We should change this name to setMetadata or set MetaProperty as I am using the function setmetadata so the current name would be misleading

component/distributed_cache/storage_callback_impl.go
```diff
@@ -0,0 +1,111 @@
...
+
+package distributed_cache
+
+import (
+	"github.com/Azure/azure-storage-fuse/v2/internal"
+	"github.com/Azure/azure-storage-fuse/v2/internal/dcache"
+)
+
+// StorageCallbackImpl is a struct that implements the Storage interface
+type StorageCallbackImpl struct {
```
vibhansa-msft, line 42: Why do we need both next component and storage ?

jainakanksha-msft: For File System calls, flow will go through nextComponent.
For cache metadata related calls, flow will directly communicate with azstorage. That's why 2 different component.

vibhansa-msft: For regular file system calls, which we do not need to take any action, shall be handled directly by component and not via callbacks ideally. only the lib shall be using callbacks.

jainakanksha-msft: Yes, FileLayoutManager/FileManager will need this during normal flows, which will be part of our dccahe_lib

jainakanksha-msft: It will not be component who will directly performing anything. It will bypass the call to particular manager only.

internal/dcache/models.go
```diff
@@ -0,0 +1,97 @@
...
+
+type RawVolume struct {
+	NodeId         string    `json:"node_id,omitempty"`
+	IPAddress      string    `json:"ipaddr,omitempty"`
+	FSID           string    `json:"fsid,omitempty"`
+	FDID           string    `json:"fdid,omitempty"`
+	State          StateEnum `json:"state,omitempty"`
+	TotalSpace     uint64    `json:"total_space,omitempty"`
+	AvailableSpace uint64    `json:"available_space,omitempty"`
+	LocalCachePath string    `json:"local_cache_path,omitempty"`
```
linuxsmiths, line 49: this is just for debugging? because this has significance only on the node hosting the RV

linuxsmiths (approved):

internal/dcache/models.go
jainakanksha-msft: Yes, for debugging, we can remove it later

PR #1689, opened 2025-04-15
Added Cluster Manager#4

jainakanksha-msft:


vibhansa-msft (approved):

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -0,0 +1,88 @@
...
+	GetDegradedMVs() []dcache.MirroredVolume
+
+	//It will return all the RVs for particular mv name as per local cache copy of cluster map
+	GetRVs(mvName string) []dcache.RawVolume
+
+	//It will check if the given nodeId is online as per local cache copy of cluster map
+	IsAlive(nodeId string) bool
+
+	//It will evaluate the lowest number of RVs for given rvs
+	LowestNumberRV(rvs []string) []string
```
linuxsmiths, line 66: nit, 
rvs -> rvNames
to be clear that those are rv names and not rv ids

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -0,0 +1,117 @@
...
+
+type ClusterManagerImpl struct {
+	storageCallback dcache.StorageCallbacks
+}
+
+// GetRVs implements ClusterManager.
+func (c *ClusterManagerImpl) GetRVs(mvName string) []dcache.RawVolume {
+	panic("unimplemented")
+}
+
```
linuxsmiths, line 48: general comment.
Why are we not following the same order of methods as in the interface?
ANy reason for this order?

linuxsmiths (approved):

ashruti-msft (approved):

internal/dcache/cluster_manager/cluster_manager_impl.go
jainakanksha-msft: With the time things got changed. Earlier we had few apis , then few others got added. Like that. No manual implementation added. I can fix as of now. But in future again it can happen.

PR #1693, opened 2025-04-15
Punch HeartBeat #3

jainakanksha-msft:


internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -266,6 +281,39 @@ func evaluateReadOnlyState() bool {
 	return false
 }
 
+func (cmi *ClusterManagerImpl) punchHeartBeat(rvList []dcache.RawVolume) {
+	hostname, err := os.Hostname()
```
linuxsmiths, line 285: we can store hostname also in ClusterManagerImpl, populated once in start().
Then the cluster manager can export functions for others to query the nodeid, hostname, IPAddr, etc

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -266,6 +281,39 @@ func evaluateReadOnlyState() bool {
...
+	// Create and update heartbeat file in storage with <nodeId>.hb
+	err = mm.UpdateHeartbeat(cmi.nodeId, data)
+	common.Assert(err != nil, "Error updating heartbeat file with nodeId %s in storage: %v", cmi.nodeId, err)
+	log.Trace("AddHeartBeat: Heartbeat file updated successfully")
+}
+
+func listMyRVs(rvList []dcache.RawVolume) {
+	for index, rv := range rvList {
+		_, availableSpace, err := common.GetDiskSpaceMetricsFromStatfs(rv.LocalCachePath)
+		common.Assert(err != nil, "Error getting disk space metrics for path %s for punching heartbeat: %v", rv.LocalCachePath, err)
```
linuxsmiths, line 310: sorry the Assert cannot take a printf like string (I remember I misled you).
It does take variable parameters and logs all, if you want a formatted string, pls use sprintf and then pass that string to assert as a single param.

and this error, though unlikely, can happen, so we should also handle the error. Assert is good fro debug builds to catch if it happens, but in prod if it happens we need to handle it.
A good way to handle would be to set available space to 0, if we cannot query it.

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -266,6 +281,39 @@ func evaluateReadOnlyState() bool {
...
+		rvList[index].AvailableSpace = availableSpace
+		// TODO{Akku}: If available space is less than 10% of total space, set state to offline
```
linuxsmiths, line 312: no, we should not mark tit offline. It's still good for serving reads.

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -266,6 +281,39 @@ func evaluateReadOnlyState() bool {
...
+		NodeID:        cmi.nodeId,
+		Hostname:      hostname,
+		LastHeartbeat: uint64(time.Now().Unix()),
+		RVList:        rvList,
+	}
+
+	// Marshal the data into JSON
+	data, err := json.MarshalIndent(hbData, "", "  ")
+	//Adding Assert because error capturing can just log the error and continue because it's a schedule method
+	common.Assert(err != nil, "Error marshalling heartbeat data %+v : error - %v", hbData, err)
```
linuxsmiths, line 299: pls use Sprintf to pass a formatted string to Assert.
also, since MarshalIndent is an external function, though it should not fail, but if it fails for whatever reason, we should handle it. In this case, we should not send the heartbeat, and log an error.

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -114,13 +116,25 @@ func (cmi *ClusterManagerImpl) start(dCacheConfig *dcache.DCacheConfig, rvs []dc
...
+			cmi.punchHeartBeat(rvs)
+		}
+	}()
 	//Schedule clustermap update at storage and local copy
 	return nil
 }
 
 // Stop implements ClusterManager.
-func (c *ClusterManagerImpl) Stop() error {
+func (cmi *ClusterManagerImpl) Stop() error {
```
linuxsmiths, line 132: why is Stop exported?

jainakanksha-msft: Stop will get call from distributed cache component to gracefully stop cluster manager

PR #1696, opened 2025-04-16
Set Metadata added in azstorage

ashruti-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description

## Checklist

## Related Links
- [Issues](<link>)


component/azstorage/block_blob.go
```diff
@@ -1698,3 +1698,19 @@ func (bb *BlockBlob) SetFilter(filter string) error {
 	bb.Config.filter = &blobfilter.BlobFilter{}
 	return bb.Config.filter.Configure(filter)
 }
+
+// SetMetadata : Set metadata property of the blob
+func (bb *BlockBlob) SetMetadata(name string, attr *internal.ObjAttr) (err error) {
```
jainakanksha-msft, line 1703: Is it only name or proper file path passed to this method

component/azstorage/block_blob.go
```diff
@@ -1698,3 +1698,19 @@ func (bb *BlockBlob) SetFilter(filter string) error {
...
+	log.Trace("BlockBlob::GetAttr : name %s", name)
```
jainakanksha-msft, line 1704: Fix the logging GetAttr

component/azstorage/block_blob.go
syeleti-msft, line 1703: is name parameter not same as attr.Name or attr.path? how does it differ

component/azstorage/datalake.go
```diff
@@ -624,3 +624,8 @@ func (dl *Datalake) SetFilter(filter string) error {
 
 	return dl.BlockBlob.SetFilter(filter)
 }
+
+// SetMetadata : Set metadata property of path
+func (dl *Datalake) SetMetadata(name string, attr *internal.ObjAttr) (err error) {
```
vibhansa-msft, line 629: I suggest we send the new metadata as parameter and not part of attr. Reason being, there might be more metadata that exists in a blob and setting just what you sent migth overwrite the existing one. Rather if we just send a list of metadata we can fetch the properties once merge both metadata and then set it.

component/azstorage/datalake.go
```diff
@@ -624,3 +624,8 @@ func (dl *Datalake) SetFilter(filter string) error {
...
+	return dl.BlockBlob.SetMetadata(name, attr)
```
vibhansa-msft, line 630: There shall be one flag where  we can say overwrite or not. If overwrite then you can skip getting and merging with old metadata and just call the setproperties.

component/azstorage/datalake.go
vibhansa-msft, line 629: We generally do not rely on this kind of interface where we send attr and assume its metadata is correct, rather explicitly send the metadata.

ashruti-msft: The reason i had sent attr obj was because we will need Etag as well, but I can send it as a separate parameter so that's fine. 
In the call from metamanager impl we will getproperties and edit the metalist's value opencount and send new metadata list. So I don't think we need to get properties again here?

ashruti-msft: Will add the getproperties call which will only work if you don't want to overwrite using the falg suggested by you

component/azstorage/block_blob.go
```diff
@@ -1698,3 +1700,29 @@ func (bb *BlockBlob) SetFilter(filter string) error {
...
+	blobClient := bb.Container.NewBlockBlobClient(filePath)
+	// Set the metadata
+	_, err = blobClient.SetMetadata(context.Background(), newMetadata, &blob.SetMetadataOptions{
+		AccessConditions: &blob.AccessConditions{
+			ModifiedAccessConditions: &blob.ModifiedAccessConditions{
+				IfMatch: etag,
+			},
+		},
+	})
+	return err
```
vibhansa-msft, line 1727: We need to convert the err to storage error and then handle 404 and other errors just like all other apis.

component/azstorage/block_blob.go
```diff
@@ -1698,3 +1700,29 @@ func (bb *BlockBlob) SetFilter(filter string) error {
 	bb.Config.filter = &blobfilter.BlobFilter{}
 	return bb.Config.filter.Configure(filter)
 }
+
+// SetMetadata : Set metadata property of the blob
+func (bb *BlockBlob) SetMetadata(filePath string, newMetadata map[string]*string, etag *azcore.ETag, overwrite bool) (err error) {
+	log.Trace("BlockBlob::SetMetadata : name %s", filePath)
```
vibhansa-msft, line 1706: You shall add UTs for this method and validate all combinations.

PR #1698, opened 2025-04-19
Added metadata apis

ashruti-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description

## Checklist

## Related Links
- [Issues](<link>)


internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -34,62 +34,223 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/bloberror"
+	"github.com/Azure/azure-storage-fuse/v2/common/log"
+	"github.com/Azure/azure-storage-fuse/v2/internal"
 	"github.com/Azure/azure-storage-fuse/v2/internal/dcache"
 )
 
 // BlobMetadataManager is the implementation of MetadataManager interface
 type BlobMetadataManager struct {
-	cacheDir string
+	cacheDir         string
```
linuxsmiths, line 52: call this mdRoot (for metadata root)

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -34,62 +34,223 @@
...
+	storageCallbacks dcache.StorageCallbacks
 }
 
 // NewMetadataManager creates a new implementation of the MetadataManager interface
-func NewMetadataManager(cacheDir string) (MetadataManager, error) {
+func NewMetadataManager(cacheDir string) (*BlobMetadataManager, error) {
```
linuxsmiths, line 57: as mentioned in the chat, don't export this function, instead call it from the init() function in this file, so that users can use it as 

import mm "path/to/metadata/manager/package"

mm.CreateFileInit()

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -34,62 +34,223 @@
...
 	return &BlobMetadataManager{
 		cacheDir: cacheDir,
 	}, nil
 }
 
 // CreateFileInit creates the initial metadata for a file
 func (m *BlobMetadataManager) CreateFileInit(filePath string, fileMetadata *dcache.FileMetadata) error {
```
linuxsmiths, line 64: what's the reason we take FileMetadata and not []byte, like the heartbeat and other APIs?
Unless we have a need to interpret that inside the metadata manager (which it should not) let's be consistent and use []byte for file APIs too.

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -34,62 +34,223 @@
...
+
+	err = m.storageCallbacks.PutBlobInStorage(internal.WriteFromBufferOptions{
+		Name:                   filepath.Join(m.cacheDir, "Objects", filePath),
+		Metadata:               metadata,
+		Data:                   jsonData,
+		IsNoneMatchEtagEnabled: true,
+		EtagMatchConditions:    "",
+	})
+	if err != nil {
+		log.Debug("CreateFileInit :: Failed to put blob in storage: %v", err)
```
linuxsmiths, line 85: error log, and log the file path too.
also we should distinguish between failures due to if-none-match condition or other errors.
The former can be a warning and latter would be error.

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -34,62 +34,223 @@
...
+		return err
+	}
+	// Store the open-count in the metadata blob property
+	openCount := "0"
+	metadata := map[string]*string{
+		"opencount": &openCount,
+	}
+
+	err = m.storageCallbacks.PutBlobInStorage(internal.WriteFromBufferOptions{
+		Name:                   filepath.Join(m.cacheDir, "Objects", filePath),
```
linuxsmiths, line 78: store it in  a variable so that it can be logged

internal/dcache/metadata_manager/metadata_manager_impl.go
linuxsmiths: Also, thinking more about it, this should return the etag of the newly created blob, so that caller can then pass that to CreateFileFinalize() to be sure it's not finalizing a file that someone else created. This would be useful to handle cases where some node creates the initial file, then maybe goes silent for a while and some other node takes over and creates a fresh init file and then the original node came up and wants to finalize the file.

For now you can add a TODO and we can address it later.

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -34,62 +34,223 @@
...
+	return count, nil
 }
 
 // CloseFile decrements the open count for a file and returns the updated count
 func (m *BlobMetadataManager) CloseFile(filePath string) (int64, error) {
-	// Dummy implementation
-	return 0, nil
+	count, err := m.updateHandleCount(filePath, false)
+	if err != nil {
+		log.Debug("CloseFile :: Failed to update file close count: %v", err)
```
linuxsmiths, line 160: close count -> open count

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -34,62 +34,223 @@
...
+	const maxBackoff = 30           // Maximum backoff time in seconds
+	backoff := 1 * time.Millisecond // Initial backoff time in milliseconds
+	var openCount int
+	for {
+		// Get the current handle count
+		attr, err := m.storageCallbacks.GetPropertiesFromStorage(internal.GetAttrOptions{
+			Name: filepath.Join(m.cacheDir, "Objects", filePath),
+		})
+		if err != nil {
+			log.Debug("updateHandleCount :: Failed to get handle count: %v", err)
```
linuxsmiths, line 176: error log and must log filepath.
pls audit all such logs and fix

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -34,62 +34,223 @@
...
+			return -1, err
+		}
+
+		if attr.Metadata["opencount"] != nil {
```
linuxsmiths, line 180: since not having "opencount" metadata indicates some issue, we should bail out
Remove this check and add before this

```
// We never create file metadata blob w/o opencount property set.
if !attr.Metadata["opencount"] {
    log.Error("updateHandleCount :: File metadata blob found w/o opencount property: %s", filePath)
    return -1, <some appropriate error>
}

```

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -34,62 +34,223 @@
...
+			if err != nil {
+				log.Debug("GetFileOpenCount :: Failed to parse handle count: %v", err)
+				return -1, err
+			}
+			if increment {
+				openCount++
+			} else {
+				openCount--
+			}
+			if openCount < 0 {
```
linuxsmiths, line 191: we must not silently swallow such errors, these indicate some issue and we should cry out aloud.
should discuss this

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -34,62 +34,223 @@
...
+	if err != nil {
+		log.Debug("CloseFile :: Failed to update file close count: %v", err)
+		return -1, err
+	}
+	return count, nil
+}
+
+func (m *BlobMetadataManager) updateHandleCount(filePath string, increment bool) (int64, error) {
+	const maxBackoff = 30           // Maximum backoff time in seconds
+	backoff := 1 * time.Millisecond // Initial backoff time in milliseconds
```
linuxsmiths, line 168:
```
        backoff := 1 * time.Millisecond // Initial backoff time
        maxBackoff := 1 * time.Second // Max backoff time
```

30 secs is a lot for backoff. If everyone backs off so much we won;t make much progress. 1 sec should be fine.

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -34,62 +34,223 @@
...
+			Etag:      to.Ptr(azcore.ETag(attr.ETag)),
+			Overwrite: true,
+		})
+		if err != nil {
+			if bloberror.HasCode(err, bloberror.ConditionNotMet) {
+				log.Warn("updateHandleCount :: SetPropertiesInStorage failed due to ETag mismatch, retrying...")
+
+				// Apply exponential backoff
+				log.Debug("updateHandleCount :: Retrying in %d milliseconds...", backoff)
+				time.Sleep(time.Duration(backoff) * time.Millisecond)
```
linuxsmiths, line 211: time.Sleep(backoff)

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -34,62 +34,223 @@
...
+
+				// Double the backoff time, but cap it at maxBackoff
+				backoff *= 2
+				if backoff > maxBackoff {
+					backoff = maxBackoff
+				}
```
linuxsmiths, line 217: we should have some exit condition too, to avoid indefinite wait.
Retrying till 1 minute and then exiting is probably safe

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -34,62 +34,223 @@
...
+	prop, err := m.storageCallbacks.GetPropertiesFromStorage(internal.GetAttrOptions{
+		Name: filepath.Join(m.cacheDir, "Objects", filePath),
+	})
+	if err != nil {
+		log.Debug("GetFileOpenCount :: Failed to get handle count: %v", err)
+		return -1, err
+	}
+	openCount, ok := prop.Metadata["opencount"]
+	if !ok {
+		log.Debug("GetFileOpenCount :: openCount not found in metadata")
```
linuxsmiths, line 241: error log and also assert

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -34,62 +34,223 @@
...
 
 // GetFileOpenCount returns the current open count for a file
 func (m *BlobMetadataManager) GetFileOpenCount(filePath string) (int64, error) {
-	// Dummy implementation
-	return 0, nil
+	prop, err := m.storageCallbacks.GetPropertiesFromStorage(internal.GetAttrOptions{
+		Name: filepath.Join(m.cacheDir, "Objects", filePath),
+	})
+	if err != nil {
+		log.Debug("GetFileOpenCount :: Failed to get handle count: %v", err)
```
linuxsmiths, line 236: error log

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -125,46 +161,209 @@ func GetClusterMap() ([]byte, *azcore.ETag, error) {
...
+		}
+
+		// We never create file metadata blob w/o opencount property set.
+		if attr.Metadata["opencount"] == nil {
+			log.Err("updateHandleCount :: File metadata blob found w/o opencount property: %s", path)
+			return -1, fmt.Errorf("Opencount property not found in metadata for path %s. Issue needs to be debugged.", path)
+		}
+		openCount, err = strconv.Atoi(*attr.Metadata["opencount"])
+		if err != nil {
+			log.Err("GetFileOpenCount :: Failed to parse handle count for path %s : %v", path, err)
```
linuxsmiths, line 288: log ``*attr.Metadata["opencount"]`` too

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -125,46 +161,209 @@ func GetClusterMap() ([]byte, *azcore.ETag, error) {
...
+		})
+		if err != nil {
+			log.Err("updateHandleCount :: Failed to get handle count for %s : %v", path, err)
+			return -1, err
+		}
+
+		// We never create file metadata blob w/o opencount property set.
+		if attr.Metadata["opencount"] == nil {
+			log.Err("updateHandleCount :: File metadata blob found w/o opencount property: %s", path)
+			return -1, fmt.Errorf("Opencount property not found in metadata for path %s. Issue needs to be debugged.", path)
```
linuxsmiths, line 284: for such cases, log and return error, Akanksha has added a utility funciton . LogAndReturnError (or similar), check with here and use that for all such cases where you want to log and also return the same error string to the caller.

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -125,46 +161,209 @@ func GetClusterMap() ([]byte, *azcore.ETag, error) {
...
-	// Dummy implementation
-
-	return 0, nil
+	path := filepath.Join(m.mdRoot, "Objects", filePath)
+	count, err := m.updateHandleCount(path, true)
+	if err != nil {
+		log.Err("OpenFile :: Failed to update file open count for path %s : %v", path, err)
+		return -1, err
+	}
+	return count, nil
```
linuxsmiths, line 252: add a debug log before this telling that file successfully opened, and the new opencount is

same for close

linuxsmiths: // A file successfully opened, must have an opencount > 0.
common.Assert(count > 0, "Open file cannot have count <= 0", count)

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -125,46 +165,209 @@ func GetClusterMap() ([]byte, *azcore.ETag, error) {
...
+		Name: path,
+	})
+	if err != nil {
+		log.Err("GetFileOpenCount :: Failed to get handle count for path %s : %v", path, err)
+		return -1, err
+	}
+	openCount, ok := prop.Metadata["opencount"]
+	if !ok {
+		log.Err("GetFileOpenCount :: openCount not found in metadata for path %s : Error %v", path, err)
+		// TODO :: Add asserts
```
linuxsmiths, line 358: common.Assert(false, fmt.Sprintf("openCount not found in metadata for path %s", path))

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -264,12 +274,15 @@ func (m *BlobMetadataManager) closeFile(filePath string) (int64, error) {
 		log.Err("CloseFile :: Failed to update file open count for path %s : %v", path, err)
 		return -1, err
 	}
+	// TODO :: Add assert file open count is not negative "file cannot have count < 0"
```
linuxsmiths, line 277: common.Assert(count >= 0, "File cannot have -ve opencount", count)

linuxsmiths (approved):

PR #1700, opened 2025-04-21
MvRvMapping 

ashruti-msft:


internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -508,7 +510,130 @@ func (cmi *ClusterManagerImpl) updateStorageClusterMapIfRequired() {
 
 }
 
-func (cmi *ClusterManagerImpl) updateMVList(clusterMapRVMap map[string]dcache.RawVolume, clusterMapMVMap map[string]dcache.MirroredVolume) {
+func updateMVList(rvMap map[string]dcache.RawVolume, existingMVMap map[string]dcache.MirroredVolume, NumReplicas int, MvsPerRv int) map[string]dcache.MirroredVolume {
```
linuxsmiths, line 513: apart from adding MVs, there's one more task that updateMVList() must do - changing state of existing component RVs

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -508,7 +510,130 @@ func (cmi *ClusterManagerImpl) updateStorageClusterMapIfRequired() {
...
+		for _, n := range nodeToRvs {
+			availableNodes = append(availableNodes, n)
+
+		}
+
+		max := (len(rvMap) * MvsPerRv) / NumReplicas
+		// End of MV generation if we have enough MVs or if we have exhausted all the nodes
+		if len(availableNodes) < NumReplicas || len(existingMVMap) >= max {
+			break
+		}
```
linuxsmiths, line 572: by using the correct slot values we shouldn't need this.
Infact we can instead assert for these.

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -508,7 +510,155 @@ func (cmi *ClusterManagerImpl) updateStorageClusterMapIfRequired() {
 
 }
 
-func (cmi *ClusterManagerImpl) updateMVList(clusterMapRVMap map[string]dcache.RawVolume, clusterMapMVMap map[string]dcache.MirroredVolume) {
+func (cmi *ClusterManagerImpl) updateMVList(rvMap map[string]dcache.RawVolume, existingMVMap map[string]dcache.MirroredVolume, NumReplicas int, MvsPerRv int) map[string]dcache.MirroredVolume {
+
```
linuxsmiths, line 514: update the models.go

```
type MirroredVolume struct {
    RVsWithState   map[string]string `json:"rvs,omitempty"`
    State          StateEnum         `json:"state,omitempty"`
}
```

and then update usage here

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -508,7 +510,155 @@ func (cmi *ClusterManagerImpl) updateStorageClusterMapIfRequired() {
...
+	// TODO :: Handle scenarios for degraded and fix scenarios
+	// Degraded - When any of the rv's in a mv is offline make mv as degraded [Done]
+	// Fix - Replace any rv which is offline with an available rv and mark rv as out-of-sync while mv's state will be degraded
+	// Sync - This will be handled by replica manager and it will change mv state to syncing
+	// Offline - Mark a mv as offline when all the rv's within it are offline [Done]
+
+	// Populate the RV struct and node struct
+	for rvId, rvInfo := range rvMap {
+		if rvInfo.State == dcache.StateOffline {
+			// Skip RVs that are offline
```
linuxsmiths, line 537: usually comments that say why there are doing something are more useful than comments that say what they are doing.
so

// Skip offline RVs as they cannot contribute to any MV.

linuxsmiths: also let's assert that RV state can only be Offline or Online

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -508,7 +510,155 @@ func (cmi *ClusterManagerImpl) updateStorageClusterMapIfRequired() {
...
+			continue
+		}
+		if _, exists := nodeToRvs[rvInfo.NodeId]; exists {
+			// If the node already exists, append the RV to its list
+			node := nodeToRvs[rvInfo.NodeId]
```
linuxsmiths, line 542: why not assign node in the above statement only

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -508,7 +510,155 @@ func (cmi *ClusterManagerImpl) updateStorageClusterMapIfRequired() {
...
+				rvs:    []rv{{rvName: rvId, slots: MvsPerRv}},
+			}
+		}
+	}
+
+	// Phase 1 : Check and update slots of Rv's used in existing MVs
+	for mvName, mv := range existingMVMap {
+		offlineRv := 0
+		for rvName := range mv.RVWithStateMap {
+			if rvMap[rvName].State == dcache.StateOffline {
```
linuxsmiths, line 561: need to be careful with RVId and RVName.
rvMap is indexed using RVId while what we have is RVName

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -508,7 +510,155 @@ func (cmi *ClusterManagerImpl) updateStorageClusterMapIfRequired() {
 
 }
 
-func (cmi *ClusterManagerImpl) updateMVList(clusterMapRVMap map[string]dcache.RawVolume, clusterMapMVMap map[string]dcache.MirroredVolume) {
+func (cmi *ClusterManagerImpl) updateMVList(rvMap map[string]dcache.RawVolume, existingMVMap map[string]dcache.MirroredVolume, NumReplicas int, MvsPerRv int) map[string]dcache.MirroredVolume {
```
linuxsmiths, line 513: add this comment at the beginning

```
    //
    // Approach:
    //
    // We make a list of nodes each having a list of RVs hosted by that node. This is
    // typically one RV per node, but it can be higher.
    // Each RV starts with a slot count equal to MvsPerRv. This is done so that we can
    // assign one RV to MvsPerRv MVs.
    // In Phase#1 we go over existing MVs and deduct slot count for all the RVs used
    // by the existing MVs. After that's done, we are left wih RVs with updated slot
    // count signifying how many more MVs they can host.
    // Now in Phase#2 we create as many new MVs as we can, continuing with the next
    // available MV name, each MV is assigned one RV from a different node, upto
    // NumReplicas for each MV.
    // This continues till we do not have enough RVs (from distinct nodes) for creating
    // a new MV.
    //
```

internal/dcache/cluster_manager/cluster_manager_impl.go
ashruti-msft: should we change map[string]string to map[string]StateEnum?

internal/dcache/cluster_manager/cluster_manager_impl.go
ashruti-msft: rvMap is map[string]dcache.RawVolume and so its indexed by rvName and rvId is a var in the struct RawVolume

ashruti-msft: I have called rvName as rvId in some places mistakenly will fix that once we can clear this part

PR #1706, opened 2025-04-22
RPC APIs server-side implementation

souravgupta-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description
Server-side implementation of the RPC APIs.

## How Has This Been Tested?
Will add UTs in other PR

## Checklist


internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,743 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+}
+
+type rvInfo struct {
+	rvID     string       // rv0, rv1, etc.
+	cacheDir string       // cache dir path for this RV
+	mvMap    sync.Map     // sync map of MV id against mvInfo
+	mvCount  atomic.Int64 // count of MVs for this RV, this should be updated whenever a MV is added or removed from the sync map
+}
+
+type mvInfo struct {
```
linuxsmiths, line 70: mvInfo must have the mvName as the first field, followed by all the component RVs.
I think it will be useful to store two values

localRv    string
remoteRVs []string

so that we know what is the local and remote RVs for this MV.
Let me look at the rest of code to see what makes better sense.

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,743 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	syncInfo                  // sync info for this MV
+	peerRVs      []string     // sorted list of peer RVs for this MV
+	chunkOpsCnt  atomic.Int64 // count of chunk operations (get, put or remove) for this MV in the given RV
+	blockOpsFlag atomic.Bool  // block chunk operations for this MV in the given RV. This is done when the .sync dir contents are being moved to the regular MV dir
```
linuxsmiths, line 74: why do you say "in the given RV", that's implicit and confusing.
we will not have same mv on two different RVs on the same node. That's against resiliency rules.

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,743 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	}
+}
+
+// check if the given mv is valid
+func (rv *rvInfo) isMvValid(mvPath string) error {
+	if !common.DirectoryExists(mvPath) {
+		return fmt.Errorf("MV path %s does not exist", mvPath)
+	}
+
+	mvID := filepath.Base(mvPath)
```
linuxsmiths, line 105: mvName

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,743 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	_, err := rv.getMVInfo(mvID)
+	return err
+}
+
+func (rv *rvInfo) getPeerRVForMV(mvID string) []string {
```
linuxsmiths, line 110: getPeerRVsForMV - plural RVs to make it clear that it may return more than one RV

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,743 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	return 10
+}
+
+// check the if the chunk address is valid
+// - check if the fsID is valid
+// - check if the cache dir exists
+// - check if the MV is valid
+func (h *ChunkServiceHandler) checkValidChunkAddress(address *models.Address) error {
+	if address == nil || address.FileID == "" || address.FsID == "" || address.MvID == "" {
+		log.Err("ChunkServiceHandler::checkValidChunkAddress: Invalid chunk address")
```
linuxsmiths, line 220: best to log the address using %+v

linuxsmiths (commented):
This is best reviewed online.

common/util.go
```diff
@@ -464,6 +464,30 @@ func GetDiskSpaceMetricsFromStatfs(path string) (uint64, uint64, error) {
...
+		// Sudo  has mounted
+		availableSpace = stat.Bfree * uint64(stat.Frsize)
+	} else {
+		// non Sudo has mounted
+		availableSpace = stat.Bavail * uint64(stat.Frsize)
+	}
+
+	return int64(availableSpace), nil
+}
+
```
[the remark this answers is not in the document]
souravgupta-msft: Done, removed

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,743 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	}
+
+	return nil
 }
 
 func (h *ChunkServiceHandler) Hello(ctx context.Context, req *models.HelloRequest) (*models.HelloResponse, error) {
-	return nil, nil
+	if req == nil {
+		log.Err("ChunkServiceHandler::Hello: Received nil Hello request")
+		return nil, rpc.NewResponseError(rpc.InvalidRequest, "received nil Hello request")
```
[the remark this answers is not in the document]
souravgupta-msft: Done

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,743 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	}
+
+	log.Debug("ChunkServiceHandler::Hello: Received Hello request from %s to %s at %v, sender's RV = %v, shared MV = %v", req.SenderNodeID, req.ReceiverNodeID, req.Time, req.RV, req.MV)
```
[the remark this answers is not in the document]
souravgupta-msft: Done, updated

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,743 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+
+	// get the RV list from the fsIDMap
+	rvList := make([]string, 0)
```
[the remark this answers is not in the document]
souravgupta-msft: Done, updated

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,743 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+		Time:           time.Now().UnixMicro(),
+		RV:             rvList,
+		MV:             req.MV,
+	}, nil
 }
 
 func (h *ChunkServiceHandler) GetChunk(ctx context.Context, req *models.GetChunkRequest) (*models.GetChunkResponse, error) {
-	return nil, nil
+	if req == nil || req.Address == nil {
+		log.Err("ChunkServiceHandler::GetChunk: Received nil GetChunk request")
```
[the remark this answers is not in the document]
souravgupta-msft: Done at all places

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,743 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
 // ChunkServiceHandler struct implements the ChunkService interface
 type ChunkServiceHandler struct {
-	locks *common.LockMap
-	// more fields can be added here as needed
+	locks   *common.LockMap
+	fsIDMap map[string]*rvInfo // map to store the fsID to rvInfo mapping
+}
+
+type rvInfo struct {
+	rvID     string       // rv0, rv1, etc.
```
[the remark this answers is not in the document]

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,743 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	cacheDir string       // cache dir path for this RV
+	mvMap    sync.Map     // sync map of MV id against mvInfo
```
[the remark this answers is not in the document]
souravgupta-msft: Done, updated

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,743 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	mvCount  atomic.Int64 // count of MVs for this RV, this should be updated whenever a MV is added or removed from the sync map
+}
+
+type mvInfo struct {
+	syncInfo                  // sync info for this MV
+	peerRVs      []string     // sorted list of peer RVs for this MV
```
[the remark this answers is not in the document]
souravgupta-msft: Updated to componentRVs

internal/dcache/rpc/server/handler.go
souravgupta-msft: Added mvName

internal/dcache/rpc/server/handler.go

internal/dcache/rpc/server/handler.go

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done

internal/dcache/rpc/models.thrift
```diff
@@ -88,9 +88,9 @@ struct LeaveMVResponse {
 
 struct StartSyncRequest {
     1: string MV,
-    2: list<string> peerRV,
-    3: string sourceRV,
-    4: string targetRV,
+    2: string sourceRV, // source RV is the lowest index online RV. The node hosting this RV will send the start sync call to the peer RVs
+    3: string targetRV, // target RV is the RV which has to be synced
```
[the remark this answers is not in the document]
souravgupta-msft: Updated

internal/dcache/rpc/server/handler.go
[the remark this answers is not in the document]
linuxsmiths: sorry, my comment was a suggestion to update the comment, so

```
mvMap    sync.Map     // all MVs this RV is part of, indexed by MV name (e.g., "mv0"), updated by JoinMV and LeaveMV
```

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,861 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	mu           sync.Mutex
+	mvName       string       // mv0, mv1, etc.
+	componentRVs []string     // sorted list of component RVs for this MV
+	chunkOpsCnt  atomic.Int64 // count of chunk operations (get, put or remove) for this MV
+	blockOpsFlag atomic.Bool  // block chunk operations (get, put or remove) for this MV. This is done when the .sync dir contents are being moved to the regular MV dir
+	syncInfo                  // sync info for this MV
+}
+
+type syncInfo struct {
+	isSyncing bool   // is the MV in syncing state
```
linuxsmiths, line 81: I'd encourage you to be generous about writing good descriptive comments.
f.e.

```
	isSyncing bool   // Is the MV in syncing state?
	                 // An MV enters syncing state after StartSync command is successfully executed.
	                 // In syncing state, PutChunk requests corresponding to client writes will be
	                 // saved in the mv#.sync folder. Similarly a successful EndSync will take an
	                 // MV out of syncing state.
	syncID    string // sync ID for this MV if it is in syncing state.
	                 // This is returned in the StartSync response and EndSync should carry this.
```

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,861 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+
+	// sourceRV is the RV that is syncing the target RV in the MV.
+	// This is used when the source RV goes offline and is replaced by the cluster manager with a new RV.
+	// In this case, the sync will need to be restarted from the new RV.
+	sourceRV string // source RV for syncing this MV
 }
 
 func NewChunkServiceHandler() *ChunkServiceHandler {
+	// TODO:: integration: get fsID, rvName and cache dir path for different RVs for the node from cluster manager
+	// below will be call to cluster manager to get the information
```
linuxsmiths, line 92: yes, and when you add that do add a good comment explaining that 

rvIDMap contains any and all cluster awareness information that a node needs to have, things like what all RVs are hosted by this node, state of each of those RVs, what all MVs are hosted by these RVs, state of those MVs, etc. It is initially created from the clustermap which is the source of truth regarding cluster information, and once the cluster is working it's updated using various RPCs. Note that any time Cluster Manager needs to update clustermap, before publishing the updated clustermap, it'll send our one or more RPCs to update the rvIDMap info in all the affected nodes, thus rvIDMap always contains the latest info and hence is used by RVs to fail requests which might be sent by nodes having a stale clustermap.

internal/dcache/rpc/response_error.go
```diff
@@ -0,0 +1,61 @@
...
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+   SOFTWARE
+*/
+
+package rpc
+
+const (
+	InvalidRequest      = iota + 1 // invalid rpc request
+	InvalidFSID                    // RV does not have the FSID
```
linuxsmiths, line 38: RVID

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,861 @@ var _ service.ChunkService = &ChunkServiceHandler{}
 
 // ChunkServiceHandler struct implements the ChunkService interface
 type ChunkServiceHandler struct {
-	locks *common.LockMap
-	// more fields can be added here as needed
+	locks   *common.LockMap
+	fsIDMap map[string]*rvInfo // map to store the fsID to rvInfo mapping
```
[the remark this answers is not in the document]
souravgupta-msft: Done, updated

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done

internal/dcache/rpc/response_error.go
souravgupta-msft: Done

internal/dcache/rpc/server/handler.go
```diff
@@ -46,44 +56,861 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+func getNodeUUID() string {
+	return "node-uuid" // TODO: get the node uuid from the config
+}
+
+// check the if the chunk address is valid
+// - check if the fsID is valid
+// - check if the cache dir exists
+// - check if the MV is valid
+func (h *ChunkServiceHandler) checkValidChunkAddress(address *models.Address) error {
+	if address == nil || address.FileID == "" || address.FsID == "" || address.MvID == "" {
```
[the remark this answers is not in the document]
souravgupta-msft: Done

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+		return nil, fmt.Errorf("MV %s is invalid for RV %s", mvName, rv.rvName)
+	}
+
+	common.Assert(mvName == mvInfo.mvName, "MV name mismatch in mv", mvName, mvInfo.mvName)
+
+	return mvInfo, nil
+}
+
+func (rv *rvInfo) addToMVMap(mvName string, val *mvInfo) {
+	rv.mvMap.Store(mvName, val)
```
linuxsmiths, line 164: we can add a comment that caller is expected to ensure that the mv path does exist, since we later rely on that.

linuxsmiths (commented):
still reviewing. publishing in case you get time to address them

internal/dcache/rpc/server/handler.go
linuxsmiths: Also, since we should only call addToMVMap() on receiving a JoinMV request and join-mv must be called only by cluster manager when it's adding a new rv to an mv and never add the same mv again, we can assert that the mv is not already present. In prod we should fail such a request since clustermap is definitely not inline with our understanding of membership.

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	rv.mvCount.Add(-1)
+}
+
+// return the current sync state of the MV
+func (mv *mvInfo) getSyncStatus() bool {
+	mv.mu.Lock()
+	defer mv.mu.Unlock()
+
+	return mv.isSyncing
+}
```
linuxsmiths, line 179: we can make isSyncing atomic, then we can avoid this function and directly use mv.isSyncing in all places w/o the risk of deadlocking with mu.Lock already held in the calling context.

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+
+	// count of chunk operations (get, put or remove) for this MV.
+	// This is used to block the end sync call till all the ongoing chunk operations are completed.
+	chunkOpsCnt atomic.Int64
+
+	// flag to block chunk operations (get, put or remove) for this MV.
+	// This flag is enabled in the end sync call to pause new chunk operations in the MV.
+	// When the end sync call is completed, this flag is disabled.
+	blockOpsFlag atomic.Bool
+
```
linuxsmiths, line 95: one more useful information to store for an mv is the totalChunkSize i.e., the total amount of space used up inside an mv by all the chunks stored in it. This can be easily updated in PutChunk and DeleteChunk and is useful for finding candidate RVs when we are looking to replace one. We need an RV with at least that much space.
Didn't you need it for JoinMV as it has the "space to be reserved" as one of the params and that comes from this field.

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+}
+
+// update the sync state of the MV
+func (mv *mvInfo) updateSyncState(isSyncing bool, syncID string, sourceRV string) error {
+	mv.mu.Lock()
+	defer mv.mu.Unlock()
+
+	if isSyncing && mv.isSyncing {
+		// if the source RV is different from the current source RV, it means that the current source RV's node is offline
+		// and is replaced by a new RV. So, the sync process needs to be restarted with the new RV.
```
linuxsmiths, line 188: to keep things simple, let's have a rule that every sync started using StartSync MUST be ended using EndSync. If needed we can later optimize things in part as we see fit, but starting with simple rules is always best.
In that case we should not have the MV already in sync state. Some communication gaps can cause thus rule to break (if we are not able to call EndSync on an RV which went offline), and we should handle it, but we should start with asserts to catch any violation of such rules. Later if seen too often due to commn issues we can think more about it, whether to remove the assert or do something else.

SUggested function

```
func (mv *mvInfo) updateSyncState(isSyncing bool, syncID string, sourceRV string) error {
	mv.mu.Lock()
	defer mv.mu.Unlock()

    //
    // EndSync.
    // Must be received after a matching StartSync, i.e., with the same syncID.
    // Due to connectivity issues we may miss some of the requests, but we still assert in debug builds
    // as those are not common and we would like to understand, to handle those better.
    //
	if !isSyncing {
	    if mv.syncID == "" {
	        msg := fmt.Sprintf("EndSync(%s) received w/o StartSync", syncID)
            common.Assert(false, msg)
            return fmt.Errorf(msg)
	    }

	    if mv.syncID != syncID {
	        msg := fmt.Sprintf("Unexpected EndSync(%s) received, expected syncID %s ", syncID, mv.syncID)
            common.Assert(false, msg)
            return fmt.Errorf(msg)
	    }

        // sourceRV must be valid.
	    common.Assert(mv.sourceRV != "")

	    mv.isSyncing = false
	    mv.syncID = ""
	    mv.sourceRV = ""

	    return nil
	}

	// StartSync.
	// - Must not already be in syncing state.
	common.Assert(syncID != "")
	common.Assert(sourceRV != "")
	
	if mv.isSyncing {
	   msg := fmt.Sprintf("Got StartSync(%s) while already in syncing state with syncID %s", mv.syncID, syncID))
       common.Assert(false, msg)
       return fmt.Errorf(msg)
    }

    common.Assert(mv.syncID == "")
    common.Assert(mv.sourceRV == "")

	mv.isSyncing = isSyncing
	mv.syncID = syncID
	mv.sourceRV = sourceRV

	return nil
}
```

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+// return the current sync state of the MV
+func (mv *mvInfo) getSyncStatus() bool {
+	mv.mu.Lock()
+	defer mv.mu.Unlock()
+
+	return mv.isSyncing
+}
+
+// update the sync state of the MV
+func (mv *mvInfo) updateSyncState(isSyncing bool, syncID string, sourceRV string) error {
```
linuxsmiths, line 182: is it source RV name or ID, pls use more descriptive variable names

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	componentRVs []string // sorted list of component RVs for this MV
+
+	// count of chunk operations (get, put or remove) for this MV.
+	// This is used to block the end sync call till all the ongoing chunk operations are completed.
+	chunkOpsCnt atomic.Int64
+
+	// flag to block chunk operations (get, put or remove) for this MV.
+	// This flag is enabled in the end sync call to pause new chunk operations in the MV.
+	// When the end sync call is completed, this flag is disabled.
+	blockOpsFlag atomic.Bool
```
linuxsmiths, line 94: I suggest naming this flag suggesting quiescing (blocking is not a very accurate term here)

```
    //
    // Two MV states are interesting from an IO standpoint.
    // An online MV is the happy case where all RVs are online and sync'ed. In this state there won't be any
    // resync Writes, and client Writes if any will be replicated to all the RVs, each of them storing the chunks
    // in their respective mv folders. This is the normal case.
    // A syncing MV is interesting. In this case there are resync writes going and possibly client writes too.
    // Client chunks are saved in a special mv.sync folder while sync writes are saved in the regular mv folder.
    // During EndSync, the client writes are moved from mv.sync to the regular mv folder and then we have the MV
    // back in online state.
    // The short period when an MV moves in and out of syncing state is important. We need to quiesce any IOs
    // to make sure we don't miss resyncing any chunk.
    // Both StartSync and EndSync will quiesce IOs just before they move the mv into and out of syncing state, and
    // resume IOs once the MV is safely moved into the new state.
    //
    // This boolean flag will be set by StartSync and EndSync to put the mv in quiesce state and it'll be read
    // and honored by the various chunk related APIs - GetChunk, PutChunk, etc.
    //
    quiesceIOs atomic.Bool
```

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	mv.chunkOpsCnt.Add(-1)
+}
+
+func (mv *mvInfo) enableBlockChunkOps() {
+	mv.blockOpsFlag.Store(true)
+}
+
+func (mv *mvInfo) disableBlockChunkOps() {
+	mv.blockOpsFlag.Store(false)
+}
```
linuxsmiths, line 250: these and blockSyncOps() can be replaced with

```
func (mv *mvInfo) quiesceIOsStart() error {
    // Set IO quiescing in the mv.
    // Now GetChunk, PutChunk, will not allow any new IO.
    mv.quiesceIOs.Store(true)

    // Wait for any ongoing IOs to complete.
    now := time.Now()
    maxWait := 30 * time.Second

    for {
        if mv.chunkOpsCnt.Load() > 0 {
            time.Sleep(1 * time.Millisecond)

            elapsed := time.Since(now)
            if elapsed > maxWait {
                msg := fmt.Sprintf("%d ongoing IOs still pending after waiting for %s",
                                   mv.chunkOpsCnt.Load(), maxWait) 
                common.Assert(false, msg)
                return fmt.Errorf(msg)
            }
        } else {
            log.Info("%s quiesced successfully!", mv.mvName)
            break
        }
    }

    // Quiesced successfully, no ongoing IOs and no new IOs will be allowed.
    return nil
}

func (mv *mvInfo) quiesceIOsEnd() {
    // quiesceIOsEnd() must be called only after quiesceIOsStart().
    common.Assert(mv.quiesceIOs.Load() == true)
    mv.quiesceIOs.Store(false)
}
```

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+// block new chunk operations for this MV till the sync is in progress
+func (mv *mvInfo) blockChunkOps() {
+	for {
+		if mv.blockOpsFlag.Load() {
+			time.Sleep(100 * time.Microsecond) // TODO: check if this is optimal
+		} else {
+			break
+		}
+	}
+}
```
linuxsmiths, line 221:
```
// Block the calling thread if this MV is currently quiesced, by StartSync or EndSync.
func (mv *mvInfo) blockIOIfMVQuiesced() error {
    // Common path.
    if ! mv.quiesceIOs.Load() {
        return nil
    }

    // Wait till MV is quiesced.
    now := time.Now()
    maxWait := 30 * time.Second

    for {
        if mv.quiesceIOs.Load() {
            time.Sleep(1 * time.Millisecond)

            elapsed := time.Since(now)
            if elapsed > maxWait {
                msg := fmt.Sprintf("%s still quiesced after %s", mv.mvName, maxWait)
                common.Assert(false, msg)
                return fmt.Errorf(msg)
            }
        } else {
            break
        }
    }

    return nil
}
```

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
 }
 
+// increment the chunk operation (get, put or remove) count for this MV
+func (mv *mvInfo) incrementChunkOps() {
+	mv.chunkOpsCnt.Add(1)
+}
+
+// ddecrement the chunk operation (get, put or remove) count for this MV
+func (mv *mvInfo) decrementChunkOps() {
+	mv.chunkOpsCnt.Add(-1)
```
linuxsmiths, line 241: before the Add() assert that count is > 0

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+// - check if the MV is valid
+func (h *ChunkServiceHandler) checkValidChunkAddress(address *models.Address) error {
+	if address == nil || address.FileID == "" || address.RvID == "" || address.MvName == "" {
+		log.Err("ChunkServiceHandler::checkValidChunkAddress: Invalid chunk address %+v", *address)
+		return rpc.NewResponseError(rpc.InvalidRequest, fmt.Sprintf("invalid chunk address %+v", *address))
+	}
+
+	// check if the rvID is valid
+	rvInfo, ok := h.rvIDMap[address.RvID]
+	if !ok || rvInfo == nil {
```
linuxsmiths, line 294: if ok is true, rvInfo cannot be nil, let's assert

common.Assert(!ok || rvInfo !=nil, fmt.Sprintf("rvInfo nil for rvID %s", address.RvID)

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+		log.Err("ChunkServiceHandler::checkValidChunkAddress: Invalid rvID %s", address.RvID)
+		return rpc.NewResponseError(rpc.InvalidRVID, fmt.Sprintf("invalid rvID %s", address.RvID))
+	}
+
+	cacheDir := rvInfo.cacheDir
+	if cacheDir == "" || !common.DirectoryExists(cacheDir) {
+		log.Err("ChunkServiceHandler::checkValidChunkAddress: Cache dir not found for RV %s", rvInfo.rvName)
+		return rpc.NewResponseError(rpc.InternalServerError, fmt.Sprintf("cache dir not found for RV %s", rvInfo.rvName))
+	}
```
linuxsmiths, line 303: if this happens this is not client error, but it's our bug and we should assert for it, instead of handling the error.
If someone has really deleted the dir, it'll fail later

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	return nil
+}
+
+// get the RVInfo from the RV name
+func (h *ChunkServiceHandler) getRVInfoFromRvName(rvName string) *rvInfo {
+	var rvInfo *rvInfo
+	for _, info := range h.rvIDMap {
+		if info == nil {
+			continue
+		}
```
linuxsmiths, line 321: we should assert for it, instead of handling it

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+			break
+		}
+	}
+}
+
+// block the sync operation for this MV till the ongoing chunk operations are completed
+func (mv *mvInfo) blockSyncOps() {
+	for {
+		if mv.chunkOpsCnt.Load() > 0 {
+			time.Sleep(100 * time.Microsecond) // TODO: check if this is optimal
```
linuxsmiths, line 227: better to avoid extremely short sleeps. 1ms should be fine. See suggested code.

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	if !common.DirectoryExists(path) {
+		err := os.MkdirAll(path, 0700)
+		if err != nil {
+			log.Err("ChunkServiceHandler::createMVDirectory: Failed to create MV directory %s [%v]", path, err.Error())
+			return err
+		}
+	}
+
+	return nil
+}
```
linuxsmiths, line 347: don't need lock or the exist check, simply do

```
func (h *ChunkServiceHandler) createMVDirectory(path string) error {
    log.Debug("ChunkServiceHandler::createMVDirectory: Creating MV directory %s", path)

    if err := os.MkdirAll(path, 0700); err != nil && !os.IsExist(err) {
        return fmt.Errorf("MkdirAll(%s) failed: %v", path, err)
    }

    return nil
}
```

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+
+	fh, err := os.Open(chunkPath)
+	if err != nil {
+		log.Err("ChunkServiceHandler::GetChunk: Failed to open chunk file %s [%v]", chunkPath, err.Error())
+		return nil, rpc.NewResponseError(rpc.InternalServerError, fmt.Sprintf("failed to open chunk file %s [%v]", chunkPath, err.Error()))
+	}
+	defer fh.Close()
+
+	// TODO:: integration: call cluster manager to get chunk size
+	chunkSize := getChunkSize()
```
linuxsmiths, line 424: note that technically chunk size for different files can be different, so the only way to know the size is to do a stat() call

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	defer flock.Unlock()
+
+	cacheDir := rvInfo.cacheDir
+	chunkPath, hashPath := getChunkAndHashPath(cacheDir, req.Address.MvName, req.Address.FileID, req.Address.OffsetInMB)
+	log.Debug("ChunkServiceHandler::GetChunk: chunk path %s, hash path %s", chunkPath, hashPath)
+
+	fh, err := os.Open(chunkPath)
+	if err != nil {
+		log.Err("ChunkServiceHandler::GetChunk: Failed to open chunk file %s [%v]", chunkPath, err.Error())
+		return nil, rpc.NewResponseError(rpc.InternalServerError, fmt.Sprintf("failed to open chunk file %s [%v]", chunkPath, err.Error()))
```
linuxsmiths, line 419: how about "chunk not found"? That should be a specific error response not server error.

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	_, err = fh.ReadAt(data, req.Offset)
+	// TODO: should we handle EOF error
+	if err != nil {
+		log.Err("ChunkServiceHandler::GetChunk: Failed to read chunk file %s [%v]", chunkPath, err.Error())
+		return nil, rpc.NewResponseError(rpc.InternalServerError, fmt.Sprintf("failed to read chunk file %s [%v]", chunkPath, err.Error()))
+	}
+
+	// get hash if requested for entire chunk
+	hash := ""
+	if req.Offset == 0 && req.Length == chunkSize {
```
linuxsmiths, line 440: req.Length == chunkSize is not the right check for last chunks which may not be full size.
Better check would be to see if we read the full chunk from offset=0 till the end. For this we can use the file size calculated above using a stat() call.

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+		hashData, err := os.ReadFile(hashPath)
+		if err != nil {
+			log.Err("ChunkServiceHandler::GetChunk: Failed to read hash file %s [%v]", hashPath, err.Error())
+			return nil, rpc.NewResponseError(rpc.InternalServerError, fmt.Sprintf("failed to read hash file %s [%v]", hashPath, err.Error()))
+		}
+		hash = string(hashData)
+	}
+
+	lmt, err := getLMT(fh)
```
linuxsmiths, line 449: this makes a stat() call.
we can utilize this to get the mtime as well as size too.
make this before the read.

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	// validate the component RVs list
+	slices.Sort(req.ComponentRV)
+	if !isComponentRVsValid(mvInfo.componentRVs, req.ComponentRV) {
+		log.Err("ChunkServiceHandler::PutChunk: Request component RVs %v are invalid for MV %s component RVs %v", req.ComponentRV, req.Chunk.Address.MvName, mvInfo.componentRVs)
+		common.Assert(false, fmt.Sprintf("invalid component RVs for MV %s", req.Chunk.Address.MvName), req.ComponentRV, mvInfo.componentRVs)
+		return nil, rpc.NewResponseError(rpc.ComponentRVsInvalid, fmt.Sprintf("request component RVs %v are invalid for MV %s component RVs %v", req.ComponentRV, req.Chunk.Address.MvName, mvInfo.componentRVs))
+	}
+
+	// check if the sync has started. If yes, block the chunk operations till the sync is completed
+	mvInfo.blockChunkOps()
```
linuxsmiths, line 497: blockIOIfMVQuiesced()
also the comment is misleading, we do not wait till the sync completes but only for a very small time when StartSynd and EndSync are updating the sync info.

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
 	}
 }
 
+// increment the chunk operation (get, put or remove) count for this MV
+func (mv *mvInfo) incrementChunkOps() {
+	mv.chunkOpsCnt.Add(1)
+}
+
+// ddecrement the chunk operation (get, put or remove) count for this MV
+func (mv *mvInfo) decrementChunkOps() {
```
linuxsmiths, line 240: incOngoingIOs and decOngoingIOs are more readable. These can be confused for general stats.

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+
+	if createSyncDir {
+		//create sync directory if not present
+		syncDir := filepath.Join(cacheDir, req.Chunk.Address.MvName+".sync")
+		err = h.createMVDirectory(syncDir)
+		if err != nil {
+			log.Err("ChunkServiceHandler::PutChunk: Failed to create sync directory %s [%v]", syncDir, err.Error())
+			return nil, rpc.NewResponseError(rpc.InternalServerError, fmt.Sprintf("failed to create sync directory %s [%v]", syncDir, err.Error()))
+		}
+	}
```
linuxsmiths, line 549: this is best created in StartSync (and deleted in EndSync, after moving the chunks to the regular mv folder).
If StartSync succeeds, it effectively means that the RV is ready to store chunks from that MV. This includes having a directory to hold them (apart from having enough space)

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	// decrement the chunk operation count for this MV when the function returns
+	defer mvInfo.decrementChunkOps()
+
+	startTime := time.Now()
+
+	// acquire lock for the chunk address to prevent concurrent writes
+	chunkAddress := getChunkAddress(req.Chunk.Address.FileID, req.Chunk.Address.RvID, req.Chunk.Address.MvName, req.Chunk.Address.OffsetInMB)
+	flock := h.locks.Get(chunkAddress)
+	flock.Lock()
+	defer flock.Unlock()
```
linuxsmiths, line 511: let's discuss the locking, whether we need it

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+
+	// flag to block chunk operations (get, put or remove) for this MV.
+	// This flag is enabled in the end sync call to pause new chunk operations in the MV.
+	// When the end sync call is completed, this flag is disabled.
+	blockOpsFlag atomic.Bool
+
+	syncInfo // sync info for this MV
+}
+
+type syncInfo struct {
```
linuxsmiths, line 99: we should also store reservedSpace inside the syncInfo struct. This must be reduced from available space of the RV to get true available space. This means while sync is going on we might be overestimating the space utilization, but I think that's ok and later we can make it more accurate.

internal/dcache/rpc/server/handler.go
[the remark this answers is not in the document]
souravgupta-msft: Updated, thanks

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	// map to store the rvID to rvInfo mapping
+	// rvIDMap contains any and all cluster awareness information that a node needs to have,
+	// things like what all RVs are hosted by this node, state of each of those RVs,
+	// what all MVs are hosted by these RVs, state of those MVs, etc.
+	// It is initially created from the clustermap which is the source of truth regarding cluster information,
+	// and once the cluster is working it's updated using various RPCs.
+	// Note that any time Cluster Manager needs to update clustermap, before publishing the updated clustermap,
+	// it'll send our one or more RPCs to update the rvIDMap info in all the affected nodes,
+	// thus rvIDMap always contains the latest info and hence is used by RVs to fail requests
+	// which might be sent by nodes having a stale clustermap.
```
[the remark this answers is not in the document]
souravgupta-msft: Done, updated

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	rvName   string       // rv0, rv1, etc.
+	cacheDir string       // cache dir path for this RV
+	mvMap    sync.Map     // sync map of MV id against mvInfo
+	mvCount  atomic.Int64 // count of MVs for this RV, this should be updated whenever a MV is added or removed from the sync map
+}
+
+type mvInfo struct {
+	mu           sync.Mutex
+	mvName       string   // mv0, mv1, etc.
+	componentRVs []string // sorted list of component RVs for this MV
```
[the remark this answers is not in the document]
souravgupta-msft: Updated lock to `RWMutex` and using `RLock()` in `getComponentRVs` method

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+		locks:   common.NewLockMap(),
+		rvIDMap: rvIDMap,
+	}
+}
+
+// check if the given mv is valid
+func (rv *rvInfo) isMvValid(mvPath string) error {
+	if !common.DirectoryExists(mvPath) {
+		return fmt.Errorf("MV path %s does not exist", mvPath)
+	}
```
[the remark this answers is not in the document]
souravgupta-msft: Done

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	return mvInfo, nil
+}
+
+func (rv *rvInfo) addToMVMap(mvName string, val *mvInfo) {
+	rv.mvMap.Store(mvName, val)
+	rv.mvCount.Add(1)
+}
+
+func (rv *rvInfo) deleteFromMVMap(mvName string) {
+	rv.mvMap.Delete(mvName)
```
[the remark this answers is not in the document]
souravgupta-msft: Done, added asserts

internal/dcache/rpc/server/handler.go
souravgupta-msft: The caller of this method is `JoinMV` which already checks if the RV is part of the given MV, before calling this method. So, we don't need to check this again.
Also, added comment that the caller of this method must ensure that the RV is not part of the given MV.

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done, made the variable atomic

internal/dcache/rpc/server/handler.go
souravgupta-msft: In JoinMV, I am calling the cluster manager API/helper method to get the available the disk space for the given cache directory.
Even if we add `totalChunkBytes` field in the `mvInfo`, we will know the total space utilized by the RV (by calculating the sum of all MVs totalChunkBytes value). We would however need the total available space in the RV for it to join a MV. I am in this case relying on the cluster manager to give me that information, as it is already calculating it.

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done updated.

souravgupta-msft: > Some communication gaps can cause thus rule to break (if we are not able to call EndSync on an RV which went offline), and we should handle it

Do you mean we should this handle the case later?

internal/dcache/rpc/server/handler.go
souravgupta-msft: Updated to `sourceRVName`

internal/dcache/rpc/server/handler.go
linuxsmiths: this is not for finding RV free space, but it's for finding the space requirement of an MV when we have to find a replacement RV for it, the replacement RV must have this much free space.

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+
+	return nil
+}
+
+// update the component RVs for the MV
+func (mv *mvInfo) updateComponentRVs(componentRVs []string) {
+	mv.mu.Lock()
+	defer mv.mu.Unlock()
+
+	mv.componentRVs = componentRVs
```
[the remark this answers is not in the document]
souravgupta-msft: This should be safe, since we are not modifying (changing, appending or removing items) the list.
I have added a TODO to confirm this behavior.

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done

internal/dcache/rpc/server/handler.go
souravgupta-msft: Updated, added asserts instead of handling the error

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done, updated

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done, updated

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done

internal/dcache/rpc/server/handler.go
souravgupta-msft: req.Length is updated to chunkSize above using,
```
chunkSize := fInfo.Size() // got from stat

if req.Length == -1 {
	common.Assert(chunkSize > req.Offset, fmt.Sprintf("chunkSize %d is less than req.Offset %d", chunkSize, req.Offset))
	req.Length = chunkSize - req.Offset
}
```

So, this logic will work for last chunk also

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done, updated

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done, added the sync dir creation and deletion in StartSync and EndSync calls

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done, updated

internal/dcache/rpc/server/handler.go
souravgupta-msft: Updated by the method `quiesceIOsStart`

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done

internal/dcache/rpc/server/handler.go
linuxsmiths: when an mv is in syncing state we save incoming chunks to mv.sync while when not in syncing we save them to mv folder. i.e. depending on the syncing state the incoming chunks need to be saved o different folder.
So, when syncing state is changing (off to on or on to off) we need to be quiesce IOs so as not to add chunks to the wrong folder. I feel it's safe to do that.

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,987 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	componentRVs []string // sorted list of component RVs for this MV
+
+	// count of in-progress chunk operations (get, put or remove) for this MV.
+	// This is used to block the end sync call till all the ongoing chunk operations are completed.
+	chunkIOInProgress atomic.Int64
+
+	// flag to block chunk operations (get, put or remove) for this MV.
+	// This flag is enabled in the end sync call to pause new chunk operations in the MV.
+	// When the end sync call is completed, this flag is disabled.
+	blockOpsFlag atomic.Bool
```
linuxsmiths, line 97: not needed anymore

internal/dcache/rpc/server/handler.go
linuxsmiths: this assert should be placed before the if check, else it won't reach here

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,987 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+
+	// validate the component RVs list
+	slices.Sort(req.ComponentRV)
+	componentRVsInMV := mvInfo.getComponentRVs()
+	if !isComponentRVsValid(componentRVsInMV, req.ComponentRV) {
+		log.Err("ChunkServiceHandler::PutChunk: Request component RVs %v are invalid for MV %s component RVs %v", req.ComponentRV, req.Chunk.Address.MvName, componentRVsInMV)
+		common.Assert(false, fmt.Sprintf("invalid component RVs for MV %s", req.Chunk.Address.MvName), req.ComponentRV, componentRVsInMV)
+		return nil, rpc.NewResponseError(rpc.ComponentRVsInvalid, fmt.Sprintf("request component RVs %v are invalid for MV %s component RVs %v", req.ComponentRV, req.Chunk.Address.MvName, componentRVsInMV))
+	}
+
```
linuxsmiths, line 584: as explained in the write_mv() pseudo code in the doc, it's very important that the client sends the exact MV state including the RVs and their state, that it knows of (from the clustermap). We should then match it with what the RV feels and if there's a mismatch we should fail with an error code signifying NeedToRefreshClusterMap. The client on receiving this error know that it's copy of the clustermap is stale and then it fetches the latest clustermap copy and then retries,

internal/dcache/rpc/server/handler.go
souravgupta-msft: Updated

internal/dcache/rpc/server/handler.go
souravgupta-msft: Removed

internal/dcache/rpc/server/handler.go
souravgupta-msft: Updated the error code

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,980 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	rvName   string       // rv0, rv1, etc. [readonly]
+	cacheDir string       // cache dir path for this RV [readonly]
+	mvMap    sync.Map     // all MVs this RV is part of, indexed by MV name (e.g., "mv0"), updated by JoinMV and LeaveMV
+	mvCount  atomic.Int64 // count of MVs for this RV, this should be updated whenever a MV is added or removed from the sync map
+}
+
+type mvInfo struct {
+	rwMutex      sync.RWMutex
+	mvName       string   // mv0, mv1, etc.
+	componentRVs []string // sorted list of component RVs for this MV
```
linuxsmiths, line 88: make this a map[string]string

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,980 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	// This is used when the source RV goes offline and is replaced by the cluster manager with a new RV.
+	// In this case, the sync will need to be restarted from the new RV.
+	sourceRVName string // source RV name for syncing this MV
 }
 
 func NewChunkServiceHandler() *ChunkServiceHandler {
+	// TODO:: integration: get rvID, rvName and cache dir path for different RVs for the node from cluster manager
+	// below will be call to cluster manager to get the information
+	rvIDMap := getRvIDMap()
+
```
linuxsmiths, line 140: we can have cluster manager call the start method of this server, passing
(dCacheConfig *dcache.DCacheConfig, rvs []dcache.RawVolume)

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,980 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+		locks:   common.NewLockMap(),
+		rvIDMap: rvIDMap,
+	}
+}
+
+// check if the given mv is valid
+func (rv *rvInfo) isMvPathValid(mvPath string) bool {
+	mvName := filepath.Base(mvPath)
+	mvInfo := rv.getMVInfo(mvName)
+	common.Assert(mvInfo != nil || common.DirectoryExists(mvPath), fmt.Sprintf("mvPath %s MUST be present", mvPath))
```
linuxsmiths, line 151: common.Assert(mvInfo == nil || common.DirectoryExists(mvPath), fmt.Sprintf("mvPath %s MUST be present", mvPath))

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,980 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+		}
+
+		if mv.syncID != syncID {
+			msg := fmt.Sprintf("Unexpected EndSync(%s) received, expected syncID %s ", syncID, mv.syncID)
+			common.Assert(false, msg)
+			return fmt.Errorf("%s", msg)
+		}
+
+		// sourceRV must be valid.
+		common.Assert(mv.sourceRVName != "")
```
linuxsmiths, line 224: IsRVNameValid()

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,980 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+		mv.sourceRVName = ""
+
+		return nil
+	}
+
+	// StartSync.
+	// - Must not already be in syncing state.
+	common.Assert(syncID != "")
+	common.Assert(sourceRVName != "")
+
```
linuxsmiths, line 237: IsValidRVName()
IsValidUUID()

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,980 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+		log.Err("ChunkServiceHandler::GetChunk: Invalid chunk address [%s]", err.Error())
+		return nil, err
+	}
+
+	rvInfo := h.rvIDMap[req.Address.RvID]
+	mvInfo := rvInfo.getMVInfo(req.Address.MvName)
+
+	// Block the calling thread if this MV is currently quiesced
+	err = mvInfo.blockIOIfMVQuiesced()
+	common.Assert(err == nil, fmt.Sprintf("failed to block IO for MV %s", mvInfo.mvName))
```
linuxsmiths, line 476: TODO: [Race] After blockIOIfMVQuiesced() decides to proceed and before we increment the ongoing IO count below, if quiesceIOsStart() is run it will succesfully mark the IOs as quiesced. This leads us to a state where IOs are running while they should be quiesced.

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,980 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+func getNodeUUID() string {
+	return "node-uuid" // TODO: get the node uuid from the config
+}
+
+// check the if the chunk address is valid
+// - check if the rvID is valid
+// - check if the cache dir exists
+// - check if the MV is valid
+func (h *ChunkServiceHandler) checkValidChunkAddress(address *models.Address) error {
+	if address == nil || address.FileID == "" || address.RvID == "" || address.MvName == "" {
```
linuxsmiths, line 376: Add todo for IsValidUUID(), IsValidMVName(), ...

internal/dcache/rpc/response_error.go
```diff
@@ -0,0 +1,65 @@
...
+   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+   SOFTWARE
+*/
+
+package rpc
+
+const (
+	InvalidRequest      = iota + 1 // invalid rpc request
+	InvalidRVID                    // RV does not have the rvID
+	MVNotHostedByRV                // RV is not a member of the given MV
```
linuxsmiths, line 39: NeedToRefreshClusterMap

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,980 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	log.Debug("ChunkServiceHandler::PutChunk: chunk path %s, hash path %s", chunkPath, hashPath)
+
+	// check if the chunk file is already present
+	_, err = os.Stat(chunkPath)
+	if err == nil {
+		log.Err("ChunkServiceHandler::PutChunk: chunk file %s already exists", chunkPath)
+		return nil, rpc.NewResponseError(rpc.ChunkAlreadyExists, fmt.Sprintf("chunk file %s already exists", chunkPath))
+	}
+
+	err = os.WriteFile(chunkPath, req.Chunk.Data, 0400)
```
linuxsmiths, line 630: write to .tmp file and then rename

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,980 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
-	return nil, nil
+	if req == nil {
+		log.Err("ChunkServiceHandler::JoinMV: Received nil JoinMV request")
+		common.Assert(false, "received nil JoinMV request")
+		return nil, rpc.NewResponseError(rpc.InvalidRequest, "received nil JoinMV request")
+	}
+
+	log.Debug("ChunkServiceHandler::JoinMV: Received JoinMV request: %+v", *req)
+
+	if req.MV == "" || req.RV == "" || len(req.ComponentRV) == 0 {
```
linuxsmiths, line 738: req.RV -> req.RVName

internal/dcache/rpc/server/handler.go
souravgupta-msft: Discuss, made this array of pair having name and state.

internal/dcache/rpc/server/handler.go

internal/dcache/rpc/server/handler.go
souravgupta-msft: Added TODO

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done, added TODO

internal/dcache/rpc/response_error.go
souravgupta-msft: Updated

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,980 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	// decrement the chunk operation count for this MV when the function returns
+	defer mvInfo.decOngoingIOs()
+
+	startTime := time.Now()
+
+	// check if the chunk file is being updated in parallel by some other thread
+	chunkAddress := getChunkAddress(req.Address.FileID, req.Address.RvID, req.Address.MvName, req.Address.OffsetInMB)
+	flock := h.locks.Get(chunkAddress)
+	flock.Lock()
+	defer flock.Unlock()
```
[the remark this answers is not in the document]
souravgupta-msft: Updated the flow and removed lock

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done

internal/dcache/rpc/server/handler.go

internal/dcache/rpc/server/handler.go
```diff
@@ -133,7 +133,16 @@ type syncInfo struct {
...
 }
 
+var handler *ChunkServiceHandler
+
+// NewChunkServiceHandler creates a new ChunkServiceHandler instance
+// This is a singleton instance and is created only once.
+// Subsequent calls to this function will return the same instance.
 func NewChunkServiceHandler() *ChunkServiceHandler {
+	if handler != nil {
+		return handler
```
linuxsmiths, line 143: if we don't have a usecase of calling this more than once, let's assert here

internal/dcache/rpc/server/handler.go
```diff
@@ -647,6 +671,13 @@ func (h *ChunkServiceHandler) PutChunk(ctx context.Context, req *models.PutChunk
 
 	// TODO: should we verify the hash after writing the chunk
 
+	// rename the .tmp file to the final file
+	err = os.Rename(tmpChunkPath, chunkPath)
+	if err != nil {
+		log.Err("ChunkServiceHandler::PutChunk: Failed to rename chunk file %s to %s [%v]", tmpChunkPath, chunkPath, err.Error())
+		return nil, rpc.NewResponseError(rpc.InternalServerError, fmt.Sprintf("failed to rename chunk file %s to %s [%v]", tmpChunkPath, chunkPath, err.Error()))
```
linuxsmiths, line 678: let's assert to catch early

internal/dcache/rpc/server/utils.go
```diff
@@ -73,28 +76,73 @@ func getChunkAddress(fileID string, rvID string, mvName string, offsetInMB int64
...
-			return false
+	sortComponentRVs(rvInReq)
+
+	isValid := true
+	for i := 0; i < len(rvInMV); i++ {
+		common.Assert(rvInMV[i] != nil, "Component RV is nil")
+		common.Assert(rvInReq[i] != nil, "Component RV is nil")
+
+		if rvInMV[i].Name != rvInReq[i].Name || rvInMV[i].State != rvInReq[i].State {
+			isValid = false
```
linuxsmiths, line 110: a debug log would be helpful here to see what is different between the sender's component RVs and reciever's

linuxsmiths (approved):

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done

internal/dcache/rpc/server/handler.go
souravgupta-msft: We can call this more than once if we need information about the reserved space in an RV. I have made this singleton, so that if cluster manager wants, it can refer the same instance and get the required information about the RV

internal/dcache/rpc/server/handler.go
souravgupta-msft: Added TODO

internal/dcache/rpc/server/handler.go
souravgupta-msft: Added TODO

internal/dcache/rpc/server/handler.go
souravgupta-msft: Removed locks

internal/dcache/rpc/server/handler.go
souravgupta-msft: Added assert

internal/dcache/rpc/server/handler.go
souravgupta-msft: Added TODO, will add `totalChunkBytes` and `reserveSpace` in another PR

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done, added in StartSync

internal/dcache/rpc/server/handler.go
```diff
@@ -47,43 +57,894 @@ var _ service.ChunkService = &ChunkServiceHandler{}
...
+	}
+
+	// validate the component RVs list
+	slices.Sort(req.ComponentRV)
+	if !isComponentRVsValid(mvInfo.componentRVs, req.ComponentRV) {
+		log.Err("ChunkServiceHandler::PutChunk: Request component RVs %v are invalid for MV %s component RVs %v", req.ComponentRV, req.MV, mvInfo.componentRVs)
+		common.Assert(false, fmt.Sprintf("invalid component RVs for MV %s", req.MV), req.ComponentRV, mvInfo.componentRVs)
+		return nil, rpc.NewResponseError(rpc.ComponentRVsInvalid, fmt.Sprintf("request component RVs %v are invalid for MV %s component RVs %v", req.ComponentRV, req.MV, mvInfo.componentRVs))
+	}
+
```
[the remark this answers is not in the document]
souravgupta-msft: Added quiesceIOsStart() and quiesceIOsEnd() at both StartSync and EndSync

internal/dcache/rpc/server/handler.go
souravgupta-msft: Will do in next PR

internal/dcache/rpc/server/handler.go
souravgupta-msft: One more information it can get the totalChunkBytes inside an MV when making JoinMV call

PR #1709, opened 2025-04-24
File Manager Implementation

syeleti-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description

- Implementation of file manager
- Add Dcache File handle to handlemap that was used in the blobfuse to interact with the libfuse
- Move file structs to models package inside file manager as they were causing some dependency issues while integrating it with the handlemap package

## How Has This Been Tested?
More UT's to be added

## Checklist


internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,240 @@
...
+	bp := NewBufferPool(bufSize, maxBuffers)
+	fileIOMgr.wp = wp
+	fileIOMgr.bp = bp
+}
+
+func EndFileIOManager() {
+	fileIOMgr.wp.endWorkerPool()
+}
+
+func ReadChunkSync(offset int64, file *File) (*models.Chunk, error) {
```
linuxsmiths, line 139: dcache file_manager should export the following functions which the libfuse component should call.

```
func (f *File) Read(offset int64, data []byte) (int64, error)
func (f *File) Readahead(length int64) error
func (f *File) Write(offset int64, data []byte) error
```

Few things to note:
- *File manages its own chunk and we should not have that chunk information exposed outside the filemanager.
- Users of dcache filemanager should get a clean read/write interface with standard parameters and return values.
- Mapping those details to chunks should be hidden inside the filemanager

So, f.e. what Read() should do is, make sure the chunk(s) on which the read falls are present in file.Chunks. If not allocate new chunk(s) and add to file.Chunks. This should be done using an internal function named ensureChunks(offset int64, length int64) whose responsibility is to ensure that all chunks needed to cover this part of the file are present in file.Chunks.
Then it should check if the requested data is already present in the Chunks. Note that this means that each stored chunk in file.Chunks must also note whether it has the data or it's waiting for it to be downloaded.
Once we have the data, then copy the required range of data from the relevant Chunk to the []byte slice passed by Read().

Similarly, Write() should call ensureChunks to make sure that file.Chunks has the required chunk(s) allocated. 
Post that copy the data from []byte slice passed by Write() to the relevant part of the chunk. We should have a  copiedTill (or something appropriate) field which will tell what part of the chunk has been copied application data.
Once this covers full chunk, then we should schedule an upload of the chunk.

component/distributed_cache/distributed_cache.go
```diff
@@ -353,7 +355,26 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
...
+
+	// The metadata file was successfully created for the file
+	handle := handlemap.NewHandle(options.Name)
+	handle.DcacheFObj = file
+
+	return handle, nil
+}
+
 func (dc *DistributedCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Handle, error) {
+	if options.Flags&os.O_WRONLY != 0 || options.Flags&os.O_RDWR != 0 {
```
linuxsmiths, line 374: dcache can only support O_WRONLY and O_RDONLY.

internal/dcache/file_manager/buffer.go
```diff
@@ -0,0 +1,86 @@
...
+func NewBufferPool(bufSize int, maxBuffers int) *bufferPool {
+	var buf []byte
+	return &bufferPool{
+		pool: sync.Pool{
+			New: func() any {
+				return buf
+			},
+		},
+		bufSize:    bufSize,
+		emptyBuf:   make([]byte, bufSize),
```
linuxsmiths, line 61: why do we need emptyBuf?
Pool.New function should be defined as
```
                        func() any {
				return make([]byte, bp.bufSize)
			},
```

internal/dcache/file_manager/buffer.go
```diff
@@ -0,0 +1,86 @@
...
+		maxBuffers: int64(maxBuffers),
+	}
+}
+
+func (bp *bufferPool) getBuffer() ([]byte, error) {
+	if atomic.LoadInt64(&bp.numRequestedBuffers) > bp.maxBuffers {
+		log.Info("Distributed Cache::getBuffer : Buffers Exhausted")
```
linuxsmiths, line 68: we should assert here. The point being, this should be enforced through some other means and we should avoid a case where we ask for more buffers than the max.
Need to discuss.

internal/dcache/file_manager/buffer.go
```diff
@@ -0,0 +1,86 @@
...
+		return nil, errors.New("Buffers Exhausted")
+	}
+
+	buf := bp.pool.Get().([]byte)
+	if buf == nil {
+		buf = make([]byte, bp.bufSize)
+	} else {
+		// This can be remove in the future.
+		copy(buf, bp.emptyBuf)
+	}
```
linuxsmiths, line 78: simply use the value returned by pool.Get(). It handles the case where there are no objects in the pool and calls the New() method provided. You can do an assert that the slice returned by Get() has a length of 0 and capacity of bufSize.

You will need to reset the slice length to 0 before returning to the pool, in putBuffer()

internal/dcache/file_manager/buffer.go
```diff
@@ -0,0 +1,86 @@
...
+	bp.numRequestedBuffers++
+	return buf, nil
+}
+
+func (bp *bufferPool) putBuffer(buf []byte) {
+	bp.pool.Put(buf)
+	bp.numRequestedBuffers--
```
linuxsmiths, line 85: before decrement assert that numRequestedBuffers > 0

internal/dcache/file_manager/buffer.go
syeleti-msft: >why do we need emptyBuf?
It is used to reset the buffer when put into the pool and also used for filling holes when writing the file.

component/distributed_cache/distributed_cache.go
syeleti-msft: O_WRONLY must always come from the create file callback while writing.
Hence OpenFile can only handle O_RDONLY

internal/dcache/file_manager/buffer.go
syeleti-msft: Done

internal/dcache/file_manager/buffer.go
linuxsmiths: why would you need to do that?
All you need is to reset the slice, i.e., it should look exactly like how the New() method creates.
You can achieve that by the following

buf = buf[:0]

at L79

internal/dcache/file_manager/models/models.go
```diff
@@ -0,0 +1,58 @@
...
+}
+
+type Chunk struct {
+	Idx              int64      // chunk index
+	ChnkOffset       int64      // offset within chunk, used when reading partial chunk
+	Buf              []byte     // buf size == chunkSize
+	Err              chan error // Any Error while downloading/uploading
+	ScheduleDownload chan struct{}
+	ScheduleUpload   chan struct{}
+}
```
linuxsmiths, line 58: Let's call it FileChunk to distinguish from the regular notion of chunk that dcache has.
A chunk is an immutable entity, but here we are talking about the in-core entity which is meant to hold a chunk read from a file or which is being prepared for writing to dcache.
We need the following additional data to describe the FileChunk

```
	Len              int64      // valid bytes in Buf.
```

so a FileChunk would start with Len = 0  and then for the read case where FileChunk is populated from data read from dcache, Len will be set to the number of bytes read (this will mostly be the size of the actual dcache Chunk).
For the write case, again, Len starts at 0 and then we copy whatever fuse write handler passes us and update Len accordingly. We fail the write if fuse doesn't pass us data offset exactly matching FileChunk's Len (i.e. new data should write exactly after the last data written). And, once Len becomes equal to len(Buf) we schedule the chunk for upload.

linuxsmiths: I don't see how you plan to use ChnkOffset. With Len, do you feel you still need ChnkOffset?
Let's discuss.

internal/dcache/file_manager/buffer.go
linuxsmiths: We should not need that. Copying is the biggest perf killer, let's avoid it unless really needed.
I think we should set caller's expectation that the returned buffer is of length equal to the file chunk size and nothing about the content. The caller will anyway maintain how much data it has written in the supplied buffer and only use that much. See my comment on adding Len to FileChunk.

internal/dcache/file_manager/buffer.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+type bufferPool struct {
+	pool                sync.Pool
+	numRequestedBuffers int64
+	maxBuffers          int64
+	bufSize             int
+	emptyBuf            []byte // Used to reset the buffers and also used as a buf to
+	//fill the holes inside the file while writing
+}
```
linuxsmiths, line 52: general note on readability. Struct fields should be in the order of relevance. 
Suggest following refactor

```
type bufferPool struct {
	pool                sync.Pool   // sync.Pool to relieve GC
	bufSize             int         // size of buffers in this pool
	maxBuffers          int64       // max allocated buffers allowed
	curBuffers          int64       // buffers currently allocated
}
```

internal/dcache/file_manager/buffer.go
```diff
@@ -0,0 +1,83 @@
...
+
+func (bp *bufferPool) getBuffer() ([]byte, error) {
+	if atomic.LoadInt64(&bp.numRequestedBuffers) > bp.maxBuffers {
+		return nil, errors.New("Buffers Exhausted")
+	}
+
+	buf := bp.pool.Get().([]byte)
+	common.Assert(len(buf) == bp.bufSize, fmt.Sprintf("Allocated Buf Size %d != Required Buf Size %d", len(buf), bp.bufSize))
+	bp.numRequestedBuffers++
```
linuxsmiths, line 74: this needs to be atomically incremented and decremented

internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,372 @@
...
+	if offset >= file.FileMetadata.Size {
+		return 0, io.EOF
+	}
+	endOffset := min(offset+int64(len(buf)), file.FileMetadata.Size)
+	bufOffset := 0
+	for offset < endOffset {
+		//  Currently calling direct readaahead for the chunk assuming sequential workflow
+		chnkIdx := getChunkIdxFromFileOffset(offset, file.FileMetadata.FileLayout)
+		// todo : Detection of the sequential nature of the handle can be found from storing the offset
+		// in the handle.
```
linuxsmiths, line 102: Good seq/random pattern detection is more complicated than just comparing with the last offset read.
We can do it properly later. For now add a TODO asking for adding sequential pattern detection

internal/dcache/file_manager/buffer.go
syeleti-msft: I agree

internal/dcache/file_manager/buffer.go
syeleti-msft: addressed

internal/dcache/file_manager/buffer.go
syeleti-msft: makes sense

component/distributed_cache/distributed_cache.go
linuxsmiths: add a log stating the files in distributed cache are immutable and writing to them is not allowed.

internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,371 @@
...
+func (file *DcacheFile) ReadFile(offset int64, buf []byte) (bytesRead int, err error) {
+	log.Debug("DistributedCache::ReadFile : offset : %d, bufSize : %d, file : %s", offset, len(buf), file.FileMetadata.Filename)
+	if offset >= file.FileMetadata.Size {
+		return 0, io.EOF
+	}
+	endOffset := min(offset+int64(len(buf)), file.FileMetadata.Size)
+	bufOffset := 0
+	for offset < endOffset {
+		//  Currently calling direct readaahead for the chunk assuming sequential workflow
+		chnkIdx := getChunkIdxFromFileOffset(offset, file.FileMetadata.FileLayout)
```
linuxsmiths, line 100: chnk -> chunk

internal/dcache/file_manager/buffer.go
```diff
@@ -0,0 +1,83 @@
...
+	return buf, nil
+}
+
+func (bp *bufferPool) putBuffer(buf []byte) {
+	bp.pool.Put(buf)
+	common.Assert(bp.getCurBuffersCnt() > 0, fmt.Sprintf("Number of buffers are less than zero:  %d", bp.getCurBuffersCnt()))
+	bp.curBuffers.Add(-1)
+}
+
+func (bp *bufferPool) getCurBuffersCnt() int64 {
```
linuxsmiths, line 81: we can use load directly

internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,371 @@
...
+	log.Debug("DistributedCache::WriteFile : offset : %d, bufSize : %d, file : %s", offset, len(buf), file.FileMetadata.Filename)
+	if offset < file.lastWriteOffset {
+		log.Info("DistributedCache[FM]::WriteFile: File handle is not writing in sequential manner, current Offset : %d, Prev Offset %d", offset, file.lastWriteOffset)
+		return syscall.ENOTSUP
+	}
+	endOffset := (offset + int64(len(buf)))
+	bufOffset := 0
+	for offset < endOffset {
+		chnkIdx := getChunkIdxFromFileOffset(offset, file.FileMetadata.FileLayout)
+		chnk, err := file.writeChunk(chnkIdx)
```
linuxsmiths, line 128: createStagedChunk

internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,371 @@
...
+		if err != nil {
+			return err
+		}
+		chnkOffset := getChunkOffsetFromFileOffset(offset, file.FileMetadata.FileLayout)
+		copied := copy(chnk.Buf[chnkOffset:], buf[bufOffset:])
```
linuxsmiths, line 133: once the ful chunk has been copied to, then schedule an upload

component/distributed_cache/distributed_cache.go
```diff
@@ -353,7 +355,52 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
...
+		log.Debug("DistributedCache::CreateFile : Path is having Azure subcomponent, path : %s", options.Name)
+		options.Name = rawPath
+		_, err := dc.NextComponent().CreateFile(options)
+		if err != nil {
+			return nil, err
+		}
+	} else if isDcachePath {
+		log.Debug("DistributedCache::CreateFile : Path is having Dcache subcomponent, path : %s", options.Name)
+		var err error
+		file, err = fm.NewDcacheFile(rawPath)
```
linuxsmiths, line 371: here also we need to set options.Name = rawPath, as we use that in L391.
Better to do
options.Name = rawPath
and then use fm.NewDcacheFile(options.Name)
this makes it consistent with the azure case.

also, same for the else block.

component/distributed_cache/distributed_cache.go
```diff
@@ -353,7 +355,52 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
...
+			return nil, err
+		}
+		// todo : if one is success and other is failure, get to the previous state by removing the
+		// created entries for the files.
+	}
+
+	handle := handlemap.NewHandle(options.Name)
+	if isDcachePath {
+		handle.IFObj = file
+	}
```
linuxsmiths, line 394: we can do this unconditionally, so that we assign file for the else case also.

internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,361 @@
...
+	select {
+	case <-chunk.ScheduleDownload:
+	default:
+		close(chunk.ScheduleDownload)
+		t := &task{
+			file:      file,
+			chunk:     chunk,
+			get_chunk: true,
+		}
+		fileIOMgr.wp.tasks <- t
```
linuxsmiths, line 345: better to export a function QueueWork from workerpool and then call

fileIOMgr.wp.QueueWork(file, chunk, true)

and 

fileIOMgr.wp.QueueWork(file, chunk, false)

at L359

internal/dcache/file_manager/models.go
```diff
@@ -0,0 +1,43 @@
...
+
+package filemanager
+
+type StagedChunk struct {
+	Idx              int64      // chunk index
+	Buf              []byte     // buf size == chunkSize
+	Len              int64      // valid bytes in Buf
+	Err              chan error // Any Error while downloading/uploading
+	ScheduleDownload chan struct{}
+	ScheduleUpload   chan struct{}
```
linuxsmiths, line 42: I don't see how these are useful. Pls check the comment in scheduleDownload()

internal/dcache/file_manager/models.go
```diff
@@ -0,0 +1,43 @@
...
+   SOFTWARE
+*/
+
+package filemanager
+
+type StagedChunk struct {
+	Idx              int64      // chunk index
+	Buf              []byte     // buf size == chunkSize
+	Len              int64      // valid bytes in Buf
+	Err              chan error // Any Error while downloading/uploading
```
linuxsmiths, line 40:
```
Err              chan error // Status of downloading/uploading, available after upload/download completes, nil means sucess. 
```

internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,361 @@
...
+func NewFileIOManager(workers int, numReadAheadChunks int, numStagingChunks int, bufSize int, maxBuffers int) {
+	fileIOMgr = fileIOManager{
+		numReadAheadChunks: numReadAheadChunks,
+		numStagingChunks:   numStagingChunks,
+	}
+	wp := NewWorkerPool(workers)
+	wp.startWorkerPool()
+	bp := NewBufferPool(bufSize, maxBuffers)
+	fileIOMgr.wp = wp
+	fileIOMgr.bp = bp
```
linuxsmiths, line 71: nit,
why not directly assign to fileIOMgr.wp and fileIOMgr.bp

```
	fileIOMgr.wp := NewWorkerPool(workers)
	fileIOMgr.wp.startWorkerPool()
	fileIOMgr.bp := NewBufferPool(bufSize, maxBuffers)
```

internal/dcache/file_manager/worker.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+}
+
+func (wp *workerPool) startWorkerPool() {
+	log.Info("DistributedCache[FM]::startWorkerPool : Starting worker Pool, num workers : %d", wp.workers)
+	wp.wg.Add(wp.workers)
+	for range wp.workers {
+		go wp.worker()
+	}
+}
```
linuxsmiths, line 69: we don't have a usecase where we may need the workerpool w/o starting it for sometime, so no need to have this method, fold it inside NewWorkerPool

```
func NewWorkerPool(workers int) *workerPool {
    common.Assert(workers > 0)

    // Create the workerpool.
	wp := &workerPool{
		workers: workers,
		close:   make(chan struct{}),
		tasks:   make(chan *task, workers*2),
	}

    // Start the workers.
    log.Info("DistributedCache[FM]::NewWorkerPool : Starting worker Pool, num workers : %d", wp.workers)
	wp.wg.Add(wp.workers)
	for range wp.workers {
		go wp.worker()
	}

	return wp
}
```

internal/dcache/file_manager/worker.go
```diff
@@ -0,0 +1,123 @@
...
+
+func (wp *workerPool) endWorkerPool() {
```
linuxsmiths, line 71: now call this destroy() (opposite of new)

internal/dcache/file_manager/worker.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// Read From the Dcache
+	//Call MvRead method for reading the chunk.
+	// err = rm.MVRead()
+	if err == nil {
+		close(task.chunk.Err)
+		return
+	} else {
+		log.Info("DistrubuteCache[FM]::readChunk : Download of chunk to Dcache failed chnk idx : %d, file %s, err : %s",
+			task.chunk.Idx, task.file.FileMetadata.Filename, err.Error())
```
linuxsmiths, line 104: log.Err

internal/dcache/file_manager/utils.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"encoding/json"
+
+	"github.com/Azure/azure-storage-fuse/v2/common/log"
+	"github.com/Azure/azure-storage-fuse/v2/internal/dcache"
+	mm "github.com/Azure/azure-storage-fuse/v2/internal/dcache/metadata_manager"
+)
+
+func getChunkStartOffsetFromFileOffset(offset int64, fileLayout *dcache.FileLayout) int64 {
+	return (offset / fileLayout.ChunkSize) * fileLayout.ChunkSize
```
linuxsmiths, line 45: nit, following is more readable

```
return getChunkIdxFromFileOffset(offset, fileLayout) * fileLayout.ChunkSize
```

internal/dcache/file_manager/utils.go
```diff
@@ -0,0 +1,108 @@
...
+		log.Debug("DistributedCache::NewDcacheFile : File Creation failed for file :  %s with err : %s", fileName, err.Error())
+		return nil, err
+	}
+	return &DcacheFile{
+		FileMetadata: fileMetadata,
+	}, nil
+}
+
+// Creates the chunk and allocates the chunk buf
+func NewChunk(idx int64, file *DcacheFile) (*StagedChunk, error) {
```
linuxsmiths, line 95: NewStagedChunk  is more readable.

internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,361 @@
...
+	}
+	endOffset := min(offset+int64(len(buf)), file.FileMetadata.Size)
+	bufOffset := 0
+	for offset < endOffset {
+		//  Currently calling direct readaahead for the chunk assuming sequential workflow
+		// todo : Add Sequential pattern detection.
+		// todo : Support Partial read of chunk, useful for random read scenarios.
+		chunk, err := file.readChunkWithReadAhead(offset)
+		if err != nil {
+			return 0, nil
```
linuxsmiths, line 103: why treat it as eof?
couldn't this be a real read error?

internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,361 @@
...
+		}
+		chunkOffset := getChunkOffsetFromFileOffset(offset, &file.FileMetadata.FileLayout)
+		copied := copy(buf[bufOffset:], chunk.Buf[chunkOffset:chunk.Len])
```
linuxsmiths, line 106: we cannot copy chunk.Len, caller may have asked for less.
do this

Instead of defining endOffset and then comparing offset < endOffset, define a new variable

bytesToCopy := min(int64(len(buf)), file.FileMetadata.Size - offset)

and then do

copied := copy(buf[bufOffset:], chunk.Buf[chunkOffset:min(chunk.Len, bytesToCopy)])

internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,361 @@
...
+	common.Assert(offset <= file.FileMetadata.Size, "Read beyond the file size")
+	return bufOffset, nil
+}
+
+// Writes the file to the corresponing chunk(s) from the buf.
+func (file *DcacheFile) WriteFile(offset int64, buf []byte) error {
+	log.Debug("DistributedCache[FM]::WriteFile : offset : %d, bufSize : %d, file : %s", offset, len(buf), file.FileMetadata.Filename)
+
+	if offset < file.lastWriteOffset {
+		log.Info("DistributedCache[FM]::WriteFile: File handle is not writing in sequential manner, current Offset : %d, Prev Offset %d", offset, file.lastWriteOffset)
```
linuxsmiths, line 119: Pls use error logs for error conditions

internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,361 @@
...
+		return syscall.ENOTSUP
+	}
+
+	endOffset := (offset + int64(len(buf)))
+	bufOffset := 0
+	for offset < endOffset {
+		chunk, err := file.CreateOrGetStagedChunk(offset)
+		if err != nil {
+			return err
```
linuxsmiths, line 128: log or return a formatted error

internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,361 @@
...
+
+// Sync Buffers for the file with dcache/azure
+// This call can come when user application calls fsync()/close()
+func (file *DcacheFile) SyncFile() error {
+	log.Debug("DistributedCache[FM]::SyncFile : Sync File for %s", file.FileMetadata.Filename)
+	var err error
+	file.StagedChunks.Range(func(chunkIdx any, Ichunk any) bool {
+		chunk := Ichunk.(*StagedChunk)
+		scheduleUpload(chunk, file)
+		err = <-chunk.Err
```
linuxsmiths, line 169: his will cause upload to be synchronized. We should try to do parallel uploads.
add a TODO for now, since sync is not a common case for us

internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,361 @@
...
+	})
+	return err
+}
+
+// Release all allocated buffers for the file
+func (file *DcacheFile) ReleaseFile() error {
+	log.Debug("DistributedCache[FM]::ReleaseFile :Releasing buffers for File for %s", file.FileMetadata.Filename)
+	file.StagedChunks.Range(func(chunkIdx any, Ichunk any) bool {
+		chunk := Ichunk.(*StagedChunk)
+		file.releaseChunk(chunk)
```
linuxsmiths, line 183: releasing w/o upload could lead to data corruption, so first do a SyncFile()

internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,361 @@
...
+}
+
+func (file *DcacheFile) getChunkForWrite(chunkIdx int64) (*StagedChunk, error) {
+	log.Debug("DistributedCache::getChunkForWrite : getChunk for Write chunkIdx : %d, file : %s", chunkIdx, file.FileMetadata.Filename)
+	chunk, loaded, err := file.getChunk(chunkIdx)
+	if err == nil && !loaded {
+		close(chunk.ScheduleDownload)
+	}
+	return chunk, err
+}
```
linuxsmiths, line 224: since I don't understand ScheduleUpload and ScheduleDownload, I don't understand why we need thise and not just getChunk()

component/distributed_cache/distributed_cache.go
syeleti-msft: done

internal/dcache/file_manager/models.go
syeleti-msft: modified

internal/dcache/file_manager/file_manager.go
syeleti-msft: modified

internal/dcache/file_manager/worker.go
syeleti-msft: modified

internal/dcache/file_manager/worker.go
syeleti-msft: modified

internal/dcache/file_manager/utils.go
syeleti-msft: modified

internal/dcache/file_manager/file_manager.go
syeleti-msft: It's a read error, EOF is handled explicitly at the start of this func.
modified by Returning the error

internal/dcache/file_manager/file_manager.go
syeleti-msft: Sure, modified

internal/dcache/file_manager/file_manager.go
```diff
@@ -0,0 +1,361 @@
...
+		copied := copy(chunk.Buf[chunkOffset:], buf[bufOffset:])
+		offset += int64(copied)
+		bufOffset += copied
+		chunk.Len += int64(copied)
+
+		common.Assert(chunk.Len == getChunkOffsetFromFileOffset(offset, &file.FileMetadata.FileLayout),
+			fmt.Sprintf("Acutal Chunk Len : %d is modified incorrectly, Expected chunkLen : %d",
+				chunk.Len, getChunkStartOffsetFromFileOffset(offset, &file.FileMetadata.FileLayout)))
+
+		if chunk.Len == file.FileMetadata.FileLayout.ChunkSize {
```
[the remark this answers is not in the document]
syeleti-msft: modified

internal/dcache/file_manager/worker.go
syeleti-msft: I agree, modified

internal/dcache/file_manager/utils.go
syeleti-msft: modified

internal/dcache/file_manager/file_manager.go
syeleti-msft: > better to export a function QueueWork from workerpool and then call

Sure

> I don't see the point of the channel ScheduleDownload (and ScheduleUpload).

Currently these channels were used for scheduling the upload/download only once per chunk

internal/dcache/file_manager/file_manager.go
syeleti-msft: It is guaranteed that release call will only be called after syncFile.

internal/dcache/file_manager/models.go
syeleti-msft: Currently these channels were used for scheduling the upload/download only once per chunk

internal/dcache/file_manager/file_manager.go
syeleti-msft: we can discuss this.

internal/dcache/file_manager/file_manager.go
linuxsmiths: let's assert for that

internal/dcache/file_manager/models.go
linuxsmiths: ok, add a TODO that we should replace these with an atomic flag

linuxsmiths (approved):

PR #1711, opened 2025-04-25
Replication manager initial code

souravgupta-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description
Added ReadMV and WriteMV workflows of replication manager.

## Checklist


internal/dcache/replication_manager/models.go
```diff
@@ -0,0 +1,61 @@
...
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+   SOFTWARE
+*/
+
+package replication_manager
+
+type ReadMvRequest struct {
+	FileID        string // unique guid of the file stored in metadata blob
+	RvID          string // id of the RV; can be filesystem guid
```
linuxsmiths, line 38: we don't want File Manager (rather anyone outside Replication Manager, and for some maintenance things, even Cluster Manager) to be aware of RVs. It should only know about MVs and express  address of data to read/write as offset and length within the MV. RM should then find the best RV, make appropriate chunk address and issue a GetChunk requests.
So I think what you have is good, just drop RvID

Also, why

internal/dcache/replication_manager/models.go
```diff
@@ -0,0 +1,61 @@
...
+	MvName        string // mv0, mv1, etc.
+	Offset        int64  // Offset in bytes
```
linuxsmiths, line 40: clarify that it's offset within MV and not offset within the file

internal/dcache/replication_manager/models.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+type ReadMvResponse struct {
+	Data []byte
+	Hash string
+}
+
+type WriteMvRequest struct {
+	FileID        string // unique guid of the file stored in metadata blob
+	RvID          string // id of the RV; can be filesystem guid
```
linuxsmiths, line 52: remove

internal/dcache/replication_manager/models.go
souravgupta-msft: It is not being used anywhere. Removed it

internal/dcache/replication_manager/models.go
souravgupta-msft: Updated, added `ChunkIndex` and `OffsetInChunk` parameters

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -0,0 +1,213 @@
...
+		log.Err("ReplicationManager::WriteMV: No component RVs found for the given MV %s", req.MvName)
+		common.Assert(false, "no component RVs found for the given MV", req.MvName)
+		return nil, fmt.Errorf("no component RVs found for the given MV %s", req.MvName)
+	}
+
+	// check if the mv is online. This is done by checking if all the component RVs are online.
+	// If any of the component RVs are offline, then fail this request.
+	// The caller of WriteMV should make sure that the MV is online at the time calling this function.
+	// This does not ensure that the MV will be online at the time of writing.
+	// If any of the PutChunk RPC calls fail, then the degradeMV method should be called.
```
linuxsmiths, line 157: We do allow writing to a degraded or syncing MV, so online MV is not a requirement.
The only requirement is that MV must not be offline (offline MV is one whose all component RVs are offline), writes to offline MV fail rightaway.

Regd'ing RVs, we have the following rules for writing:
- For an "online" and "syncing" RVs we can do PutChunk
- For "offline" and "outofsync" RVs write need to wait till it becomes "syncing". That will happen once the fix-mv and resync-mv workflow is run. So we will need a way to queue these writes and be woken up once those workflows run

Also, if a write to an "online" or "syncing" RV fails with an error signifying RV is offline, then we need to retry the write after running the degrade-mv + fix-mv + resync-mv workflow.

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -0,0 +1,213 @@
...
+	if !rvsValid {
+		log.Err("ReplicationManager::WriteMV: Not all component RVs are online for the given MV %s", req.MvName)
+		common.Assert(false, "not all component RVs are online for the given MV", req.MvName)
+		return nil, fmt.Errorf("not all component RVs are online for the given MV %s", req.MvName)
+	}
+
+	log.Debug("ReplicationManager::WriteMV: Component RVs for the given MV %s are: %v", req.MvName, componentRVs)
+
+	// get hash of the data in the request
+	hash := getMD5Sum(req.Data)
```
linuxsmiths, line 168: we can omit the hashing part right now, as we will need to use config for hashing also also

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -0,0 +1,213 @@
...
+
+	resp := &ReadMvResponse{
+		// TODO: update this filed after bytes read in response
+		BytesRead: int64(len(rpcResp.Chunk.Data)),
+	}
+
+	return resp, nil
+}
+
+func WriteMV(req *WriteMvRequest) (*WriteMvResponse, error) {
```
linuxsmiths, line 131: Pls take a look at the write_mv() pseudo code added to the doc

internal/dcache/replication_manager/models.go
```diff
@@ -0,0 +1,181 @@
...
+	MvName string // name of the MV which has this chunk. For example, mv0, mv1, etc.
+
+	// ChunkIndex is the name of the chunk to be read. It should be multiple of chunk size.
+	// For example, if chunk size is 4MB, then chunk index should be 0, 4, 8, etc.
+	// The chunks are stored in RV as,
+	// <mvName>/<fileID>.<chunkIndex>.data and
+	// <mvName>/<fileID>.<chunkIndex>.hash
+	ChunkIndex    int64
+	OffsetInChunk int64 // read offset within the chunk. This should not be greater than chunk size
+	Length        int64 // Length in bytes of data to be read. If length is -1, then read till the end of the chunk from offsetInChunk
```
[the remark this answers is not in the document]
linuxsmiths: Pls consider the following. This has good comments and clearly explains how caller can use it

```
type ReadMvRequest struct {
	FileID string // unique guid of the file, as stored in metadata blob
	MvName string // name of the MV to be read, e.g., "mv0", "mv1", etc.

	//
	// An MV can hold multiple chunks of the file (one from each stripe).
	// ChunkIndex identifies the exact chunk to be read. It is the index of the chunk
	// within the file and is 0 based.
	// e.g., for stripe size of 16MiB and ChunkSizeInMiB == 4,
	// file offset of 1MiB would translate to the following:
	//  MvName: "mv0"   (assuming first MV in the file's mv list is "mv0")
	//  ChunkIndex: 0
	//  OffsetInChunk: 1048576 (1MiB)
	// while a file offset of 17MiB would translate to the following:
	//  MvName: "mv0"
	//  ChunkIndex: 4
	//  OffsetInChunk: 1048576 (1MiB)
	//
	// The chunks are stored in RV as,
	// <MvName>/<FileID>.<ChunkIndex * ChunkSizeInMiB>.data and
	// <MvName>/<FileID>.<ChunkIndex * ChunkSizeInMiB>.hash
	//
	ChunkIndex    int64
	OffsetInChunk int64     // read offset within the chunk. This should not be greater than ChunkSizeInMiB
	Length        int64     // Length in bytes of data to be read
	ChunkSizeInMiB int64    // Chunk size in MiB

	Data []byte             // buffer to store the data read from the chunk. Must be at least Length bytes
}
```

linuxsmiths (approved):

PR #1712, opened 2025-04-25
CI Pipeline 1ES migration

vibhansa-msft:
CI Pipeline migration to 1ES template.

### DIFF
--- a/azure-pipeline-templates/build.yml
+++ b/azure-pipeline-templates/build.yml
@@ -187,10 +187,10 @@ steps:
     retryCountOnTaskFailure: 3
 
   # azcli login
-  - script: |
-      az login --identity --username $(ACLI_BLOBFUSE_MSI_APP_ID)
-    displayName: 'Azure CLI login'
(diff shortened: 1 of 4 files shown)


azure-pipeline-templates/build.yml
```diff
@@ -187,10 +187,10 @@ steps:
     retryCountOnTaskFailure: 3
 
   # azcli login
-  - script: |
-      az login --identity --username $(ACLI_BLOBFUSE_MSI_APP_ID)
-    displayName: 'Azure CLI login'
-    condition: eq('${{ parameters.skip_azcli }}', 'false')
+  # - script: |
```
vibhansa-msft, line 190: MSI login using username field is depricated and if I migrate this step to work with client-id it fails with 400 error. Tried this step manually on vm as well but it consistently fails to login. Hence for now we are skipping this part and cli login based test.

jainakanksha-msft (approved):

PR #1715, opened 2025-04-28
Cluster Manager API Implementation

jainakanksha-msft:


common/util.go
```diff
@@ -610,10 +610,7 @@ func GetNodeUUID() (string, error) {
 			return "", fmt.Errorf("fail to read UUID File at :%s with error %s", uuidFilePath, err)
 		}
 		stringData := string(data)
-		isValidGUID, err := IsValidUUID(stringData)
-		if err != nil {
-			return "", fmt.Errorf("IsValidUUID failed for %s: %v", stringData, err)
-		}
+		isValidGUID := IsValidUUID(stringData)
```
linuxsmiths, line 613: nit, pls use UUID consistently, so name the local variable also isValidUUID

common/util.go
```diff
@@ -633,13 +630,10 @@ func GetNodeUUID() (string, error) {
...
-	valid, err := regexp.MatchString(
+func IsValidUUID(guid string) bool {
+	valid, _ := regexp.MatchString(
 		"^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$", guid)
-	if err != nil {
-		return false, err
-	}
-	return valid, nil
+	return valid
 }
```
linuxsmiths, line 637: for completeness, let's not ignore err

```
func IsValidUUID(guid string) bool {
    valid, err := regexp.MatchString(
        "^[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}$", guid)
    // valid should not be true if there's an error.
    Assert((err == nil) || !valid)

    return (err != nil) && valid
}
```

internal/dcache/cluster_manager/utils.go
```diff
@@ -0,0 +1,136 @@
...
+import (
+	"fmt"
+
+	"github.com/Azure/azure-storage-fuse/v2/common"
+	"github.com/Azure/azure-storage-fuse/v2/internal/dcache"
+)
+
+func IsValidClusterMap(cm dcache.ClusterMap) (bool, string) {
+	// top‐level fields
+	if cm.CreatedAt <= 0 {
```
linuxsmiths, line 45: better to also check if the time epoch is in the future.
we can safely check for 2050

```
minEpoch := 1735689600  // Jan 1 2025, safe lowerlimit for epoch validity check
maxEpoch := 2524608000  // Jan 1 2050, safe upperlimit for epoch validity check

```

and then check 
if cm.CreatedAt < minEpoch || cm.CreatedAt > maxEpoch

same for other epochs

internal/dcache/cluster_manager/utils.go
```diff
@@ -0,0 +1,136 @@
...
+			default:
+				return false, fmt.Sprintf("ClusterMap:: [%+v] MVMap[%q].Rvs[%q]: Invalid RV state %q in MV ", cm, name, rvName, state)
+			}
+		}
+	}
+
+	return true, ""
+}
+
+func IsValidRVMap(rVMap map[string]dcache.RawVolume) (bool, bool, string) {
```
linuxsmiths, line 102: why two boolean returns? it's confusing and doesn't seem to add value.
Why don't we have the usual (bool, err) return. The string can be returned as a formatted error, which is more readable.

internal/dcache/cluster_manager/utils.go
```diff
@@ -0,0 +1,136 @@
...
+	seen := make(map[string]string, len(rVMap))
+	for rvName, rv := range rVMap {
+		if prev, ok := seen[rv.RvId]; ok {
+			return true, false, fmt.Sprintf(
+				"ClusterMap::RVMap [%+v] duplicate RvId %q found in RVMap entries %q and %q", rVMap, rv.RvId, prev, rvName,
+			)
+		}
+		seen[rv.RvId] = rvName
+		if rv.RvId == "" {
+			return true, false, fmt.Sprintf("ClusterMap::RVMap [%+v] RVMap[%q]: empty RvId", rVMap, rvName)
```
linuxsmiths, line 112: no need to special case this

internal/dcache/cluster_manager/utils.go
```diff
@@ -0,0 +1,136 @@
...
+			if rvName == "" {
+				return false, fmt.Sprintf("ClusterMap:: [%+v] MVMap[%q].Rvs: empty key", cm, name)
+			}
+			switch state {
+			case string(dcache.StateOnline), string(dcache.StateOffline):
+			default:
+				return false, fmt.Sprintf("ClusterMap:: [%+v] MVMap[%q].Rvs[%q]: Invalid RV state %q in MV ", cm, name, rvName, state)
+			}
+		}
+	}
```
linuxsmiths, line 97: for consistency, move this to IsValidMVMap()

internal/dcache/cluster_manager/utils.go
```diff
@@ -0,0 +1,136 @@
...
+		}
+		for rvName, state := range mv.Rvs {
+			if rvName == "" {
+				return false, fmt.Sprintf("ClusterMap:: [%+v] MVMap[%q].Rvs: empty key", cm, name)
+			}
+			switch state {
+			case string(dcache.StateOnline), string(dcache.StateOffline):
+			default:
+				return false, fmt.Sprintf("ClusterMap:: [%+v] MVMap[%q].Rvs[%q]: Invalid RV state %q in MV ", cm, name, rvName, state)
+			}
```
linuxsmiths, line 95: just checking that rvName is prsent in cm.RVMap should be sufficient as we have already verified that all RVs in cm.RVMap are valid.

but RVs in MV list can have the following additional states

outofsync, syncing

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -88,4 +91,7 @@ type ClusterManager interface {
 
 	//Update RV state to offline and update MVs
 	reportRVFull(rvName string) error
+
+	//Notify consumer about cluster manager Event
+	notifyUpdates() <-chan dcache.ClusterManagerEvent
```
linuxsmiths, line 96: getClusterMapUpdateNotificationChannel() is a more appropriate name, as the caller will call this to get the notification channel for getting clustermap update notifications and then read from the channel for any notifications.

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -228,61 +259,182 @@ func (cmi *ClusterManagerImpl) stop() error {
...
+	}
 	return nil
 }
 
 // getActiveMVs implements ClusterManager.
-func (c *ClusterManagerImpl) getActiveMVs() []dcache.MirroredVolume {
-	return make([]dcache.MirroredVolume, 0)
+func (cmi *ClusterManagerImpl) getActiveMVs() map[string]dcache.MirroredVolume {
+	if cmi.localMap == nil {
+		log.Debug("ClusterManagerImpl::getActiveMVs: local cluster map not loaded")
```
linuxsmiths, line 271: this shouldn't happen and should be asserted.
simply do

``
common.Assert(cmi.localMap != nil)
```

File: internal/dcache/cluster_manager/cluster_manager_impl.go [hunk 10]
```diff
@@ -228,61 +259,182 @@ func (cmi *ClusterManagerImpl) stop() error {
...
-	return make([]dcache.MirroredVolume, 0)
+func (cmi *ClusterManagerImpl) getDegradedMVs() map[string]dcache.MirroredVolume {
+
+	if cmi.localMap == nil {
+		log.Debug("ClusterManagerImpl::getDegradedMVs: local cluster map not loaded")
+		return nil
+	}
+	degradedMVs := make(map[string]dcache.MirroredVolume)
+	for mvName, mv := range cmi.localMap.MVMap {
+		if mv.State != dcache.StateOnline {
```
linuxsmiths, line 301: MV state might be online, offline, degraded, syncing

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -228,61 +259,182 @@ func (cmi *ClusterManagerImpl) stop() error {
...
 	return false
 }
 
 // lowestNumberRV implements ClusterManager.
-func (c *ClusterManagerImpl) lowestNumberRV(rvNames []string) []string {
-	return make([]string, 0)
+func (c *ClusterManagerImpl) lowestNumberRV(rvNames []string) string {
+	log.Debug("ClusterManagerImpl::lowestNumberRV: evaluating lowest number rvName %v", rvNames)
+	lowestNumberRv := ""
+	var min int
```
linuxsmiths, line 356: min must be set to math.MaxInt32 else following code won't work

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -228,61 +259,182 @@ func (cmi *ClusterManagerImpl) stop() error {
...
+	for _, name := range rvNames {
+		num, _ := strconv.Atoi(strings.TrimPrefix(name, "rv"))
```
linuxsmiths, line 358: any function that returns error and some other value, we cannot check the other value w/o making sure err is nil

internal/dcache/cluster_manager/cluster_manager_impl.go
```diff
@@ -228,61 +259,182 @@ func (cmi *ClusterManagerImpl) stop() error {
...
+		if num < min {
+			min = num
+			lowestNumberRv = name
+		}
+	}
+	log.Debug("ClusterManagerImpl::lowestNumberRV: lowest number rvName is %s", lowestNumberRv)
```
linuxsmiths, line 364:
```
log.Debug("ClusterManagerImpl::lowestNumberRV: lowest number rvName in %v is %s", rvNames, lowestNumberRv)
```

jainakanksha-msft: I have already printed the list earlier in Debug log.

PR #1719, opened 2025-04-30
Implement FUSE IO callbacks

syeleti-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description

Implement FUSE callbacks

## Checklist


component/distributed_cache/models.go
```diff
@@ -0,0 +1,70 @@
...
+	cacheAccessAzure  uint16 = iota // tells handle that file can access Azure.
+	cacheAccessDcache               // tells handle that file can access dcache.
+)
+
+type file struct {
+	dcFile *filemanager.DcacheFile // Dcache file
+	access common.BitMap16         // access flags that handles its access permissions for azure/dcache
+}
+
+func (file *file) SetAzureAccessFlag() {
```
linuxsmiths, line 51: SetFsAzure
SetFsDcache
SetFsDefault

IsFsAzure
IsFsDcache

component/distributed_cache/distributed_cache.go
```diff
@@ -152,6 +154,8 @@ func (dc *DistributedCache) Start(ctx context.Context) error {
 		return log.LogAndReturnError(errString)
 	}
 	log.Info("DistributedCache::Start : component started successfully")
+	// todo : Replace the hardcoded values with user config values.
```
linuxsmiths, line 157: also add a todo for moving this to a init() function in file_manager.go

syeleti-msft: added

component/distributed_cache/distributed_cache.go
```diff
@@ -356,68 +361,265 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
...
 		}
 
-		_, err = dc.NextComponent().CreateFile(options)
+		handle, err = dc.NextComponent().CreateFile(options)
 		if err != nil {
+			log.Err("DistributedCache::CreateFile : File Creation failed with err : %s Azure subcomponent, path : %s", err.Error(), options.Name)
 			return nil, err
 		}
 		// todo : if one is success and other is failure, get to the previous state by removing the
 		// created entries for the files.
```
linuxsmiths, line 398: Create should fail if either of the create fails. Since user didn't explicitly specify the namespace, our semantics is that we will create the file both in cache as well as azure, so failing to create in either should cause create to fail.

component/distributed_cache/distributed_cache.go
```diff
@@ -356,68 +361,265 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
...
+		dcFile, err = fm.OpenDcacheFile(rawPath)
+		if err != nil {
+			log.Err("DistributedCache::OpenFile : File Open failed with err : %s Dcache subcomponent, path : %s", err.Error(), options.Name)
+			return nil, err
+		}
+		handle, err = dc.NextComponent().OpenFile(options)
+		if err != nil {
+			log.Err("DistributedCache::OpenFile : File Open failed with err : %s Azure subcomponent, path : %s", err.Error(), options.Name)
+			return nil, err
+		}
```
linuxsmiths, line 460: opening file for read with no explicit namespace specified, must have the following semantics.
It should first check the file in dcache, if present, read from dcache, else check in azure if present, read from azure, else fail the open.
So we should do this here

open in dcache
if it succeeds, call SetFsDcache() and set handle.IFObj (to indicate this file must be read from dcache) and return.
if this fails with enoent then try to open in azure, else if it fails with any other error, fail the open
if azure open fails, fail the open call, else call SetFsAzure() to indicate this file must be read from azure.

component/distributed_cache/distributed_cache.go
```diff
@@ -356,68 +361,265 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
...
+		bytesRead, dcacheErr = dcFile.ReadFile(options.Offset, options.Data)
+		if dcacheErr == nil || dcacheErr == io.EOF {
+			return bytesRead, dcacheErr
+		}
+		common.Assert(bytesRead == 0)
+		log.Err("DistributedCache::ReadInBuffer : Failed to read the file from the Dcache, offset : %d, file : %s",
+			options.Offset, options.Handle.Path)
+		// Let's try to read from the azure, if the handle has the access
+	}
+	if options.Handle.IsFsAzure() {
```
linuxsmiths, line 500: see the comment in OpenFile().
The decision to read from dcache or azure is already made in OpenFile().
In ReadInBuffer(), one and only one of fs azure or dcache must be set and we should just read from that.

component/distributed_cache/distributed_cache.go
```diff
@@ -356,68 +361,265 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
...
+	err := errors.Join(dcacheErr, azureErr)
+	common.Assert(err != nil)
+	return 0, err
 }
 
 func (dc *DistributedCache) WriteFile(options internal.WriteFileOptions) (int, error) {
-	return 0, syscall.ENOTSUP
+	log.Err("DistributedCache::WriteFile : WriteFile, offset : %d, buf size : %d, file : %s",
+		options.Offset, len(options.Data), options.Handle.Path)
+
```
linuxsmiths, line 516: handle/assert 0 byte writes?

component/distributed_cache/distributed_cache.go
```diff
@@ -356,68 +361,265 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
...
-	handle.IFObj = file
+	// Only the files created with dcache needs flush.
+	handle.Flags.Set(handlemap.HandleFlagDcacheAllowWrites)
 
 	return handle, nil
 }
 func (dc *DistributedCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Handle, error) {
 	if options.Flags&os.O_WRONLY != 0 || options.Flags&os.O_RDWR != 0 {
-		log.Info("DistributedCache::OpenFile: Writing to an existing File is not allowed, file : %s", options.Name)
+		log.Err("DistributedCache::OpenFile: Writing to an exisiting File is not allowed, file : %s", options.Name)
```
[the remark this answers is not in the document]
linuxsmiths: Also, we should fail this only if the write is to dcache file. Writes to fs=azure should be allowed, no?

component/distributed_cache/distributed_cache.go
```diff
@@ -356,68 +361,265 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
...
 }
 
 func (dc *DistributedCache) WriteFile(options internal.WriteFileOptions) (int, error) {
-	return 0, syscall.ENOTSUP
+	log.Err("DistributedCache::WriteFile : WriteFile, offset : %d, buf size : %d, file : %s",
+		options.Offset, len(options.Data), options.Handle.Path)
+
+	if !options.Handle.Flags.IsSet(handlemap.HandleFlagDcacheAllowWrites) {
+		return 0, syscall.EIO
+	}
```
linuxsmiths, line 519: we should assert for this?
since we would have failed OpenFile() for dcache files with write flags, fuse shouldn't call write for such an handle.
Also, if the write is for an azure file, we shouldn't disallow.

component/distributed_cache/distributed_cache.go
```diff
@@ -356,68 +361,265 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
...
+
+	// Set the handle is dirty to get the flush call.
```
linuxsmiths, line 521: pls add this comment above this

    //
    // When user wants to write to a default path (no explicit fs=azure/fs=dcache namespace specified)
    // we have mulitple possible semantics:
    // 1. Write through
    //    In this mode every application write is written to both the dcache as well as Azure, as if
    //    user explicitly wrote to either of them. If any of these write fails, the application write
    //    is failed.
    // 2. Write back on close
    //    In this mode application writes are sent to dcache and only on close() the entire dcache
    //    file is written to Azure as well.
    // 3. Write back on eviction
    //    In this mode application writes are sent to dcache and only if.when the dcache file is evicted,
    //    we write it to Azure.
    //
    // For now we implement the "Write through" semantics.
    //

component/distributed_cache/distributed_cache.go
```diff
@@ -356,68 +361,265 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
...
+	}
+	return len(options.Data), nil
+}
+
+func (dc *DistributedCache) SyncFile(options internal.SyncFileOptions) error {
+	log.Debug("DistributedCache::SyncFile : SyncFile file : %s", options.Handle.Path)
+
+	var dcacheErr, azureErr error
+	if options.Handle.IsFsDcache() {
+		common.Assert(options.Handle.IFObj != nil)
```
linuxsmiths, line 551: also assert that HandleFlagDcacheAllowWrites must be set.

component/distributed_cache/distributed_cache.go
```diff
@@ -356,68 +361,265 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
...
-	return syscall.ENOTSUP
+	log.Debug("DistributedCache::FlushFile : Close file : %s", options.Handle.Path)
+	// Allow only one Flush/close call per file when writing, if user application duplicates the fd
+	// then the writes after fist close would fail.
+
+	var dcacheErr, azureErr error
+
+	if !options.Handle.Flags.IsSet(handlemap.HandleFlagDcacheAllowWrites) {
+		return nil
+	}
```
linuxsmiths, line 577: same comment as WriteFile

internal/dcache/file_manager/file_manager.go
```diff
@@ -191,7 +209,24 @@ func (file *DcacheFile) ReleaseFile() error {
 	return nil
 }
 
-// Gets the existing chunk from the chunks
+// This method is called when all the File IO operations are successful
+// and user wants to sync the file
+func (file *DcacheFile) finalizeFile() error {
+	file.FileMetadata.State = dcache.Ready
```
linuxsmiths, line 215: we should set the file size also

internal/dcache/file_manager/utils.go
```diff
@@ -63,27 +67,67 @@ func isOffsetChunkStarting(offset int64, fileLayout *dcache.FileLayout) bool {
...
-		MVList:     []string{"mv0", "mv1", "mv2"},
+		ChunkSize:  int64(chunkSize),
+		StripeSize: int64(stripeSize),
+		MVList:     make([]string, numMVs),
 	}
+
+	// Get active MV's from the clustermap
+	activeMvs := clustermap.GetActiveMVs()
+	common.Assert(len(activeMvs) >= int(numMVs))
+	// todo : is there any better policy to pick?
```
linuxsmiths, line 99: random would be better

internal/dcache/models.go
```diff
@@ -100,8 +100,17 @@ type DCacheConfig struct {
 	CacheAccess            string `json:"cache-access,omitempty"`
 }
 
+type fileState int
```
linuxsmiths, line 103: this should be string, as defined in the doc
"ready", "writing", "deleting"

internal/handlemap/handle_map.go
```diff
@@ -55,6 +55,10 @@ const (
 	HandleFlagDirty          // File has been modified with write operation or is a new file
 	HandleFlagFSynced        // User has called fsync on the file explicitly
 	HandleFlagCached         // File is cached in the local system by blobfuse2
+	// Following are the Dcache Flags
+	HandleFSAzure
+	HandleFSDcache
+	HandleFlagDcacheAllowWrites // Handle can only write If this flag is set
```
linuxsmiths, line 61: HandleFSAzure               // Handle refers to a file in Azure
	HandleFSDcache              // Handle refers to a file in Distributed Cache.
	                            // Both HandleFSAzure and HandleFSDcache will be set for handles corresponding
	                            // to file paths without an explicit fs=azure/fs=dcache namespace specified.
	HandleFlagDcacheAllowWrites // Write to Distributed Cache through this handle is only allowed if this flag is set
	HandleFlagDcacheAllowReads  // Read from Distributed Cache through this handle is only allowed if this flag is set

internal/handlemap/handle_map.go
```diff
@@ -156,6 +160,30 @@ func (handle *Handle) Cleanup() {
 	}
 }
 
+// **********************Dcache Related Methods Start******************************
+func (handle *Handle) SetFsAzure() {
+	handle.Flags.Set(HandleFSAzure)
```
linuxsmiths, line 165: add these asserts before setting

// Must be set once and only once.
common.Assert(!handle.IsFsAzure())
common.Assert(!handle.IsFsDcache())

add the same in SetFsDcache() and SetFsDefault() too.

internal/handlemap/handle_map.go
```diff
@@ -156,6 +160,30 @@ func (handle *Handle) Cleanup() {
...
+}
+
+func (handle *Handle) IsFsAzure() bool {
+	return handle.Flags.IsSet(HandleFSAzure)
+}
+
+func (handle *Handle) IsFsDcache() bool {
+	return handle.Flags.IsSet(HandleFSDcache)
+}
+
```
linuxsmiths, line 184: Also add the following helpers too

```
func (handle *Handle) SetDcacheAllowWrites() {
    // Must be set only once.
    common.Assert(!handle.IsDcacheAllowWrites())
    // Read and write to dcache are not allowed from the same handle.
    common.Assert(!handle.IsDcacheAllowReads())
    // Using a handle we can write to Azure, DCache or both.
    common.Assert(handle.IsFsAzure() || handle.IsFsDcache())

	handle.Flags.Set(HandleFlagDcacheAllowWrites )
}

func (handle *Handle) SetDcacheAllowReads() {
    // Must be set only once.
    common.Assert(!handle.IsDcacheAllowReads())
    // Read and write to dcache are not allowed from the same handle.
    common.Assert(!handle.IsDcacheAllowWrites())
    // Using a handle we can read from Azure or DCache but never both.
    common.Assert(handle.IsFsAzure() || handle.IsFsDcache())
    common.Assert(!(handle.IsFsAzure() && handle.IsFsDcache()))

	handle.Flags.Set(HandleFlagDcacheAllowReads )
}

func (handle *Handle) IsDcacheAllowWrites() bool {
	allowWrites := handle.Flags.IsSet(HandleFlagDcacheAllowWrites)
	allowReads := handle.Flags.IsSet(HandleFlagDcacheAllowReads)
    // Read and write to dcache are not allowed from the same handle.
	common.Assert(!(allowWrites && allowReads))
	return allowWrites
}

func (handle *Handle) IsDcacheAllowReads() bool {
	allowReads := handle.Flags.IsSet(HandleFlagDcacheAllowReads)
	allowWrites := handle.Flags.IsSet(HandleFlagDcacheAllowWrites)
    // Read and write to dcache are not allowed from the same handle.
	common.Assert(!(allowWrites && allowReads))
	return allowReads
}
```

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -238,7 +239,12 @@ func (m *BlobMetadataManager) getFile(filePath string) (*dcache.FileMetadata, er
 		log.Debug("GetFile :: Failed to unmarshal JSON data: %v", err)
 		return nil, err
 	}
+
 	log.Debug("GetFile :: Successfully unmarshaled JSON data for file %s", path)
+	if metadata.State == dcache.Writing || metadata.State == dcache.Deleting {
+		return nil, errors.New(fmt.Sprintf("File is in %d state", metadata.State))
+	}
```
linuxsmiths, line 246: metadata manager should not be checking the content of the file and take actions based on that.
See the todo @ L224, we will change it to return a []byte

internal/dcache/file_manager/worker.go
```diff
@@ -104,11 +105,21 @@ func (wp *workerPool) queueWork(file *DcacheFile, chunk *StagedChunk, get_chunk
 
 func (wp *workerPool) readChunk(task *task) {
 	log.Info("DistributedCache::readChunk : Reading chunk idx : %d, file: %s", task.chunk.Idx, task.file.FileMetadata.Filename)
```
linuxsmiths, line 107: let's make this Debug

internal/dcache/file_manager/worker.go
```diff
@@ -122,10 +133,21 @@ func (wp *workerPool) readChunk(task *task) {
 
 func (wp *workerPool) writeChunk(task *task) {
 	log.Info("DistributedCache::writeChunk : Writing chunk idx : %d, file: %s", task.chunk.Idx, task.file.FileMetadata.Filename)
```
linuxsmiths, line 135: let's make this Debug

internal/dcache/file_manager/worker.go
```diff
@@ -104,11 +105,21 @@ func (wp *workerPool) queueWork(file *DcacheFile, chunk *StagedChunk, get_chunk
...
-	var err error
-
 	// Read From the Dcache
-	//Call MvRead method for reading the chunk.
-	// err = rm.MVRead()
+	readMVReq := &rm.ReadMvRequest{
+		FileID:         task.file.FileMetadata.FileID,
+		MvName:         getMVForChunk(task.chunk, &task.file.FileMetadata.FileLayout),
+		ChunkIndex:     task.chunk.Idx,
+		OffsetInChunk:  task.chunk.Idx * task.file.FileMetadata.FileLayout.ChunkSize,
```
linuxsmiths, line 113: this is offset within the chunk not within the file, so if you are reading full chunk, this should be 0.

internal/dcache/file_manager/worker.go
```diff
@@ -104,11 +105,21 @@ func (wp *workerPool) queueWork(file *DcacheFile, chunk *StagedChunk, get_chunk
...
+		Length:         task.chunk.Len,
+		ChunkSizeInMiB: int64(len(task.chunk.Buf)),
+		Data:           task.chunk.Buf,
+	}
+
+	readMVresp, err := rm.ReadMV(readMVReq)
+	common.Assert(err == nil)
+	common.Assert(readMVresp.BytesRead == task.chunk.Len)
```
linuxsmiths, line 121: move this assert inside "err == nil" block below, at L124

then move the err==nil assert above after the log at L128. That way if it fails we get the error log also, which will help debugging.

internal/dcache/file_manager/worker.go
```diff
@@ -122,10 +133,21 @@ func (wp *workerPool) readChunk(task *task) {
...
+		ChunkIndex:     task.chunk.Idx,
+		Data:           task.chunk.Buf,
+		ChunkSizeInMiB: int64(len(task.chunk.Buf)),
+		IsLastChunk:    isLastChunk,
+	}
 
 	//Call MvWrite method for reading the chunk.
-	// err = rm.MVWrite()
+	_, err := rm.WriteMV(writeMVReq)
+	common.Assert(err == nil)
```
linuxsmiths, line 149: move this assert after log at L156

internal/dcache/file_manager/utils.go
```diff
@@ -63,27 +67,67 @@ func isOffsetChunkStarting(offset int64, fileLayout *dcache.FileLayout) bool {
...
+		FileMetadata: fileMetadata,
+	}, nil
+}
+
+// Does all init process for opening the file.
+func OpenDcacheFile(fileName string) (*DcacheFile, error) {
+	fileMetadata, err := mm.GetFile(fileName)
+	if err != nil {
 		return nil, err
 	}
```
linuxsmiths, line 132: in the else block we should check if the state is not "ready" then we should fail with an error that signifies noent to the caller, so that we can fail the fuse open call with ENOENT

internal/dcache/file_manager/utils.go
```diff
@@ -63,27 +67,67 @@ func isOffsetChunkStarting(offset int64, fileLayout *dcache.FileLayout) bool {
 	return (offset%fileLayout.ChunkSize == 0)
 }
 
+func getMVForChunk(chunk *StagedChunk, fileLayout *dcache.FileLayout) string {
+	numMvs := int64(len(fileLayout.MVList))
```
linuxsmiths, line 71: few asserts in the beginning.

```
// Must have full strip worth of MVs.
common.Assert(numMvs == (fileLayout.StripeSize / fileLayout.ChunkSize))
// For writes file size won't be set yet, for reads we must be reading within the file.
common.Assert((fileLayout.Size == -1) ||
              ((chunk.Idx * fileLayout.ChunkSize) < fileLayout.Size))
```

internal/dcache/file_manager/utils.go
```diff
@@ -63,27 +67,67 @@ func isOffsetChunkStarting(offset int64, fileLayout *dcache.FileLayout) bool {
...
-		ChunkSize:  4 * 1024 * 1024,
-		StripeSize: 16 * 1024 * 1024,
-		MVList:     []string{"mv0", "mv1", "mv2"},
+		ChunkSize:  int64(chunkSize),
+		StripeSize: int64(stripeSize),
+		MVList:     make([]string, numMVs),
 	}
+
+	// Get active MV's from the clustermap
+	activeMvs := clustermap.GetActiveMVs()
```
linuxsmiths, line 97: pls use MVs and not Mvs

linuxsmiths: Instead of using GetActiveMVs(), let's add the following to clustermap and then use GetActiveMVNames() here

```
func GetActiveMVNames() []string {
        return clustermapImpl.getActiveMVNames()
}

func (c *ClusterMapImpl) getActiveMVNames() []string {
        common.Assert(c.localMap != nil)

        i := 0
        activeMVNames := make([]string, len(c.localMap.MVMap))
        for mvName, mv := range c.localMap.MVMap {
                if mv.State == dcache.StateOnline {
                        activeMVNames[i] = mvName
                        i++
                }
        }
        return activeMVNames[:i]
}
```

internal/dcache/file_manager/utils.go
```diff
@@ -63,27 +67,67 @@ func isOffsetChunkStarting(offset int64, fileLayout *dcache.FileLayout) bool {
...
+	for k, _ := range activeMvs {
+		fileMetadata.FileLayout.MVList[i] = k
+		i++
+		if i == int(numMVs) {
+			break
+		}
+	}
+
+	log.Debug("DistributedCache[FM]::NewDcacheFile : Choose MV's are %s, %s, %s,  file: %s",
+		fileMetadata.FileLayout.MVList[0], fileMetadata.FileLayout.MVList[1], fileMetadata.FileLayout.MVList[2], fileName)
```
linuxsmiths, line 110: don't assume replication factor of 3, use this

```
    log.Debug("DistributedCache[FM]::NewDcacheFile : Chosen MV's are: %v,  file: %s",
              fileMetadata.FileLayout.MVList, fileName)
```

internal/dcache/file_manager/file_manager.go
linuxsmiths, line 215: before setting state to Ready, assert the current state is "writing"

internal/dcache/file_manager/file_manager.go
```diff
@@ -191,7 +209,24 @@ func (file *DcacheFile) ReleaseFile() error {
...
+	fileMetadataBytes, err := json.Marshal(file.FileMetadata)
+	if err != nil {
+		log.Err("DistributedCache[FM]::finalizeFile : FileMetadata marshalling fail, file: %s", file.FileMetadata.Filename)
+		return err
+	}
+	err = mm.CreateFileFinalize(file.FileMetadata.Filename, fileMetadataBytes)
+	if err != nil {
+		log.Err("DistributedCache[FM]::finalizeFile : File Finalize failed for file :  %s with err : %s",
+			file.FileMetadata.Filename, err.Error())
```
linuxsmiths, line 224: also log file.FileMetadata as %+v

internal/dcache/file_manager/file_manager.go
```diff
@@ -191,7 +209,24 @@ func (file *DcacheFile) ReleaseFile() error {
...
+	}
+	return err
```
linuxsmiths, line 226: before returning, log this

```
    log.Debug("DistributedCache[FM]::finalizeFile : Final metadata for file: %s, %+v",
              file.FileMetadata.Filename, file.FileMetadata)
```

internal/dcache/file_manager/file_manager.go
```diff
@@ -175,6 +177,22 @@ func (file *DcacheFile) SyncFile() error {
...
 	})
+	common.Assert(err != nil)
+	return err
+}
+
+// Close and Finalize the file. writes are failed after this operation
+func (file *DcacheFile) CloseFile() error {
+	log.Debug("DistributedCache[FM]::CloseFilee : Close File for %s", file.FileMetadata.Filename)
+	err := file.SyncFile()
+	common.Assert(err == nil)
```
linuxsmiths, line 188: sync must be called only when writing to file, so check for IsDcacheAllowWrites()
and add a comment why we are doing sync

// We stage application writes into StagedChunk and upload only when we have a full chunk.
// In case of last chunk being partial, we need to upload it now.

internal/dcache/file_manager/file_manager.go
```diff
@@ -175,6 +177,22 @@ func (file *DcacheFile) SyncFile() error {
 		}
 		return true
 	})
+	common.Assert(err != nil)
```
linuxsmiths, line 180: debug log at L172 logging the chunk index and length, would be helpful

linuxsmiths (approved):

linuxsmiths (approved):

PR #1722, opened 2025-04-30
RPC: Integration with cluster map

souravgupta-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description
Integration of RPC and replication manager with the cluster map.

### DIFF
--- a/internal/dcache/replication_manager/utils.go
+++ b/internal/dcache/replication_manager/utils.go
@@ -41,7 +41,7 @@ import (
 	"github.com/Azure/azure-storage-fuse/v2/common"
 	"github.com/Azure/azure-storage-fuse/v2/common/log"
 	"github.com/Azure/azure-storage-fuse/v2/internal/dcache"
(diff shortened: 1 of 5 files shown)


internal/dcache/replication_manager/utils.go
```diff
@@ -41,7 +41,7 @@ import (
 	"github.com/Azure/azure-storage-fuse/v2/common"
 	"github.com/Azure/azure-storage-fuse/v2/common/log"
 	"github.com/Azure/azure-storage-fuse/v2/internal/dcache"
-	cm "github.com/Azure/azure-storage-fuse/v2/internal/dcache/cluster_manager"
+	cmap "github.com/Azure/azure-storage-fuse/v2/internal/dcache/clustermap"
```
linuxsmiths, line 44: let's use clustermap itself, as we are already using that in clustermanager.
Moreover, clustermap is not very long and it's more readable.

souravgupta-msft: Done

linuxsmiths (approved):

Issue #1662, opened 2025-03-20
Question about the file /tmp/blobfuse2.pid


yangchengs:
A cx reported one issue that one file /tmp/blobfuse2.pid(xxxxx) took 10GB size and it caused the /tmp folder out of space.

In mount.go, 

		log.Info("mount: Mounting blobfuse2 on %s", options.MountPath)
		if !options.Foreground {
			pidFile := strings.Replace(options.MountPath, "/", "_", -1) + ".pid"
			pidFileName := filepath.Join(os.ExpandEnv(common.DefaultWorkDir), pidFile)

			pid := os.Getpid()
			fname := fmt.Sprintf("/tmp/blobfuse2.%v", pid)

			dmnCtx := &daemon.Context{
				PidFileName: pidFileName,
				PidFilePerm: 0644,
				Umask:       022,
				LogFileName: fname, // this will redirect stderr of child to given file
			}

Can anyone explain how can I generate a /tmp/blobfuse2.pid with an error?

syeleti-msft:
The /tmp/blobfuse2.pid file gets deleted as soon as the mount is successful,
Could you please share the debug logs. 
you can enable the logging by pasting the following at the top of your blobfuse2 configuration file. 
```
logging:
  type: base
  level: log_debug
  file-path: ~/path/to/log/file
```
Questions:
1. is the mount got successful when this happened?

syeleti-msft:
Closing this issue as removing the fuse-trace option from the config would not increase the file size.

PR #1723, opened 2025-05-01
Fix logs using up all the space of /tmp folder

syeleti-msft:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

- **Bug Fix**: (Brief description of the feature or issue being addressed)
 /tmp folder is getting full due to the space used by the deleted file /tmp/blobfuse2.pid when the fuse-trace option is enabled and mounted without foreground option. 
 **Reason:**
 Currently we use stdout for emitting the libfuse logs. Now when user don't use foreground option while mounting, generally the mounted process have its stdout and stderr linked to the file /tmp/blobfuse2.log so these logs are redirected to that file. But this file is deleted by the parent process after the mount become successful. so now this is a deleted file with an open file descriptor. Size of this file gets bigger and bigger until user unmounts. sad part is user cannot delete this file as it's already deleted by the parent process of our mount. space taken by this /tmp/blobfuse2.pid file is released when user unmounts.
 
 **Output:** "lsof | grep 'deleted' "when fuse logs are enabled and blobfuse is mounted as daemon.
blobfuse2 2938722 2938730 blobfuse2          fantom    2w      REG                8,1       598       1739 /tmp/blobfuse2.2938713 (deleted)
blobfuse2 2938722 2938738 blobfuse2          fantom    1w      REG                8,1      1505       1739 /tmp/blobfuse2.2938713 (deleted)

You can see in the output that the size of the file is getting increased as we do operations on mountpoint.

Fix : Disable fuse-trace option when user mounts as daemon.

## Checklist

## Related Links
- [Issues](<https://github.com/Azure/azure-storage-fuse/issues/1662>)

### DIFF
--- a/component/libfuse/libfuse.go
+++ b/component/libfuse/libfuse.go
@@ -332,6 +332,14 @@ func (lf *Libfuse) Configure(_ bool) error {
 		return fmt.Errorf("%s config error %s", lf.Name(), err.Error())
 	}
 
+	// Disable libfuse logs if the mount is not running in foreground.
+	// Currently as of 01-05-2025, we emmit the libfuse logs only to the stdout.
+	if !common.ForegroundMount {
+		if lf.traceEnable {
+			lf.traceEnable = false
+		}
+	}
+
 	log.Crit("Libfuse::Configure : read-only %t, allow-other %t, allow-root %t, default-perm %d, entry-timeout %d, attr-time %d, negative-timeout %d, ignore-open-flags %t, nonempty %t, direct_io %t, max-fuse-threads %d, fuse-trace %t, extension %s, disable-writeback-cache %t, dirPermission %v, mountPath %v, umask %v",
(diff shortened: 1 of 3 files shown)


component/libfuse/libfuse.go
```diff
@@ -332,6 +332,14 @@ func (lf *Libfuse) Configure(_ bool) error {
 		return fmt.Errorf("%s config error %s", lf.Name(), err.Error())
 	}
 
+	// Disable libfuse logs if the mount is not running in foreground.
```
vibhansa-msft, line 335: We can not disable this. When we encounter any error during fuse init, we mount in foreground and enable to fuse logs. That's how we can debug what parameter we are sending to fuse, which it does not like or any specific error it can emit out. With this dsiabled there is no way to debug any fuse level logs.

component/libfuse/libfuse.go
```diff
@@ -332,6 +332,14 @@ func (lf *Libfuse) Configure(_ bool) error {
...
+	// Currently as of 01-05-2025, we emit the libfuse logs only to the stdout.
+	if !common.ForegroundMount {
+		if lf.traceEnable {
```
vibhansa-msft, line 338: You shall never be using fuse trace enable unless there is a failure in init of the fuse itself. This is not a general purpose option that customers are suppose to use, its for very high level debugging to look at any change in fuse driver causing the mount failure.

vibhansa-msft (approved):

PR #1727, opened 2025-05-03
Vibhansa/xload fixes

vibhansa-msft:
Few additional logs in preload to debug issues better.


component/xload/splitter.go
```diff
@@ -210,10 +210,19 @@ func (ds *downloadSplitter) Process(item *WorkItem) (int, error) {
...
 				}
+
+				// send the download status to stats manager
+				ds.GetStatsManager().AddStats(&StatsItem{
+					Component:        SPLITTER,
+					Name:             item.Path,
+					Success:          false,
+					Download:         false,
+					BytesTransferred: respSplitItem.DataLen,
+				})
```
souravgupta-msft, line 225: Data manager already sends the stats for bytes downloaded to the Stats Manager

component/xload/stats_manager.go
```diff
@@ -210,11 +221,14 @@ func (sm *StatsManager) calculateBandwidth() {
 	filesPending := sm.totalFiles - filesProcessed
 	percentCompleted := (float64(filesProcessed) / float64(sm.totalFiles)) * 100
 	bandwidthMbps := float64(bytesTransferred*8) / (timeLapsed * float64(MB))
+	diskSpeedMbps := float64(sm.bytesWritten*8) / (timeLapsed * float64(MB))
 
+	max, pr, reg, waiting := sm.pool.GetUsageDetails()
 	log.Crit("statsManager::calculateBandwidth : timestamp %v, %.2f%%, %v Done, %v Failed, "+
-		"%v Pending, %v Total, Bytes transferred %v, Throughput (Mbps): %.2f",
+		"%v Pending, %v Total, Bytes transferred %v, Throughput (Mbps): %.2f, Disk Speed (Mbps): %.2f, Cache usage: %v%%, (%v / %v / %v : %v), Time: %.2f",
```
souravgupta-msft, line 228: This should be blockpool usage

```suggestion
		"%v Pending, %v Total, Bytes transferred %v, Throughput (Mbps): %.2f, Disk Speed (Mbps): %.2f, Blockpool usage: %v%%, (%v / %v / %v : %v), Time: %.2f",
```

component/xload/stats_manager.go
```diff
@@ -160,7 +167,11 @@ func (sm *StatsManager) statsProcessor() {
 
 		case SPLITTER:
 			// log.Debug("statsManager::statsProcessor : splitter: Name %v, success %v, download %v", item.name, item.success, item.download)
-			sm.updateSuccessFailedCtr(item.Success)
+			if item.Download {
+				sm.updateSuccessFailedCtr(item.Success)
+			} else {
+				sm.updateDiskStats(item.BytesTransferred)
+			}
```
souravgupta-msft, line 174: We should add some boolean field in `StatsItem` to indicate that this stat is for the bytes written to disk.

component/xload/stats_manager.go
```diff
@@ -210,11 +221,14 @@ func (sm *StatsManager) calculateBandwidth() {
 	filesPending := sm.totalFiles - filesProcessed
 	percentCompleted := (float64(filesProcessed) / float64(sm.totalFiles)) * 100
 	bandwidthMbps := float64(bytesTransferred*8) / (timeLapsed * float64(MB))
+	diskSpeedMbps := float64(sm.bytesWritten*8) / (timeLapsed * float64(MB))
```
souravgupta-msft, line 224: How much difference are you seeing between `bandwidthMbps` and `diskSpeedMbps`

component/azstorage/utils.go
```diff
@@ -138,6 +138,11 @@ func getAzDatalakeServiceClientOptions(conf *AzStorageConfig) (*serviceBfs.Clien
 
 // getLogOptions : to configure the SDK logging policy
 func getSDKLogOptions() policy.LogOptions {
+	// If BLOBFUSE_DISABLE_SDK_LOG env var is set to true, then disable the SDK logging
+	if os.Getenv("BLOBFUSE_DISABLE_SDK_LOG") == "true" {
+		return policy.LogOptions{}
+	}
+
```
souravgupta-msft, line 145: Add this in the `setSDKLogListener` method

component/xload/splitter.go
vibhansa-msft: thats data downloaded stats, I wanted to cover the disk writing speeds also. data downloaded will always have "download" set to true so here I am using false to indicate that this is disk writing bytes.

component/xload/stats_manager.go
vibhansa-msft: as of now I see them almost same, disk being little low compared to download. But these new stats cleared that there are no more blocks left in the blockpool and many data manager instances are waiting. This means most of the blocks are in the responce queue only, and that will happen only if disk writing is slower then download speed.

component/azstorage/utils.go
vibhansa-msft: done

component/xload/stats_manager.go
vibhansa-msft: done

component/azstorage/utils.go
```diff
@@ -154,6 +159,10 @@ func getSDKLogOptions() policy.LogOptions {
 //   - logging type is silent
 //   - logging level is less than debug
 func setSDKLogListener() {
+	if os.Getenv("BLOBFUSE_DISABLE_SDK_LOG") == "true" {
+		return
+	}
+
 	if log.GetType() == "silent" || log.GetLogLevel() < common.ELogLevel.LOG_DEBUG() {
 		// reset listener
 		azlog.SetListener(nil)
```
souravgupta-msft:
```suggestion
	if os.Getenv("BLOBFUSE_DISABLE_SDK_LOG") == "true" || log.GetType() == "silent" || log.GetLogLevel() < common.ELogLevel.LOG_DEBUG() {
		// reset listener
		azlog.SetListener(nil)
```

component/xload/lister.go
```diff
@@ -191,10 +192,13 @@ func (rl *remoteLister) Process(item *WorkItem) (int, error) {
 					}
 
 					// push the directory to input pool for its listing
-					rl.Schedule(&WorkItem{
+					err := rl.Schedule(&WorkItem{
 						CompName: rl.GetName(),
 						Path:     name,
 					})
+					if err != nil {
+						return
```
souravgupta-msft, line 201: Log the error here

component/xload/lister.go
```diff
@@ -212,6 +216,9 @@ func (rl *remoteLister) Process(item *WorkItem) (int, error) {
 					Mtime:    entry.Mtime,
 					MD5:      entry.MD5,
 				})
+				if err != nil {
+					return 0, err
```
souravgupta-msft, line 222: Log the error

bhansalivikas: done

ashruti-msft (approved):

PR #1732, opened 2025-05-05
Clustermap audit

linuxsmiths:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

## Checklist

## Related Links
- [Issues](<link>)


internal/dcache/clustermap/clustermap.go
```diff
@@ -33,55 +33,353 @@
...
+        return clusterMap.getMyRVs()
+}
+
+// It will return all the RVs Map <rvName, rvState> for the particular mvName as per local cache copy of cluster map.
+func GetRVs(mvName string) map[string]dcache.StateEnum {
+        return clusterMap.getRVs(mvName)
+}
+
+// It will check if the given nodeId is online as per local cache copy of cluster map.
+func IsOnline(nodeId string) bool {
```
vibhansa-msft, line 85: Why do we need these wrappers? If dcache code is going to create an object of ClusterMap then they can directly call the member functions instead of calling some globally exposed methods which internally then do nothing but to forward the call to cluster-map object. Is the only objective here is to make this singleton ?

linuxsmiths: Encapsulation is the primary objective here. Components other than clustermap (and clustermanager) do not need to worry about the clustermap, and how to get the latest one, etc. They simply call these functions exported by the clustermap package to find the info they want to know about the cluster. Internally, clustermap package takes care of refreshing the local clustermap object periodically (infact it's the clustermanager that refreshes and then notifies clustermap). So you can see how all those details are hidden from components which will use the clustermap package as a clustermanager client.

jainakanksha-msft (approved):

PR #1735, opened 2025-05-05
RPC: Refactoring

souravgupta-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description
Refactoring of some methods in RPC package

### DIFF
--- a/internal/dcache/rpc/client/functions.go
+++ b/internal/dcache/rpc/client/functions.go
@@ -314,7 +314,7 @@ func EndSync(ctx context.Context, targetNodeID string, req *models.EndSyncReques
 
 // cleanup closes all the RPC node client pools
 func Cleanup() error {
-	log.Debug("rpc_client::Cleanup: Closing all node client pools")
+	log.Info("rpc_client::Cleanup: Closing all node client pools")
 	err := cp.closeAllNodeClientPools()
 	if err != nil {
 		log.Err("rpc_client::Cleanup: Failed to close all node client pools [%v]", err.Error())
@@ -324,6 +324,6 @@ func Cleanup() error {
 }
 
 func init() {
-	log.Debug("rpc_client::init: package initialized, create client pool")
+	log.Info("rpc_client::init: package initialized, create client pool")
 	cp = newClientPool(defaultMaxPerNode, defaultMaxNodes, defaultTimeout)
 }
--- a/internal/dcache/rpc/client/functions.go
+++ b/internal/dcache/rpc/client/functions.go
@@ -86,26 +86,27 @@ func Hello(ctx context.Context, targetNodeID string, req *models.HelloRequest) (
 }
 
 func GetChunk(ctx context.Context, targetNodeID string, req *models.GetChunkRequest) (*models.GetChunkResponse, error) {
-	log.Debug("rpc_client::GetChunk: Sending GetChunk request to node %s: %+v", targetNodeID, *req)
+	reqStr := rpc.GetChunkRequestToString(req)
+	log.Debug("rpc_client::GetChunk: Sending GetChunk request to node %s: %v", targetNodeID, reqStr)
 
 	// get RPC client from the client pool
 	client, err := cp.getRPCClient(targetNodeID)
 	if err != nil {
-		log.Err("rpc_client::GetChunk: Failed to get RPC client for node %s [%v] : %+v", targetNodeID, err.Error(), *req)
+		log.Err("rpc_client::GetChunk: Failed to get RPC client for node %s [%v] : %v", targetNodeID, err.Error(), reqStr)
 		return nil, err
 	}
 	defer func() {
 		// release RPC client back to the pool
 		err = cp.releaseRPCClient(client)
 		if err != nil {
-			log.Err("rpc_client::GetChunk: Failed to release RPC client for node %s [%v] : %+v", targetNodeID, err.Error(), *req)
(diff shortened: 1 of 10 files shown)


internal/dcache/rpc/client/functions.go
```diff
@@ -307,6 +314,7 @@ func EndSync(ctx context.Context, targetNodeID string, req *models.EndSyncReques
 
 // cleanup closes all the RPC node client pools
 func Cleanup() error {
+	log.Debug("rpc_client::Cleanup: Closing all node client pools")
```
linuxsmiths, line 317: this can be Info, as it's rare and it's good to know
same for L327

internal/dcache/rpc/server/server.go
```diff
@@ -36,39 +36,35 @@ package rpc_server
...
-// - address: address of the node having the format <ip>:<port>
-// - rvs: map of raw volumes of this node
-// - dCacheConfig: dCache configuration
-func NewNodeServer(address string, rvs map[string]dcache.RawVolume, config *dcache.DCacheConfig) (*NodeServer, error) {
-	// TODO: add assert for IsValidIP
-	common.Assert(address != "", "node address cannot be empty")
+// NewNodeServer creates a Thrift server for the node
+func NewNodeServer() (*NodeServer, error) {
+	nodeID, err := common.GetNodeUUID()
+	common.Assert(err == nil, "failed to get node ID: %v", err)
```
linuxsmiths, line 54: for whatever reason if we cannot get nodeid we cannot proceed, so return failure to prevent RPC server startup in an invalid state

internal/dcache/rpc/server/server.go
```diff
@@ -36,39 +36,35 @@ package rpc_server
...
+
+	address := rpc.GetNodeAddressFromID(nodeID)
+	rvs := clustermap.GetMyRVs()
```
linuxsmiths, line 57: add a TODO to call IsValidRVMap() to test rvs list and fail if it's invalid.
this can be done only after we move IsValidRVMap() to clustermap

internal/dcache/rpc/server/server.go
linuxsmiths: similarly w/o a valid address we cannot safely proceed, so check with IsValidHostPort() and fail if invalid

internal/dcache/rpc/utils.go
```diff
@@ -0,0 +1,132 @@
...
+}
+
+// convert *models.RVNameAndState to string
+// used for logging
+func ComponentRVsToString(rvs []*models.RVNameAndState) string {
+	str := strings.Builder{}
+	str.WriteString("[")
+	for _, rv := range rvs {
+		common.Assert(rv != nil, "Component RV is nil")
+		str.WriteString(fmt.Sprintf("%+v", *rv))
```
linuxsmiths, line 73: probably leaving a space between each RV may be more readable, "%+v "
likewise add a space after starting "[" at L70

internal/dcache/rpc/utils.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	str.WriteString("]")
+	return str.String()
+}
+
+// convert *models.GetChunkRequest to string
+// used for logging
+func GetChunkRequestToString(req *models.GetChunkRequest) string {
+	return fmt.Sprintf("Address %+v, OffsetInChunk %v, Length %v, ComponentRV %v",
+		*req.Address, req.OffsetInChunk, req.Length, ComponentRVsToString(req.ComponentRV))
```
linuxsmiths, line 83: In all of these it'll be better to use {} to mark the begin and end of the info, will be helpful when these are used in a context where we print multiple of these or these are mixed with other logs, so

fmt.Sprintf("{Address %+v, OffsetInChunk %v, Length %v, ComponentRV %v}",
		*req.Address, req.OffsetInChunk, req.Length, ComponentRVsToString(req.ComponentRV))

internal/dcache/rpc/utils.go
```diff
@@ -0,0 +1,132 @@
...
+func LeaveMVRequestToString(req *models.LeaveMVRequest) string {
+	return fmt.Sprintf("MV %v, RVName %v, ComponentRV %v",
+		req.MV, req.RVName, ComponentRVsToString(req.ComponentRV))
+}
+
+// convert *models.StartSyncRequest to string
+// used for logging
+func StartSyncRequestToString(req *models.StartSyncRequest) string {
+	return fmt.Sprintf("MV %v, SourceRVName %v, TargetRVName %v, ComponentRV %v, DataLength %v",
+		req.MV, req.SourceRVName, req.TargetRVName, ComponentRVsToString(req.ComponentRV), req.DataLength)
```
linuxsmiths, line 124: DataLength doesn't sound right. Let's call it SyncSize

linuxsmiths (approved):
LG with some suggestions.

internal/dcache/rpc/client/functions.go

internal/dcache/rpc/server/server.go
souravgupta-msft: Done, updated

internal/dcache/rpc/server/server.go
souravgupta-msft: Added TODO

internal/dcache/rpc/utils.go
souravgupta-msft: Done

PR #1738, opened 2025-05-06
RPC: Fix race condition between chunk IO and sync operations

souravgupta-msft:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description
Fix the race condition between chunk IO (get, put or remove chunk) and sync (start or end sync) operations. Sync operation will wait for ongoing IO operations to complete and will block new IO operations.

### DIFF
--- a/internal/dcache/rpc/server/handler.go
+++ b/internal/dcache/rpc/server/handler.go
@@ -115,14 +115,12 @@ type mvInfo struct {
 	// Both StartSync and EndSync will quiesce IOs just before they move the mv into and out of syncing state, and
 	// resume IOs once the MV is safely moved into the new state.
 	//
-	// opMutex is used to ensure that only one operation, chunk IO (get, put or remove chunk) or
+	// syncOpMutex is used to ensure that only one operation, chunk IO (get, put or remove chunk) or
 	// sync (start sync or end sync) is in progress at a time.
-	// syncOpMutex is rw mutex. IO operations like get, put or remove chunk takes
-	// read lock on syncOpMutex, and sync operations like StartSync or EndSync takes write lock on it.
-	// opMutex is released immediately after acquiring the syncOpMutex.
+	// IO operations like get, put or remove chunk takes read lock on syncOpMutex, and sync operations
+	// like StartSync or EndSync takes write lock on it.
 	// This ensures that the sync operation waits for the ongoing IO operations to complete.
-	// It also makes sure that no new IO operations will start as it has already acquired the opMutex.
-	opMutex     sync.Mutex
+	// It also makes sure that no new IO operations till the sync is complete.
 	syncOpMutex sync.RWMutex
 
 	syncInfo // sync info for this MV
@@ -370,9 +368,7 @@ func (mv *mvInfo) decTotalChunkBytes(bytes int64) {
 // but will block sync operations like StartSync or EndSync,
 // until the read lock is released.
 func (mv *mvInfo) acquireSyncOpReadLock() {
-	mv.opMutex.Lock()
 	mv.syncOpMutex.RLock()
-	mv.opMutex.Unlock()
 }
 
 // release the read lock on the syncOpMutex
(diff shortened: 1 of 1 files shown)


internal/dcache/rpc/server/handler.go
```diff
@@ -119,9 +115,15 @@ type mvInfo struct {
...
-	// This boolean flag will be set by StartSync and EndSync to put the mv in quiesce state and it'll be read
-	// and honored by the various chunk related APIs - GetChunk, PutChunk, etc.
-	quiesceIOs atomic.Bool
+	// opMutex is used to ensure that only one operation, chunk IO (get, put or remove chunk) or
+	// sync (start sync or end sync) is in progress at a time.
+	// syncOpMutex is rw mutex. IO operations like get, put or remove chunk takes
+	// read lock on syncOpMutex, and sync operations like StartSync or EndSync takes write lock on it.
+	// opMutex is released immediately after acquiring the syncOpMutex.
+	// This ensures that the sync operation waits for the ongoing IO operations to complete.
+	// It also makes sure that no new IO operations will start as it has already acquired the opMutex.
```
linuxsmiths, line 124: you will get the same behaviour just with RWMutex. Most RW locks take care of lot of readers starving occasional writer (as that's the usecase rwlocks are typically used for). I checked go implementation also does that

this is the comment from go explaining that

// If any goroutine calls [RWMutex.Lock] while the lock is already held by
// one or more readers, concurrent calls to [RWMutex.RLock] will block until
// the writer has acquired (and released) the lock, to ensure that
// the lock eventually becomes available to the writer.
// Note that this prohibits recursive read-locking.

This is also present in the auto-generated doc, which makes this an official behaviour that we can rely on

https://pkg.go.dev/sync#RWMutex

souravgupta-msft: Done, using only rw mutex

internal/dcache/rpc/server/handler.go
```diff
@@ -119,9 +115,13 @@ type mvInfo struct {
...
 	//
-	// This boolean flag will be set by StartSync and EndSync to put the mv in quiesce state and it'll be read
-	// and honored by the various chunk related APIs - GetChunk, PutChunk, etc.
-	quiesceIOs atomic.Bool
+	// opMutex is used to ensure that only one operation, chunk IO (get, put or remove chunk) or
+	// sync (start sync or end sync) is in progress at a time.
+	// IO operations like get, put or remove chunk takes read lock on opMutex, and sync operations
+	// like StartSync or EndSync takes write lock on it.
+	// This ensures that the sync operation waits for the ongoing IO operations to complete.
+	// It also makes sure that no new IO operations till the sync is complete.
```
linuxsmiths, line 123: It also makes sure that if start/end sync is waiting for the write lock, no new IO operations are started till the start/end sync gets the write lock and completes. This ensures that a continuous flow of IOs will not delay the start/end sync indefinitely.

internal/dcache/rpc/server/handler.go
```diff
@@ -1063,17 +999,16 @@ func (h *ChunkServiceHandler) StartSync(ctx context.Context, req *models.StartSy
 		return nil, rpc.NewResponseError(rpc.NeedToRefreshClusterMap, fmt.Sprintf("request component RVs are invalid for MV %s [%v]", req.MV, err.Error()))
 	}
 
-	// Set IO quiescing in the mv. Now GetChunk, PutChunk, will not allow any new IO.
+	// acquire write lock on the opMutex for this MV. Now GetChunk, PutChunk and RemoveChunk will not allow any new IO.
 	// Also wait for any ongoing IOs to complete.
-	err := mvInfo.quiesceIOsStart()
-	common.Assert(err == nil, fmt.Sprintf("failed to quiesce IOs for MV %s [%v]", req.MV, err))
+	mvInfo.acquireSyncOpWriteLock()
```
linuxsmiths, line 1005: you may want a debug log after the sync acquires the lock

linuxsmiths (approved):

internal/dcache/rpc/server/handler.go

internal/dcache/rpc/server/handler.go
souravgupta-msft: Added debug logs

PR #1739, opened 2025-05-06
Updated test case and few bug fixes

jainakanksha-msft:


internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -1153,54 +1153,19 @@ func (cmi *ClusterManager) updateRVList(existingRVMap map[string]dcache.RawVolum
 	}
 
 	// Both these maps are indexed by RV id.
-	rVsByRvIdFromHB := make(map[string]dcache.RawVolume)
-	rvLastHB := make(map[string]uint64)
+	rVsByRvIdFromHB, rvLastHB, err := collectHBForGivenNodeIds(nodeIds)
+	if err != nil {
+		return false, err
+	}
+
```
linuxsmiths, line 1160: let's add the following assert

```
// Both the RV and the RV HB map must have the exact same RVs.
common.Assert(len(rVsByRvIdFromHB) == len(rvLastHB), len(rVsByRvIdFromHB), len(rvLastHB))
```

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -1153,54 +1153,19 @@ func (cmi *ClusterManager) updateRVList(existingRVMap map[string]dcache.RawVolum
 	}
 
 	// Both these maps are indexed by RV id.
-	rVsByRvIdFromHB := make(map[string]dcache.RawVolume)
-	rvLastHB := make(map[string]uint64)
+	rVsByRvIdFromHB, rvLastHB, err := collectHBForGivenNodeIds(nodeIds)
+	if err != nil {
+		return false, err
```
linuxsmiths, line 1158: add the following assert here

```
// For onlyMyRVs==true caller would call us only after punching our hb, so we must find the hb.
common.Assert(false, "[BUG] Did not find our own heartbeat"))

```

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -1210,51 +1175,62 @@ func (cmi *ClusterManager) updateRVList(existingRVMap map[string]dcache.RawVolum
...
-		if rvHb, found := rVsByRvIdFromHB[rvInClusterMap.RvId]; found {
-			lastHB := rvLastHB[rvHb.RvId]
+	// (1.a) If it is only myRV, verify if it is not already present in existingRVMap.
+	if onlyMyRVs {
+		for rvName, rvInClusterMap := range existingRVMap {
+			if _, found := rVsByRvIdFromHB[rvInClusterMap.RvId]; found {
+				log.Warn("ClusterManager::updateRVList: RV %s is already present in clusterMap, skipping it",
+					rvName)
+				return false, nil
+			}
```
linuxsmiths, line 1185: since there could be more than one RV, we cannot just return if we find one.
It'll be simpler to let onlyMyRvs case also run through the following loops and only skip the "RV present in existingRVMap, but missing from rVsByRvIdFromHB" case.

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -1210,51 +1175,62 @@ func (cmi *ClusterManager) updateRVList(existingRVMap map[string]dcache.RawVolum
...
+		}
+	} else {
+		// (1.b) Update RVs present in existingRVMap and which have changed State or AvailableSpace.
+		for rvName, rvInClusterMap := range existingRVMap {
+
+			if rvHb, found := rVsByRvIdFromHB[rvInClusterMap.RvId]; found {
+				lastHB := rvLastHB[rvHb.RvId]
```
linuxsmiths, line 1192:
```
lastHB, found  := rvLastHB[rvHb.RvId]

// If an RV is present in rVsByRvIdFromHB, it MUST have a valid HB in rvLastHB.
common.Assert(found)
```

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -1289,6 +1265,51 @@ func (cmi *ClusterManager) updateRVList(existingRVMap map[string]dcache.RawVolum
 	return changed, nil
 }
 
+// For all the given NodeIds, fetch the heartbeat and return the map of RVs and map of their last heartbeat by RVId.
+func collectHBForGivenNodeIds(nodeIds []string) (map[string]dcache.RawVolume, map[string]uint64, error) {
+	rVsByRvIdFromHB := make(map[string]dcache.RawVolume)
+	rvLastHB := make(map[string]uint64)
+	for _, nodeId := range nodeIds {
```
linuxsmiths, line 1272: nit, pls leave a blank line after a variable decl block and the first code line

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -1289,6 +1265,51 @@ func (cmi *ClusterManager) updateRVList(existingRVMap map[string]dcache.RawVolum
...
+				nodeId, err)
+		}
+
+		var hbData dcache.HeartbeatData
+		if err := json.Unmarshal(bytes, &hbData); err != nil {
+			common.Assert(false, err)
+			return nil, nil, fmt.Errorf("ClusterManager::collectHBForGivenNodeIds: Failed to parse heartbeat bytes (%d) for node %s: %v",
+				len(bytes), nodeId, err)
+		}
+
```
linuxsmiths, line 1288: this is the only place where we process heartbeats. Let's make sure heartbeats are valid before we process and act.
Let's add a IsValidHeartbeat() function and use it here for checking and skip the heartbeat if it's not valid. Assert for debug builds.
IsValidHeartbeat() should check that hostname is valid (not empty), ipaddress is valid, nodeid is valid and LastHeartbeat is within minUnixEpoch and maxUnixEpoch.
We can also assert than it has at least one RV (with a TODO for supporting nodes with no RVs)

internal/dcache/clustermap/utils.go
```diff
@@ -60,11 +60,11 @@ var (
...
 
-	minNumReplicas int64 = 1
-	maxNumReplicas int64 = 256
+	MinNumReplicas int64 = 1
+	MaxNumReplicas int64 = 256
 
-	minMvsPerRv int64 = 10
-	maxMvsPerRv int64 = 100
+	MinMvsPerRv int64 = 10
+	MaxMvsPerRv int64 = 100
```
linuxsmiths, line 68: export all these limits, not just these

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -785,7 +785,6 @@ func (m *BlobMetadataManager) getClusterMap() ([]byte, *string, error) {
 	})
 	if err != nil {
 		log.Err("GetClusterMap:: Failed to get cluster map properties %s: %v", clustermapPath, err)
-		common.Assert(false, err)
```
linuxsmiths, line 788: Good!

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -818,11 +818,11 @@ func (cmi *ClusterManager) updateStorageClusterMapIfRequired() error {
...
-	if changed {
-		cmi.updateMVList(clusterMap.RVMap, clusterMap.MVMap)
-	} else {
-		log.Debug("ClusterManager::updateStorageClusterMapIfRequired: No changes in RV mapping")
-	}
+	// if changed {
+	cmi.updateMVList(clusterMap.RVMap, clusterMap.MVMap)
+	// } else {
+	// log.Debug("ClusterManager::updateStorageClusterMapIfRequired: No changes in RV mapping")
+	// }
```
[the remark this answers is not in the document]
jainakanksha-msft: Yes, a Temp changes as of now. we have to fix the flow atleast for initial RV addition.

linuxsmiths (approved):

PR #1740, opened 2025-05-06
fix mv workflow

ashruti-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

Need to add join Mv calls 

## Checklist

## Related Links
- [Issues](<link>)

### DIFF
--- a/internal/dcache/cluster_manager/cluster_manager.go
+++ b/internal/dcache/cluster_manager/cluster_manager.go
@@ -847,6 +847,37 @@ func (cmi *ClusterManager) updateStorageClusterMapIfRequired() error {
 	return nil
 }
 
+func (cmi *ClusterManager) fixMV(rvToDelete string, mvName string, existingMVMap map[string]dcache.MirroredVolume, rvMap map[string]dcache.RawVolume, updatedRvMap map[string]dcache.StateEnum) string {
+	log.Debug("ClusterManager::fixMV: Replacing RV %s in MV %s", rvToDelete, mvName)
+
+	// Store all node ids of component Rvs in a map.
+	// This is done to ensure that we do not pick a new RV from a node already present in Mv.
+	nodeList := make(map[string]bool)
+	for rVName := range existingMVMap[mvName].RVs {
+		nodeList[rvMap[rVName].NodeId] = true
+	}
+	// Ensure the node of newly selected Rv's is not used again
+	for rvName := range updatedRvMap {
+		nodeList[rvMap[rvName].NodeId] = true
+	}
+
+	maxAvailableSpace := uint64(0)
+	replaceRv := ""
+	for rvName, rvInfo := range rvMap {
+		if rvInfo.State == dcache.StateOnline {
+			if _, ok := nodeList[rvInfo.NodeId]; !ok {
+				if maxAvailableSpace < rvInfo.AvailableSpace {
+					maxAvailableSpace = rvInfo.AvailableSpace
+					replaceRv = rvName
+				}
+			}
+		}
+	}
+
+	updatedRvMap[replaceRv] = dcache.StateOutOfSync
+	return replaceRv
+}
+
 // Takes rvMap which is a set of all known RVs (existing RV list from clustermap, and updated as per the most recent
 // heartbeats), it is indexed by RV name and contains complete info about the RV, and existingMVMap which is the set
 // of MVs present in the clustermap, indexed by MV name and contains complete info about the MV.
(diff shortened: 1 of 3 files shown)


internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -977,6 +1019,9 @@ func (cmi *ClusterManager) updateMVList(rvMap map[string]dcache.RawVolume, exist
 			if rvMap[rvName].State == dcache.StateOffline {
 				offlineRv++
 				mv.RVs[rvName] = dcache.StateOffline
+				// Run the fix-mv workflow to replace the offline rv with an online rv
+				// Storing newly selected rv as rvName so slot count for it is updated
+				rvName = cmi.fixMV(rvName, mvName, existingMVMap, rvMap, updatedRvMap)
```
linuxsmiths, line 1024: fix-mv workflow has to run after the current loop (that runs upgrade-mv/offline-mv workflows).
we need to run a separate loop after this, going over only degraded MVs and then running the fix-mv workflow for those.
It's an exercise for you to find out why?

linuxsmiths: this is a rough outline of the code that must run, it's primitive and needs to be discussed and is there to just give an idea.
Note that for the fix-mv workflow we need to call JoinMV for the new RV and UpdateMV for the existing ones.

```
        //
        // Now fix the degraded MVs by replacing their offline RVs with good ones.
        // This is the fix-mv workflow.
        //
        // Note that we can/must only fix degraded MVs, offline MVs cannot be fixed as there's no good component
        // RV to copy chunks from. Once an MV is offline it won't be used by File Manager to put any file's data.
        // Offline MVs will just be lying around like satelite debris in space.
        //
        // TODO: See if we need delete-mv workflow to clean those up.
        //
        for mvName, mv := range existingMVMap {
                offlineRv := 0

                if mv.State != dcache.StateDegraded {
                    continue
                }

                for rvName := range mv.RVs {
                        // Only valid RVs can be used as component RVs for an MV.
                        _, exists := rvMap[rvName]
                        common.Assert(exists)

                        if rvMap[rvName].State != dcache.StateOffline {
                            continue
                        }

                        delete(rvMap, rvName)

                        rvName = findReplacementRV(mv.RVs, rvMap, nodeToRvs)
                        mv.RVs[rvName] = dcache.StateOutOfSync
                        existingMVMap[mvName] = mv
                }
        }
```

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -1124,10 +1186,73 @@ func (cmi *ClusterManager) updateMVList(rvMap map[string]dcache.RawVolume, exist
...
+			MV:           mvName,
+			RVName:       rv.Name,
+			ReserveSpace: 0,
+			ComponentRV:  componentRVs,
+		}
+		timeout := 2 * time.Second
+		ctx, cancel := context.WithTimeout(context.Background(), timeout)
+		defer cancel()
+		// TODO :: Check if this call is correct
+		_, err := rpc_server.JoinMV(ctx, joinMvReq)
```
linuxsmiths, line 1248: we need to call the JoinMV call in rpc_client not rpc_server

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -1124,10 +1186,73 @@ func (cmi *ClusterManager) updateMVList(rvMap map[string]dcache.RawVolume, exist
 				break
 			}
 		}
+
+		// Call join mv and check if all rv's are able to join the mv with rv list
+		deleteRv, err := cmi.joinMV(mvName, existingMVMap[mvName])
+		if err != nil {
+			// If RPC server has not started we delete the new mv added and remove the node from the nodeToRvs map
+			// and redo distribution
+			// TODO :: Would prefer if we could validate this behaviour before allocating RVs to the Mv
```
linuxsmiths, line 1195: Since we need to send all the component RVs in JoinMV, we have to do it after all RVs are assigned.. 
we can start with this.. mostly JoinMV will not fail so we should be fine, perf wise

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -1124,10 +1186,73 @@ func (cmi *ClusterManager) updateMVList(rvMap map[string]dcache.RawVolume, exist
...
+			}
+		}
 	}
 
 	log.Debug("ClusterManager::updateMVList: existing MV map after phase#2: %v", existingMVMap)
 	// TODO :: arrange the map entries in lexicographical order
+	return nil
+}
+
+func (cmi *ClusterManager) joinMV(mvName string, rvs dcache.MirroredVolume) (string, error) {
```
linuxsmiths, line 1213: why do you call dcache.MirroredVolume type as 'rvs'?
JoinMV() should take just dcache.MirroredVolume as parameter. This described the MV completely including the component RVs, and that's what we need to make a JoinMV RPC call.

internal/dcache/cluster_manager/cluster_manager.go
linuxsmiths: also, add a note that we should try to send these RPCs in parallel

linuxsmiths:
this change was already included in some other PR

PR #1744, opened 2025-05-07
Updated the clustermap call to synchronously

jainakanksha-msft:


internal/dcache/clustermap/clustermap.go
```diff
@@ -56,6 +56,11 @@ func Update() {
 	clusterMap.update()
 }
 
+// UpdateSync will load the local clustermap synchronously.
+func UpdateSync() {
+	clusterMap.loadLocalMap()
```
linuxsmiths, line 61: this assert here
```
        //
        // Sync update is called only once in the beginning, when we don't have the localMap set.
        // It can technically be called later, but we should not call, catch inadvertent bad callers.
        //
        common.Assert(c.localMap == nil)

```

 
and following assert inside update()

```
        //
        // You don't want to call the async update in the beginning when setting localMap for the first time.
        // That time you call the sync update function UpdateSync(), catch inadvertent bad callers.
        //
        common.Assert(c.localMap != nil)

```

linuxsmiths (approved):

PR #1747, opened 2025-05-08
Replication Manager: Resync MV initial implementation

souravgupta-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description
Initial implementation of ResyncMV workflow in replication manager.

### DIFF
--- a/internal/dcache/replication_manager/replication_manager.go
+++ b/internal/dcache/replication_manager/replication_manager.go
@@ -483,7 +483,7 @@ func syncComponentRV(mvName string, lioRV string, targetRVName string, syncSize 
 		mvName, lioRV, targetRVName, syncSize, rpc.ComponentRVsToString(componentRVs))
 
 	common.Assert(lioRV != targetRVName, lioRV, targetRVName)
-	common.Assert(syncSize > 0, syncSize)
+	common.Assert(syncSize >= 0, syncSize)
 
 	sourceNodeID := getNodeIDFromRVName(lioRV)
 	common.Assert(common.IsValidUUID(sourceNodeID))
--- a/internal/dcache/replication_manager/replication_manager.go
+++ b/internal/dcache/replication_manager/replication_manager.go
@@ -328,8 +328,8 @@ retry:
 
 					errRV := updateComponentRVState(req.MvName, rv.Name, dcache.StateOffline, componentRVs)
 					if errRV != nil {
-						errStr := fmt.Sprintf("Failed to update component RV %s to %s state for MV %s [%v]",
-							rv.Name, string(dcache.StateOffline), req.MvName, errRV)
+						errStr := fmt.Sprintf("Failed to update %s/%s state to offline [%v]",
+							rv.Name, req.MvName, req.MvName, errRV)
 						log.Err("ReplicationManager::WriteMV: %s", errStr)
 					}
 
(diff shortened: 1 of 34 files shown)


internal/dcache/replication_manager/replication_manager.go
```diff
@@ -256,9 +261,310 @@ retry:
...
+	// TODO: make this parallel
+	for mvName, degradedMV := range degradedMVs {
+		common.Assert(degradedMV.State == dcache.StateDegraded, fmt.Sprintf("MV %s is not in degraded state", mvName))
+		err := ResyncMV(mvName, degradedMV)
+		if err != nil {
+			// TODO: discuss error handling in this scenario
+			log.Err("ReplicationManager::ResyncDegradedMVs: Failed to resync MV %s [%v]", mvName, err.Error())
+		}
+	}
+
```
linuxsmiths, line 294: Things will be very clear if we look at the resync process a bit differently. The basic ideas are the same, but here is my suggestion on how we should look at it. I'll state a series of fact before stitching the overall idea.

SyncMV is a composite workflow which can comprise of one or more SyncComponentRV workflows.
SyncComponentRV is responsible for sync'ing just one component RV. We need one SyncComponentRV for each component RV (of the given MV) which is OutOfSync.
Every SyncComponentRV runs as a separate go routine.
StartSync/EndSync are run in the scope of SyncComponentRV and not SyncMV, which means SyncId belongs to SyncComponentRV and not SyncMV. SyncId corresponds to a "sync pair", i.e., for every OutOfSync component RV we will have a different SyncId.
Since one component RV can be acting as source for multiple SyncComponentRV jobs, we should support multiple simultaneous SyncIds for the same component RV.
One component RV can be the target of at most one SyncId (but as mentioned above, it can be the source for multiple SyncIds).

With those facts, let's now see how the overall flow will work.

ResyncDegradedMVs() will be called everytime RM gets notified that there are one or more degraded MVs that need attention.
It queries the list of degraded MVs, for each degraded MV it check if it hosts the "lowest online RV" (LOR). If it doesn't then it just skips that degraded MV. 
If it does host the LOR for a degraded MV, it does the following:
- For all component RVs of that MV which are in OutOfSync state it, it has to send a StartSync message with Source RV as the LOR and Target RV as the OutOfSync RV. The StartSync message will be sent to the LOR (we can start simple by sending the message to only the LOR where it's really needed, and then extend it to others for more resilience).  Once it gets a successful reply for all the StartSync messages, it then goes ahead and makes changes to the persistent MV info in clustermap. Specifically it changes the OutOfSYnc RVs to Synving and MV state to Syncing. Once the MV changes are persisted, it posts a SyncJob in a channel. The SyncJob will contain the details required to perform the sync, namely the MV, the target component RV it has to sync, the SyncId corresponding to this sync job, etc.
- One of the go routine picks this SyncJob from the channel and calls ResyncComponentRV() to perform the actual chunk transfer. Once the xfer is complete it sends the EndSync (with the SyncId corresponding to the SyncJob) message to the Source RV/LOR. On successful EndSync response, it posts a completion for the SyncJob.
- ResyncDegradedMVs() will be monitoring this channel and as soon as it sees a successful job it then updates the persistent state of the MV in clustermap. The target component RV corresponding to this SyncJob is changed from "syncing" to "online" and if there's no other syncjob running, then the MV state is also set to online. The MV state is now persisted in the global clustermap.

linuxsmiths (approved):

PR #1757, opened 2025-05-13
streamdir changes for unqualified path

syeleti-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

### DIFF
--- a/component/distributed_cache/distributed_cache.go
+++ b/component/distributed_cache/distributed_cache.go
@@ -390,29 +390,44 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
 		log.Debug("DistributedCache::StreamDir : Path is having Debug subcomponent, path : %s", options.Name)
 		return debug.StreamDir(options)
 	} else {
+		// When enumerating a fresh directory, options.IsFsDcache must be true.
+		common.Assert(token != "" || *options.IsFsDcache == true)
+
+		// When enumerating a fresh directory, options.DcacheEntries must be empty.
+		common.Assert(token != "" || len(options.DcacheEntries) == 0)
 		//
-		// Semantics for Readdir for unquailified path, If the dirent exist in both Dcache and Azure filesystem, then dirent
-		// present in the dcache takes the precedence over azure, also both the entries would be listed.
-		// To know which virtual fs we are currently in we use options.IsFsDcache.
+		// Semantics for Readdir for unquailified path, if a dirent exists in both Dcache and Azure filesystem,
+		// then dirent present in the dcache takes the precedence over Azure and the entry in Azure is masked and
+		// only the entry in dcache is listed. This is to match user expectation when they actually read such a
+		// file, the file from dcache is read, hence it makes sense to list the same.
+		// To know which virtual fs we are currently in we use options.IsFsDcache. This is set to true in opendir()
+		// and thus we start by enumerating from dcache. Once dcache enumeration hits eod (signified by empty token
+		// return) we set options.IsFsDcache to false so that on receiving the next Streamdir() call from fuse we
+		// start enumerating from Azure. We store all the entries enumerated from dcache in options.DcacheEntries
+		// map which allows us to skip any entry already listed from dcache from Azure's listing.
 		//
 		if *options.IsFsDcache { // List from dcache.
-			log.Debug("DistributedCache::StreamDir : Listing on Unquailified path, listing from dcache, path : %s", options.Name)
+			log.Debug("DistributedCache::StreamDir : Listing on Unqualified path, listing from dcache, path : %s", options.Name)
 			rawPath = filepath.Join(mm.GetMdRoot(), "Objects", rawPath)
 			options.Name = rawPath
 			if dirList, token, err = dc.NextComponent().StreamDir(options); err != nil {
 				return dirList, token, err
 			}
+
 			dirList = parseDcacheMetadataForDirEntries(dirList)
 			for _, attr := range dirList {
 				options.DcacheEntries[attr.Name] = struct{}{}
 			}
+
 			if token == "" {
-				// Now start listing from the Azure.
+				// Empty token signifies end-of-directory for dcache listing, start listing from Azure.
+				// We set token to the special non-empty value to prevent fuse from treating this as
+				// end-of-directory.
 				*options.IsFsDcache = false
 				token = dcacheDirContToken
 			}
 		} else { // List from Azure.
-			log.Debug("DistributedCache::StreamDir : Listing on Unquailified path, listing from Azure, path : %s", options.Name)
+			log.Debug("DistributedCache::StreamDir : Listing on Unqualified path, listing from Azure, path : %s", options.Name)
 			// Reset the token if it's starting to iterate from start.
 			if options.Token == dcacheDirContToken {
 				options.Token = ""
--- a/component/distributed_cache/distributed_cache.go
+++ b/component/distributed_cache/distributed_cache.go
@@ -410,6 +410,9 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
 			}
 		} else {
 			log.Debug("DistributedCache::StreamDir : Listing on Unquailified path, listing from Azure, path : %s", options.Name)
+			if token == dcacheDirContToken {
+				token = ""
+			}
 			options.Name = rawPath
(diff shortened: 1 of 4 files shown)


component/distributed_cache/distributed_cache.go
```diff
@@ -365,39 +366,75 @@ func (dc *DistributedCache) GetAttr(options internal.GetAttrOptions) (*internal.
...
 		log.Debug("DistributedCache::StreamDir : Path is having Debug subcomponent, path : %s", options.Name)
 		return debug.StreamDir(options)
 	} else {
 		//
 		// Semantics for Readdir for unquailified path, If the dirent exist in both Dcache and Azure filesystem, then dirent
-		// present in the dcache takes the precedence.
+		// present in the dcache takes the precedence over azure, also both the entries would be listed.
+		// To know which virtual fs we are currently in we use options.IsFsDcache.
 		//
-		// TODO: Implement this, rn serving from Azure.
```
linuxsmiths, line 383:
```
		//
		// Semantics for Readdir for unquailified path, if a dirent exists in both Dcache and Azure filesystem,
		// then dirent present in the dcache takes the precedence over Azure and the entry in Azure is masked and
		// only the entry in dcache is listed. This is to match user expectation when they actually read such a
		// file, the file from dcache is read, hence it makes sense to list the same.
		// To know which virtual fs we are currently in we use options.IsFsDcache. This is set to true in opendir()
		// and thus we start by enumerating from dcache. Once dcache enumeration hits eod (signified by empty token
		// return) we set options.IsFsDcache to false so that on receiving the next Streamdir() call from fuse we
		// start enumerating from Azure. We store all the entries enumerated from dcache in options.DcacheEntries
		// map which allows us to skip any entry already listed from dcache from Azure's listing.
		//
```

component/distributed_cache/distributed_cache.go
```diff
@@ -365,39 +366,75 @@ func (dc *DistributedCache) GetAttr(options internal.GetAttrOptions) (*internal.
...
-		common.Assert(rawPath == options.Name, rawPath, options.Name)
-	}
+		if *options.IsFsDcache { // List from dcache.
+			log.Debug("DistributedCache::StreamDir : Listing on Unquailified path, listing from dcache, path : %s", options.Name)
```
linuxsmiths, line 399: Unqualified

component/distributed_cache/distributed_cache.go
```diff
@@ -365,39 +366,75 @@ func (dc *DistributedCache) GetAttr(options internal.GetAttrOptions) (*internal.
...
+			}
+			dirList = parseDcacheMetadataForDirEntries(dirList)
+			for _, attr := range dirList {
+				options.DcacheEntries[attr.Name] = struct{}{}
+			}
+			if token == "" {
+				// Now start listing from the Azure.
+				*options.IsFsDcache = false
+				token = dcacheDirContToken
+			}
```
linuxsmiths, line 413: pls leave a new line after every block., so

```
			rawPath = filepath.Join(mm.GetMdRoot(), "Objects", rawPath)
			options.Name = rawPath
			if dirList, token, err = dc.NextComponent().StreamDir(options); err != nil {
				return dirList, token, err
			}

			dirList = parseDcacheMetadataForDirEntries(dirList)
			for _, attr := range dirList {
				options.DcacheEntries[attr.Name] = struct{}{}
			}

			if token == "" {
				// Empty token signifies end-of-directory for dcache listing, start listing from Azure.
				// We set token to the special non-empty value to prevent fuse from treating this as
				// end-of-directory.
				*options.IsFsDcache = false
				token = dcacheDirContToken
			}
```

pls copy the above as it has a comment suggestion too

component/distributed_cache/distributed_cache.go
```diff
@@ -365,39 +366,75 @@ func (dc *DistributedCache) GetAttr(options internal.GetAttrOptions) (*internal.
...
 	} else {
 		//
 		// Semantics for Readdir for unquailified path, If the dirent exist in both Dcache and Azure filesystem, then dirent
-		// present in the dcache takes the precedence.
+		// present in the dcache takes the precedence over azure, also both the entries would be listed.
+		// To know which virtual fs we are currently in we use options.IsFsDcache.
 		//
-		// TODO: Implement this, rn serving from Azure.
-		common.Assert(rawPath == options.Name, rawPath, options.Name)
-	}
```
linuxsmiths, line 385: add these asserts in the beginning

```
// When enumerating a fresh directory, options.IsFsDcache must be true.
common.Assert(token != "" || *options.IsFsDcache == true)

// When enumerating a fresh directory, options.DcacheEntries must be empty.
common.Assert(token != "" || len(options.DcacheEntries) == 0)
```

component/distributed_cache/distributed_cache.go
```diff
@@ -365,39 +366,75 @@ func (dc *DistributedCache) GetAttr(options internal.GetAttrOptions) (*internal.
...
+			for _, attr := range dirList {
+				options.DcacheEntries[attr.Name] = struct{}{}
+			}
+			if token == "" {
+				// Now start listing from the Azure.
+				*options.IsFsDcache = false
+				token = dcacheDirContToken
+			}
+		} else { // List from Azure.
+			log.Debug("DistributedCache::StreamDir : Listing on Unquailified path, listing from Azure, path : %s", options.Name)
```
linuxsmiths, line 415: same typo as above

component/distributed_cache/distributed_cache.go
```diff
@@ -365,39 +366,75 @@ func (dc *DistributedCache) GetAttr(options internal.GetAttrOptions) (*internal.
...
+		// To know which virtual fs we are currently in we use options.IsFsDcache.
 		//
-		// TODO: Implement this, rn serving from Azure.
-		common.Assert(rawPath == options.Name, rawPath, options.Name)
-	}
+		if *options.IsFsDcache { // List from dcache.
+			log.Debug("DistributedCache::StreamDir : Listing on Unquailified path, listing from dcache, path : %s", options.Name)
+			rawPath = filepath.Join(mm.GetMdRoot(), "Objects", rawPath)
+			options.Name = rawPath
+			if dirList, token, err = dc.NextComponent().StreamDir(options); err != nil {
```
linuxsmiths, line 402: if a directory has exactly 5000 entries, then will it return empty token when returning 5000 entries in a call signifying eod.
If it doesn't then the next time when fuse calls us will StreamDir() above return an empty dirList. In such case we will return a non-empty token (dcacheDirContToken) and no entry. Will that confuse fuse or it can handle it?
We should test with exact 5000 entries in a dir

component/distributed_cache/distributed_cache.go
```diff
@@ -365,39 +366,75 @@ func (dc *DistributedCache) GetAttr(options internal.GetAttrOptions) (*internal.
...
-	// todo: exclude the files which are in deleting state.
-	if isDcachePath {
-		dirList = parseDcacheMetadataForDirEntries(dirList)
+			// Ignore the dirent if it's already retured by the dcache listing.
+			var modifiedDirList []*internal.ObjAttr = make([]*internal.ObjAttr, 0, len(dirList))
+			for _, attr := range dirList {
+				if _, ok := options.DcacheEntries[attr.Name]; !ok {
+					modifiedDirList = append(modifiedDirList, attr)
+				}
+			}
```
linuxsmiths, line 432: again, if this results in an empty modifiedDirList  (as all entries are present in both azure and dcache), will fuse be able to handle an empty dirlist return

syeleti-msft: There is no explicit way to signal the fuse, but when the buf is passed to us in readdir call, we call the [filler](https://github.com/libfuse/libfuse/blob/84a27de6a644163eda97b7052e064774ec73f925/include/fuse.h#L87) function to fill the dirent into the buf. Now when the filler function returns 1 we know that buf is full, return from the cur readdir call and we get the next readdir call. but if the filler function return 0 after adding the dirent and we returned from the readdir call this means we reached EOD.

This needs some logic change, when we are transitioning from dcache to azure we need to assure that enough entries are there in childcachedir structure to not stop listing after dcache.

component/distributed_cache/distributed_cache.go
syeleti-msft: I'll check this

syeleti-msft: The server is returning empty string for exactly 5000 entries

PR #1770, opened 2025-05-20
Add unlink support

syeleti-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

### DIFF
--- a/component/distributed_cache/distributed_cache.go
+++ b/component/distributed_cache/distributed_cache.go
@@ -865,7 +865,61 @@ func (dc *DistributedCache) CloseFile(options internal.CloseFileOptions) error {
 }
 
 func (dc *DistributedCache) DeleteFile(options internal.DeleteFileOptions) error {
-	return syscall.ENOTSUP
+	log.Debug("DistributedCache::DeleteFile: Delete file: %s", options.Name)
+	var dcacheErr, azureErr error
+
+	isAzurePath, isDcachePath, isDebugPath, rawPath := getFS(options.Name)
+
+	if isDcachePath {
+		log.Debug("DistributedCache::DeleteFile: Delete for Dcache file: %s", options.Name)
+		err := fm.DeleteDcacheFile(rawPath)
+		if err != nil {
+			log.Err("DistributedCache::DeleteFile: Delete failed for Dcache file: %s", options.Name)
+			return err
+		}
+	} else if isAzurePath {
+		log.Debug("DistributedCache::DeleteFile: Delete Azure file: %s", options.Name)
+		options.Name = rawPath
+		err := dc.NextComponent().DeleteFile(options)
+		if err != nil {
+			log.Err("DistributedCache::DeleteFile: Delete failed for Azure file: %s", options.Name)
+			return err
+		}
+	} else if isDebugPath {
+		return syscall.EROFS
+	} else {
+		//
+		// Semantics for Unqualified Path, Delete from both Azure and Dcache. If file is present in only  onequalified
+		// path, then delete only from that path. If the call has come here it already means that the file is present in
+		// atleast one qualified path as stat would be checked before doing deletion.
+		//
+		log.Debug("DistributedCache::DeleteFile: Delete Dcache file for Unqualified Path: %s", options.Name)
+
+		dcacheErr = fm.DeleteDcacheFile(rawPath)
+
+		if dcacheErr != nil {
+			log.Err("DistributedCache::DeleteFile: Delete failed for Unqualified Path Dcache file: %s", options.Name)
+			// Continue only if the above dcacheError is valid, ex: blob not found. Else fail the delete.
+			if dcacheErr == syscall.ENOENT {
+				dcacheErr = nil
+			} else {
+				return dcacheErr
+			}
+		}
+
+		options.Name = rawPath
+		log.Debug("DistributedCache::DeleteFile: Delete Azure file for Unqualified Path: %s", options.Name)
+		azureErr = dc.NextComponent().DeleteFile(options)
+		if azureErr != nil {
+			log.Err("DistributedCache::DeleteFile: Delete failed for Unqualified Path Azure file: %s", options.Name)
+		}
+
+		if azureErr == syscall.ENOENT {
+			azureErr = nil
+		}
+	}
+
+	return errors.Join(dcacheErr, azureErr)
 }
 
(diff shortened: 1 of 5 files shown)


component/distributed_cache/distributed_cache.go
```diff
@@ -865,7 +865,61 @@ func (dc *DistributedCache) CloseFile(options internal.CloseFileOptions) error {
...
+	log.Debug("DistributedCache::DeleteFile: Delete file: %s", options.Name)
+	var dcacheErr, azureErr error
+
+	isAzurePath, isDcachePath, isDebugPath, rawPath := getFS(options.Name)
+
+	if isDcachePath {
+		log.Debug("DistributedCache::DeleteFile: Delete for Dcache file: %s", options.Name)
+		err := fm.DeleteDcacheFile(rawPath)
+		if err != nil {
+			log.Err("DistributedCache::DeleteFile: Delete failed for Dcache file: %s:%v", options.Name, err)
```
linuxsmiths, line 877: log.Err("DistributedCache::DeleteFile: Delete failed for Dcache file %s: %v", options.Name, err)

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -418,13 +432,64 @@ func (m *BlobMetadataManager) getFile(filePath string) ([]byte, int64, error) {
...
+		"state":               &state,
+	}
+
+	err := m.storageCallback.PutBlobInStorage(internal.WriteFromBufferOptions{
+		Name:                   path,
+		Data:                   fileMetadata,
+		Metadata:               metadata,
+		IsNoneMatchEtagEnabled: false,
+		EtagMatchConditions:    "",
+	})
```
linuxsmiths, line 485: we should only set the state property to deleting and nothing else - none of the other properties and also not file data. So you may want to call m.storageCallback.SetMetaPropertiesInStorage(). It should get the properties and the only update the state to deleting.
Just like updateHandleCount() updates only the opencount property.

internal/dcache/file_manager/utils.go
```diff
@@ -213,6 +215,40 @@ func OpenDcacheFile(fileName string) (*DcacheFile, error) {
 	}, nil
 }
 
+func DeleteDcacheFile(fileName string) error {
+	fileMetadataBytes, fileSize, fileState, err := mm.GetFile(fileName)
```
linuxsmiths, line 219: we won't need the fileMetadataBytes to update the file state (to deleting). See below.

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -386,30 +398,32 @@ func (m *BlobMetadataManager) createFileFinalize(filePath string, fileMetadata [
...
-func (m *BlobMetadataManager) getFile(filePath string) ([]byte, int64, error) {
+func (m *BlobMetadataManager) getFile(filePath string) ([]byte, int64, dcache.FileState, error) {
 	common.Assert(len(filePath) > 0)
 
 	path := filepath.Join(m.mdRoot, "Objects", filePath)
 	// Get the metadata content from storage.
 	data, prop, err := m.getBlobSafe(path)
 	if err != nil {
 		log.Debug("GetFile:: Failed to get metadata file content for file %s: %v", path, err)
-		return nil, -1, err
```
linuxsmiths, line 397: pls add this

+               //
+               // getBlobSafe() should only fail when blob is non-existent.
+               // Assert to catch any other error.
+               //
+               common.Assert(errors.Is(err, syscall.ENOENT), err)

internal/dcache/file_manager/utils.go
syeleti-msft: not needed anymore

component/distributed_cache/utils.go
```diff
@@ -229,28 +230,57 @@ func parseDcacheMetadata(attr *internal.ObjAttr) error {
...
 		log.Err("DistributedCache::GetAttr: %v", err)
 		common.Assert(false, err)
+		return err
+	}
+
+	// parse file state.
+	if state, ok := attr.Metadata["state"]; ok {
+		if !(*state == string(dcache.Writing) || *state == string(dcache.Ready)) {
+			err = fmt.Errorf("File: %s, has invalid state:  [%s] metadata property", attr.Name, *state)
+			log.Err("DistributedCache::GetAttr: %v", err)
```
jainakanksha-msft, line 246: Correct all the loggings.
Logging pattern which we are following is like fileName::MethodName:following the error message

syeleti-msft: Done

component/distributed_cache/distributed_cache.go
```diff
@@ -906,46 +947,51 @@ func (dc *DistributedCache) DeleteFile(options internal.DeleteFileOptions) error
...
 		log.Debug("DistributedCache::DeleteFile: Delete Azure file for Unqualified Path: %s", options.Name)
 		azureErr = dc.NextComponent().DeleteFile(options)
 		if azureErr != nil {
-			log.Err("DistributedCache::DeleteFile: Delete failed for Unqualified Path Azure file %s: %v", options.Name, azureErr)
-		}
-
-		if azureErr == syscall.ENOENT {
-			azureErr = nil
+			log.Err("DistributedCache::DeleteFile: Delete failed for Unqualified Path Azure file %s: %v",
+				options.Name, azureErr)
```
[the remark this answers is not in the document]
linuxsmiths: yes, but only if both are not noent

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -386,45 +399,92 @@ func (m *BlobMetadataManager) createFileFinalize(filePath string, fileMetadata [
...
+	common.Assert(len(filePath) > 0)
+	path := filepath.Join(m.mdRoot, "Objects", filePath)
+	log.Debug("renameFileToDeleting:: %s", path)
+
+	renamedPath := path + dcache.DcacheDeletingFileNameSuffix
+
+	err := m.storageCallback.RenameFileInStorage(internal.RenameFileOptions{
+		Src: path,
+		Dst: renamedPath,
+	})
```
linuxsmiths, line 479: does this fail/succeed if the target file exists?

linuxsmiths (approved):

PR #1771, opened 2025-05-20
AI Scripts from build prep

vibhansa-msft:
Adding some AI scripts


test/ai/checkpointLoad.sh
```diff
@@ -0,0 +1,132 @@
+
```
jainakanksha-msft, line 1: Kindly add conclusive context, what these individual scripts are doing.
We might take some reference in future.

souravgupta-msft (approved):

jainakanksha-msft (approved):

souravgupta-msft (approved):

PR #1773, opened 2025-05-20
Reborn failure correction and retry to mount

vibhansa-msft:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description
We create a .pid file in working directory when blobfuse mounts. If the last unmount was unsuccessful there is a chance that process no longer exists but the .pid file is still under lock. In this situation if we try to remount to same path the mount will fail with "failed to daemonize the application" error. We can delete the .pid file at this stage and retry and mount shall be successful.
If the delete of .pid fails then we need to error out the mount command.

## How Has This Been Tested?
No automation possible, have verified it manually. To validate this just create the .pid file respective to your mount path manually and lock the file using "flock" command (with sleep) and then try to mount the path. Before fix the mount will fail, and post fix mount will delete the .pid file and retry the mount.

## Checklist

## Related Links
NA


cmd/mount.go
```diff
@@ -484,6 +484,7 @@ var mountCmd = &cobra.Command{
 
 		log.Info("mount: Mounting blobfuse2 on %s", options.MountPath)
 		if !options.Foreground {
+			// pidFile := strings.Replace(options.MountPath, "/", "_", -1) + fmt.Sprintf("%d", os.Getpid()) + ".pid"
```
syeleti-msft, line 487: remove this comment

syeleti-msft (approved):

cmd/mount.go
```diff
@@ -512,10 +512,18 @@ var mountCmd = &cobra.Command{
...
-				log.Err("mount : failed to daemonize application [%v]", err)
-				return Destroy(fmt.Sprintf("failed to daemonize application [%s]", err.Error()))
+				log.Err("mount : failed to daemonize application [%s], trying auto cleanup", err.Error())
+				rmErr := os.Remove(pidFileName)
+				if rmErr != nil {
+					log.Err("mount : auto cleanup failed [%v]", rmErr.Error())
+					return Destroy(fmt.Sprintf("failed to daemonize application [%s]", err.Error()))
+				}
+				goto retry
 			}
```
souravgupta-msft, line 527: Retry should be done only twice or fixed number of times.

vibhansa-msft: There is no retry here. Only if remove of file succeeds then we retry to daemonize which will succeed. If remove of pid file fails then we just error out. Ideally it shall never retry more than once.

PR #1781, opened 2025-05-22
List fix for Azure Qualified Path

syeleti-msft:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

### DIFF
--- a/component/distributed_cache/distributed_cache.go
+++ b/component/distributed_cache/distributed_cache.go
@@ -390,6 +390,7 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
 		}
 		dirList = parseDcacheMetadataForDirEntries(dirList)
 	} else if isAzurePath {
+	listAzurePath:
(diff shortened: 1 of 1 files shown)


component/distributed_cache/distributed_cache.go
```diff
@@ -400,6 +401,12 @@ func (dc *DistributedCache) StreamDir(options internal.StreamDirOptions) ([]*int
 		if isMountPointRoot(rawPath) {
 			dirList = hideCacheMetadata(dirList)
 		}
+
+		// Retry listing when the list has no dirents after hiding the root cache dir.
+		if (len(dirList) == 0) && token != "" {
+			options.Token = token
+			goto listAzurePath
+		}
```
linuxsmiths, line 409: this is a common condition we must ensure before we exit the function, i.e., we should never exit the function with an empty dirList and not-empty token.
So, move L477-486 to L488 and then simply goto L383.

This way we will never exit with that buggy condition.

linuxsmiths (approved):

PR #1786, opened 2025-05-23
Override ownership of clustermap

jainakanksha-msft:


internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -961,6 +964,98 @@ func (cmi *ClusterManager) updateStorageClusterMapWithMyRVs() error {
...
+	clusterMap.LastUpdatedAt = time.Now().Unix()
+	updatedClusterMapBytes, err := json.Marshal(clusterMap)
+	if err != nil {
+		log.Err("ClusterManager::overrideClusterMapIfStale: Marshal failed for clustermap: %v %+v",
+			err, clusterMap)
+		common.Assert(false, err)
+		return false, err
+	}
+
+	if err = mm.UpdateClusterMapEnd(updatedClusterMapBytes); err != nil {
```
linuxsmiths, line 1034: since UpdateClusterMapEnd() is not atomic we cannot safely call it, some other node may race with us.
We have to follow the pattern for updating clustermap
1. get cluster map with etag
2. cluster map start with etag
3. if successful, cluster map end

linuxsmiths (approved):

PR #1803, opened 2025-05-30
Duplicate RVId handling

jainakanksha-msft:
This change will block the user to start blobfuse with distributed_cache in case the different cache-dir have the same fileSYstem UUID.


internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -735,6 +740,17 @@ func (cmi *ClusterManager) safeCleanupMyRVs(myRVs []dcache.RawVolume) (bool, err
...
+		// Ensure no two different cache-dir report the same rvId(filesystem GUID)
+		for _, inputRV := range myRVs {
+			for _, cmRV := range myRVsFromClustermap {
+				if inputRV.RvId == cmRV.RvId && inputRV.LocalCachePath != cmRV.LocalCachePath {
+					return false, fmt.Errorf(
+						"ClusterManager::safeCleanupMyRVs: duplicate RVid (filesystem GUID) %s reported for paths %s and %s",
+						inputRV.RvId, cmRV.LocalCachePath, inputRV.LocalCachePath)
+				}
+			}
+		}
```
linuxsmiths: This check should be done against *all* RVs and not just our RVs.
We don't want to add an RV with the same RVid as some other nodes' RV, that will cause problems.
Use cm.RvIdToName() and cm.RVNameToNodeId(), etc for performing the check against existing RVIds from othr nodes

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -724,23 +728,34 @@ func (cmi *ClusterManager) safeCleanupMyRVs(myRVs []dcache.RawVolume) (bool, err
...
+			}
 		}
 
 		rvStillOnline := false
 		for _, rv := range myRVs {
 			log.Info("ClusterManager::safeCleanupMyRVs: Checking my RV %+v", rv)
 
 			// Check online status for this RV.
-			for rvName, rvInfo := range myRVsFromClustermap {
+			for rvName, rvInfo := range rvsFromClustermap {
```
linuxsmiths, line 758: we want to do this only for our RVs and not for all RVs.
Only the check for dup rvid, we need to do for all RVs, the offline check we need to do only for our RVs as that's all we care about.

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -724,23 +728,34 @@ func (cmi *ClusterManager) safeCleanupMyRVs(myRVs []dcache.RawVolume) (bool, err
...
-		if len(myRVsFromClustermap) > 0 {
-			log.Info("ClusterManager::safeCleanupMyRVs: Got %d of my RV(s) from clustermap %+v",
-				len(myRVsFromClustermap), myRVsFromClustermap)
+		if len(rvsFromClustermap) > 0 {
+			log.Info("ClusterManager::safeCleanupMyRVs: Got %d of RV(s) from clustermap %+v",
+				len(rvsFromClustermap), rvsFromClustermap)
 		} else {
-			log.Info("ClusterManager::safeCleanupMyRVs: No my RV(s) in clustermap")
+			log.Info("ClusterManager::safeCleanupMyRVs: No RV(s) in clustermap")
+		}
```
linuxsmiths: some of these checks make better sense for my RVs

jainakanksha-msft: I think that is not making any difference.
Like if there is no RV in clustermap, then also it is delivering same information.

linuxsmiths (approved):

PR #1805, opened 2025-05-30
Return correct permission and times in stat for clustermap proc file

syeleti-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

This will return the last-updated-at field in clusetermap as atime, mtime, ctime and also file perm as 0444

### DIFF
--- a/internal/dcache/debug/debug_fs.go
+++ b/internal/dcache/debug/debug_fs.go
@@ -39,6 +39,7 @@ import (
 	"sync"
 	"sync/atomic"
 	"syscall"
+	"time"
 
 	"github.com/Azure/azure-storage-fuse/v2/common"
 	"github.com/Azure/azure-storage-fuse/v2/internal"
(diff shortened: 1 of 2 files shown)


internal/dcache/debug/debug_fs.go
```diff
@@ -86,13 +89,20 @@ func init() {
...
+			Path:  options.Name,
+			Mode:  0444,
+			Mtime: time.Now(),
+			Atime: time.Now(),
+			Ctime: time.Now(),
+			Size:  0,
 		}
-		return &attr, nil
+
+		pFile.getAttr(pFile, attr)
```
linuxsmiths, line 103: this should be called only if pFile.getAttr is not nil

syeleti-msft: done

PR #1807, opened 2025-06-02
Misc perf fixes

linuxsmiths:
Set most of the perf config values with logical explanation. Use callers thread for the last replica write.
getRPCClientNoLock() will wait till a free client is available. Will panic if a free client is not available till 60 secs.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

## Checklist

## Related Links
- [Issues](<link>)

### DIFF
--- a/internal/dcache/file_manager/file_manager.go
+++ b/internal/dcache/file_manager/file_manager.go
@@ -65,24 +65,85 @@ type fileIOManager struct {
 
 var fileIOMgr fileIOManager
 
-func NewFileIOManager(workers int, numReadAheadChunks int, numStagingChunks int, bufSize int, maxBuffers int) {
+func NewFileIOManager() error {
+	//
+	// A worker runs either readChunk() or writeChunk(), so this is the number of chunks we can be
+	// reading/writing in parallel. fileIOManager is one for the entire blobfuse process so these
+	// chunks can be spread across multiple files and served from multiple nodes/RVs.
+	// Note that fir writeChunk one worker is used up regardless of the NumReplicas setting. Each
+	// replica write uses us one ReplicationManager worker but only one fileIOManager worker per
+	// MV write (not replica write).
+	// Since all these reads/writes may be served from different nodes/RVs, the limiting factor would
+	// be the n/w b/w of this node. We need enough parallel readChunk/writeChunk for maxing out the
+	// n/w b/w of the node. Keeping small files in mind, and given that go routines are not very
+	// expensive, we keep 1000 workers.
+	//
+	workers := 1000
+
+	//
+	// How many chunks will we readahead per file.
+	// To achieve high sequential read throughput, this number should be kept reasonably high.
+	// With 4MiB chunk size, 64 readahead chunks will use up 256MiB of memory per file.
+	//
+	numReadAheadChunks := 64
+
+	//
+	// How many writeback chunks per file.
+	// These many chunks we will store per file before we put back pressure on the writer application.
+	// Obviously we start upload of chunks as soon as we have a full chunk, so only those chunks will
+	// eat up the writeback space which are not completely written to the target node.
+	// Hopefully we won't be writing too many large files simultaneously, so we can keep this number
+	// high enough to give 1GiB writeback space per file.
+	//
+	numStagingChunks := 256
+
+	//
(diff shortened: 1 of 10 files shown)


internal/dcache/file_manager/file_manager.go
```diff
@@ -65,35 +65,102 @@ type fileIOManager struct {
...
+	// chunks can be spread across multiple files and served from multiple nodes/RVs.
+	// Note that fir writeChunk one worker is used up regardless of the NumReplicas setting. Each
+	// replica write uses us one ReplicationManager worker but only one fileIOManager worker per
+	// MV write (not replica write).
+	// Since all these reads/writes may be served from different nodes/RVs, the limiting factor would
+	// be the n/w b/w of this node. We need enough parallel readChunk/writeChunk for maxing out the
+	// n/w b/w of the node. Keeping small files in mind, and given that go routines are not very
+	// expensive, we keep 1000 workers.
+	//
+	workers := 1000
```
souravgupta-msft, line 81: We can define a constant for this and other variables instead of hardcoding the values.

internal/dcache/file_manager/file_manager.go
```diff
@@ -65,35 +65,102 @@ type fileIOManager struct {
...
-	common.Assert(numReadAheadChunks > 0)
-	common.Assert(numStagingChunks > 0)
+func NewFileIOManager() error {
+	//
+	// A worker runs either readChunk() or writeChunk(), so this is the number of chunks we can be
+	// reading/writing in parallel. fileIOManager is one for the entire blobfuse process so these
+	// chunks can be spread across multiple files and served from multiple nodes/RVs.
+	// Note that fir writeChunk one worker is used up regardless of the NumReplicas setting. Each
+	// replica write uses us one ReplicationManager worker but only one fileIOManager worker per
+	// MV write (not replica write).
```
souravgupta-msft:
```suggestion
	// Note that for writeChunk one worker is used up regardless of the NumReplicas setting. Each
	// replica write uses more than one ReplicationManager worker but only one fileIOManager worker per
	// MV write (not replica write).
```

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -367,13 +373,17 @@ retry:
...
+			// inline and save one threadpool thread.
 			//
+			isLastComponentRV := componentRVIdx == (len(componentRVs) - 1)
 			rm.tp.schedule(&workitem{
 				targetNodeID: targetNodeID,
 				rvName:       rv.Name,
 				putChunkReq:  rpcReq,
 				respChannel:  responseChannel,
-			})
+			}, isLastComponentRV /* runInline */)
```
souravgupta-msft, line 386: Why are we adding an extra parameter for running the operation inline?

internal/dcache/file_manager/file_manager.go
linuxsmiths: these are defined and explained here, defining an extra level of variables will only make things less clearer.

internal/dcache/replication_manager/replication_manager.go
linuxsmiths: how else do we tell threadpool to run it inline instead of scheduling a thread for it?
another way of doing it would be to add that parameter to workitem, but that imo would be less clearer, as it's an additional guidance to schedule() telling it whether it should run the workitem inline or schedule it to be run by another thread. It's not a property of workitem per se.

souravgupta-msft: My question was more around why we are running the operation inline for the last component RV. We can just send it to the scheduler like how we do for the other RVs. Any reason for doing this for the last RV?

souravgupta-msft (approved):

internal/dcache/replication_manager/replication_manager.go
linuxsmiths: oh, this is what I explained in the call earlier today. It's an optimization to save an extra context switch (for NumReplicas=3 it's 1/3rd of the context switches). Note that if we schedule another RM thread the current fileIoMgr thread who made the WriteMV() call is anyway going to wait for all the RM threads to complete, so we can as well let it run one of the replica writes thus saving on some context switches.

PR #1815, opened 2025-06-03
Undo reserved space and sync jobs added in case of failed RV state ch…

linuxsmiths:
…anges

JoinMV sets reserved space on RV. This must be undone if the fix-mv workflow doesn't complete. We do this when refreshFromClusterMap() finds out some RV in outofsync state in rvInfo but it's offline in clustermap.

Similarly there are other state transitions, f.e., StartSync adds sync job to mvInfo, but if sync wokflow cannot complete and this RV is reverted back to outofsync state by refreshFromClusterMap(), we must purge the sync jobs.

Also made changes to synchronize calls to UpdateComponentRVState() to avoid huge number of wasted retries when lot of parallel sync jobs result in too many calls to UpdateComponentRVState() in close time proximity.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

## Checklist

## Related Links
- [Issues](<link>)

### DIFF
--- a/internal/dcache/rpc/server/handler.go
+++ b/internal/dcache/rpc/server/handler.go
@@ -135,7 +135,7 @@ type mvInfo struct {
 	// added by new-mv workflow. Put another way, this will be non-zero only for MV replicas which are in
 	// outofsync or syncing state. On an EndSync request, that converts syncing state to online state
 	// mvInfo.reservedSpace is cleared and its value is added to mvInfo.totalChunkBytes, also this is
-	// reduced from rvInfo.reservedSpace as this space is no longer reseved but rather actual space is now
+	// reduced from rvInfo.reservedSpace as this space is no longer reserved but rather actual space is now
 	// used by chunks stored in the RV.
 	// If an MV replica cannot complete resync, this must be reduced from rvInfo.reservedSpace.
 	// This means while an MV replica is being sync'ed the space used on the RV may be overcompensated, this
@@ -301,7 +301,7 @@ func (rv *rvInfo) addToMVMap(mvName string, mv *mvInfo, reservedSpace int64) {
 	// Note that the actual space reservation is done in rvInfo.reservedSpace, while mvInfo.reservedSpace
 	// is used for undoing rvInfo.reservedSpace in case the sync does not complete.
 	//
-	common.Assert(mv.reservedSpace.Load() == 0, mv.reservedSpace, rv.rvName, mvName)
+	common.Assert(mv.reservedSpace.Load() == 0, mv.reservedSpace.Load(), rv.rvName, mvName)
 	mv.reservedSpace.Store(reservedSpace)
 	rv.incReservedSpace(reservedSpace)
 
@@ -331,11 +331,8 @@ func (rv *rvInfo) incReservedSpace(bytes int64) {
 // Decrement the reserved space for this RV.
 func (rv *rvInfo) decReservedSpace(bytes int64) {
 	common.Assert(bytes > 0)
-	//
-	// TODO: Uncomment this only after clustermanager joinMV() correctly reserves space for MV replica.
-	//
-	// rv.reservedSpace.Add(-bytes)
-	// common.Assert(rv.reservedSpace.Load() >= 0, rv.rvName, rv.reservedSpace.Load())
+	rv.reservedSpace.Add(-bytes)
+	common.Assert(rv.reservedSpace.Load() >= 0, rv.rvName, rv.reservedSpace.Load())
 	log.Debug("rvInfo::decReservedSpace: reserved space for RV %s is %d", rv.rvName, rv.reservedSpace.Load())
 }
 
@@ -524,10 +521,7 @@ func (mv *mvInfo) isSourceOrTargetOfSync() (isSource bool, isTarget bool) {
 		}
 	})
 
-	// Unreachable code.
-	common.Assert(false)
-
-	return false, false
+	return isSource, isTarget
 }
 
 // Get component RVs for this MV.
@@ -797,7 +791,7 @@ func (mv *mvInfo) refreshFromClustermap() error {
 		if oldRvNameAndState.State == string(dcache.StateOutOfSync) && rvState == dcache.StateOffline {
 			common.Assert(mv.rv.reservedSpace.Load() >= mv.reservedSpace.Load(),
 				mv.rv.reservedSpace.Load(), mv.reservedSpace.Load())
-			mv.rv.reservedSpace.Add(-mv.reservedSpace.Load())
+			mv.rv.decReservedSpace(mv.reservedSpace.Load())
 			mv.reservedSpace.Store(0)
 		}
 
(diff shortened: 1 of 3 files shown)


internal/dcache/rpc/server/handler.go
```diff
@@ -128,6 +128,20 @@ type mvInfo struct {
...
 	totalChunkBytes atomic.Int64
 
+	// Amount of space reserved for this MV replica, on the hosting RV.
+	// When a new mvInfo is created by JoinMV() this is set to the ReserveSpace parameter to JoinMV.
+	// This is also added to rvInfo.reservedSpace to reserve space in the RV.
+	// This is non-zero only for MV replicas which are added by the fix-mv workflow and not for MV replicas
+	// added by new-mv workflow. Put another way, this will be non-zero only for MV replicas which are in
+	// outofsync or syncing state. On an EndSync request, that converts syncing state to online state
+	// mvInfo.reservedSpace is cleared and its value is added to mvInfo.totalChunkBytes, also this is
+	// reduced from rvInfo.reservedSpace as this space is no longer reseved but rather actual space is now
```
souravgupta-msft:
```suggestion
	// reduced from rvInfo.reservedSpace as this space is no longer reserved but rather actual space is now
```

internal/dcache/rpc/server/handler.go
```diff
@@ -670,9 +767,75 @@ func (mv *mvInfo) refreshFromClustermap() error {
...
+		//
+		if mv.rv.rvName != rvName {
+			continue
+		}
+
+		//
+		// Must have the hosting RV in the componentRVs list.
+		//
+		oldRvNameAndState := mv.getComponentRVNameAndState(rvName)
+		common.Assert(oldRvNameAndState != nil, rvName, mv.mvName, rpc.ComponentRVsToString(mv.componentRVs))
```
souravgupta-msft, line 776: This can fail in case of incorrect membership info in the server. For example, local component rvs are rv0, rv1 and rv2.
But after clustermap refresh, new rvs are rv1, rv2 and rv3. For rv3, this assert will fail as it is not present in the local rv list.

internal/dcache/rpc/server/handler.go
```diff
@@ -1458,7 +1621,7 @@ refreshFromClustermapAndRetry:
 	// JoinMV would have reserved this space before starting sync.
 	//
 	if len(req.SyncID) > 0 {
-		rvInfo.decReservedSpace(req.Length)
+		common.Assert(rvInfo.reservedSpace.Load() >= req.Length, rvInfo.reservedSpace.Load(), req.Length)
```
souravgupta-msft: Just above this line, we have this code.
```
// Increment the total chunk bytes for this MV.
mvInfo.incTotalChunkBytes(req.Length)
```

Now that we are adding `totalChunkBytes` of an MV in the EndSync call. So, for PutChunk(sync) calls we can ignore updating this count. We can have something like,

```
// Increment the total chunk bytes for this MV for client PutChunk calls.
// For sync PutChunk calls, the MV's totalChunkBytes will be updated in the EndSync call,
// indicating the successful sync operation for this RV/MV.
if len(req.SyncID) == 0 {
	mvInfo.incTotalChunkBytes(req.Length)
}
```

Also, in GetMVSize() RPC, we need to send total chunk bytes + the reserved space in MV.
```
return &models.GetMVSizeResponse{
		MvSize: mvInfo.totalChunkBytes.Load() + mvInfo.reservedSpace.Load(),
	}, nil
```

internal/dcache/rpc/server/handler.go
```diff
@@ -271,15 +289,25 @@ func (rv *rvInfo) getMVs() []string {
...
 	common.Assert(common.DirectoryExists(mvPath), mvPath)
+	common.Assert(mv.rv == rv, mv.rv.rvName, rv.rvName)
 
-	rv.mvMap.Store(mvName, val)
+	//
+	// Set reservedSpace for this MV and increment the reserved space for the hosting RV.
+	// Note that the actual space reservation is done in rvInfo.reservedSpace, while mvInfo.reservedSpace
+	// is used for undoing rvInfo.reservedSpace in case the sync does not complete.
+	//
+	common.Assert(mv.reservedSpace.Load() == 0, mv.reservedSpace, rv.rvName, mvName)
```
souravgupta-msft:
```suggestion
	common.Assert(mv.reservedSpace.Load() == 0, mv.reservedSpace.Load(), rv.rvName, mvName)
```

internal/dcache/rpc/server/handler.go
```diff
@@ -670,9 +767,75 @@ func (mv *mvInfo) refreshFromClustermap() error {
...
+		// mvInfo.totalChunkBytes will be increased by mvInfo.reservedSpace, and rvInfo.reservedSpace will
+		// be reduced by mvInfo.reservedSpace and mvInfo.reservedSpace will be set to 0.
+		// If the mvInfo has the state as StateOutOfSync while clustermap has the state as StateOffline,
+		// this means the fix-mv workflow didn't complete successfully so we need to rollback the reserved
+		// space changes.
+		//
+		if oldRvNameAndState.State == string(dcache.StateOutOfSync) && rvState == dcache.StateOffline {
+			common.Assert(mv.rv.reservedSpace.Load() >= mv.reservedSpace.Load(),
+				mv.rv.reservedSpace.Load(), mv.reservedSpace.Load())
+			mv.rv.reservedSpace.Add(-mv.reservedSpace.Load())
```
souravgupta-msft: Can you use the incReservedSpace() method to be consistent.
```
mv.rv.incReservedSpace(-mv.reservedSpace.Load())
```

or, use
```
mv.rv.decReservedSpace(mv.reservedSpace.Load())
```

internal/dcache/rpc/server/handler.go
```diff
@@ -2006,6 +2176,14 @@ func (h *ChunkServiceHandler) EndSync(ctx context.Context, req *models.EndSyncRe
 	// Update the state of target RV in this MV replica from syncing to online.
 	mvInfo.updateComponentRVState(req.TargetRVName, dcache.StateSyncing, dcache.StateOnline)
 
+	// As sync has completed, clear reservedSpace and commit it in totalChunkBytes.
+	mvInfo.totalChunkBytes.Add(mvInfo.reservedSpace.Load())
+	common.Assert(rvInfo.reservedSpace.Load() >= mvInfo.reservedSpace.Load(),
+		rvInfo.reservedSpace.Load(), mvInfo.reservedSpace.Load(), rvInfo.rvName,
+		req.MV, rpc.EndSyncRequestToString(req))
+	rvInfo.reservedSpace.Add(-mvInfo.reservedSpace.Load())
```
souravgupta-msft:
```
rvInfo.incReservedSpace(-mvInfo.reservedSpace.Load())

or

rvInfo.decReservedSpace(mvInfo.reservedSpace.Load())
```

internal/dcache/rpc/server/handler.go
```diff
@@ -402,30 +491,38 @@ func (mv *mvInfo) isSourceOrTargetOfSync() (isSource bool, isTarget bool) {
...
+			/* MV replica is source for one syncJob */
+			isSource = true
+			isTarget = false
+			return false
 		}
-	}
+	})
 
 	// Unreachable code.
 	common.Assert(false)
```
souravgupta-msft: This will always result in panic as we are not returning after iterating the syncJobs map.

internal/dcache/rpc/server/handler.go
souravgupta-msft: Ok, it won't fail as the rvName corresponds the local/hosting RV

internal/dcache/rpc/server/handler.go
```diff
@@ -670,9 +767,75 @@ func (mv *mvInfo) refreshFromClustermap() error {
...
+			// only one syncJob, clear that.
+			//
+			common.Assert(mv.syncJobsCount.Load() == 1, rvName, mv.mvName, mv.syncJobsCount.Load(),
+				mv.getSyncJobs(), rpc.ComponentRVsToString(mv.componentRVs))
+
+			log.Warn("mvInfo::refreshFromClustermap: %s/%s, clearing old syncJob left from previous incomplete sync attempt, syncJobs: %+v",
+				rvName, mv.mvName, mv.getSyncJobs())
+
+			mv.deleteAllSyncJobs()
+		}
```
souravgupta-msft, line 828: We should also decrement the rvInfo.reservedSpace by mvInfo.reservedSpace and set mvInfo.reservedSpace to 0.

internal/dcache/rpc/server/handler.go
linuxsmiths: yes, thanks

internal/dcache/rpc/server/handler.go
linuxsmiths: since the space is reserved in JoinMV when we move an RV to StateOutOfSync, and we are still in StateOutOfSync, we should not decrease the space.

souravgupta-msft (approved):

PR #1817, opened 2025-06-04
Inject error

ashruti-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

## Checklist

## Related Links
- [Issues](<link>)


internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -321,6 +339,11 @@ func (m *BlobMetadataManager) createFileInit(filePath string, fileMetadata []byt
 		EtagMatchConditions:    "",
 	})
 
+	injectErr := common.InjectError(common.PROB_MODERATE)
+	if injectErr != nil {
+		err = injectErr
+	}
```
linuxsmiths, line 348: this will cause the assert at L364 to fire.
We need to make sure that when we inject error, there are no asserts in that entire code path

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -366,6 +389,12 @@ func (m *BlobMetadataManager) createFileFinalize(filePath string, fileMetadata [
 	if common.IsDebugBuild() {
 		prop, err := m.storageCallback.GetPropertiesFromStorage(
 			internal.GetAttrOptions{Name: path})
+
+		injectErr := common.InjectError(common.PROB_MODERATE)
+		if injectErr != nil {
+			err = injectErr
+		}
+
```
linuxsmiths, line 404: don't inject error here as it's only for debug builds.
also you have an assert fail right after this.

**As I said, it's very important you make sure each of the error injections you add are verified.**

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -395,6 +424,11 @@ func (m *BlobMetadataManager) createFileFinalize(filePath string, fileMetadata [
 		EtagMatchConditions:    "",
 	})
 
+	injectErr := common.InjectError(common.PROB_LOW)
+	if injectErr != nil {
+		err = injectErr
+	}
```
linuxsmiths, line 441: assert at L434 will fail

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -250,6 +256,12 @@ func (m *BlobMetadataManager) getBlobSafe(blobPath string) ([]byte, *internal.Ob
 		data, err := m.storageCallback.GetBlobFromStorage(internal.ReadFileWithNameOptions{
 			Path: blobPath,
 		})
+
+		injectErr = common.InjectError(common.PROB_LOW)
+		if injectErr != nil {
+			err = injectErr
+		}
```
linuxsmiths, line 264: this will cause assert at L267 to fail.
We need to test each of the error injections that we add.
Also, instead of removing the asserts in the path, we should change them to fail for errors not caused by error injection but caused by real errors.

PR #1820, opened 2025-06-05
Make joinmv idempotent, and other fixes

linuxsmiths:
## Type of Change

- [ x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

## Checklist

## Related Links
- [Issues](<link>)

### DIFF
--- a/internal/dcache/rpc/server/handler.go
+++ b/internal/dcache/rpc/server/handler.go
@@ -146,7 +146,7 @@ type mvInfo struct {
 	// updated for each request but once the sync completes, in the EndSync handler, we add
 	// mvInfo.reservedSpace to totalChunkBytes thus accounting all the chunks that were copied
 	// to this MV as a result of the sync operations. This means that totalChunkBytes will be
-	// 0 for outofsync MV replicas while it can non-zero for online and even syncing replicas.
+	// 0 for outofsync MV replicas while it can be non-zero for online and even syncing replicas.
 	// syncing MV replicas will have non-zero totalChunkBytes only if there are client writes
 	// during the sync. This will happen if an MV replica went offline during a file write and
 	// a new one had to be picked.
--- a/internal/dcache/rpc/server/handler.go
+++ b/internal/dcache/rpc/server/handler.go
@@ -1803,9 +1803,27 @@ func (h *ChunkServiceHandler) JoinMV(ctx context.Context, req *models.JoinMVRequ
 	// Thrift should not be calling us with nil req.
 	common.Assert(req != nil)
 
+	//
+	// See if it's a new-mv request (and not a fix-mv request)
+	//
+	newMV := false
+	found := false
+	for _, rv := range req.ComponentRV {
+		if rv.Name == req.RVName {
+			// Must be either online for new-mv and outofsync for fix-mv.
+			common.Assert(rv.State == string(dcache.StateOnline) ||
+				rv.State == string(dcache.StateOutOfSync), rv.Name, req.MV, rv.State)
+
+			newMV = (rv.State == string(dcache.StateOnline))
+			found = true
+		}
+	}
+	common.Assert(found, req.RVName, req.MV, rpc.JoinMVRequestToString(req))
+
 	// TODO:: discuss: changing type of component RV from string to RVNameAndState
 	// requires to call componentRVsToString method as it is of type []*models.RVNameAndState
-	log.Debug("ChunkServiceHandler::JoinMV: Received JoinMV request: %v", rpc.JoinMVRequestToString(req))
+	log.Debug("ChunkServiceHandler::JoinMV: Received JoinMV request (newMV: %v): %v",
+		newMV, rpc.JoinMVRequestToString(req))
 
 	if !common.IsValidUUID(req.SenderNodeID) ||
 		!cm.IsValidMVName(req.MV) ||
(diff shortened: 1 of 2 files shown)


jainakanksha-msft (approved):

internal/dcache/rpc/server/handler.go
```diff
@@ -756,6 +804,20 @@ func (mv *mvInfo) refreshFromClustermap() error {
...
+		// following achieves that w/o a clustermap update being forced after degrade-mv.
+		// Note that this is one of those "safe deductions" that we can do while taking the risk of
+		// deviating away from the actual clustermap, but note that clustermap will soon be updated
+		// to reflect it.
+		//
+		if cm.GetRVState(rvName) == dcache.StateOffline && rvState != dcache.StateOffline {
+			log.Warn("mvInfo::refreshFromClustermap: %s/%s state is %s while RV state is offline, marking component RV state as offline",
+				rvName, mv.mvName, rvState)
+			rvState = dcache.StateOffline
+		}
```
souravgupta-msft, line 819: When can we get this scenario?
We get rvState from newRVs list which we get from `cm.GetRVs(mv.mvName)`. So, `cm.GetRVState(rvName)` should be same as the rvState.

internal/dcache/rpc/server/handler.go
```diff
@@ -789,8 +851,9 @@ func (mv *mvInfo) refreshFromClustermap() error {
 		// space changes.
 		//
 		if oldRvNameAndState.State == string(dcache.StateOutOfSync) && rvState == dcache.StateOffline {
-			common.Assert(mv.rv.reservedSpace.Load() >= mv.reservedSpace.Load(),
-				mv.rv.reservedSpace.Load(), mv.reservedSpace.Load())
```
souravgupta-msft, line 793: Why did we remove this assert? We are decrementing the rv's reserved space by the mv's reserved space and we don't want to have reserved space as negative.

internal/dcache/rpc/server/handler.go
```diff
@@ -1740,9 +1803,27 @@ func (h *ChunkServiceHandler) JoinMV(ctx context.Context, req *models.JoinMVRequ
...
+			// Must be either online for new-mv and outofsync for fix-mv.
+			common.Assert(rv.State == string(dcache.StateOnline) ||
+				rv.State == string(dcache.StateOutOfSync), rv.Name, req.MV, rv.State)
+
+			newMV = (rv.State == string(dcache.StateOnline))
+			found = true
+		}
+	}
+	common.Assert(found, req.RVName, req.MV, rpc.JoinMVRequestToString(req))
+
```
souravgupta-msft, line 1822: We should add this code after logging the JoinMV request in L1825.

internal/dcache/rpc/server/handler.go
```diff
@@ -1773,18 +1854,40 @@ func (h *ChunkServiceHandler) JoinMV(ctx context.Context, req *models.JoinMVRequ
...
+		// for an MV and fixMV() sends JoinMV request to each of the outofsync RVs. If one or
+		// more of these fail, the joinMV() will treat it as a failure and not update clustermap.
+		// Next time when fixMV() is called it'll again attempt fixing and again send JoinMV.
+		// Note that for proper handling we need to ensure that the reservedSpace remains
+		// same across both calls. Also if an RV is joined but never used later (maybe joinMV()
+		// picked a new RV in the next iteration), we should time out and undo the reservedSpace.
+		// This one is a TODO.
+		//
+		errStr := fmt.Sprintf("Double join for %s/%s, prev join at: %s, by: %s",
+			req.RVName, req.MV, mvInfo.joinedWhen, mvInfo.joinedBy)
```
souravgupta-msft, line 1868: We can have asserts for `joinedWhen` and `joinedBy` when it is a multiple join scenario.
Also, should we mention "multiple join" instead of "double join" in the log?

internal/dcache/rpc/server/handler.go
```diff
@@ -1773,18 +1854,40 @@ func (h *ChunkServiceHandler) JoinMV(ctx context.Context, req *models.JoinMVRequ
...
-		return &models.JoinMVResponse{}, nil
+		// Refresh our mvInfo state as per the latest clustermap.
+		// This will undo the changes made by the prev incomplete JoinMV, updating mvInfo as if the
+		// previous JoinMV never happened. If refreshFromClustermap() fails, we cannot safely proceed.
+		// For newMV, we won't have the MV in clustermap yet, so no need to refresh.
+		//
+		if !newMV {
+			err := mvInfo.refreshFromClustermap()
+			if err != nil {
+				errStr = fmt.Sprintf("%s, refreshFromClustermap() failed, aborting JoinMV", errStr)
```
souravgupta-msft:
```suggestion
				errStr = fmt.Sprintf("%s, refreshFromClustermap() failed, aborting JoinMV: %v", errStr, err)
```

internal/dcache/rpc/server/handler.go
```diff
@@ -1773,18 +1854,40 @@ func (h *ChunkServiceHandler) JoinMV(ctx context.Context, req *models.JoinMVRequ
...
+				log.Err("ChunkServiceHandler::JoinMV: %s", errStr)
+				common.Assert(false, errStr)
+				return nil, rpc.NewResponseError(models.ErrorCode_InvalidRV, errStr)
```
souravgupta-msft: We should send the error code as InternalServerError instead of InvalidRV here

internal/dcache/rpc/server/handler.go
```diff
@@ -1773,18 +1854,40 @@ func (h *ChunkServiceHandler) JoinMV(ctx context.Context, req *models.JoinMVRequ
...
 		//
-		log.Warn("ChunkServiceHandler::JoinMV: %s is already part of %s, ignoring", req.RVName, req.MV)
-		return &models.JoinMVResponse{}, nil
+		// Refresh our mvInfo state as per the latest clustermap.
+		// This will undo the changes made by the prev incomplete JoinMV, updating mvInfo as if the
+		// previous JoinMV never happened. If refreshFromClustermap() fails, we cannot safely proceed.
+		// For newMV, we won't have the MV in clustermap yet, so no need to refresh.
+		//
+		if !newMV {
+			err := mvInfo.refreshFromClustermap()
```
souravgupta-msft, line 1879: After this block, we are deleting the mvInfo from mvMap. So, what is the use of refreshing the clustermap here?
```
// Remove the MV replica, we will add a fresh one later down.
rvInfo.deleteFromMVMap(req.MV)
mvInfo = nil
```

internal/dcache/rpc/server/handler.go
linuxsmiths: cm.GetRVs(mv.mvName) gives the component RVs so rvState is the component RV state, while cm.GetRVState(rvName) is the state of that RV.

internal/dcache/rpc/server/handler.go
linuxsmiths: the same assert is inside mv.rv.decReservedSpace()

internal/dcache/rpc/server/handler.go
linuxsmiths: rvInfo.getMVInfo() has those asserts
"double join" is better as it emphasizes a join for something already joined.. multiple is not the key

internal/dcache/rpc/server/handler.go
linuxsmiths: we want to update rv.reservedSpace and any other "undo" that it may do.. it's best to behave as if we went back to the clustermap state, before applying the new update

PR #1822, opened 2025-06-05
Batch the Updates for component RV

jainakanksha-msft:


internal/dcache/clustermap/clustermap.go
```diff
@@ -248,42 +248,22 @@ func ReportRVOffline(rvName string) error {
...
-	// TODO: This still results in as many GetBlob/PutBlob calls as the number of threads calling this
-	//       function. We should try to batch the updates that come in close time proximity.
-	//       We can have a channel where we add the required updates and one go routine can then update the
-	//       global clustermap once every few seconds.
-	//
-	updateLock.Lock()
-	defer updateLock.Unlock()
-
-	return componentRVStateUpdater(mvName, rvName, rvNewState)
+	updateComponentRVStateChannel <- dcache.ComponentRVUpdateMessage{MvName: mvName, RvName: rvName, RvNewState: rvNewState}
```
linuxsmiths: UpdateComponentRVState() cannot return till the requested change is saved in the clustermap.
The batching change should not change the semantics.
Multiple threads calling UpdateComponentRVState() with disjoint changes will have their changes queued to a list and then that  list will be periodically collated and published into clustermap, but only after a given change is published can the corresponding UpdateComponentRVState() can return.
Also, the status of the update must be returned by UpdateComponentRVState(), as it was doing before.

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -2724,13 +2750,63 @@ func refreshMyRVs(myRVs []dcache.RawVolume) {
...
+	log.Info("ClusterManager::batchUpdateComponentRVState: Received batch of %d component RV updates", len(batchUpdates))
+
+	// Merge  multiple updates for the same MV/RV by priority ---
+	// priority: OutOfSync(1) < Syncing(2) < Online(3) < Offline(4)
+	priority := map[dcache.StateEnum]int{
+		dcache.StateOutOfSync: 1,
+		dcache.StateSyncing:   2,
+		dcache.StateOnline:    3,
+		dcache.StateOffline:   4,
+	}
```
linuxsmiths: we don't need to make it complex. 
The idea of batching is to batch multiple unrelated RV updates, something that we see when lots of sync jobs each of which is targetting a different target RV.
We can simplify it by not allowing multiple updates to the same rv/mv in the same batch. We won't get it and we shouldn't handle it.

linuxsmiths (approved):
Audited/Updated/Tested

PR #1825, opened 2025-06-09
Garbage collection of chunks for  dcache

syeleti-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

### DIFF
--- a/internal/dcache/replication_manager/replication_manager.go
+++ b/internal/dcache/replication_manager/replication_manager.go
@@ -290,7 +290,7 @@ retry:
 func WriteMV(req *WriteMvRequest) (*WriteMvResponse, error) {
 	common.Assert(req != nil)
 
-	log.Debug("ReplicationManager::WriteMV: Received WriteMV request: %v", req.toString())
+	log.Debug("ReplicationManager::WriteMV: Received WriteMV request (%v): %v", PutChunkStyle, req.toString())
 
 	if err := req.isValid(); err != nil {
 		err = fmt.Errorf("invalid WriteMV request %s [%v]", req.toString(), err)
@@ -302,14 +302,24 @@ func WriteMV(req *WriteMvRequest) (*WriteMvResponse, error) {
 	var rvsWritten []string
 	retryCnt := 0
 
+	//
+	// Flag to check if we have a BrokenChain error in the PutChunkDC response.
+	// If we have a BrokenChain error for an RV, it means that the PutChunkDC request was not
+	// forwarded to it as the nexthop RV was down/offline. We will get ThriftError for the nexthop RV
+	// and BrokenChain error for the subsequent RVs.
+	// In case of BrokenChain error, we retry the WriteMV() operation skipping the nexthop RV and refreshing
+	// the clustermap (if needed).
+	//
+	brokenChain := false
+
 	// TODO: TODO: hash validation will be done later
 	// get hash of the data in the request
 	// hash := getMD5Sum(req.Data)
 
 retry:
-	if retryCnt > 0 {
-		log.Info("ReplicationManager::WriteMV: [%d] Retrying WriteMV %v after clustermap refresh, RVs written in prev attempt: %v",
-			retryCnt, req.toString(), rvsWritten)
+	if retryCnt > 0 || brokenChain {
+		log.Info("ReplicationManager::WriteMV: [%d] Retrying WriteMV %v after clustermap refresh or broken chain (%v), RVs written in prev attempt: %v",
+			retryCnt, req.toString(), brokenChain, rvsWritten)
 	}
 
 	//
@@ -338,7 +348,46 @@ retry:
 	//
 	rvsWritten = nil
 
-	for componentRVIdx, rv := range componentRVs {
+	//
+	// Allocate the PutChunkDCRequest for orchestrating the required PutChunk calls as per the PutChunkStyle
+	// selected. If PutChunkStyle is OriginatorSendsToAll then we will send the PutChunk request to each
+	// component RV in putChunkDCReq.Request and putChunkDCReq.NextRVs, while for PutChunkStyle DaisyChain, we
+	// will just send the PutChunkDC request to the first RV (in putChunkDCReq.Request).
+	//
+	// We set the "MaybeOverwrite" flag to true in PutChunkRequest to let the server know that this
+	// could potentially be an overwrite of a chunk that we previously wrote, so that it relaxes its
+	// overwrite checks. To be safe we set MaybeOverwrite to true when retryCnt > 0 or we are retrying
+	// because of brokenChain error for one of the RVs.
+	//
+	putChunkDCReq := &models.PutChunkDCRequest{
+		Request: &models.PutChunkRequest{
+			Chunk: &models.Chunk{
+				Address: &models.Address{
+					FileID:      req.FileID,
+					RvID:        "", // will be set later down
+					MvName:      req.MvName,
+					OffsetInMiB: req.ChunkIndex * req.ChunkSizeInMiB,
+				},
+				Data: req.Data,
+				Hash: "", // TODO: hash validation will be done later
+			},
+			Length:         int64(len(req.Data)),
+			SyncID:         "", // this is regular client write
+			ComponentRV:    componentRVs,
+			MaybeOverwrite: retryCnt > 0 || brokenChain,
+		},
+		NextRVs: make([]string, 0), // will be added later down, if needed
+	}
+
+	//
+	// Go over all the component RVs and populate the putChunkDCReq.
+	// If any of the component RVs is local, then putChunkDCReq.Request refers to that and
+	// putChunkDCReq.NextRVs contains all the other RVs. If none of the component RVs is local, then
+	// putChunkDCReq.Request refers to the first component RV in the componentRVs list and
+	// putChunkDCReq.NextRVs contains all the other RVs.
+	// We don't write to offline and outofsync component RVs, so they won't be added to putChunkDCReq.
+	//
+	for _, rv := range componentRVs {
 		//
 		// Omit writing to RVs in “offline” or “outofsync” state. It’s ok to omit them as the chunks not
 		// written to them will be copied to them when the mv is (soon) resynced.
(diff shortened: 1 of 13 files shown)


internal/dcache/gc/gc.go
```diff
@@ -0,0 +1,259 @@
...
+				chunkIdx := int64(i)
+
+				// Resume the progress of deletion of the chunks if it's a retry.
+				if resumeChunkIdx, ok := deleteProgressForRvs[rv.Name]; ok {
+					common.Assert(resumeChunkIdx >= chunkIdx && retryCnt > 0,
+						retryCnt, chunkIdx, resumeChunkIdx, file.Filename, rv, rvs)
+					chunkIdx = max(chunkIdx, resumeChunkIdx)
+				}
+
+				// TODO: remove all the chunks corresponding to an RV in one RPC call.
```
linuxsmiths: this cannot be a TODO. We need to implement this first.
If you omit OffsetInMiB in RemoveChunkRequest, it should imply "remove all chunks for the file".
Also RemoveChunkResponse should be modified to return number of chunks deleted. This will be max 1 if OffsetInMiB is specified, but if OffsetInMiB is not specified it can be more than 1.
Also, if RemoveChunkResponse completes with success and number of chunks deleted is 0, that should indicate to the caller that all chunks have been deleted from the mv.

internal/dcache/gc/gc.go
```diff
@@ -0,0 +1,259 @@
...
+		//
+
+		for _, rv := range rvs {
+			if rv.State == string(dcache.StateOffline) || rv.State == string(dcache.StateOutOfSync) {
+				log.Info("GC::removeAllChunksForFile: skip deleting the chunks from rv: %s, rv state: %s, file: %s",
+					rv, rv.State, file.Filename)
+				continue
+			}
+
+			if rv.State == string(dcache.StateOnline) || rv.State == string(dcache.StateSyncing) {
```
linuxsmiths: GC should be simple and rugged.
If it fails to do something, it should simply omit and try again in the next iteration (when GC is called again after some time). Only when the chunks are fully removed it should remove the file from the list of to-be-gced files

internal/dcache/gc/gc.go
```diff
@@ -0,0 +1,259 @@
...
+							OffsetInMiB: chunkIdx * file.FileLayout.ChunkSize / common.MbToBytes,
+						},
+						ComponentRV: rvs,
+					}
+
+					ctx, cancel := context.WithTimeout(context.Background(), RPCClientTimeout*time.Second)
+					defer cancel()
+
+					rpcResp, err := rpc_client.RemoveChunk(ctx, targetNodeId, rpcReq)
+					if err != nil {
```
linuxsmiths: same comment, no need to complex error handling.

internal/dcache/gc/gc.go
```diff
@@ -0,0 +1,259 @@
...
+		return
+	}
+
+}
+
+func AsyncFileChunkGarbageCollector(file *dcache.FileMetadata) {
+	if file.Size != 0 {
+		gc.deletedFileQueue <- file
+	}
+}
```
linuxsmiths: while this will be the common way to delete chunks for a file, we should also support the case where the node that deletes a file dies, so it cannot gc the file chunks using this mechanism.
We should periodically check for .deleting files and try to GC them. We could have used that as the only mechanism, but it might have overhead of enumerating potentially large number of files, so let's support both for now.

internal/dcache/gc/gc.go
```diff
@@ -0,0 +1,259 @@
...
+	numGcWorkers     int
+	deletedFileQueue chan *dcache.FileMetadata
+	wg               sync.WaitGroup
+}
+
+var gc *GcInfo
+
+func NewGC() {
+	gc = &GcInfo{
+		numGcWorkers:     100, // TODO: This number should be decided.
```
linuxsmiths: these threads don't put too much load, so we can have them higher in number for scaling better.
let's start with 100, later we can increase upto 1000

internal/dcache/gc/gc.go
syeleti-msft: done

internal/dcache/gc/gc.go
syeleti-msft: I agree, will add this in the next PR

internal/dcache/gc/gc.go
syeleti-msft: Yes, I agree

component/distributed_cache/distributed_cache.go
```diff
@@ -187,6 +188,8 @@ func (dc *DistributedCache) Start(ctx context.Context) error {
 		return log.LogAndReturnError(fmt.Sprintf("DistributedCache::Start error [Failed to start fileio manager : %v]", err))
 	}
 
+	gc.NewGC()
```
linuxsmiths: call it Start(), NewGC looks like as if it returns something.

component/distributed_cache/distributed_cache.go
```diff
@@ -279,6 +282,7 @@ func (dc *DistributedCache) Stop() error {
 
 	dc.pw.destroyParallelWriter()
 	fm.EndFileIOManager()
+	gc.EndGC()
```
linuxsmiths: call this End().

internal/dcache/gc/gc.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+var gc *GcInfo
+
+type gcFile struct {
+	file *dcache.FileMetadata
+	//
+	// When deletion failed for one or more MVs, it will be retired from this here.
+	//
+	retryMvList []string
```
linuxsmiths: this doesn't seem to be used

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -614,6 +614,142 @@ retry:
...
+	// The following set would be used while retring the skipping the repeated deletion from the RV's which have already
+	// got success in removing their chunks, while retrying for its clustermap update. RvName->struct{}.
+	//
+	deleteProgressForRvs := make(map[string]struct{})
+	retryCnt := 0
+
+retry:
+
+	log.Debug("ReplicationManager::RemoveMV: RemoveMV request: %v, retryCnt: %d, FileID: %s, MV: %s",
+		req.FileID, req.MvName, retryCnt)
```
linuxsmiths: missing params, and also mis-ordered.
need to add a toString() method to print the request

internal/dcache/rpc/models.thrift
```diff
@@ -81,7 +81,12 @@ struct RemoveChunkResponse {
     // When a RemoveChunkResponse carries a status of success and numChunksDeleted==0, it would indicate
     // to the caller that all chunks of the file are deleted from the specified rv/mv directory.
     //
-    4: i64 numChunksDeleted
+    4: i64 numChunksDeleted,
+    //
+    // The following flag is set when not all chunks to the corresponding fileID are not deleted from the RV. The
+    // caller needs to retry the request in this case.
+    //
+    5: bool needRetry
```
linuxsmiths: no need for an additional flag for this, see comment at L81/82, numChunksDeleted along with the status should be able to convey that

internal/dcache/gc/utils.go
```diff
@@ -0,0 +1,111 @@
...
+func shiftOnlineRVsToStart(componentRVs []*models.RVNameAndState) {
+	i := 0
+	for j := 0; j < len(componentRVs); j++ {
+		if componentRVs[j].State == string(dcache.StateOnline) {
+			componentRVs[i], componentRVs[j] = componentRVs[j], componentRVs[i]
+			i++
+		}
+	}
+}
+
```
linuxsmiths: not used

internal/dcache/gc/utils.go
```diff
@@ -0,0 +1,111 @@
...
+		common.Assert(cm.IsValidRVName(rvName), rvName)
+		common.Assert(cm.IsValidComponentRVState(rvState), rvName, rvState)
+
+		componentRVs = append(componentRVs,
+			&models.RVNameAndState{Name: rvName, State: string(rvState)})
+	}
+
+	common.Assert(len(componentRVs) == int(getNumReplicas()),
+		mvName, len(componentRVs), getNumReplicas(), rpc.ComponentRVsToString(componentRVs))
+
```
linuxsmiths: can we avoid dup'ing them and other functions

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -614,6 +614,142 @@ retry:
...
+					if errCM != nil {
+						log.Err("ReplicationManager::RemoveMV: Failed to refresh the cluster map, RV: %s, file ID: %s: %v",
+							rv.Name, req.FileID, errCM)
+						common.Assert(false, req.FileID, errCM)
+						return nil, errCM
+					}
+
+					retryCnt++
+					goto retry
+				}
```
linuxsmiths: let's not add the refresh clustermap logic here.. we should refresh clustermap at the start unconditionally and then treat all errors as "remove not completed, need to retry"

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -614,6 +614,142 @@ retry:
...
+	mvState, rvs, lastClusterMapEpoch := getComponentRVsForMV(req.MvName)
+
+	shiftOnlineRVsToStart(rvs)
+
+	//
+	// Initially delete all the chunks from the all the online RV's first then move on to the syncing rv's.
+	// If we delete the chunk from all the online rvs then this chunk may get listed in the sync job when syncing
+	// the online rv to other outofsync rvs. So It is necessary to also make a removeChunk rpc request to all the
+	// other rvs which are having the state syncing.
+	//
```
linuxsmiths: no need to do this.
Simply do this..
do not delete chunks for component rvs not in online state, so syncing RVs we will wait for them to come back online..
till we have any RV that's not offline and not successfully removed, file removal cannot be claimed done.

internal/dcache/gc/utils.go
syeleti-msft: Now these functions became obsolete as the logic has moved to replication manager for removing the chunks from the mv. will remove them all from here.

internal/dcache/rpc/models.thrift
syeleti-msft: Added a boolean to avoid the confusion of using the same variable to represent 2 different things

internal/dcache/gc/gc.go
syeleti-msft: addressed

internal/dcache/replication_manager/replication_manager.go
syeleti-msft: Okay

internal/dcache/rpc/models.thrift
linuxsmiths: having multiple variables to mean the same causes confusion, now we have to unnecessarily assert that when needRetry is set numChunksDeleted!=0. We can as well use that directly, and it's not confusing to have that semantics.. success with numChunksDeleted==0 to mean that nothing to delete.. this is like when read() returns 0 it means eof.

Let me check what you have

syeleti-msft: reverted the code back

internal/dcache/gc/gc.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func (gc *GcInfo) removeAllChunksForFile(gcFile *gcFile) {
+	if gcFile.retryCnt > 20 {
+		//
+		// The periodic GC thread will soon schedule this gcFile again for the deletion while listing the deleted files.
+		//
+		log.Warn("GC::removeAllChunksForFile: file: %s, retryCnt: %d, skipping the deletion of the chunks as max retries used",
+			gcFile.file.Filename, gcFile.retryCnt)
+	}
```
linuxsmiths: did u want to return in this case?

I think we should not return.. anyways we are scheduling the delete by adding the file back to the end of the list, so it cannot cause starvation of other files needing to be deleted

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -614,6 +614,115 @@ retry:
...
+				Address: &models.Address{
+					FileID:      req.FileID,
+					RvID:        rvId,
+					MvName:      req.MvName,
+					OffsetInMiB: -1,
+				},
+				ComponentRV: rvs,
+			}
+
+			ctx, cancel := context.WithTimeout(context.Background(), RPCClientTimeout*time.Second)
```
linuxsmiths: RemoveChunk RPC can take longer than other RPCs, so we should use a higher timeout.
Also, we must test to make sure we work fine even if this times out.

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -614,6 +614,115 @@ retry:
...
+			if err == nil {
+				//
+				// numChunksdeleted would be zero if the call dont need any retries. GC must retry for this MV again if
+				// it is non zero.
+				//
+				if rpcResp.NumChunksDeleted != 0 {
+					err = fmt.Errorf("Delete partially success for fileID: %s, RV: %s, node %s, NumChunksDeleted: %d",
+						req.FileID, rv.Name, targetNodeId, rpcResp.NumChunksDeleted)
+					log.Err("ReplicationManager::RemoveMV: %v", err)
+					common.Assert(false, err)
```
linuxsmiths: we should not assert this, as this is expected once for every file..

internal/dcache/rpc/server/handler.go
```diff
@@ -1997,6 +2000,13 @@ func (h *ChunkServiceHandler) RemoveChunk(ctx context.Context, req *models.Remov
 		availableSpace = 0
 	}
 
+	//
+	// If there is no retry needed, Set the numChunksDeleted to zero to tell success to the caller.
+	//
+	if !needRetry {
+		numChunksDeleted = 0
+	}
```
linuxsmiths: oh I didn't mean this.. I meant RemoveChunk should truthfully return the chunks deleted.
When it's called and it doesn't find any chunks, then the caller can infer as success

PR #1830, opened 2025-06-10
Added few test cases

jainakanksha-msft:


test/distributed_cache/test-cluster.sh
```diff
@@ -200,8 +200,13 @@ stop_blobfuse_on_node()
...
 
-    echo "Stopping blobfuse @ $(date)" >> $logfile
-    ssh $vm ~/stop-blobfuse.sh >> $logfile 2>&1
+    (
+        echo "Stopping blobfuse @ $(date)" >> $logfile
+        ssh $vm ~/stop-blobfuse.sh >> $logfile 2>&1
+    )&
+
+    # Give some time for blobfuse process to stop.
+    sleep 1
```
linuxsmiths: I had done it like this because unlike start-blobfuse.sh, sto-blobfuse.sh waits till the blobfuse is stopped, so once it returns we know that blobfuse has stopped and we don't need to wait.

test/distributed_cache/test-cluster.sh
```diff
@@ -360,6 +360,15 @@ rv_count=$(get_rv_count "$cm")
...
 log_status $? "is $rv_count"
 
+becho -n "Cluster must be readonly"
+readonly_status=$(echo "$cm" | jq '."readonly"')
+if [ "$rv_count" -ge "$MIN_NODES" ]; then
+    [ "$readonly_status" == "false" ]
+else
+    [ "$readonly_status" == "true" ]
+fi
+log_status $? "rv_count is $rv_count, min_nodes is $MIN_NODES, readonly is $readonly_status"
```
linuxsmiths: min-nodes actually has a relation to num-replicas. We should not allow min-nodes less than num-replicas.
We should update IsValidDcacheConfig() to reflect that. Once we do that then we can safely check only against min-nodes, else it doesn't look right for a num-replicas=3 cluster to become writeable after 1 node is up (say min-node is 1).

linuxsmiths: the check above is not completely correct
Note that we should compare with number of nodes and not number of RVs.
Also, the status check string should be correct. It should say "Cluster must be readonly" or "Cluster must not be readonly" depending on the node count vs min-nodes comparision.
also make the message like

"Cluster must be readonly [num nodes ($num_nodes) < min-nodes ($MIN_NODES)]"

similar for the -ve case

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -898,7 +898,7 @@ func (cmi *ClusterManager) ensureInitialClusterMap(dCacheConfig *dcache.DCacheCo
 	clusterMap = dcache.ClusterMap{
 		Readonly:      true,
 		State:         dcache.StateReady,
-		Epoch:         1,
+		Epoch:         0,
```
linuxsmiths: why do you change this?

internal/dcache/models.go
```diff
@@ -74,8 +74,8 @@ type ClusterMap struct {
 	LastUpdatedAt int64                     `json:"last_updated_at"`
 	LastUpdatedBy string                    `json:"last_updated_by"`
 	Config        DCacheConfig              `json:"config"`
-	RVMap         map[string]RawVolume      `json:"rv-map"`
-	MVMap         map[string]MirroredVolume `json:"mv-map"`
+	RVMap         map[string]RawVolume      `json:"rv-list"`
+	MVMap         map[string]MirroredVolume `json:"mv-list"`
```
linuxsmiths, line 100: and this?
it's a map and not a list, so let it be called map

test/distributed_cache/test-cluster.sh
```diff
@@ -6,19 +6,53 @@
...
 #    checking clustermap and performing filesystem operations from various
 #    cluster nodes.
 #
 
+# Check if number of nodes is provided
+if [ $# -ne 1 ]; then
+    echo "Usage: $0 <number_of_nodes>"
+    echo "Example: $0 5"
+    exit 1
+fi
```
linuxsmiths: why do we need to take number of nodes from user?
why can't we deduce it from clustermap.
I'd not want to take any additional info from the user. He should be able to point the test script at a cluster and it should be able to test it by gathering the reqd infro from the clustermap.

test/distributed_cache/test-cluster.sh
```diff
@@ -165,11 +199,11 @@ get_node_id()
     fi
 }
 
-wait_till_next_epoch()
+wait_till_next_scheduled_epoch()
 {
     local next_epoch=$(expr $LAST_UPDATED_AT + $CLUSTERMAP_EPOCH)
     local now=$(date +%s)
-    local secs_to_next_epoch=$(expr $next_epoch - $now + 2)
+    local secs_to_next_epoch=$(expr $next_epoch - $now + 5)
```
linuxsmiths: why did you have to change this?
did it fail and if so did you analyze why 2 sec buffer was not sufficient.

test/distributed_cache/test-cluster.sh
```diff
@@ -181,6 +215,42 @@ wait_till_next_epoch()
...
+    local now=$(date +%s)
+    local secs_to_next_epoch=$(expr $next_epoch - $now)
+   # Check if we're close to the heartbeat timeout boundary
+    local heartbeat_timeout=$(expr $HB_SECONDS \* $HB_TILL_NODE_DOWN - 2)
+    local secs_to_next_epoch_with_buffer
+
+    if [ $secs_to_next_epoch -lt $heartbeat_timeout ]; then
+    # Add an additional epoch period if we're close to timeout
+    secs_to_next_epoch_with_buffer=$(expr $secs_to_next_epoch + $CLUSTERMAP_EPOCH + 5)
+    echo "Close to heartbeat timeout boundary, adding extra epoch wait time"
```
linuxsmiths, line 307: general comment.
pl;s indent uniformly across the file, use 4 spaces

test/distributed_cache/test-cluster.sh
```diff
@@ -254,6 +325,46 @@ read_clustermap_from_node()
     ssh $vm "cat $clustermap_path" 2>>$logfile | tee -a $logfile
 }
 
+write_data_in_dcache()
+{
+    local vm=$1
+    local file_name=$2
+    local block_size=$3
+    local count=$4
+    ssh $vm "dd if=/dev/urandom of=$MOUNTDIR/fs=dcache/$file_name bs=$block_size count=$count"
```
linuxsmiths: add conv=fsync to be safe
to all dds

test/distributed_cache/test-cluster.sh
```diff
@@ -254,6 +325,46 @@ read_clustermap_from_node()
     ssh $vm "cat $clustermap_path" 2>>$logfile | tee -a $logfile
 }
 
+write_data_in_dcache()
+{
+    local vm=$1
+    local file_name=$2
+    local block_size=$3
+    local count=$4
```
linuxsmiths, line 415: just take size_in_GB as param and deduce these two internally

test/distributed_cache/test-cluster.sh
```diff
@@ -254,6 +325,46 @@ read_clustermap_from_node()
...
+write_data_in_azure()
+{
+    local vm=$1
+    local file_name=$2
+    local block_size=$3
+    local count=$4
+    ssh $vm "dd if=/dev/urandom of=$MOUNTDIR/fs=azure/$file_name bs=$block_size count=$count"
+}
+
+write_data_in_mountdir()
```
linuxsmiths: call it write_data_on_both()
, since that's what is the intent..
for others too use on in place of in

internal/dcache/cluster_manager/cluster_manager.go
jainakanksha-msft: This is needed to start with version 1.
Right now, we don't use version 1

internal/dcache/models.go
jainakanksha-msft: I made it consistent throughout.
fs=debug consider it as rv-list mv-list, which makes the things inconsistent. While reading data in automation script.

PR #1831, opened 2025-06-11
Add "removeAllChunks" option to the removeChunk RPC Request

syeleti-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Checklist

## Related Links
- [Issues](<link>)

### DIFF
--- a/internal/dcache/rpc/models.thrift
+++ b/internal/dcache/rpc/models.thrift
@@ -64,11 +64,11 @@ struct PutChunkResponse {
     3: list<RVNameAndState> componentRV
 }
 
+// Remove chunks belonging to a file.
 struct RemoveChunkRequest {
     1: string senderNodeID,
     2: Address address,
-    3: list<RVNameAndState> componentRV, // used to validate the component RV for the MV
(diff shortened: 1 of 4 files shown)


internal/dcache/rpc/models.thrift
```diff
@@ -67,14 +67,16 @@ struct PutChunkResponse {
 struct RemoveChunkRequest {
     1: string senderNodeID,
     2: Address address,
-    3: list<RVNameAndState> componentRV // used to validate the component RV for the MV
+    3: list<RVNameAndState> componentRV, // used to validate the component RV for the MV
+    4: bool removeAllChunks
```
linuxsmiths: I don't see any usecase for removing a specific chunk.
To avoid confusion, we should stop supporting that, which means we should not add this argument and update the comments to suggest that it removes all chunks for a given file.
The caller should not identify a specific chunk and must set OffsetInMiB to -1.

For now leave it, but later we can rename this to RemoveChunksForFile

PR #1832, opened 2025-06-11
PutChunk daisy chain implementation

souravgupta-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description

Added PutChunkEx RPC to write chunks to the component RVs in daisy chain manner. When PutChunkEx call is received, it writes the chunk to its local RV and parallelly forwards the call to the next RVs. The response returned by this RPC is the individual PutChunk response for each RV mapped by the RV name.

### DIFF
--- a/internal/dcache/replication_manager/models.go
+++ b/internal/dcache/replication_manager/models.go
@@ -215,6 +215,80 @@ func (req *WriteMvRequest) isValid() error {
 type WriteMvResponse struct {
 }
 
+type WriteMvExRequest struct {
+	FileID string // unique guid of the file, as stored in metadata blob
+	MvName string // name of the MV where this chunk will be written, e.g., "mv0", "mv1", etc.
+
+	// An MV can hold multiple chunks of the file (one from each stripe).
+	// ChunkIndex identifies the exact chunk to be read. It is the index of the chunk
+	// within the file and is 0 based.
+	// e.g., for stripe size of 16MiB and ChunkSizeInMiB == 4,
+	// file offset of 1MiB would translate to the following:
+	//  MvName: "mv0"   (assuming first MV in the file's mv list is "mv0")
+	//  ChunkIndex: 0
+	//  OffsetInChunk: 1048576 (1MiB)
+	// while a file offset of 17MiB would translate to the following:
+	//  MvName: "mv0"
+	//  ChunkIndex: 4
+	//  OffsetInChunk: 1048576 (1MiB)
+	//
+	// The chunks are stored in RV as,
+	//  <MvName>/<FileID>.<ChunkIndex * ChunkSizeInMiB>.data and
+	//  <MvName>/<FileID>.<ChunkIndex * ChunkSizeInMiB>.hash
+	//
+	ChunkIndex     int64
+	Data           []byte // Data to be written
+	ChunkSizeInMiB int64  // Chunk size in MiB
+	IsLastChunk    bool   // boolean flag to indicate if this is the last chunk
+}
+
+// helper method which can be used for logging the request contents except the data buffer
+// Use this instead of %+v to avoid printing the data buffer
+func (req *WriteMvExRequest) toString() string {
+	return fmt.Sprintf("{FileID: %s, MvName: %s, ChunkIndex: %d, ChunkSizeInMiB: %d, IsLastChunk: %v, Data buffer size: %d}",
+		req.FileID, req.MvName, req.ChunkIndex, req.ChunkSizeInMiB, req.IsLastChunk, len(req.Data))
+}
+
+// check if the request is valid
+func (req *WriteMvExRequest) isValid() error {
+	common.Assert(common.IsValidUUID(req.FileID))
+	common.Assert(cm.IsValidMVName(req.MvName))
+
+	if (req.ChunkSizeInMiB*common.MbToBytes) < cm.MinChunkSize ||
+		(req.ChunkSizeInMiB*common.MbToBytes) > cm.MaxChunkSize {
+		reqStr := req.toString()
+		err := fmt.Errorf("ChunkSizeInMiB is invalid in request: %s", reqStr)
+		log.Err("WriteMvExRequest::isValid: %v", err)
+		return err
+	}
+
+	if req.ChunkIndex < 0 || req.ChunkIndex > ChunkIndexUpperBound {
+		reqStr := req.toString()
+		err := fmt.Errorf("ChunkIndex is invalid in request: %s", reqStr)
+		log.Err("WriteMvExRequest::isValid: %v", err)
+		return err
+	}
+
+	if len(req.Data) == 0 || len(req.Data) > int(req.ChunkSizeInMiB*common.MbToBytes) {
+		reqStr := req.toString()
+		err := fmt.Errorf("data buffer is invalid in request: %s", reqStr)
+		log.Err("WriteMvExRequest::isValid: %v", err)
+		return err
+	}
+
+	if !req.IsLastChunk && len(req.Data) != int(req.ChunkSizeInMiB*common.MbToBytes) {
+		reqStr := req.toString()
+		err := fmt.Errorf("data buffer length is not equal to chunk size in request: %s", reqStr)
+		log.Err("WriteMvExRequest::isValid: %v", err)
+		return err
+	}
+
+	return nil
+}
+
+type WriteMvExResponse struct {
+}
+
 // syncJob tracks resync of one MV replica, srcRVName/mvName -> destRVName/mvName.
 type syncJob struct {
 	mvName       string                   // name of the MV to be synced
--- a/internal/dcache/replication_manager/replication_manager.go
+++ b/internal/dcache/replication_manager/replication_manager.go
@@ -614,6 +614,368 @@ retry:
 	return &WriteMvResponse{}, nil
 }
 
+func WriteMVEx(req *WriteMvExRequest) (*WriteMvExResponse, error) {
+	common.Assert(req != nil)
+
+	log.Debug("ReplicationManager::WriteMVEx: Received WriteMVEx request: %v", req.toString())
+
+	if err := req.isValid(); err != nil {
+		err = fmt.Errorf("invalid WriteMVEx request %s [%v]", req.toString(), err)
+		log.Err("ReplicationManager::WriteMVEx: %v", err)
+		common.Assert(false, err)
+		return nil, err
+	}
+
+	var rvsWritten []string
+	retryCnt := 0
+
+	// TODO: TODO: hash validation will be done later
+	// get hash of the data in the request
+	// hash := getMD5Sum(req.Data)
+
+retry:
+	if retryCnt > 0 {
+		log.Info("ReplicationManager::WriteMVEx: [%d] Retrying WriteMVEx %v after clustermap refresh, RVs written in prev attempt: %v",
+			retryCnt, req.toString(), rvsWritten)
+	}
+
+	// Get component RVs for MV, from clustermap and also the corresponding clustermap epoch.
+	// If server returns NeedToRefreshClusterMap, we will ask cm.RefreshClusterMap() to update
+	// the clustermap to a value higher than this epoch.
+	mvState, componentRVs, lastClusterMapEpoch := getComponentRVsForMV(req.MvName)
+
+	log.Debug("ReplicationManager::WriteMVEx: Component RVs for %s (%s) are: %v",
+		req.MvName, mvState, rpc.ComponentRVsToString(componentRVs))
+
+	//
+	// List of RVs to which the chunk was written successfully in this WriteMVEx attempt, used for logging.
+	// Note that everytime we refresh clustermap we need to write all the replicas according to the
+	// latest component RVs. An MV write should be considered a transaction that is applied to the cluster
+	// in a given state. Once a transaction is applied successfully, then we are guaranteed that any change
+	// to the MV composition will ensure that any chunk written will be correctly synchronized.
+	// Note that rvInfo/mvInfo and the NeedToRefreshClusterMap error returned by the target RV(s), helps
+	// check if the transaction can be safely applied.
+	//
+	rvsWritten = nil
+
+	var rpcResp *models.PutChunkExResponse
+
+	//
+	// When we send the PutChunkEx RPC request to the target node, it sends the PutChunk call to its
+	// local RV and parallelly forwards the call to the next RVs in the list. This is the ideal case where
+	// error returned by the PutChunkEx RPC is nil. The errors of the individual PutChunk calls to the
+	// component RVs are returned in the PutChunkExResponse where the response or error of each RV is
+	// mapped to the RV name.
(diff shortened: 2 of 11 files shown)


internal/dcache/replication_manager/models.go
```diff
@@ -215,6 +215,80 @@ func (req *WriteMvRequest) isValid() error {
 type WriteMvResponse struct {
 }
 
+type WriteMvExRequest struct {
```
linuxsmiths: There's no need to define a new request type. WriteMVEx() does the same thing (writing to an MV) but does it differently, so it should use the same request type WriteMvRequest and returns the same response type WriteMvResponse.

souravgupta-msft: Done, updated to use WriteMvRequest and WriteMvResponse.

internal/dcache/rpc/server/handler.go
```diff
@@ -1878,6 +1879,241 @@ refreshFromClustermapAndRetry:
...
+			currRV.Name, req.Chunk.Address.MvName, currRV.State)
+
+		rpcResp = h.forwardPutChunk(ctx, req, nextRVs)
+		common.Assert(rpcResp != nil)
+		common.Assert(rpcResp.Responses != nil) // rpcResp.Responses should not be nil
+
+		rpcResp.Responses[currRV.Name] = nil // nil PutChunkResponseOrError for skipped RVs
+
+	} else if currRV.State == string(dcache.StateOnline) || currRV.State == string(dcache.StateSyncing) {
+
```
linuxsmiths: These checks must be done only at the originator, once the originator decides who all to fwd to, from that point on, all that the intermediate nodes do is - write to their local RV and blindly forward to the next RV (with properly updated nextRVs list). On the response side, it splices the status of its local PutChunk() and the slice response it got from the next RV it fwded to and relays back the response to its sender. So eventually the originator will get a sliced response, for each RV.

internal/dcache/rpc/models.thrift
```diff
@@ -64,6 +64,21 @@ struct PutChunkResponse {
     3: list<RVNameAndState> componentRV
 }
 
+struct PutChunkExRequest {
+    1: PutChunkRequest request,
+    2: list<RVNameAndState> nextRVs
```
linuxsmiths: Since we don't want the forwarding nodes to have any intelligence wrt handling RV states, this should be just a list of RV names.

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -614,6 +614,376 @@ retry:
 	return &WriteMvResponse{}, nil
 }
 
+func WriteMVEx(req *WriteMvRequest) (*WriteMvResponse, error) {
```
linuxsmiths: I wanted to see how different will this be from WriteMV().
Now that we have the complete code, I feel that we should have a single WriteMV() function to work for both cases.
I'll submit a commit to highlight what I'm thinking.
That code may not be complete, pls complete it.

internal/dcache/rpc/client/functions.go
```diff
@@ -359,6 +359,96 @@ func PutChunk(ctx context.Context, targetNodeID string, req *models.PutChunkRequ
 		targetNodeID, reqStr)
 }
 
+func PutChunkEx(ctx context.Context, targetNodeID string, req *models.PutChunkExRequest) (*models.PutChunkExResponse, error) {
```
linuxsmiths: Now that we don't have WriteMVEx() we can call this with the more appropriate name PutChunkDC() (DC for daisy chain).

internal/dcache/rpc/server/handler.go
```diff
@@ -1878,6 +1879,241 @@ refreshFromClustermapAndRetry:
...
+			//
+			// If the PutChunkEx call fails for the current RV, it means that the call to the
+			// next RVs were not forwarded.
+			// Note, the errors of the PutChunk calls to the individual RVs are stored in the
+			// PutChunkExResponse.
+			// If the error returned by the PutChunkEx call is nil, it means that the PutChunk call
+			// was made to the local RV and also forwarded to the next RVs successfully.
+			// For non-nil error, we will have to forward the call to the next RVs here.
+			//
```
linuxsmiths: this should have none of these complications. It should be a dumb forwarder, forwarding request to the next hop (along with nextRVs) and relaying the response array (one for each RV) back.

internal/dcache/rpc/server/handler.go
```diff
@@ -1878,6 +1879,241 @@ refreshFromClustermapAndRetry:
...
+// If the PutChunkExRequest is for an RV that is not hosted by this node, it will return an InvalidRVID error.
+func (h *ChunkServiceHandler) PutChunkEx(ctx context.Context, req *models.PutChunkExRequest) (*models.PutChunkExResponse, error) {
+	// Thrift should not be calling us with nil req.
+	common.Assert(req != nil)
+
+	// Thrift should not be calling us with nil Request, Chunk or Address.
+	common.Assert(req.Request != nil)
+	common.Assert(req.Request.Chunk != nil)
+	common.Assert(req.Request.Chunk.Address != nil)
+
```
linuxsmiths, line 1942: also assert that req.NextRVs is not empty. Caller should call us PutChunkEx only if it needs to be fwded to at least one, else it should simply call PutChunk

internal/dcache/rpc/server/handler.go
```diff
@@ -1878,6 +1879,241 @@ refreshFromClustermapAndRetry:
...
+	var wg sync.WaitGroup
+	wg.Add(1)
+
+	//
+	// Parallelly forward the PutChunkEx request to the next RVs in the list.
+	//
+	go func() {
+		defer wg.Done()
+
+		rpcResp = h.forwardPutChunk(ctx, req.Request, req.NextRVs)
```
linuxsmiths, line 1981: we should not need forwardPutChunk(), instead PutChunkEx should be implemented recursively as

PutChunk for the Request entry
PutChunkEx with nextRVs[0] copied to the Request and nextRVs[1:] as the new nextRV. If there's only one element left in nextRV, then just make a PutChunk call for that.

Response should be gathered and sent for all the RVs (one in Request and those in nextRVs)

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -399,27 +530,46 @@ retry:
...
+			common.Assert(putChunkExResp == nil)
 		} else {
-			common.Assert(false, "Unexpected RV state", rv.State, rv.Name, req.MvName)
+			log.Debug("ReplicationManager::WriteMVEx: Received PutChunkEx response for %s/%s from node %s: %s",
+				rv.Name, req.MvName, targetNodeID, rpc.PutChunkExResponseToString(putChunkExResp))
+			common.Assert(putChunkExResp != nil)
 		}
 	}
 
+processResponses:
```
linuxsmiths, line 577: you will need to handle ThriftError also, treating it as transport error from the nexthop RV

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -375,18 +434,90 @@ retry:
...
+		//
+		// Write to the next hop component RV.
+		//
+
+		// TODO: Add rvName to Address to avoid potentially expensive search for RV name.
+		rvName := getRvNameFromRvID(putChunkExReq.Request.Address.RvID)
+
+		targetNodeID := getNodeIDFromRVName(rvName)
+		common.Assert(common.IsValidUUID(targetNodeID))
+		common.Assert(targetNodeID == common.GetNodeUUID(), targetNodeID, common.GetNodeUUID())
```
souravgupta-msft: This cannot be true if there is no local component RV and only one RV is online/syncing.
I will remove this in my changes, added here fyi.

internal/dcache/rpc/server/handler.go

internal/dcache/rpc/models.thrift
souravgupta-msft: Updated to have just RV name

internal/dcache/replication_manager/replication_manager.go

internal/dcache/rpc/client/functions.go
souravgupta-msft: Done, rename to PutChunkDC

internal/dcache/rpc/server/handler.go

internal/dcache/rpc/server/handler.go
souravgupta-msft: Added

internal/dcache/rpc/server/handler.go
souravgupta-msft: Updated

internal/dcache/replication_manager/replication_manager.go
souravgupta-msft: Done

internal/dcache/rpc/utils.go
```diff
@@ -143,6 +143,34 @@ func PutChunkResponseToString(resp *models.PutChunkResponse) string {
...
+// used for logging
+func PutChunkDCResponseToString(response *models.PutChunkDCResponse) string {
+	str := strings.Builder{}
+	str.WriteString("[\n")
+
+	for rvName, resp := range response.Responses {
+		common.Assert(resp != nil)
+		if resp == nil {
+			// PutChunkResponseOrError can be nil if the state of the RV is outofsync or offline.
+			str.WriteString(fmt.Sprintf("{%s : <nil>}\n", rvName))
```
linuxsmiths: is it still possible?

internal/dcache/rpc/utils.go
```diff
@@ -143,6 +143,34 @@ func PutChunkResponseToString(resp *models.PutChunkResponse) string {
...
+		} else {
+			str.WriteString(fmt.Sprintf("{%s : {PutChunkResponse %s, Error %s}}\n",
+				rvName, PutChunkResponseToString(resp.Response), resp.Error.String()))
```
linuxsmiths: one of resp.Response or resp.Error will be nil, so this will crash

internal/dcache/replication_manager/threadpool.go
```diff
@@ -77,10 +77,6 @@ type workitem struct {
 }
 
 type responseItem struct {
-	// Node ID of the target node that processed the request.
-	// Used for logging purpose.
-	targetNodeID string
-
```
linuxsmiths, line 83: why did u remove this?

internal/dcache/replication_manager/utils.go
```diff
@@ -204,6 +243,37 @@ func updateLocalComponentRVState(rvs []*models.RVNameAndState, rvName string,
...
+		common.Assert(resp != nil)
+		common.Assert(len(responseChannel) < int(getNumReplicas()),
+			len(responseChannel), getNumReplicas())
+
+		var err error
+
+		if resp.Error != nil {
+			common.Assert(resp.Response == nil)
+			err = resp.Error
+		}
```
linuxsmiths: this is needed as an interface only compares to till when both the dynamic type and value of that concrete type is nil.
I think we are better off having PutChunkResponseOrError use Error of type error and not of ResponseError.
After all it carries the outcome of a PutChunk call potentially made by one of the intermediate nodes.
It's better to return the exact error, so that we can do the same checks as we do in the case of OriginatorSendsToAll, o/w we need to be aware of the special case - when a node directly performs PutChunk the way to detect thrift/transport error is different and when it's done by an intermediate node and conveyed through PutChunkDCResponse, it's different. This confusion is not good.
With this we can get rid of the new ThriftError that you had to define for this. The BrokednLink error will still be helpful as it's conveying a new condition which does not even exist in OriginatorSendsToAll case.
Let me know if you want to discuss this.

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -464,12 +621,13 @@ retry:
...
+		log.Err("ReplicationManager::WriteMV: PutChunk to %s/%s failed [%v]",
+			respItem.rvName, req.MvName, respItem.err)
 
-		log.Err("ReplicationManager::WriteMV: PutChunk to %s/%s, node %s failed [%v]",
-			respItem.rvName, req.MvName, respItem.targetNodeID, respItem.err)
+		common.Assert(respItem.putChunkResp == nil)
 
 		rpcErr := rpc.GetRPCResponseError(respItem.err)
-		if rpcErr == nil {
+		if rpcErr == nil || rpcErr.GetCode() == models.ErrorCode_ThriftError {
```
linuxsmiths, line 633: once we make PutChunkResponseOrError.Error as type error, then we don't need this special check.

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -517,13 +675,20 @@ retry:
...
+				respItem.rvName, req.MvName)
 			continue
 		}
 
+		//
 		// The error is RPC error of type *rpc.ResponseError.
-		if rpcErr.GetCode() == models.ErrorCode_NeedToRefreshClusterMap {
+		//
+		if rpcErr.GetCode() == models.ErrorCode_BrokenChain {
+			brokenChain = true
```
linuxsmiths, line 699: see if we can easily assert that this error won't happen for nexthop

linuxsmiths (approved):
I'm approving it, pls validate thoroughly before merging.

internal/dcache/replication_manager/threadpool.go
souravgupta-msft: This was just used for logging. In PutChunkDC response, we don't send the target node id for the RV to which the PutChunk call was made.

internal/dcache/replication_manager/replication_manager.go
souravgupta-msft: This can happen in the next try if there are nodes going down in incremental order. For example, in the first try rv1 went down, so we marked that as offline and skipped the PutChunk for next RVs. In the next retry, suppose rv2 went offline. So, we can get broken chain error for RVs after rv2 and this can go on. This is why I am not incrementing the retry count in case of broken chain error.

internal/dcache/replication_manager/utils.go
souravgupta-msft: PutChunkResponseOrError is defined in the Thrift IDL.
```
struct PutChunkResponseOrError {
    1: optional PutChunkResponse response,
    2: optional ResponseError error
}
```
Thrift is language independent. So, we cannot define GO's in-built error type in the Thrift's IDL.
This is why we have defined a custom exception `ResponseError`, which implements the GO's error interface, and all the RPC throws this exception. So, in the client side we can directly return this as error type.

internal/dcache/replication_manager/replication_manager.go
souravgupta-msft: Same as above comment.

internal/dcache/replication_manager/utils.go
linuxsmiths: oh, this is unfortunate that we cannot return an error type of our choice.. anyways, if it's difficult to make thrift use that, we can live with this.

internal/dcache/replication_manager/replication_manager.go
linuxsmiths: What I mean is - we should never get this error from our nexthop to which we sent the PutChunkDC request

PR #1838, opened 2025-06-18
RPC calls for RV local to the node

souravgupta-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description
When making calls to RV which is local to the node, directly call the RPC method using the local server's handler. If the RV to which the call is made is not local, then call it using the Thrift RPC client.


internal/dcache/replication_manager/replication_manager.go
```diff
@@ -230,7 +230,15 @@ retry:
 		ctx, cancel := context.WithTimeout(context.Background(), RPCClientTimeout*time.Second)
 		defer cancel()
 
-		rpcResp, err = rpc_client.GetChunk(ctx, targetNodeID, rpcReq)
+		//
+		// If the selected RV is local, then we directly call the GetChunk() method using the
+		// local server's handler. Else we call the GetChunk() RPC via the Thrift RPC client.
+		//
+		if isLocalRV {
+			rpcResp, err = rpc_server.GetChunk(ctx, targetNodeID, rpcReq)
```
linuxsmiths: let's call this GetChunkLocal to highlight the fact that this only allows to get a local chunk. Same for other requests.
Also, passing targetNodeId to GetChunkLocal sounds confusing, so remove that.
If you want to assert that we are calling GetChunkLocal() only for local requests, you can assert at the call site (if it makes sense).

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -428,8 +443,12 @@ retry:
 			if targetNodeID == rpc.GetMyNodeUUID() {
 				// Only one component RV can be local.
 				common.Assert(putChunkDCReq.Request.Chunk.Address.RvID == "",
-					putChunkDCReq.Request.Chunk.Address.String())
+					putChunkDCReq.Request.Chunk.Address.String(), rv.Name, rvID)
+				common.Assert(!isLocalRV,
```
linuxsmiths: the prev assert is (slightly) better as it's asserting RvID is not set before writing to it.
Infact let's not set the variable isLocalRV here, instead right at the callsite we should do the check for that (targetNodeID== "my node id"). That makes it easier to read and less error prone.
Extra uuid comparison is fine.
We need not return isLocalRV from getReaderRV() also.
if it was not for the import loop issue, I'd have loved rpc_client.GetChunk() to make the local bs non-local decision, so the callers can be completely unaware and it works uniformly for all requests,

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -491,6 +512,7 @@ retry:
 		rm.tp.schedule(&workitem{
 			targetNodeID: targetNodeID,
 			rvName:       rvName,
+			isLocalRV:    isLocalRV,
```
linuxsmiths: also remove from here and let the the check be made right at the call site

internal/dcache/replication_manager/replication_manager.go

internal/dcache/replication_manager/replication_manager.go

internal/dcache/replication_manager/replication_manager.go
souravgupta-msft: I have made the changes of removing the `isLocalRV` variable and checking at the caller's end using,
`targetNodeID == rpc.GetMyNodeUUID()`

I do see one problem here which is if two or more Blobfuse processes are running in the same VM, each having their own RVs. This check will then fail because we would need IPC via Thrift in this case to read a chunk from some other Blobfuse process in the same node.

linuxsmiths: you are saying two different things here?
Two different blobfuse processes trying to read each others' chunks?
That's not a valid scenario. We can have two different blbofuse processes but only when that node is taking part in two different caches, in which case it'll not try to access RVs from other process.
If they have to access their chunk only, the check works fine.

souravgupta-msft: Ok, makes sense. We should then restrict users from not using the same cache id for two different Blobfuse processes in the same VM, if this check is not yet present.

PR #1845, opened 2025-06-26
Direct IO read write

souravgupta-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description
Added option for direct IO read and write.

### DIFF
--- a/common/types.go
+++ b/common/types.go
@@ -322,18 +322,6 @@ func GetIdLength(id string) int64 {
 	return int64(len(existingBlockId))
 }
 
-// Align up the size of the file to the file system block size
-// This is required when doing direct IO operations.
-func AlignToBlockSize(size int64) int64 {
-	Assert(size >= 0, size)
-
-	if size%FS_BLOCK_SIZE == 0 {
-		return size
-	}
-
-	return ((size / FS_BLOCK_SIZE) + 1) * FS_BLOCK_SIZE
-}
-
 func init() {
 	val, present := os.LookupEnv("HOME")
 	if !present {
--- a/common/types.go
+++ b/common/types.go
@@ -68,8 +68,13 @@ const (
 	BfuseStats    = "blobfuse_stats"
 	BlockIDLength = 16
 
-	// File system block size
-	FS_BLOCK_SIZE = 4096
+	//
+	// File system block size.
+	// IO size and offset must be aligned to this, for direct IOs.
+	// TODO: Though 512 bytes should work for most common filesystems, but let's see if we need to
+	//       query it using statx(STATX_DIOALIGN).
+	//
+	FS_BLOCK_SIZE = 512
 
 	FuseAllowedFlags = "invalid FUSE options. Allowed FUSE configurations are: `-o attr_timeout=TIMEOUT`, `-o negative_timeout=TIMEOUT`, `-o entry_timeout=TIMEOUT` `-o allow_other`, `-o allow_root`, `-o umask=PERMISSIONS -o default_permissions`, `-o ro`"
 
--- a/common/types.go
+++ b/common/types.go
@@ -68,6 +68,9 @@ const (
 	BfuseStats    = "blobfuse_stats"
 	BlockIDLength = 16
 
+	// File system block size
+	FS_BLOCK_SIZE = 4096
+
 	FuseAllowedFlags = "invalid FUSE options. Allowed FUSE configurations are: `-o attr_timeout=TIMEOUT`, `-o negative_timeout=TIMEOUT`, `-o entry_timeout=TIMEOUT` `-o allow_other`, `-o allow_root`, `-o umask=PERMISSIONS -o default_permissions`, `-o ro`"
 
 	UserAgentHeader = "User-Agent"
--- a/common/types.go
+++ b/common/types.go
@@ -322,6 +322,18 @@ func GetIdLength(id string) int64 {
 	return int64(len(existingBlockId))
 }
 
+// Align up the size of the file to the file system block size
+// This is required when doing direct IO operations.
+func AlignToBlockSize(size int64) int64 {
+	Assert(size >= 0, size)
+
+	if size%FS_BLOCK_SIZE == 0 {
+		return size
+	}
+
+	return ((size / FS_BLOCK_SIZE) + 1) * FS_BLOCK_SIZE
+}
+
 func init() {
 	val, present := os.LookupEnv("HOME")
 	if !present {
--- a/common/util.go
+++ b/common/util.go
@@ -64,6 +64,7 @@ var RootMount bool
 var ForegroundMount bool
 var IsDistributedCacheEnabled bool
 var IsStream bool
+var MyNodeUUID string
 
 // IsDirectoryMounted is a utility function that returns true if the directory is already mounted using fuse
 func IsDirectoryMounted(path string) bool {
@@ -652,6 +653,10 @@ func UpdatePipeline(pipeline []string, component string) []string {
 }
 
 func GetNodeUUID() (string, error) {
+	if MyNodeUUID != "" {
+		return MyNodeUUID, nil
+	}
+
 	uuidFilePath := filepath.Join(DefaultWorkDir, "blobfuse_node_uuid")
 
 	// Read the UUID file.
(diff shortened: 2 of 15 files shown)


internal/dcache/file_manager/worker.go
```diff
@@ -157,11 +158,24 @@ func (wp *workerPool) writeChunk(task *task) {
...
+	if rpc.IOType == rpc.BufferedIO {
+		dataBufferSize = task.chunk.Len
+	} else if rpc.IOType == rpc.DirectIO {
+		//
+		// For DirectIO, we must align the data buffer size to the filesystem block size.
+		//
+		dataBufferSize = common.AlignToBlockSize(task.chunk.Len)
+	} else {
+		common.Assert(false, "Unexpected IOType", rpc.IOType)
+	}
```
linuxsmiths: we don't gain much by forcing direct IO by force aligning smaller chunk IOs, instead let's fallback to buffered write for unaligned IOs, which will be very few. For both reads and writes.

souravgupta-msft: I have made the change in write. In read do we ensure that we are sending the GetChunk RPC for the whole chunk, meaning the req.OffsetInChunk is aligned with the filesystem block size and the req.Length is equal to the size of the chunk stored in the cache.

souravgupta-msft: Made the changes in read also. Please review.

common/types.go
```diff
@@ -68,6 +68,9 @@ const (
 	BfuseStats    = "blobfuse_stats"
 	BlockIDLength = 16
 
+	// File system block size
+	FS_BLOCK_SIZE = 4096
```
linuxsmiths: where did you get this from?
Usually I've seen that the alignment requirement is more relaxed and most filesystems work with 512 byte alignment. In new kernels we can query the alignment using statx(), but let's use 512 for now as most filesystems should support 512 bytes.

common/util.go
```diff
@@ -652,6 +653,10 @@ func UpdatePipeline(pipeline []string, component string) []string {
 }
 
 func GetNodeUUID() (string, error) {
+	if MyNodeUUID != "" {
+		return MyNodeUUID, nil
+	}
```
linuxsmiths, line 658: string assignment/access in Go is not atomic, so this can race with some other thread writing MyNodeUUID and we might get some bad value here. To be safe we should call GetNodeUUID() once to warm the cache.

component/distributed_cache/distributed_cache.go
```diff
@@ -108,6 +109,8 @@ type DistributedCacheOptions struct {
 	SafeDeletes         bool   `config:"safe-deletes" yaml:"safe-deletes,omitempty"`
 	CacheAccess         string `config:"cache-access" yaml:"cache-access,omitempty"`
 	ClustermapEpoch     uint64 `config:"clustermap-epoch" yaml:"clustermap-epoch,omitempty"`
+	readIOMode          string `config:"read-io-mode" yaml:"read-io-mode,omitempty"`
+	writeIOMode         string `config:"write-io-mode" yaml:"write-io-mode,omitempty"`
```
linuxsmiths: start with caps, like others?

component/distributed_cache/distributed_cache.go
```diff
@@ -393,6 +396,33 @@ func (distributedCache *DistributedCache) Configure(_ bool) error {
...
+	if !config.IsSet(compName + ".read-io-mode") {
+		distributedCache.cfg.readIOMode = rpc.DirectIO
+	}
+	if !config.IsSet(compName + ".write-io-mode") {
+		distributedCache.cfg.writeIOMode = rpc.DirectIO
+	}
+
+	err = rpc.SetReadIOMode(distributedCache.cfg.readIOMode)
+	if err != nil {
+		log.Err("DistributedCache::Configure : [%v]", err)
```
linuxsmiths: we should treat this like other config failures, i.e., fail the startup

internal/dcache/rpc/client/client_pool.go
```diff
@@ -521,7 +521,8 @@ func (ncPool *nodeClientPool) createRPCClients(numClients uint32) error {
 
 	// Create RPC clients and add them to the channel.
 	for i := 0; i < int(numClients); i++ {
-		client, err := newRPCClient(ncPool.nodeID, rpc.GetNodeAddressFromID(ncPool.nodeID))
+		var client *rpcClient
+		client, err = newRPCClient(ncPool.nodeID, rpc.GetNodeAddressFromID(ncPool.nodeID))
```
linuxsmiths, line 525: Ah, did you hit it or noticed while audit?

internal/dcache/rpc/server/handler.go
```diff
@@ -1564,53 +1565,96 @@ func (h *ChunkServiceHandler) GetChunk(ctx context.Context, req *models.GetChunk
...
+	//
+	// Read the chunk using buffered IO mode if,
+	//   - The IO type is BufferedIO, or
+	//   - The size of chunk is not multiple of file system block size. This can happen only
+	//     for the last chunk of the file.
+	//   - The requested offset and length is not aligned to file system block size.
+	//
+	if rpc.ReadIOMode == rpc.BufferedIO ||
+		chunkSize%common.FS_BLOCK_SIZE != 0 ||
+		(req.OffsetInChunk+req.Length)%common.FS_BLOCK_SIZE != 0 {
```
linuxsmiths: these two checks are not correct.
direct IO has the following requirements:
1. IO Offset and Size (and not the file size) must be block (512 byte) aligned.
2. User buffer must be block aligned.

It's best to first try direct and if it fails with EINVAL, fallback to buffered IO.

internal/dcache/rpc/server/handler.go
```diff
@@ -1564,53 +1659,46 @@ func (h *ChunkServiceHandler) GetChunk(ctx context.Context, req *models.GetChunk
 	// Allocate byte slice, data from the chunk file will be read into this.
 	//
 	data := make([]byte, req.Length)
```
linuxsmiths, line 1682: this doesn't guarantee alignment of allocated buffer

linuxsmiths (approved):
Approving after my audit changes.

internal/dcache/rpc/server/handler.go
linuxsmiths: also this is something which we need to convert to sync.Pool

souravgupta-msft: Yes, I had thought about this and wanted to discuss with Srinivas if we can reuse the sync pool logic he has in file manager.

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Noticed while seeing some logs

internal/dcache/rpc/server/handler.go
```diff
@@ -1468,6 +1471,117 @@ func (h *ChunkServiceHandler) Hello(ctx context.Context, req *models.HelloReques
...
+	common.Assert(readOffset >= 0)
+
+	readLength := len(*data)
+
+	//
+	// Caller must pass data buffer aligned on FS_BLOCK_SIZE, else we have to unnecessarily perform buffered read.
+	//
+	dataAddr := unsafe.Pointer(&(*data)[0])
+	isDataBufferAligned := ((uintptr(dataAddr) % common.FS_BLOCK_SIZE) == 0)
+	common.Assert(isDataBufferAligned, uintptr(dataAddr), common.FS_BLOCK_SIZE)
```
souravgupta-msft, line 1494: We can remove this assert here and at L1759, because we are already falling back to buffered IO when the data buffer in not aligned.

internal/dcache/rpc/server/handler.go
```diff
@@ -1767,7 +1767,7 @@ func writeChunkAndHash(chunkPath, hashPath *string, data *[]byte, hash *string)
 	//
 	// Write the chunk using buffered IO mode if,
 	//   - Write IO type is configured as BufferedIO, or
-	//   - The requested offset and length is not aligned to file system block size.
+	//   - The write length (or chunk size) is not aligned to file system block size.
```
linuxsmiths, line 1770: this is not correct, it's indeed offset and length that's checked for alignment.

PR #1638, opened 2025-02-19
[IGNORE] DUMMY PR (testing somethings)

syeleti-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Code quality improvement
- [ ] Other (describe):

## Description

- **Feature / Bug Fix**: (Brief description of the feature or issue being addressed)

## Checklist

- [ ] The purpose of this PR is explained in this or a referenced issue.
- [ ] Tests are included and/or updated for code changes.
- [ ] Documentation update required.
- [ ] Updates to module CHANGELOG.md are included.
- [ ] License headers are included in each file.

## Related Links
- [Issues](<link>)

PR #1847, opened 2025-06-30
Ashruti/throttle fix

ashruti-msft:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

## Checklist

## Related Links
- [Issues](<link>)

### DIFF
--- a/component/block_cache/blockpool.go
+++ b/component/block_cache/blockpool.go
@@ -61,10 +61,11 @@ type BlockPool struct {
 	// Number of block that this pool can handle at max
 	maxBlocks uint32
 
-	// Add these:
+	// Total number of blocks allocated to the pool
 	totalBlocks uint32
-	capBlocks   uint32
-	mu          sync.Mutex
+
+	// Mutex to protect the totalBlocks count
+	mu sync.Mutex
 }
 
 // NewBlockPool allocates a new pool of blocks
@@ -81,9 +82,8 @@ func NewBlockPool(blockSize uint64, memSize uint64) *BlockPool {
 	pool := &BlockPool{
 		blocksCh:     make(chan *Block, poolBlockCount-1),
 		resetBlockCh: make(chan *Block, poolBlockCount-1),
-		maxBlocks:    uint32(poolBlockCount),
+		maxBlocks:    uint32(blockCount),
 		blockSize:    blockSize,
-		capBlocks:    uint32(blockCount),
 		totalBlocks:  uint32(poolBlockCount),
 	}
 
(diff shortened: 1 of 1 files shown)


component/block_cache/blockpool.go
```diff
@@ -109,63 +116,61 @@ func (pool *BlockPool) Terminate() {
...
 		}
-		_ = b.Delete()
+		_ = block.Delete()
 	}
 }
 
 // Usage provides % usage of this block pool
 func (pool *BlockPool) Usage() uint32 {
-	return ((pool.maxBlocks - (uint32)(len(pool.blocksCh)+len(pool.resetBlockCh))) * 100) / pool.maxBlocks
+	return ((pool.maxBlocks - (uint32)(len(pool.blocksCh)+len(pool.priorityCh))) * 100) / pool.maxBlocks
```
bhansalivikas: We shall count the resetBlockCh as well here, as there might be blocks in that channel which are waiting to be reset

component/block_cache/blockpool.go
```diff
@@ -109,63 +116,61 @@ func (pool *BlockPool) Terminate() {
...
+	var block *Block = nil
 
 	select {
-	case b = <-pool.blocksCh:
+	case block = <-pool.priorityCh:
+		break
+	case block = <-pool.blocksCh:
 		break
-
-	default:
```
syeleti-msft, line 138: Removing the default block would make the blobfuse process stuck when the blocks are all used by the other file handles. and all of them are not willing to close the file. This would make the callback stuck and wait indefinitely, ideally this is not good for the user. Also afterwards if we hold enough number of callbacks the entire blobfuse goes unresponsive that is not good. Lets keep this for now to not get other issues.

syeleti-msft (commented):
removing the issue as it is not related to this PR

component/block_cache/blockpool.go
anu-avocado: Currently we are facing a memory leak issue due to the allocation of blocks on the mustGet call leading to blobfuse crash. To fix this issue and also honour our config we have to make mustget wait in case we run out of blocks. We have created a different channel for priority blocks to help make the wait time less but at the end of the day if we run out of memory user should be responsible to increase the limit in config instead of us spilling over like we used to.

syeleti-msft: > user should be responsible to increase the limit in config instead of us spilling over like we used to.

This is the problem right if we block the read/write callbacks for indefinite amount of time, how the user would know?
The removal of the default block would make the entire blobfuse responsive in some scenarios.

syeleti-msft: All the fuse callbacks must complete in some finite amount of time, else return the return with appropriate error to tell the user application what has happened.

blobfuse2-nightly.yaml
syeleti-msft, line 1: This was taken care in this PR https://github.com/Azure/azure-storage-fuse/pull/1638, pls revert this file

anu-avocado: this caused some failure in pipeline so we are intentitonally deleting this part here

PR #1848, opened 2025-07-01
Periodic GC to reclaim the chunks of stale files.

syeleti-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description

When node responsible for deletion of the chunks went down before actually doing the task, the periodic GC go routine running on the nodes will schedule the deletes for such files.


internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -116,7 +116,8 @@ func Init(storageCallback dcache.StorageCallbacks, cacheId string) error {
 	//
 	directories := []string{metadataManagerInstance.mdRoot,
 		metadataManagerInstance.mdRoot + "/Nodes",
-		metadataManagerInstance.mdRoot + "/Objects"}
+		metadataManagerInstance.mdRoot + "/Objects",
+		metadataManagerInstance.mdRoot + "/Deleted"}
```
linuxsmiths: see the above comment, Objects folder must be created in the end.

internal/dcache/metadata_manager/metadata_manager_impl.go
```diff
@@ -551,22 +562,22 @@ func (m *BlobMetadataManager) renameFileToDeleting(filePath string, deletedFileP
...
-	path := filepath.Join(m.mdRoot, "Objects", filePath)
+	path := filepath.Join(m.mdRoot, "Deleted", fileID)
 	err := m.storageCallback.DeleteBlobInStorage(internal.DeleteFileOptions{
 		Name: path,
 	})
 	if err != nil {
 		// Treat BlobNotFound as success.
-		if bloberror.HasCode(err, bloberror.BlobNotFound) {
-			log.Warn("DeleteFile:: DeleteBlobInStorage failed since blob %s is already deleted: %v",
+		if err == syscall.ENOENT {
```
linuxsmiths, line 590: are we certain that DeleteBlobInStorage returns the exact error syscall.ENOENT and not a wrapped error?

internal/dcache/gc/gc.go
```diff
@@ -119,6 +165,42 @@ func (gc *GcInfo) removeAllChunksForFile(gcFile *gcFile) {
...
+			//
+			// This is kept to higher value to avoid frequent storage calls as there is file IO going on this handle.
+			//
+			time.Sleep(30 * time.Second)
+		}
+		gc.deletedFileQueue <- gcFile
+	}
+
+	// Refresh the OpenCount for this file.
+	if gcFile.file.OpenCount > 0 {
```
linuxsmiths, line 215: oops, we forgot this before

internal/dcache/gc/gc.go
```diff
@@ -76,8 +95,34 @@ func Start() {
 		numGcWorkers: 100,
 		// Allow more requests to be queued than workers.
 		deletedFileQueue: make(chan *gcFile, 1000),
+		done:             make(chan struct{}),
+		//
+		// The deleteTimeOut is kept to 5 minutes, The time taken to delete a file that was scheduled in worst case is:
+		// (((stripeSize / chunkSize) * numReplicas) RPC calls + timeTaken to delete Metadata file + clustermap Update) * (Queue Size / numWorkers) for GC
+		// (((16M/4M) * 3) 100ms for each RPC call + 300ms + 300ms) * (1000 / 100) = 18000ms in worst case maybe.
+		// The time taken to delete the last file that was scheduled into the GC queue in worst case would be ~18s
```
linuxsmiths: this is not accurate since the time taken by the RemoveChunk RPC will depend on the number of chunks, which can be large for large files.

internal/dcache/gc/gc.go
```diff
@@ -214,3 +302,149 @@ func ScheduleChunkDeletion(file *dcache.FileMetadata) {
...
+		}
+
+		// Schedule the file for GC.
+		gcFile := &gcFile{
+			file:         dcFile,
+			retryCnt:     0,
+			removeMVList: dcFile.FileLayout.MVList,
+		}
+
+		gc.deletedFileQueue <- gcFile
```
linuxsmiths: to prevent multiple nodes attempting to delete the same files, which might result in too many RemoveChunk RPCs, we should distribute the responsibility of specific files to specific node.
Something like a node which hosts the first RV of the first MV of the file will delete the file. if that node is not online, it can fallback to the next RV and so on.

internal/dcache/metadata_manager/metadata_manager_impl.go
syeleti-msft: Yes, looked into the code

internal/dcache/gc/gc.go
syeleti-msft: Yes, that makes sense

internal/dcache/gc/gc.go
```diff
@@ -76,8 +95,35 @@ func Start() {
...
 
+	// Start Periodic GC to reclaim the chunks for the Stale files.
+	go func() {
+		pollTicker := time.NewTicker(gc.interval)
+
+		for {
+			select {
+			case <-gc.done:
+				log.Info("GC::Stopping Periodic GC go routine")
+				break
```
linuxsmiths: this will break out of select and not for.
We can use return.
Also, better to add sync add/wait

linuxsmiths (approved):

PR #1849, opened 2025-07-01
schedule cluster map update improvements

jainakanksha-msft:


internal/dcache/models.go
```diff
@@ -74,8 +74,8 @@ type ClusterMap struct {
 	LastUpdatedAt int64                     `json:"last_updated_at"`
 	LastUpdatedBy string                    `json:"last_updated_by"`
 	Config        DCacheConfig              `json:"config"`
-	RVMap         map[string]RawVolume      `json:"rv-map"`
-	MVMap         map[string]MirroredVolume `json:"mv-map"`
+	RVMap         map[string]RawVolume      `json:"rv-list"`
+	MVMap         map[string]MirroredVolume `json:"mv-list"`
```
linuxsmiths, line 78: why do you want to change these labels?

common/log/base_logger.go
```diff
@@ -225,13 +225,20 @@ func (l *BaseLogger) logEvent(lvl string, format string, args ...interface{}) {
 	// full func name: e.g. github.com/Azure/…/common/log.(*BaseLogger).Info
 	fullName := runtime.FuncForPC(pc).Name()
 
+	// Get hostname for the log line
+	hostname, err := os.Hostname()
+	if err != nil {
+		hostname = "unknown-host"
+	}
```
linuxsmiths: instead of making the Hostname() call everytime, store it in a global variable from inti() and use that here

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -898,7 +898,7 @@ func (cmi *ClusterManager) ensureInitialClusterMap(dCacheConfig *dcache.DCacheCo
 	clusterMap = dcache.ClusterMap{
 		Readonly:      true,
 		State:         dcache.StateReady,
-		Epoch:         1,
+		Epoch:         0,
```
linuxsmiths, line 904: any specific reason for this change?

jainakanksha-msft: version 1 of epoch has never been used if we start with zero value.

common/log/base_logger.go
jainakanksha-msft: Sure.

internal/dcache/models.go
jainakanksha-msft: I made it consistent throughout.
fs=debug consider it as rv-list mv-list, which makes the things inconsistent throughout.

linuxsmiths (approved):

PR #1859, opened 2025-07-08
Inband offline detection of RVs

souravgupta-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description
If RPC calls like PutChunk to an RV fails, then instead of updating the state of the RV to "offline", we update it to "inband-offline". Marking RV's state to offline is not correct as the connection between the two nodes might be bad, whereas the RV/server node is still up. So, we update its state to "inband-offline" state. The fix-mv workflow will later pick this "inband-offline" state RV and replace it with different RV (similar to how it is done for offline RVs).


internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -1849,12 +1849,12 @@ func (cmi *ClusterManager) updateMVList(rvMap map[string]dcache.RawVolume,
...
 
 			//
-			// If this component RV is not offline, its containing node must be excluded for replacement RV(s).
+			// If this component RV is not offline/inband-offline, its containing node must be excluded for replacement RV(s).
 			// We don't exclude the node if the component RV is offline to support the case where the same node
 			// comes back up online and we may want to use the same RV or another RV from the same node, as
 			// replacement RV.
 			//
-			if mv.RVs[rvName] != dcache.StateOffline {
+			if mv.RVs[rvName] != dcache.StateOffline && mv.RVs[rvName] != dcache.StateInbandOffline {
```
linuxsmiths: actually we should leave it as before since for inband-offline we would like to exclude the node as most likely other RVs on that node are also not reachable. If we allow that node to be used for picking replacement RVs, we might get into a situation where it picks another RV from the same node and that's bad too, then next time around it picks this one and so on..
We must exclude the node but count it as offline.

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -1990,6 +1989,15 @@ func (cmi *ClusterManager) updateMVList(rvMap map[string]dcache.RawVolume,
...
+					} else if mv.RVs[rvName] == dcache.StateInbandOffline {
+						//
+						// The rv selected for replacement is the same as the one we are trying to replace.
+						// If the state of the RV in the MV is inband-offline, we do not want to replace it
+						// with itself. So, we skip this and move to the next RV.
+						//
+						log.Debug("ClusterManager::fixMV: Not replacing %s/%s with itself as it is inband-offline",
+							rvName, mvName)
+						continue
 					}
```
linuxsmiths: with the above suggested change, this should not happen as we exclude the entire node.
We can probably just assert that for inband-offline case we don't pick the same RV

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -2144,21 +2152,24 @@ func (cmi *ClusterManager) updateMVList(rvMap map[string]dcache.RawVolume,
...
+	// - If no component RV of an MV is online (they are either offline, outofsync, syncing or inband-offline),
+	//   the MV is marked as offline, else
 	// - If at least one component RV is online, the MV is marked as degraded, else
 	// - All component RVs are either online or syncing, then MV is marked as syncing.
 	//
 	// Few examples:
 	// online, online, online => online
 	// online, online, offline => degraded
 	// online, online, outofsync => degraded
+	// online, online, inband-offline => degraded
```
linuxsmiths: nit, instead of a new entry, change L2162 to
online, online, offline|inband-offline => degraded

Replace all lines with offline similarly, and in the end

offline|inband-offline, offline|inband-offline, offline|inband-offline=> offline

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -2201,21 +2215,21 @@ func (cmi *ClusterManager) updateMVList(rvMap map[string]dcache.RawVolume,
 			// RV, so we don't reduce it now.
 			//
 			if rvMap[rvName].State != dcache.StateOffline {
-				if mv.RVs[rvName] != dcache.StateOffline {
+				if mv.RVs[rvName] != dcache.StateOffline && mv.RVs[rvName] != dcache.StateInbandOffline {
```
linuxsmiths: read the comment carefully, we should not need to change this because an inband-offline component RV will never be replaced by itself.

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -2201,21 +2215,21 @@ func (cmi *ClusterManager) updateMVList(rvMap map[string]dcache.RawVolume,
...
 					consumeRVSlot(mvName, rvName)
 				}
 			}
 		}
 
-		common.Assert((onlineRVs+offlineRVs+outofsyncRVs+syncingRVs) == len(mv.RVs),
+		common.Assert((onlineRVs+offlineRVs+outofsyncRVs+syncingRVs+inbandOfflineRVs) == len(mv.RVs),
```
linuxsmiths: nit, always place inband offline right next to offline for best readabiliy.
f.e.,
common.Assert((onlineRVs+offlineRVs+inbandOfflineRVs+outofsyncRVs+syncingRVs) == len(mv.RVs),

same in all places.

also move L2181 to 2178

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -3216,6 +3230,9 @@ func (cmi *ClusterManager) batchUpdateComponentRVState(msgBatch []*dcache.Compon
 			//       offline but the RV is online in the RV list, then updateMVList()->fixMV() might pick the same
 			//       RV as a replacement RV, which would be wrong as RV, for all purposes, is offline.
 			//
+			// TODO: discuss, this will never happen as we are not calling UpdateComponentRVState() from
+			//       replication manager with offline state as the new state of the RV.
+			//
```
linuxsmiths: yes, this is not an issue now.

internal/dcache/clustermap/utils.go
```diff
@@ -91,7 +91,8 @@ func IsValidComponentRVState(rvState dcache.StateEnum) bool {
 	case dcache.StateOnline,
 		dcache.StateOffline,
 		dcache.StateOutOfSync,
-		dcache.StateSyncing:
+		dcache.StateSyncing,
+		dcache.StateInbandOffline:
```
linuxsmiths: put it close to Offline

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -869,7 +857,7 @@ func RemoveMV(req *RemoveMvRequest) (*RemoveMvResponse, error) {
 		// some time.
 		//
 		if rv.State == string(dcache.StateOffline) || rv.State == string(dcache.StateOutOfSync) ||
-			rv.State == string(dcache.StateSyncing) {
+			rv.State == string(dcache.StateSyncing) || rv.State == string(dcache.StateInbandOffline) {
```
linuxsmiths: place offline and inband-offline right one after another.
in all places

internal/dcache/rpc/server/handler.go
```diff
@@ -750,7 +750,11 @@ func (mv *mvInfo) updateComponentRVs(componentRVs []*models.RVNameAndState, forc
 				}
 
 				if oldState == string(dcache.StateOffline) && newState == string(dcache.StateOutOfSync) {
+					//
 					// Same RV (now online) being reused by fix-mv.
+					// Note: we don't check if the oldstate is inband-offline here, since in case of
+					//       inband-offline state, same RV is not resued by the fix-mv workflow.
+					//
```
linuxsmiths: assert this?

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Done

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Updated

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Updated

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Updated

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Updated

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Removed this section

internal/dcache/clustermap/utils.go
souravgupta-msft: Done

internal/dcache/replication_manager/replication_manager.go

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done

internal/dcache/replication_manager/threadpool.go
```diff
@@ -156,11 +157,15 @@ func (tp *threadpool) runItem(item *workitem) {
 
 	switch item.reqType {
 	case putChunkRequest:
-		putChunkReq := item.rpcReq.(*models.PutChunkRequest)
+		putChunkReq, ok := item.rpcReq.(*models.PutChunkRequest)
+		_ = ok
+		common.Assert(ok)
```
linuxsmiths, line 162: better to have this and L199 consistent.

souravgupta-msft: Done

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -2983,14 +3004,26 @@ func (cmi *ClusterManager) getNextComponentRVUpdateBatch() []*dcache.ComponentRV
 			common.Assert(msg.Err != nil)
 			common.Assert(len(msg.Err) == 0, len(msg.Err))
 
-			if _, ok := existing[msg.MvName+msg.RvName]; ok {
+			if rvPrevState, ok := existing[msg.MvName+msg.RvName]; ok {
+				//
+				// If we have already seen this RV/MV combination in this batch, check if the new RV state
+				// is same as the one we have already seen. If yes, we add it to te batch.
```
jainakanksha-msft: typo **te**

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -2999,14 +3004,26 @@ func (cmi *ClusterManager) getNextComponentRVUpdateBatch() []*dcache.ComponentRV
...
+				// Later on in the batchUpdateComponentRVState() we will check if there are mutiple entries
+				// for the same RV/MV combination, we will skip the duplicate updates.
+				//
+				if rvPrevState == msg.RvNewState {
+					msgBatch = append(msgBatch, &msg)
```
linuxsmiths, line 3033: this update is exactly same as the prev one, we can simply skip it. Then we don't need to worry about dup updates in batchUpdateComponentRVState()

souravgupta-msft: We add it because the caller of this method waits for this,
`return <-updateRVMessage.Err`

In the batchUpdateComponentRVState() we are handling this case of duplicate entries, and this method adds the error in the Err channel of the message to unblock the caller about the status of the RV/MV state update in the clustermap. So, I have added this to the message batch here.

linuxsmiths: oh yes, we need to respond to the caller.
Let's update the comment to clarify that in case a new reader wonders.

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -3140,30 +3179,54 @@ func (cmi *ClusterManager) batchUpdateComponentRVState(msgBatch []*dcache.Compon
...
 				//
-				if currentState == rvNewState {
-					common.Assert(currentState == dcache.StateOffline, currentState)
+				if currentState == rvNewState ||
+					(currentState == dcache.StateOffline && rvNewState == dcache.StateInbandOffline) {
+					common.Assert(currentState != rvNewState ||
+						currentState == dcache.StateInbandOffline, currentState)
+
 					log.Debug("ClusterManager::batchUpdateComponentRVState: %s/%s ignoring state change (%s -> %s)",
 						rvName, mvName, currentState, rvNewState)
```
linuxsmiths, line 3272: nyc, but we should not close the Err channel here as that will let the caller proceed even before the clustermap is updated.

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -3140,30 +3179,54 @@ func (cmi *ClusterManager) batchUpdateComponentRVState(msgBatch []*dcache.Compon
...
+				//
+				// We can also get this transition,
+				// StateOffline -> StateInbandOffline
+				// We can get this when the RV is offline and its status has been updated in the clustermap.
+				// However in inband PutChunk failure, we send to update the RV state to InbandOffline.
+				// Since, the RV is already marked offline in the clustermap, we ignore this request.
 				//
-				if currentState == rvNewState {
-					common.Assert(currentState == dcache.StateOffline, currentState)
+				if currentState == rvNewState ||
```
linuxsmiths, line 3265: with the new existingMap check added above, can this now happen?

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -3140,30 +3179,54 @@ func (cmi *ClusterManager) batchUpdateComponentRVState(msgBatch []*dcache.Compon
...
 				//
 				// Since we can have multiple PutChunk requests outstanding, all but the first one will find the
-				// currentState as StateOffline, we need to ignore such update requests.
+				// currentState as StateInbandOffline, we need to ignore such update requests.
+				//
+				// We can also get this transition,
+				// StateOffline -> StateInbandOffline
+				// We can get this when the RV is offline and its status has been updated in the clustermap.
+				// However in inband PutChunk failure, we send to update the RV state to InbandOffline.
+				// Since, the RV is already marked offline in the clustermap, we ignore this request.
```
linuxsmiths, line 3223: since currentState is the component RV state and not the state of the RV in the RV map, is it possible?

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -1841,9 +1841,15 @@ func (cmi *ClusterManager) updateMVList(rvMap map[string]dcache.RawVolume,
 			// We don't exclude the node if the component RV is offline to support the case where the same node
 			// comes back up online and we may want to use the same RV or another RV from the same node, as
 			// replacement RV.
+			// If it the component RV is inband-offline, we exclude its node because we don't want other RVs in
+			// that node to be used as replacement RV. But we still count it as offline, so that it is replaced
+			// with a good RV.
```
linuxsmiths: some typo and suggested changes

```
                       // If the component RV is inband-offline, we exclude its node because we don't want other RVs in
                       // that node to be used as replacement RV (since it's likely that the entire node is down),
                       // but we still count it as offline, as it needs to be replaced with a good RV.
```

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -2132,26 +2145,27 @@ func (cmi *ClusterManager) updateMVList(rvMap map[string]dcache.RawVolume,
 	// per rvMap, then the component RV is force marked offline. Then it sets the MV state based on the cumulative
 	// state of all of it's component RVs as follows:
 	// - If all component RVs of an MV are online, the MV is marked as online, else
-	// - If no component RV of an MV is online (they are either offline, outofsync or syncing), the MV
-	//   is marked as offline, else
+	// - If no component RV of an MV is online (they are either offline, outofsync, syncing or inband-offline),
```
linuxsmiths: nit
```
If no component RV of an MV is online (they are either offline/inband-offline, outofsync or syncing),
```

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -2983,14 +3004,26 @@ func (cmi *ClusterManager) getNextComponentRVUpdateBatch() []*dcache.ComponentRV
...
 			common.Assert(len(msg.Err) == 0, len(msg.Err))
 
-			if _, ok := existing[msg.MvName+msg.RvName]; ok {
+			if rvPrevState, ok := existing[msg.MvName+msg.RvName]; ok {
+				//
+				// If we have already seen this RV/MV combination in this batch, check if the new RV state
+				// is same as the one we have already seen. If yes, we add it to the batch.
+				// Later on in the batchUpdateComponentRVState() we will check if there are mutiple entries
+				// for the same RV/MV combination, we will skip the duplicate updates.
+				//
```
linuxsmiths: suggested change

```
                               //
                               // This RV/MV combination has already been added to the batch.
                               // We don't support multiple updates to the same RV/MV in the same batch,
                               // but if multiple updates are all updating the RV state to the same value,
                               // we can safely ignore the duplicate updates.
                               // Note that we still need to add the message to the batch, so that the batch
                               // updater later correctly notifies the caller waiting for its update to complete
                               // (by reading from the msg.Err channel).
                               // Later on in the batchUpdateComponentRVState() we will check if there are
                               // mutiple updates for the same RV/MV combination, we will skip such duplicate
                               // updates.
                               //
                               // Note that this is an optimization, w/o this multiple updates queued by inline
                               // write failures (which could be many, since we perform multiple parallel writes
                               // to different chunks for the same file) will be processed in different ticks
                               // taking lot of time to update the clustermap.
                               //
```

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -3100,6 +3133,12 @@ func (cmi *ClusterManager) batchUpdateComponentRVState(msgBatch []*dcache.Compon
 			continue
 		}
 
+		//
+		// Map to track the RV/MV combinations that we have already seen in this batch.
+		// Key for this map is "mvName+rvName", value is the new RV state.
+		//
```
linuxsmiths, line 3159: add this

```
               // See getNextComponentRVUpdateBatch() for how it can add duplicate updates for the same
               // RV/MV combination to the batch.

```

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -3100,6 +3133,12 @@ func (cmi *ClusterManager) batchUpdateComponentRVState(msgBatch []*dcache.Compon
...
+		existing := make(map[string]dcache.StateEnum)
+
```
linuxsmiths, line 3161: this comment is for L3157, add this assert also

```
                       // Individual component RV state is never moved to offline, but instead to inband-offline.
                       common.Assert(rvNewState != dcache.StateOffline, rvNewState)
```

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -3140,30 +3179,54 @@ func (cmi *ClusterManager) batchUpdateComponentRVState(msgBatch []*dcache.Compon
...
 
+			//
+			// If the RV/MV combination is already present in the batch, we can ignore this update as it
+			// was already processed in the previous iteration. We do assert that the previous state
+			// is same as the new state requested, so that we don't end up with multiple updates for the same
+			// RV/MV combination in the batch.
+			//
+			if rvPrevState, ok := existing[mvName+rvName]; ok {
+				_ = rvPrevState
+				common.Assert(rvPrevState == rvNewState)
```
linuxsmiths: we can also assert that currentState==rvNewState (prev update would have changed the state) or it's the case where currentState is offline and rvNewState is inband-offline (in which case the prev update would not have changed the state).

internal/dcache/clustermap/clustermap.go
```diff
@@ -267,6 +267,14 @@ func ReportRVOffline(rvName string) error {
 // to the clustermap. The success or failure of this batched update will decide the success/failure of each
 // of the individual updates.
 func UpdateComponentRVState(mvName string, rvName string, rvNewState dcache.StateEnum) error {
+	startTime := time.Now()
+	defer func() {
+		timeTaken := time.Since(startTime).Microseconds()
+		_ = timeTaken
+		log.Debug("ClusterMap::UpdateComponentRVState: request took %d microseconds: %s/%s -> %v",
+			timeTaken, rvName, mvName, rvNewState)
+	}()
```
linuxsmiths: but this whole thing inside if common.IsDebugBuild() so that the entire code is removed in release builds.

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -301,6 +301,13 @@ retry:
 func WriteMV(req *WriteMvRequest) (*WriteMvResponse, error) {
 	common.Assert(req != nil)
 
+	startTime := time.Now()
+	defer func() {
+		timeTaken := time.Since(startTime).Microseconds()
+		_ = timeTaken
+		log.Debug("ReplicationManager::WriteMV: WriteMV request took %d microseconds: %v", timeTaken, req.toString())
+	}()
+
```
linuxsmiths: same as above

internal/dcache/rpc/client/client_pool.go
```diff
@@ -139,19 +139,51 @@ func (cp *clientPool) getRPCClient(nodeID string) (*rpcClient, error) {
...
+	waitTime := 0
+
+	for {
+		select {
+		case client := <-ncPool.clientChan:
+			ncPool.lastUsed.Store(time.Now().Unix())
+			common.Assert(client.nodeID == nodeID, client.nodeID, nodeID)
+			ncPool.numActive.Add(1)
+			return client, nil
+		case <-waitTicker.C:
```
linuxsmiths: we can omit the ticker.

```
case <-time.After(2 * time.Second): // Timeout after 2 second
```

internal/dcache/rpc/response_error.go
```diff
@@ -117,10 +117,10 @@ func IsBrokenPipe(err error) bool {
...
-//	      If the blobfuse process goes down before the send()/write() can send the RPC request, it fails
-//	      with IsBrokenPipe().
-//	      Another important thing to note is that IsBrokenPipe() can come when the target blobfuse process
+//	     send the RPC request but before it could respond, the sender read/recv call fails with this error.
+//	     If the blobfuse process goes down before the send()/write() can send the RPC request, it fails
+//	     with IsBrokenPipe().
+//	     Another important thing to note is that IsBrokenPipe() can come when the target blobfuse process
 //		 might have restarted at some point in the past and may be running now, so it may make sense to
 //		 retry connection attempt on getting IsBrokenPipe(), whereas IsConnectionReset() means the process
 //		 just stopped/restarted, so trying connection attempt immediately may not help.
```
linuxsmiths: indentation issues

internal/dcache/rpc/server/handler.go
```diff
@@ -780,10 +780,18 @@ func (mv *mvInfo) updateComponentRVs(componentRVs []*models.RVNameAndState, forc
 				}
 
 				if oldState == string(dcache.StateOffline) && newState == string(dcache.StateOutOfSync) {
+					//
 					// Same RV (now online) being reused by fix-mv.
+					// Note: we don't check if the oldstate is inband-offline here, since in case of
+					//       inband-offline state, same RV is not resued by the fix-mv workflow.
```
linuxsmiths: resued -> reused

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Here we can close the Err channel to unblock the caller, because the state of the RV/MV is already offline, so we should skip updating it to inband-offline.

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: currentState is the state of the RV/MV in the clustermap. The check we have for the existing map is,
```
	if rvPrevState, ok := existing[mvName+rvName]; ok {
		_ = rvPrevState
		common.Assert(rvPrevState == rvNewState)
		log.Debug("ClusterManager::batchUpdateComponentRVState: %s/%s, ignoring duplicate state change (%s -> %s)",
			rvName, mvName, currentState, rvNewState)
		ignoredCount++
		continue
	}
```

This check is for ignoring duplicate entries in the msg batch. We can have a case where the current state in the clustermap is same as the new state. Consider this case when in the first batch we updated the request for an RV/MV to inband-offline. And then in the next batch, we can get the same state update for the same RV/MV.

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Done

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Updated

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Updated

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Added

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Added

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: We cannot assert `currentState==rvNewState`. For example, there are multiple duplicate entries in the batch to update rv0/mv0 to inband-offline from online state. In the first entry, we add it to the existing map. In the next entry, we ignore it since it was already processed in the previous duplicate entry. Here, currentState(online) is not equal to rvNewState(inband-offline).

internal/dcache/clustermap/clustermap.go
souravgupta-msft: Done

internal/dcache/replication_manager/replication_manager.go

internal/dcache/rpc/client/client_pool.go

internal/dcache/rpc/response_error.go
souravgupta-msft: Fixed

internal/dcache/rpc/server/handler.go
souravgupta-msft: Done

internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -3147,30 +3202,54 @@ func (cmi *ClusterManager) batchUpdateComponentRVState(msgBatch []*dcache.Compon
...
+				log.Debug("ClusterManager::batchUpdateComponentRVState: %s/%s, ignoring duplicate state change (%s -> %s)",
+					rvName, mvName, currentState, rvNewState)
+				ignoredCount++
+				continue
+			}
+
 			//
 			// and the new state requested must be valid.
 			// Note that we support only few distinct state transitions.
```
linuxsmiths: add  a comment calling out that we will reach here only for the 1st mv/rv combination

internal/dcache/rpc/client/client_pool.go
```diff
@@ -139,19 +139,48 @@ func (cp *clientPool) getRPCClient(nodeID string) (*rpcClient, error) {
...
+			// we reset the connections available for the node, which first closes the stale connections
+			// and then creates new connections. Since, the node is down, creating new RPC connection
+			// to the node fails with connection refused error. So, eventually all the connections in the
+			// pool are closed and client pool for the node is deleted. In this case, we do not wait
+			// for a client to become available and return error to the caller.
+			//
+			if ncPool.numActive.Load() == 0 && len(ncPool.clientChan) == 0 {
+				_, exists := cp.clients[nodeID]
+				_ = exists
+				common.Assert(!exists, nodeID, cp.clients)
```
linuxsmiths: remove this

internal/dcache/rpc/response_error.go
```diff
@@ -115,15 +115,15 @@ func IsBrokenPipe(err error) bool {
...
-// Note: read()/recv() fails with this, i.e., only when the blobfuse process goes down after we successfully
-//
-//	      send the RPC request but before it could respond, the sender read/recv call fails with this error.
-//	      If the blobfuse process goes down before the send()/write() can send the RPC request, it fails
-//	      with IsBrokenPipe().
-//	      Another important thing to note is that IsBrokenPipe() can come when the target blobfuse process
-//		 might have restarted at some point in the past and may be running now, so it may make sense to
-//		 retry connection attempt on getting IsBrokenPipe(), whereas IsConnectionReset() means the process
-//		 just stopped/restarted, so trying connection attempt immediately may not help.
+// Note:
```
linuxsmiths: indent

internal/dcache/rpc/server/handler.go
```diff
@@ -801,10 +801,18 @@ func (mv *mvInfo) updateComponentRVs(componentRVs []*models.RVNameAndState, forc
...
+					// Note: we don't check if the oldstate is inband-offline here, since in case of
+					//       inband-offline state, same RV is not reused by the fix-mv workflow.
+					//
 					continue
 				}
 
+				// For same RV, we cannot have old state as inband-offline.
+				common.Assert(oldState != string(dcache.StateInbandOffline),
+					mv.mvName, oldName, oldState, newState)
+
```
linuxsmiths: let's keep inband-offline state limited to the WriteMV path.. in the handler code let's always use offline.

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Discussed this offline

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Added more comments as discussed.

internal/dcache/cluster_manager/cluster_manager.go
souravgupta-msft: Added

internal/dcache/rpc/client/client_pool.go

internal/dcache/rpc/response_error.go
souravgupta-msft: Done

internal/dcache/rpc/server/handler.go
souravgupta-msft: Updated, please review

linuxsmiths (approved):

PR #1862, opened 2025-07-09
Cycle cloud Bug Fixes

jainakanksha-msft:


component/distributed_cache/distributed_cache.go
```diff
@@ -170,6 +170,12 @@ func (dc *DistributedCache) Start(ctx context.Context) error {
 		return log.LogAndReturnError("DistributedCache::Start error [azstorage component not found]")
 	}
 
+	// We are creating UUID before initialising any of the dcache component. So that UUID is available for all components
```
linuxsmiths:
```
	// Create UUID before initializing any of the dcache components, so that UUID is correctly available for all components.

```

component/distributed_cache/distributed_cache.go
```diff
@@ -170,6 +170,12 @@ func (dc *DistributedCache) Start(ctx context.Context) error {
...
+	ensureUUID()
+
+	// rpc client should be initialised before any of its user.
```
linuxsmiths: // rpc client must be initialized before any of its users.

PR #1865, opened 2025-07-09
Eliminate du usage

keith-ms:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description
Previously, in order to estimate disk usage for a specific path, the `GetUsage()` function would use the `Command` function to execute the `du` command, parse the output and return the results in megabytes. After troubleshooting an issue that involved file writes to a volume that made use of this driver causing the write to hang, I tracked down the hang to entry into the `GetUsage()` function.

This change removes the usage of `du` in favor of replicating the functionality using the `WalkDir()` function included in the Go standard library. I also added functions with clearer names, preserved the API of the `common` package, updated functions that previously called the `GetUsage()` function to use the newer functions and added additional testing.

Removing the use of the `Command` function prevents the creation of a new external process and makes the `GetUsage()` function more portable across Linux distributions.

## Checklist

## Related Links
- [Issues](<link>)

### DIFF
--- a/common/util.go
+++ b/common/util.go
@@ -363,65 +363,49 @@ func NotifyMountToParent() error {
 	return nil
 }
 
-var duPath []string = []string{"/usr/bin/du", "/usr/local/bin/du", "/usr/sbin/du", "/usr/local/sbin/du", "/sbin/du", "/bin/du"}
-var selectedDuPath string = ""
+// Defining this wrapper function to preserve the original API.
+func GetUsageInBytes(path string) (int64, error) {
+	totalUsage := int64(0)
 
-// GetUsage: The current disk usage in MB
-func GetUsage(path string) (float64, error) {
-	var currSize float64
-	var out bytes.Buffer
-
-	if selectedDuPath == "" {
-		selectedDuPath = "-"
-		for _, dup := range duPath {
-			_, err := os.Stat(dup)
-			if err == nil {
-				selectedDuPath = dup
-				break
-			}
+	err := filepath.WalkDir(path, func(path string, _ os.DirEntry, err error) error {
+		if err != nil {
+			return err
 		}
-	}
 
-	if selectedDuPath == "-" {
-		return 0, fmt.Errorf("failed to find du")
-	}
+		/*
+			Use os.Lstat() to get the file information. The previous implementation of this function
+			used du, and it did not include the -L flag to follow symlinks.
+			Using os.Stat() will resolve symlinks and return the size of the target file,
+			which will change the behavior of this function.
+		*/
+		fileInfo, err := os.Lstat(path)
+		if err != nil {
+			return err
+		}
 
-	// du - estimates file space usage
-	// https://man7.org/linux/man-pages/man1/du.1.html
-	// Note: We cannot just pass -BM as a parameter here since it will result in less accurate estimates of the size of the path
-	// (i.e. du will round up to 1M if the path is smaller than 1M).
-	cmd := exec.Command(selectedDuPath, "-sh", path)
-	cmd.Stdout = &out
+		totalUsage += fileInfo.Size()
+		return nil
+	})
 
-	err := cmd.Run()
 	if err != nil {
 		return 0, err
 	}
 
-	size := strings.Split(out.String(), "\t")[0]
-	if size == "0" {
-		return 0, nil
-	}
+	return totalUsage, nil
+}
 
-	// some OS's use "," instead of "." that will not work for float parsing - replace it
-	size = strings.Replace(size, ",", ".", 1)
-	parsed, err := strconv.ParseFloat(size[:len(size)-1], 64)
+func GetUsageInMegabytes(path string) (float64, error) {
+	return GetUsage(path)
+}
+
+// GetUsage: The current disk usage in MB using Go libraries
+// Deprecated: Use GetUsageInBytes or GetUsageInMegabytes instead, which provide more clarity on the return units.
+func GetUsage(path string) (float64, error) {
+	totalUsageInBytes, err := GetUsageInBytes(path)
 	if err != nil {
-		return 0, fmt.Errorf("failed to parse du output")
+		return 0, err
 	}
-
-	switch size[len(size)-1] {
-	case 'K':
-		currSize = parsed / float64(1024)
-	case 'M':
-		currSize = parsed
-	case 'G':
-		currSize = parsed * 1024
-	case 'T':
-		currSize = parsed * 1024 * 1024
-	}
-
-	return currSize, nil
+	return float64(totalUsageInBytes) / (1024 * 1024), nil
 }
 
 var currentUID int = -1
--- a/component/block_cache/block_cache.go
+++ b/component/block_cache/block_cache.go
@@ -1785,7 +1785,15 @@ func (bc *BlockCache) diskEvict(node *list.Element) {
 
 // checkDiskUsage : Callback to check usage of disk and decide whether eviction is needed
 func (bc *BlockCache) checkDiskUsage() bool {
-	data, _ := common.GetUsage(bc.tmpPath)
+	var data = float64(0)
+	var err error
+	if bc.tmpPath != "" {
+		data, err = common.GetUsageInMegabytes(bc.tmpPath)
+		if err != nil {
+			log.Err("BlockCache::checkDiskUsage : Failed to get disk usage for %s [%s]", bc.tmpPath, err.Error())
+			return false
+		}
+	}
 	usage := uint32((data * 100) / float64(bc.diskSize/_1MB))
 
 	if bc.maxDiskUsageHit {
@@ -1930,12 +1938,22 @@ func (bc *BlockCache) StatFs() (*syscall.Statfs_t, bool, error) {
 		return nil, false, nil
 	}
 
-	usage, _ := common.GetUsage(bc.tmpPath)
-	usage = usage * float64(_1MB)
+	// Previously, this section ignored the error returned by a call to GetUsage.
+	// In order to preserve the behavior of GetUsage, I'll check for an empty value
+	// for bc.tmpPath. If it's empty, use 0 for usage.
+	usage := int64(0)
+	var err error
+	if bc.tmpPath != "" {
+		usage, err = common.GetUsageInBytes(bc.tmpPath)
+		if err != nil {
+			log.Err("BlockCache::StatFs : Failed to get disk usage for %s [%s]", bc.tmpPath, err.Error())
+			return nil, false, err
+		}
+	}
 
-	available := (float64)(maxCacheSize) - usage
+	available := (float64)(maxCacheSize) - float64(usage)
 	statfs := &syscall.Statfs_t{}
-	err := syscall.Statfs("/", statfs)
+	err = syscall.Statfs("/", statfs)
 	if err != nil {
 		log.Debug("BlockCache::StatFs : statfs err [%s].", err.Error())
 		return nil, false, err
--- a/component/file_cache/cache_policy.go
+++ b/component/file_cache/cache_policy.go
@@ -86,7 +86,7 @@ func getUsagePercentage(path string, maxSize float64) float64 {
 		}
 	} else {
 		// We need to compuate % usage of temp directory against configured limit
-		currSize, err = common.GetUsage(path)
+		currSize, err = common.GetUsageInMegabytes(path)
 		if err != nil {
 			log.Err("cachePolicy::getUsagePercentage : failed to get directory usage for %s [%v]", path, err.Error)
 		}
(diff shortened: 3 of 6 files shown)


common/util.go
```diff
@@ -363,65 +363,49 @@ func NotifyMountToParent() error {
...
-
-	if selectedDuPath == "" {
-		selectedDuPath = "-"
-		for _, dup := range duPath {
-			_, err := os.Stat(dup)
-			if err == nil {
-				selectedDuPath = dup
-				break
-			}
+	err := filepath.WalkDir(path, func(path string, _ os.DirEntry, err error) error {
```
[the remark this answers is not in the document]
keith-ms: This comment is incorrect, the error is checked on line 390.

common/util.go
```diff
@@ -363,65 +363,49 @@ func NotifyMountToParent() error {
...
+			return err
+		}
 
-	// du - estimates file space usage
-	// https://man7.org/linux/man-pages/man1/du.1.html
-	// Note: We cannot just pass -BM as a parameter here since it will result in less accurate estimates of the size of the path
-	// (i.e. du will round up to 1M if the path is smaller than 1M).
-	cmd := exec.Command(selectedDuPath, "-sh", path)
-	cmd.Stdout = &out
+		totalUsage += fileInfo.Size()
```
bhansalivikas: We are doing a WalkDir and then for each item we are doign a os.Lstat(). I see this is a lot of overhead in terms of time and CPU usage on blobfuse. Most customers will not like this change and hence it shall not be a default behavior.

keith-ms: I ran some benchmarks comparing the use of `WalkDir` versus `Walk` versus `du`. You're right - there is certainly memory efficiency to be gained by using the `Walk` function and eliminating the extra call to `os.Lstat()`. I'll put up a change that will fix this problem. Memory usage is also lower when running the calculation in Go until about 10,000 or so files in the path and subdirectories. Go memory utilization with this method increases linearly with number of files and `du` stays constant at about 2MB of memory usage regardless of the number of files. Below about 10,000 files, Go is below 2MB of memory usage.

The big win is on CPU usage: because of the overhead of starting up a process (~1ms), calculating the usage for a path is significantly faster in Go than it is when shelling out to `du`, especially with a small number of files (<10,000).

Eliminating the usage of `du` also removes brittleness associated with parsing output from the command. Different locales or even non-GNU versions of `du` may cause the output to not conform to expectations.

An aside: I will make a change so that the result returned by the various `GetUsage*` functions return a size based on block usage, not based on logical usage. This is in line with the previous invocation of `du` without the `--apparent-use` flag.

bhansalivikas: If there is a deep nesting or the number of files in path is very large than the code will have significant performance impact. If 'du' if not usable in your environment I propose we keep the default 'du' flow as is and you add a new flow to compute the size in this new flow and keep this behind a hidden flag.

keith-ms: I ran benchmarks for up to 100,000 files. The Go program's memory utilization grows linearly with the number of files (up to about 42MB or so in this situation) versus the constant memory utilization demonstrated by `du`. I doubt that I can get the pure Go implementation down to such low memory utilization. 

I'll add a flag that disables the use of the external `du`, as you requested.

cmd/mount.go
```diff
@@ -844,6 +844,12 @@ func init() {
 	mountCmd.Flags().BoolVar(&options.AttrCache, "use-attr-cache", true, "Use attribute caching.")
 	config.BindPFlag("use-attr-cache", mountCmd.Flags().Lookup("use-attr-cache"))
 
+	mountCmd.Flags().Bool("file-cache-use-du", true, "Use du command for file cache disk usage calculation.")
```
bhansalivikas: This shall be generic flag like no-du and shall not have any "file-cache" reference in this. If this option is very specific to file-cache component than it shall be exported from file-cache component itself.

cmd/mount.go
```diff
@@ -844,6 +844,12 @@ func init() {
...
+	config.BindPFlag("file_cache.use-du", mountCmd.Flags().Lookup("file-cache-use-du"))
+
+	mountCmd.Flags().Bool("block-cache-use-du", true, "Use du command for block cache disk usage calculation.")
```
bhansalivikas: there shall be only one flag and not seperate for file-cache and block-cache. Lets generalise it to "no-du" or something like that and default behaviour shall be to use 'du' command so current flow does not change for any existing customer.

common/util.go
```diff
@@ -363,11 +363,42 @@ func NotifyMountToParent() error {
 	return nil
 }
 
-var duPath []string = []string{"/usr/bin/du", "/usr/local/bin/du", "/usr/sbin/du", "/usr/local/sbin/du", "/sbin/du", "/bin/du"}
-var selectedDuPath string = ""
+func GetUsageWithWalkInBytes(path string) (int64, error) {
+	totalUsage := int64(0)
+
+	const blockSize = 512
+	err := filepath.Walk(path, func(path string, info os.FileInfo, err error) error {
```
bhansalivikas, line 382: if there is a child directory, does walk take care of it or we need to iterate the children seperatly ?

common/util.go
```diff
@@ -363,11 +363,42 @@ func NotifyMountToParent() error {
...
+		if err != nil {
+			return err
+		}
+		if stat, ok := info.Sys().(*syscall.Stat_t); ok {
+			totalUsage += stat.Blocks * blockSize
```
bhansalivikas, line 387: should it be blocks*block-size of the file size directly ? If the file size does not align properly to the block-size then this will give a higher usage.

component/block_cache/block_cache.go
```diff
@@ -102,12 +103,14 @@ type BlockCacheOptions struct {
 	PrefetchOnOpen bool    `config:"prefetch-on-open" yaml:"prefetch-on-open,omitempty"`
 	Consistency    bool    `config:"consistency" yaml:"consistency,omitempty"`
 	CleanupOnStart bool    `config:"cleanup-on-start" yaml:"cleanup-on-start,omitempty"`
+	UseDu          bool    `config:"use-du" yaml:"use-du,omitempty"`
```
bhansalivikas: shall not be part of component level config. Rather use a common config as the same flag is needed by more than one component. You can check use of 'read-only' for reference.

component/block_cache/block_cache.go
```diff
@@ -102,12 +103,14 @@ type BlockCacheOptions struct {
...
 }
 
 const (
 	compName                = "block_cache"
 	defaultTimeout          = 120
 	defaultBlockSize        = 16
+	defaultUseDu            = true
```
bhansalivikas: no need to define the constant for true.

component/block_cache/block_cache.go
```diff
@@ -1788,9 +1797,26 @@ func (bc *BlockCache) diskEvict(node *list.Element) {
 	_ = os.Remove(localPath)
 }
 
+// getDiskUsage returns the current disk usage in MB
+func (bc *BlockCache) getDiskUsage() (float64, error) {
```
bhansalivikas, line 1811: This function looks ok but if we call this function periodically or at high interval then we are comparing the flags everytime. This can be optimised further by creating a function pointer to selected method and in this function just call whatever its poitning to,

component/file_cache/cache_policy.go
```diff
@@ -74,21 +75,28 @@ type cachePolicy interface {
 }
 
 // getUsagePercentage:  The current cache usage as a percentage of the maxSize
-func getUsagePercentage(path string, maxSize float64) float64 {
+func getUsagePercentage(path string, maxSize float64, useDu bool) float64 {
```
bhansalivikas: again this can be acheived using function pointer instead of passing down the flags and then doing if-else check.

component/file_cache/file_cache.go
```diff
@@ -951,7 +981,13 @@ func (fc *FileCache) OpenFile(options internal.OpenFileOptions) (*handlemap.Hand
 
 		if fileSize > 0 {
 			if fc.diskHighWaterMark != 0 {
-				currSize, err := common.GetUsage(fc.tmpPath)
+				var currSize float64
+				var err error
+				if fc.useDu {
```
bhansalivikas: Same comments to avoid if checks in every call.

sampleBlockCacheConfig.yaml
```diff
@@ -20,6 +20,7 @@ block_cache:
   mem-size-mb: 4096
   prefetch: 80
   parallelism: 128
+  use-du: true  # Use external 'du' command for disk usage calculation
```
bhansalivikas: if default is true and than we shouldn't be explicitly putting this flag in any of the sample configs.

setup/baseConfig.yaml
```diff
@@ -28,6 +28,7 @@ block_cache:
   disk-timeout-sec: <default disk cache eviction timeout (in sec). Default - 120 sec>
   prefetch: <number of blocks to be prefetched in serial read case. Min - 11, Default - 2 times number of CPU cores>
   parallelism: <number of parallel threads downloading the data and writing to disk cache. Default - 3 times number of CPU cores> 
+  use-du: <if true, du command will be used to calculate disk usage. Default - true>
```
bhansalivikas: as a thumb rule, we have all bool flags set to false by default. This is breaking that rule so ideally we shall have a flag called "no-du" or "du-not-present" or something.

common/util.go
keith-ms: Both `Walk` and `WalkDir` will walk the entire directory rooted at `path`:

https://pkg.go.dev/path/filepath#Walk
https://pkg.go.dev/path/filepath#WalkDir

I found it strange that the benchmarking I did showed that `Walk` performed better in terms of CPU and memory utilization than `WalkDir`. I'll try again to be sure.

common/util.go
keith-ms: That's correct and intentional. I want this output to be as similar to `du` as possible. The existing implementation that uses `du` passes in the `-sh` flags. This prints a summary and makes the output human readable. In the GNU implementation of `du`, the default output is disk usage, not the size of the file (the latter requiring the use of `--apparent-size`).

component/block_cache/block_cache.go
keith-ms: The `read-only` flag appears only as a CLI parameter, not in the configuration file. There's no generic section in `baseConfig.yaml`. Where should a generic parameter like this exist within the configuration file?

keith-ms: I added it. I'm not satisfied with having `no-du` not rooted under anything else inside of the configuration file, but it may be acceptable. Creating a section called `global` may overcomplicate the configuration file.

common/util.go
```diff
@@ -363,11 +375,42 @@ func NotifyMountToParent() error {
...
+	totalUsageInBytes, err := GetUsageWithWalkInBytes(path)
+	if err != nil {
+		return 0, err
+	}
+	return float64(totalUsageInBytes) / (1024 * 1024), nil
+}
+
+// GetUsageWithDu: The current disk usage in MB
+func GetUsageWithDu(path string) (float64, error) {
+	var duPath []string = []string{"/usr/bin/du", "/usr/local/bin/du", "/usr/sbin/du", "/usr/local/sbin/du", "/sbin/du", "/bin/du"}
```
bhansalivikas, line 411: These are static values, lets keep this as global variable only otherwise every time the function is called, it will end up creating a new string slice with these values. Just a waste of memory nothing much.

component/block_cache/block_cache.go
```diff
@@ -253,6 +254,23 @@ func (bc *BlockCache) Configure(_ bool) error {
 		bc.diskTimeout = conf.DiskTimeout
 	}
 
+	var noDu bool
+	if err := config.UnmarshalKey("no-du", &noDu); err == nil {
+		if noDu {
+			bc.diskUsageConfiguration = common.DiskUsageConfiguration{
+				DiskUsageFunction: common.GetUsageWithWalkInMegabytes,
+				UsesDu:            false,
```
bhansalivikas, line 262: If the callback method is set then do we need this flag marking use du or no du. Based on the method linked here in function pointer we have any way made a decision to use or not use du.

component/block_cache/block_cache.go
```diff
@@ -253,6 +254,23 @@ func (bc *BlockCache) Configure(_ bool) error {
...
+			}
+		} else {
+			bc.diskUsageConfiguration = common.DiskUsageConfiguration{
```
bhansalivikas, line 265: Instead of being a file-cache or a block-cache member this could be a global variable in itself where GetUsage is generalised so that in any code space in blobfuse if we try to compute the disk usage it uses the configured method. By doing it in this format we are saying file-cache and block-cache will define du or no-du but other components are still free to rely on du command. Ideally a global function pointer shall be set based on the input cli flag and all other component shall just use that function pointer to get disk space, globally that function defines whether to use du or not.

PR #1872, opened 2025-07-14
End all threads when blobfuse is unmounted

ashruti-msft:
…d hung threads

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

## Checklist

## Related Links
- [Issues](<link>)

### DIFF
--- a/component/xload/blockpool.go
+++ b/component/xload/blockpool.go
@@ -34,6 +34,8 @@
 package xload
 
 import (
+	"context"
+	"fmt"
 	"sync/atomic"
 
 	"github.com/Azure/azure-storage-fuse/v2/common/log"
@@ -55,10 +57,13 @@ type BlockPool struct {
 
 	// Number of threads waiting for block
 	waitLength atomic.Int32
+
+	// Context to cancel new block allocation
+	ctx context.Context
 }
 
 // NewBlockPool allocates a new pool of blocks
-func NewBlockPool(blockSize uint64, blockCount uint32) *BlockPool {
+func NewBlockPool(blockSize uint64, blockCount uint32, ctx context.Context) *BlockPool {
 	// Ignore if config is invalid
 	if blockSize == 0 || blockCount == 0 {
 		log.Err("BlockPool::NewBlockPool : blockSize : %v, block count : %v", blockSize, blockCount)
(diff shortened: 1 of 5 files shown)


component/xload/blockpool.go
```diff
@@ -142,8 +148,18 @@ func (pool *BlockPool) GetBlock(priority bool) *Block {
 
 // TryGet a block from the pool. If the pool is empty, wait till a block is released back to the pool
 func (pool *BlockPool) tryGet() *Block {
+	var block *Block = nil
+
+	select {
+	case <-pool.ctx.Done():
+		err := fmt.Errorf("Failed to Allocate Buffer as the process was cancelled, Len (blocksCh: %d, priorityCh: %d), MaxBlocks: %d",
+			len(pool.blocksCh), len(pool.priorityCh), pool.maxBlocks)
+		log.Warn("BlockPool::GetBlock : %v", err)
```
souravgupta-msft: This should be error log

component/xload/splitter.go
```diff
@@ -302,6 +292,22 @@ func (ds *downloadSplitter) Process(item *WorkItem) (int, error) {
...
+
+	if ds.validateMD5 {
+		err = ds.checkConsistency(item)
+		if err != nil {
+			// TODO:: xload : retry if md5 validation fails
+			log.Err("downloadSplitter::Process : unable to validate md5 for %s [%s]", item.Path, err.Error())
+			operationSuccess = false
+		}
+	}
+
```
souravgupta-msft: Any reason why this block is moved down?
Based on the checkConsistency, we update the `operationSuccess` variable, and then we check the `operationSuccess` and decide if we want to delete the file if not downloaded successfully. So, this block should be moved back up.

component/xload/blockpool.go
bhansalivikas: error that we return will be logged by the caller anyway, so we can actually skip this error log here I beleive.

component/xload/splitter.go
bhansalivikas: Even if the download fails, we tend to do chtimes and validateMD5. So order reversal looks natural, as in we shall first check whether download was successful or not and then validate such stuff.

souravgupta-msft (approved):

souravgupta-msft (approved):

vibhansa-msft (approved):

PR #1880, opened 2025-07-17
Allocate buffer from BufferPool while reading from the Local RV

syeleti-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

Total Allocations came down while using dcache with 1 replica in 1 node, where entire data of the file is served from local node.

```

|---------------+-------------------|
| Before change | Total Allocs      |
|---------------+-------------------|
| WRITE 10G     | 1297115616(1.2G)  |
| READ 10G      | 11957655696(11.9G) |
|---------------+-------------------|


|--------------+-------------------|
| After change | Total Allocs      |
|--------------+-------------------|
| WRITE 10G    | 1121690104(1.12G) |
| READ 10G     | 1130115944(1.13G) |
|--------------+-------------------|
```

### DIFF
--- a/internal/dcache/buffer.go
+++ b/internal/dcache/buffer.go
@@ -131,6 +131,9 @@ func PutBuffer(buf []byte) {
 	// Caller must free a buffer that's allocated using getBuffer().
 	common.Assert(bufPool.curBuffers.Load() > 0, bufPool.curBuffers.Load())
 
+	// This buffer must have been allocated from the pool and hence must have capacity at least bp.bufSize.
+	common.Assert(cap(buf) >= bufPool.bufSize, cap(buf), bufPool.bufSize)
+
 	// Reslice the length of the buffer to its original capacity if it got compacted.
 	buf = buf[:bufPool.bufSize]
 
--- a/internal/dcache/buffer.go
+++ b/internal/dcache/buffer.go
@@ -31,10 +31,11 @@
    SOFTWARE
 */
 
-package filemanager
+package dcache
 
 import (
 	"errors"
+	"fmt"
 	"sync"
 	"sync/atomic"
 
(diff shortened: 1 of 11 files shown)


internal/dcache/rpc/server/handler.go
```diff
@@ -1836,7 +1836,25 @@ func (h *ChunkServiceHandler) GetChunk(ctx context.Context, req *models.GetChunk
...
+			log.Err("ChunkServiceHandler::GetChunk: %s", errStr)
+			common.Assert(false, err)
+			return nil, rpc.NewResponseError(models.ErrorCode_InternalServerError, errStr)
+		}
+		// Reslice the data buffer accordingly, length of the buffer that we get from the BufferPool is of
+		// maximum size(i.e., Chunk Size)
+		data = data[:req.Length]
+	} else {
+		data = make([]byte, req.Length)
+	}
```
souravgupta-msft: For reading from local RVs this will work fine where we won't hit the GC issue. We should consider allocating the buffer from buffer pool for all cases, local read as well as via RPC. For now, you can add a TODO.
Also please check how much perf difference are we seeing for local reads with this change.

syeleti-msft: Sure, will update the perf diff.

syeleti-msft: Updated the description, pls take a look at it

internal/dcache/buffer.go
```diff
@@ -43,26 +44,68 @@ import (
...
+type BufferPool struct {
 	pool       sync.Pool    // sync.Pool to relieve GC
 	bufSize    int          // size of buffers in this pool
 	maxBuffers int64        // max allocated buffers allowed
 	curBuffers atomic.Int64 // buffers currently allocated
 }
 
-func NewBufferPool(bufSize int, maxBuffers int) *bufferPool {
-	return &bufferPool{
+func CreateBufferPool(bufSize uint64) error {
```
linuxsmiths: InitBufferPool 
and let's add some comments above this explaining that this is a singleton pool which can be used by callers to allocate buffers of 'bufSize' using the GetBuffer() and PutBuffer() functions.
Also let's keep BufPool private and make GetBuffer() and PutBuffer() package functions.

internal/dcache/buffer.go
```diff
@@ -43,26 +44,68 @@ import (
...
+		bufSize:    int(bufSize),
 		maxBuffers: int64(maxBuffers),
 	}
+
+	return nil
 }
 
-func (bp *bufferPool) getBuffer() ([]byte, error) {
+func (bp *BufferPool) GetBuffer() ([]byte, error) {
 	if bp.curBuffers.Load() > bp.maxBuffers {
```
linuxsmiths: let's add a TODO to make this wait for buffers to get freed and only fail after a timeout.

internal/dcache/buffer.go
```diff
@@ -76,12 +119,15 @@ func (bp *bufferPool) getBuffer() ([]byte, error) {
...
-func (bp *bufferPool) putBuffer(buf []byte) {
+func (bp *BufferPool) PutBuffer(buf []byte) {
 	// All buffers allocated from the pool must be of the same size.
-	common.Assert(len(buf) == bp.bufSize, len(buf), bp.bufSize)
+	common.Assert(len(buf) <= bp.bufSize, len(buf), bp.bufSize)
 	// Caller must free a buffer that's allocated using getBuffer().
 	common.Assert(bp.curBuffers.Load() > 0, bp.curBuffers.Load())
 
+	// Reslice the length of the buffer to its original capacity if it got compacted.
+	buf = buf[:bp.bufSize]
```
linuxsmiths: // This buffer must have been allocated from the pool and hence must have capacity at least bp.bufSize.
common.Assert(cap(buf) >= bp.bufSize, cap(buf), bp.bufSize)

internal/dcache/file_manager/file_manager.go
```diff
@@ -548,7 +500,7 @@ func (file *DcacheFile) finalizeFile() error {
 // Return the requested chunk from the staged chunks, if present, else create a new one and add to the staged chunks.
 // 'allocateBuf' controls if the StagedChunk returned has its buffer allocated by us. Note that ReadMV()
 // returns the buffer where the data is read by the GetChunk() RPC, so we don't want a pre-allocated buffer
-// in that case.
+// in that case. We only use pre-allocated buffer when reading from the local RV, where there is no RPC involved.
```
linuxsmiths: move this comment to L523 and explain in detail about how IsBufExternal will be always true for reads here but later when the actual chunk read is done, depending on whether the chunk was read locally....

internal/dcache/rpc/server/handler.go
```diff
@@ -1836,7 +1836,25 @@ func (h *ChunkServiceHandler) GetChunk(ctx context.Context, req *models.GetChunk
...
 	// TODO: Need to ensure this is FS_BLOCK_SIZE aligned.
 	//
-	data := make([]byte, req.Length)
+	var data []byte
+
+	if req.IsLocalRV {
+		//
+		// Allocate the buffer from the buffer pool.
+		//
+		data, err = dcache.BufPool.GetBuffer()
```
linuxsmiths: explain in more detail why

internal/dcache/rpc/server/handler.go
```diff
@@ -1873,6 +1891,10 @@ func (h *ChunkServiceHandler) GetChunk(ctx context.Context, req *models.GetChunk
 	//}
 	n, _, err = readChunkAndHash(&chunkPath, hashPathPtr, req.OffsetInChunk, &data)
 	if err != nil {
+		if req.IsLocalRV {
```
linuxsmiths: add to the above error path

internal/dcache/rpc/server/utils.go
```diff
@@ -153,6 +153,7 @@ func GetChunkLocal(ctc context.Context, req *models.GetChunkRequest) (*models.Ge
 	// Caller must not set SenderNodeID, catch misbehaving callers.
 	common.Assert(len(req.SenderNodeID) == 0, req.SenderNodeID)
 	req.SenderNodeID = rpc.GetMyNodeUUID()
+	req.IsLocalRV = true
```
linuxsmiths, line 184: write a good comment explaing the reason why we do this and why we cannot do this for the non-local case

linuxsmiths (approved):
Approved with comments.

internal/dcache/rpc/server/handler.go
souravgupta-msft: Can you add a TODO here stating we should always allocate buffer from the buffer pool irrespective of whether it is a local node or not.

PR #1882, opened 2025-07-17
Provide a mode to just disable kernel cache not the blobfuse cache

vibhansa-msft:
As of now we either we have direct-io mode where both kernel cache and blobfuse cache are disabled or you have all cache enabled. In AKS mode, few customers want to have blobfuse file level cache but no kernel cache (as kernel page cache can not be controlled as in when to wipe it out). With this change adding a new flag which will internally be treated as direct-io for libfuse component but both file-cache and attr-cache can still be controlled with independent timeouts.

### DIFF
--- a/component/libfuse/libfuse.go
+++ b/component/libfuse/libfuse.go
@@ -75,6 +75,7 @@ type Libfuse struct {
 	maxFuseThreads        uint32
 	directIO              bool
 	umask                 uint32
+	noKernelCache         bool
 }
 
 // To support pagination in readdir calls this structure holds a block of items for a given directory
@@ -194,6 +195,11 @@ func (lf *Libfuse) Validate(opt *LibfuseOptions) error {
 	lf.ownerGID = opt.Gid
 	lf.ownerUID = opt.Uid
 	lf.umask = opt.Umask
+
+	if lf.noKernelCache {
+		opt.DirectIO = true
+		lf.directIO = true
+	}
 
 	if opt.allowOther {
 		lf.dirPermission = uint(common.DefaultAllowOtherPermissionBits)
@@ -326,6 +332,8 @@ func (lf *Libfuse) Configure(_ bool) error {
 		return err
 	}
 
+	_ = config.UnmarshalKey("no-kernel-cache", &lf.noKernelCache)
+
 	err = lf.Validate(&conf)
 	if err != nil {
 		log.Err("Libfuse::Configure : config error [invalid config settings]")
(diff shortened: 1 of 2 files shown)


component/libfuse/libfuse.go
```diff
@@ -340,8 +348,8 @@ func (lf *Libfuse) Configure(_ bool) error {
...
-		lf.readOnly, lf.allowOther, lf.allowRoot, lf.filePermission, lf.entryExpiration, lf.attributeExpiration, lf.negativeTimeout, lf.ignoreOpenFlags, lf.nonEmptyMount, lf.directIO, lf.maxFuseThreads, lf.traceEnable, lf.extensionPath, lf.disableWritebackCache, lf.dirPermission, lf.mountPath, lf.umask)
+	log.Crit("Libfuse::Configure : read-only %t, allow-other %t, allow-root %t, default-perm %d, entry-timeout %d, attr-time %d, negative-timeout %d, ignore-open-flags %t, nonempty %t, direct_io %t, max-fuse-threads %d, fuse-trace %t, extension %s, disable-writeback-cache %t, dirPermission %v, mountPath %v, umask %v, kernel_cache %v",
+		lf.readOnly, lf.allowOther, lf.allowRoot, lf.filePermission, lf.entryExpiration, lf.attributeExpiration, lf.negativeTimeout, lf.ignoreOpenFlags, lf.nonEmptyMount, lf.directIO, lf.maxFuseThreads, lf.traceEnable, lf.extensionPath, lf.disableWritebackCache, lf.dirPermission, lf.mountPath, lf.umask, lf.noKernelCache)
```
anu-avocado: We should show no_kernel_cache as true in the configure log as showing kernel_Cache true causes confusion

component/libfuse/libfuse.go
```diff
@@ -195,6 +196,11 @@ func (lf *Libfuse) Validate(opt *LibfuseOptions) error {
 	lf.ownerUID = opt.Uid
 	lf.umask = opt.Umask
 
+	if lf.noKernelCache {
```
anu-avocado: We should update the log message in line 239 to 
log.Crit("Libfuse::Validate : DirectIO or noKernelCache enabled, setting fuse timeouts to 0")

component/libfuse/libfuse.go
anu-avocado: Also, in this configure log we are showing direct-io true and nokernelcache true which might be confusing for us to debug issues might be better to show direct io false in the log message unless specified by cx. Also, cx might question why we are showing direct-io true. What are your thoughts about this?

anu-avocado:
Please add a test for to validate config behaviour for the new flag and mention the flag in advancedConfig, thanks!

component/libfuse/libfuse.go
bhansalivikas: if kernel-cache is true then we can assume direct-io came through this, in either case both flag means the same for libfuse

setup/advancedConfig.yaml
```diff
@@ -67,7 +68,7 @@ libfuse:
   negative-entry-expiration-sec: <time kernel can cache attributes of non existent paths (in sec). Default - 120 sec>
   fuse-trace: true|false <enable libfuse api trace logs for debugging>
   extension: <physical path to extension library>
-  direct-io: true|false <enable to bypass the kernel cache>
+  direct-io: true|false <enable to bypass the kernel cache. Also disable blobfuse2 data and metadata cache>
```
syeleti-msft:
```suggestion
  direct-io: true|false <enable to bypass the kernel cache. It also disables blobfuse2 data and metadata caching. Default - false>
```

setup/advancedConfig.yaml
```diff
@@ -34,6 +34,7 @@ foreground: true|false <run blobfuse2 in foreground or background>
 # Common configurations
 allow-other: true|false <allow other users to access the mounted directory - used for FUSE and File Cache>
 nonempty: true|false <allow mounting on non-empty directory>
+disable-kernel-cache: true|false <disable kernel cache but keep blobfuse2 cache>
```
syeleti-msft:
```suggestion
disable-kernel-cache: true|false <disable kernel cache but keep blobfuse2 cache. Default - false>
```

syeleti-msft (approved):

PR #1886, opened 2025-07-21
Do not cleanup active MV directories when a node joins a cluster

linuxsmiths:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

## Checklist

## Related Links
- [Issues](<link>)


internal/dcache/cluster_manager/cluster_manager.go
```diff
@@ -598,12 +598,12 @@ func getMyRVsInClustermap(myRVs []dcache.RawVolume) map[string]dcache.RawVolume
...
 }
 
-// Cleanup the given RV's directory. If doNotDeleteMVs is nil all MVs are deleted else those MVs
+// Cleanup the given RV's directory. If doNotDeleteMVs is nil, all MVs are deleted else those MVs
 // are skipped and everything else is deleted.
 // It returns failure if it fails to delete even a single matching MV. This is to ensure that
 // we prevent such a node from joining the cluster.
 //
-// TODO: Once we have sufficient runin we can let it join the cluster even on partial cleanup.
+// TODO: Once we have sufficient runs we can let it join the cluster even on partial cleanup.
```
linuxsmiths: it was meant to be runin :-)
We can call it run-in

internal/dcache/rpc/client/client_pool.go
```diff
@@ -150,6 +150,7 @@ func (cp *clientPool) getRPCClient(nodeID string) (*rpcClient, error) {
 			ncPool.lastUsed.Store(time.Now().Unix())
 			common.Assert(client.nodeID == nodeID, client.nodeID, nodeID)
 			ncPool.numActive.Add(1)
+			log.Debug("clientPool::getRPCClient: Successfully reterived RPC client for node %s", nodeID)
```
linuxsmiths: is this adding any more debuggability? else we should avoid any more logging for the fast path.
also typo in retrieved

jainakanksha-msft: Yes, till the time we fix the RPCClient lock item, we need this for more clear metrics.
We will remove this as soon as we place a fix in this path.

internal/dcache/clustermap/clustermap.go
```diff
@@ -449,6 +459,46 @@ func (c *ClusterMap) getOfflineMVs() map[string]dcache.MirroredVolume {
...
+
+	rv, ok := localMap.RVMap[rvName]
+	if !ok {
+		common.Assert(false, rvName)
+		return nil
+	}
+
+	rvIsOffline := (rv.State == dcache.StateOffline)
+
+	activeMVs := make(map[string]struct{})
```
souravgupta-msft: If RV is offline in the clustermap, we can directly return nil here so that we don't loop over the MVs below.

internal/dcache/clustermap/clustermap.go
```diff
@@ -449,6 +459,46 @@ func (c *ClusterMap) getOfflineMVs() map[string]dcache.MirroredVolume {
...
+	for mvName, mv := range localMap.MVMap {
+		// RV is offline in the RV list, so it cannot have any active MVs.
+		if rvIsOffline {
+			continue
+		}
```
souravgupta-msft: This won't be needed if we return nil if the RV is offline in clustermap.

internal/dcache/clustermap/clustermap.go
```diff
@@ -449,6 +459,46 @@ func (c *ClusterMap) getOfflineMVs() map[string]dcache.MirroredVolume {
...
+
+		// If the component RV is offline/inband-offline, then the MV is not active.
+		if rvState == dcache.StateOffline || rvState == dcache.StateInbandOffline {
+			continue
+		}
+
+		// Else, it's an active MV for this RV.
+		activeMVs[mvName] = struct{}{}
```
souravgupta-msft: Should we only add the MV to the map only when the RV's state in the MV is online. We should avoid outofsync and syncing states also here.

internal/dcache/rpc/client/client_pool.go
linuxsmiths: @souravgupta-msft, fyi.

PR #1893, opened 2025-07-28
Add FIO tests to check the data integrity

syeleti-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Other (describe): Add FIO workloads

## Description

** Add fio workloads to check the data integrity of the file cache and block cache.
**  Making the stage not run by default.


azure-pipeline-templates/fio-data-integrity.yml
```diff
@@ -0,0 +1,119 @@
...
+      sudo apt-get install fio -y
+    displayName: 'Install Utils'
+
+# Run fio on 1 file 10G size SEQ read,write
+  - script: |
+      ls $(MOUNT_DIR)
+      rm -rf ./*
+      echo "******FIO JOB STARTING******"
+      fio --rw=readwrite --bs=1M --size=10g --numjobs=1 $(WORK_DIR)/test/fio/rw.fio
+    displayName: 'SEQ-[RW]-10G-1F-1T'
```
jainakanksha-msft, line 64: What is 1T in display name?

test/fio/rw.fio
```diff
@@ -0,0 +1,8 @@
+[writers]
+bs=1M
+iodepth=5
+ioengine=sync
```
jainakanksha-msft, line 3: IF ioengine is sync, having iodepth as 5 will not have any impact because 1 job have to wait for anotherjob to complete.

azure-pipeline-templates/fio-data-integrity.yml
```diff
@@ -0,0 +1,119 @@
...
+          $(WORK_DIR)/blobfuse2 mount $(MOUNT_DIR) --config-file=${{ parameters.config_file }} --default-working-dir=$(WORK_DIR) --file-cache-timeout=3200
+
+  - script:
+      sudo apt-get install fio -y
+    displayName: 'Install Utils'
+
+# Run fio on 1 file 10G size SEQ read,write
+  - script: |
+      ls $(MOUNT_DIR)
+      rm -rf ./*
```
jainakanksha-msft, line 61: Are we expecting to remove content from ./* or from MOUNT_DIR?

syeleti-msft: cwd for the script block is MOUNT_DIR, hence both are same

azure-pipeline-templates/fio-data-integrity.yml
syeleti-msft: thread count

azure-pipeline-templates/fio-data-integrity.yml
syeleti-msft: will change this as it looks confusing

test/fio/rw.fio
syeleti-msft: true, removed

blobfuse2-nightly.yaml
```diff
@@ -1069,6 +1074,83 @@ stages:
...
+          timeoutInMinutes: 60
+          strategy:
+            matrix:
+              Ubuntu-20:
+                AgentName: 'blobfuse-ubuntu20'
+                distro: 'ubuntu'
+                distro_version: 'ubuntu_20_x86'
+                fuselib: 'libfuse-dev'
+                tags: 'fuse2'
+              Ubuntu-22:
```
jainakanksha-msft, line 1092: Should we also add testing for Ubuntu-24

azure-pipeline-templates/fio-data-integrity.yml
```diff
@@ -0,0 +1,119 @@
...
+# Run fio on 10 files 1G size SEQ read, write
+  - script: |
+      ls -lrt
+      rm -rf ./*
+      echo "******FIO JOB STARTING*******"
+      fio --rw=readwrite --bs=1M --size=1g --numjobs=10 $(WORK_DIR)/test/fio/rw.fio
+    displayName: 'SEQ-[RW]-1G-10F-10T'
+    workingDirectory: $(MOUNT_DIR)
+
+# Run fio on 1 file 10GB size RAND read, write
```
jainakanksha-msft, line 76: Should we also add random read tests, after random writes, to validate cache read paths and data integrity

blobfuse2-nightly.yaml
syeleti-msft: we didn't add anything specific option in fuse that would only work on Ubuntu-24 afaik. though we can add the distro. anything specific are we looking here?

jainakanksha-msft: Here we were testing only for 20 and 22, that's why I asked to also add 24 distro.

syeleti-msft: will add them later.

azure-pipeline-templates/fio-data-integrity.yml
syeleti-msft: will add them in the next pr. when these tests would actually integrate in the daily run in the nightly pipeline

azure-pipeline-templates/fio-data-integrity.yml
```diff
@@ -0,0 +1,119 @@
...
+        VERBOSE_LOG: ${{ parameters.verbose_log }}
+
+
+  - template: 'mount.yml'
+    parameters:
+      prefix: ${{ parameters.cache_mode }}
+      mountStep:
+        script: |
+          $(WORK_DIR)/blobfuse2 mount $(MOUNT_DIR) --config-file=${{ parameters.config_file }} --default-working-dir=$(WORK_DIR) --file-cache-timeout=3200
+
```
anu-avocado, line 53: is there a reason for hardcoding file cache timeout and this value

anu-avocado: also this mount command works for block-cache also although we give a  file-cache flag which is technically wrong but not sure we can change it now as it will break existing customers. Just something to be aware of

bhansalivikas: for block-cache it will work only if you give '--block-cache' cli parameter. Without any explicit option its file-cache only. Also, in case of block-cache this is only applicable for persistance where disk is used to store the blocks and hence the name.

bhansalivikas: @syeleti-msft any particular reason for the timeout that Anubhuti has pointed out?

ashruti-msft (approved):

PR #1898, opened 2025-07-30
Removed duplicate step From Sec Pack Installation

jainakanksha-msft:


setup/vmSetupAzSecPack.sh
```diff
@@ -4,38 +4,6 @@
...
-sudo apt-get update -y
-
-# Install required packages
-sudo apt-get install apt-transport-https ca-certificates curl gnupg lsb-release -y
-
-# Create directory for Microsoft GPG key
-sudo mkdir -p /etc/apt/keyrings
-
-# Download and install Microsoft GPG key
-curl -sLS https://packages.microsoft.com/keys/microsoft.asc | gpg --dearmor | sudo tee /etc/apt/keyrings/microsoft.gpg > /dev/null
```
bhansalivikas, line 17: why we do not need these steps ?

jainakanksha-msft: These steps are already covered by curl -sL https://aka.ms/InstallAzureCLIDeb | sudo bash
So, don't require these steps again.
Reference document-
https://learn.microsoft.com/en-us/cli/azure/install-azure-cli-linux?view=azure-cli-latest&pivots=apt#install-azure-cli

vibhansa-msft (approved):

PR #1899, opened 2025-07-30
Timeout for creating new RPC client and doing RW operations

souravgupta-msft:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description
Added timeouts for RPC client creation and doing read/write operations.


internal/dcache/rpc/client/client.go
```diff
@@ -35,12 +35,21 @@ package rpc_client
...
 	"github.com/Azure/azure-storage-fuse/v2/internal/dcache/rpc/gen-go/dcache/service"
 	"github.com/apache/thrift/lib/go/thrift"
 )
 
+const (
+	// defaultConnectionTimeout is the default timeout for establishing a new connection to the node.
+	defaultConnectionTimeout = 1 * time.Second
+
+	// defaultSocketTimeout is the default read/write timeout for the underlying socket.
+	defaultSocketTimeout = 2 * time.Second
```
linuxsmiths: let's set defaultConnectionTimeout  to 10 secs and defaultSocketTimeout  to 20 secs.
add a comment explaining that we don't want to keep these very low to be resilient to occasional packet drops

internal/dcache/rpc/client/client_pool.go
```diff
@@ -48,11 +48,31 @@ import (
...
+
+	clients map[string]*nodeClientPool // map of node ID to rpc node client pool
+
+	//
+	// Map of node ID to which this node cannot create RPC clients. The value of the map is
+	// time when the new RPC client creation to the node was last attempted.
+	// This is used to prevent creating new RPC clients to the node by different threads till
+	// the negative RPC client creation timeout expires.
+	//
+	negativeClients map[string]time.Time
```
linuxsmiths: negativeNodes

internal/dcache/rpc/client/client_pool.go
```diff
@@ -64,21 +84,71 @@ type clientPool struct {
...
+		if ok {
+			timeElapsed := int64(time.Since(t).Seconds())
+			if timeElapsed < defaultNegativeTimeout {
+				err := fmt.Errorf("not creating RPC clients for node %s, negative timeout not expired yet (%d seconds elapsed, %d seconds timeout)",
+					nodeID, timeElapsed, defaultNegativeTimeout)
+				log.Err("clientPool::getNodeClientPool: %v", err)
+				return nil, err
+			} else {
+				log.Debug("clientPool::getNodeClientPool: Negative timeout expired for node %s, removing from negative clients map",
+					nodeID)
```
linuxsmiths: log timeElapsed and defaultNegativeTimeout

internal/dcache/rpc/client/client_pool.go
```diff
@@ -115,18 +192,16 @@ func (cp *clientPool) getNodeClientPool(nodeID string) (*nodeClientPool, error)
 // If the pool doesn't have any free client, it waits for 60secs for a client to become available and returns as
 // soon as an RPC client is released and added to the pool. If no client becomes available for 60secs, it
 // indicates some bug and it panics the program.
-//
-// Caller MUST NOT hold the clientPool lock.
```
linuxsmiths: revert

internal/dcache/rpc/client/client_pool.go
```diff
@@ -213,8 +290,8 @@ func (cp *clientPool) getRPCClientNoWait(nodeID string) (*rpcClient, error) {
 func (cp *clientPool) releaseRPCClient(client *rpcClient) error {
```
linuxsmiths, line 296: add 
// Note: Caller MUST NOT hold the clientPool lock.

internal/dcache/rpc/client/client_pool.go
```diff
@@ -259,6 +336,46 @@ func (cp *clientPool) closeRPCClient(client *rpcClient) error {
 	return nil
 }
 
+// deleteRPCClient deletes an RPC client from the pool.
+// It first closes the client and then removes it from the pool.
+// This is used when the client is no longer needed or when the node is down and we want to
```
linuxsmiths: or -> e.g.,

also add to the comment that if there are no active connections and no connections in the channel, it delete the node client pool

internal/dcache/rpc/client/functions.go
```diff
@@ -91,6 +91,12 @@ const (
 	// defaultTimeout is the default duration in seconds after which an idle RPC client is closed.
 	//
 	defaultTimeout = 60
+
+	//
+	// defaultNegativeTimeout is the default duration in seconds after which the node ID is deleted
+	// from the negative clients map and its RPC client creation is attempted again.
+	//
+	defaultNegativeTimeout = 60
```
linuxsmiths: 30 seconds
we want to keep it large enough so that clients trying to connect in close proximity make use of this information but small enough to not attempt connection to a node that might have now come up

internal/dcache/rpc/client/functions.go
```diff
@@ -148,6 +154,19 @@ func Hello(ctx context.Context, targetNodeID string, req *models.HelloRequest) (
 
 				// Retry Hello once more with fresh connection.
 				continue
+			} else if rpc.IsTimedOut(err) {
+				//
+				// If the error is due to a timeout (either node is down or cannot be reached over the n/w),
+				// we delete the RPC client as recreating it will most likely fail with the same error.
+				//
+				err1 := cp.deleteRPCClient(client)
```
linuxsmiths: deleteAllRPCClients

internal/dcache/rpc/client/client.go
souravgupta-msft: Done

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Updated

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Added

internal/dcache/rpc/client/client_pool.go

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Updated

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Added deleteAllRPCClients() method which will take care of deleting the node client pool if there are no active connections and no connections in the channel

internal/dcache/rpc/client/functions.go

internal/dcache/rpc/client/functions.go
souravgupta-msft: Updated, please review

internal/dcache/rpc/client/client_pool.go
```diff
@@ -48,11 +48,31 @@ import (
...
 type clientPool struct {
-	mu         sync.Mutex
-	clients    map[string]*nodeClientPool // map of node ID to rpc node client pool
-	maxPerNode uint32                     // Maximum number of open RPC clients per node
-	maxNodes   uint32                     // Maximum number of nodes for which RPC clients are open
-	timeout    uint32                     // Duration in seconds after which a RPC client is closed
+	//
+	// Mutex for the client pool to ensure that clients map is updated
+	// (new node connections are added/removed) in a thread-safe manner.
+	//
```
linuxsmiths, line 55: Also protects the negativeNodes map.

internal/dcache/rpc/client/client_pool.go
```diff
@@ -64,21 +84,71 @@ type clientPool struct {
...
+		t, ok := cp.negativeNodes[nodeID]
+		if ok {
+			timeElapsed := int64(time.Since(t).Seconds())
+			if timeElapsed < defaultNegativeTimeout {
+				err := fmt.Errorf("not creating RPC clients for node %s, negative timeout not expired yet (%d seconds elapsed, %d seconds timeout)",
+					nodeID, timeElapsed, defaultNegativeTimeout)
+				log.Err("clientPool::getNodeClientPool: %v", err)
+				return nil, err
+			} else {
+				log.Debug("clientPool::getNodeClientPool: Negative timeout expired for node %s, removing from negative clients map (%d seconds elapsed, %d seconds timeout)",
```
linuxsmiths: removing from negative clients map -> removing from negative **nodes** map

internal/dcache/rpc/client/client_pool.go
```diff
@@ -99,6 +169,13 @@ func (cp *clientPool) getNodeClientPool(nodeID string) (*nodeClientPool, error)
 		err := ncPool.createRPCClients(cp.maxPerNode)
 		if err != nil {
 			log.Err("clientPool::getNodeClientPool: createRPCClients(%s) failed: %v", nodeID, err)
+
+			//
+			// Add to negativeClients map to prevent creating new RPC clients to the node ID by other
+			// threads till the negative timeout expires.
```
linuxsmiths: add,
Note that createRPCClients() failure indicates some transport problem or the node/blobfuse is down.

internal/dcache/rpc/client/client_pool.go
```diff
@@ -259,6 +340,136 @@ func (cp *clientPool) closeRPCClient(client *rpcClient) error {
...
+
+	// client is allocated from the pool, so pool must exist.
+	common.Assert(exists)
+	common.Assert(ncPool.nodeID == client.nodeID, ncPool.nodeID, client.nodeID)
+
+	//
+	// Clients present in the pool. If it's less than cp.maxPerNode, rest are currently allocated to other
+	// callers. We cannot replenish those. Those will be deleted by their respective caller when their RPC
+	// requests fail with "timeout" error.
+	//
```
linuxsmiths:
```
	//
	// Clients present in the pool. The one that we are deleting is not in the pool so the pool can have
	// max cp.maxPerNode-1 clients. If it's less than cp.maxPerNode-1, rest are currently allocated to other
	// callers. We cannot replenish those. Those will be deleted by their respective caller when their RPC
	// requests fail with "timeout" error.
	//
```

please fix the corresponding comment for reset clients case

internal/dcache/rpc/client/client_pool.go
```diff
@@ -259,6 +340,136 @@ func (cp *clientPool) closeRPCClient(client *rpcClient) error {
...
+	numClients := len(ncPool.clientChan)
+	common.Assert(numClients < int(cp.maxPerNode), numClients, cp.maxPerNode)
+
+	//
+	// Delete this client. This closes this client and removes it from the pool.
+	// It can only fail if Thrift fails to close the connection.
```
linuxsmiths, line 419: Add
This should technically not happen, so we assert.

internal/dcache/rpc/client/client_pool.go
```diff
@@ -259,6 +340,136 @@ func (cp *clientPool) closeRPCClient(client *rpcClient) error {
...
+func (cp *clientPool) deleteRPCClient(client *rpcClient) error {
+	log.Debug("clientPool::deleteRPCClient: Deleting RPC client to %s node %s",
+		client.nodeAddress, client.nodeID)
+
+	common.Assert(cp.isLockHeld())
+
+	// Close the client first.
+	err := cp.closeRPCClient(client)
+	if err != nil {
+		return err
```
linuxsmiths, line 362: let's add an assert here

```
// We don't expect connection close to fail, let's know if it happens.
common.Assert(false, err)
```

internal/dcache/rpc/client/client_pool.go
```diff
@@ -259,6 +340,136 @@ func (cp *clientPool) closeRPCClient(client *rpcClient) error {
...
+		// have the clientPool lock.
+		//
+		common.Assert(err == nil, err)
+
+		err = cp.deleteRPCClient(client)
+		if err != nil {
+			//
+			// We have deleted at least one connection, so we don't fail the deleteAllRPCClients()
+			// call, log an error and proceed. This way even if we are not able to create new
+			// good connections, we drain and close all stale connections.
```
linuxsmiths: the last sentence doesn't apply to the close, remove it

internal/dcache/rpc/client/client_pool.go
```diff
@@ -259,6 +340,136 @@ func (cp *clientPool) closeRPCClient(client *rpcClient) error {
...
+			//
+			log.Err("clientPool::deleteAllRPCClients: Failed to delete RPC client to %s node %s: %v",
+				client.nodeAddress, client.nodeID, err)
+			//
+			// Connection refused and timeout are the only viable errors.
+			// Assert to know if anything else happens.
+			//
+			common.Assert(rpc.IsConnectionRefused(err) || rpc.IsTimedOut(err), err)
```
linuxsmiths: this is not true for close.
As I had mentioned above, it should not happen, so we assert for this

internal/dcache/rpc/client/client_pool.go
```diff
@@ -259,6 +340,136 @@ func (cp *clientPool) closeRPCClient(client *rpcClient) error {
...
+		} else {
+			numConnDeleted++
+		}
+	}
+
+	// We must have deleted at least the client we are called for (and maybe more).
+	common.Assert(numConnDeleted > 0)
```
linuxsmiths, line 466: move this assert after the log, and also add the following assert

```
// We don't expect failure closing any client connection, so there shouldn`t be any client left in the pool.
common.Assert(len(ncPool.clientChan) == 0, len(ncPool.clientChan), client.nodeAddress, client.nodeID)
```

internal/dcache/rpc/client/client_pool.go
```diff
@@ -259,6 +340,136 @@ func (cp *clientPool) closeRPCClient(client *rpcClient) error {
 	return nil
 }
 
+// deleteRPCClient deletes an RPC client from the pool.
+// It first closes the client and then removes it from the pool.
+// This is used when the client is no longer needed, e.g. when the node is down and we want to
+// remove all clients to the node.
+//
```
linuxsmiths, line 349: Since you have moved deleteNodeClientPoolIfInactive() to deleteAllRPCClients(), it's actually not safe to all this function outside of deleteAllRPCClients(), let's warn the user of that.

Note: Don't call this function outside deleteAllRPCClients().

internal/dcache/rpc/client/functions.go
```diff
@@ -91,6 +91,14 @@ const (
 	// defaultTimeout is the default duration in seconds after which an idle RPC client is closed.
 	//
 	defaultTimeout = 60
+
+	//
+	// defaultNegativeTimeout is the default duration in seconds after which the node ID is deleted
+	// from the negative clients map and its RPC client creation is attempted again.
+	// We want to keep it large enough so that clients trying to connect in close proximity make use of
+	// this information but small enough to not attempt connection to a node that might have now come up.
```
linuxsmiths:
```
    // We want to keep it large enough so that clients trying to connect in close proximity make use of
	// this information and not attempt to connect to an unreachable node, potentially delaying the
	// caller, but small enough to promptly attempt connection to a node that might have now come up.
```

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Updated

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Updated

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Added

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Updated

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Done

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Done

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Updated

internal/dcache/rpc/client/client_pool.go

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Done

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Done

internal/dcache/rpc/client/functions.go

PR #1917, opened 2025-08-11
Update lock in client pool

souravgupta-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Other (describe):

## Description
Update locking strategy in client pool. Instead of using global lock at the client pool level, we will now use lock at the node level to perform operations like create/get/release/delete RPC clients. This will ensure that multiple threads can do these operations on different nodes simultaneously. Whereas only thread can do this for a node at a time.

### DIFF
--- a/component/libfuse/libfuse.go
+++ b/component/libfuse/libfuse.go
@@ -250,12 +250,26 @@ func (lf *Libfuse) Validate(opt *LibfuseOptions) error {
 		}
 	*/
 
-	if lf.directIO {
-		lf.negativeTimeout = 0
-		lf.attributeExpiration = 0
-		lf.entryExpiration = 0
-		log.Crit("Libfuse::Validate : DirectIO enabled, setting fuse timeouts to 0")
-	}
+	//
+	// For now I'm turning off kernel attribute caching altogether, as it causes issues with
+	// new files getting created. Kernel caches the file size as -1 (when the file is created)
+	// and later when the file is completely written and its size updated, the kernel cache is
+	// not invalidated, so the file size is still -1.
+	//
+	// TODO: Later if we want to support kernel attribute caching, we have to duly invalidte the
+	//       kernel cache when the file is written and its size is updated.
+	//
+	/*
+		if lf.directIO {
+			lf.negativeTimeout = 0
+			lf.attributeExpiration = 0
+			lf.entryExpiration = 0
+			log.Crit("Libfuse::Validate : DirectIO enabled, setting fuse timeouts to 0")
+		}
+	*/
+	lf.negativeTimeout = 0
+	lf.attributeExpiration = 0
+	lf.entryExpiration = 0
 
 	if !(config.IsSet(compName+".uid") || config.IsSet(compName+".gid") ||
 		config.IsSet("lfuse.uid") || config.IsSet("lfuse.gid")) {
--- a/component/libfuse/libfuse_handler.go
+++ b/component/libfuse/libfuse_handler.go
@@ -229,8 +229,11 @@ func populateFuseArgs(opts *C.fuse_options_t, args *C.fuse_args_t) (*C.fuse_opti
 	// We want to keep as few fuse threads as possible, just large enough to handle parallel requests
 	// fairly well. Fuse has a default of 10 which should be ok but we set it to 8 to be explicit about
 	// our intent to use a small number of threads.
+	// Commenting out max_threads for now as some systems still has older libfuse which does not support
+	// this option.
+	// TODO: Make it conditional based on libfuse version.
 	//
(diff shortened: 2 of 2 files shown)


internal/dcache/rpc/client/client_pool.go
```diff
@@ -195,17 +297,33 @@ func (cp *clientPool) getNodeClientPool(nodeID string) (*nodeClientPool, error)
...
+	cp.acquireRWMutexReadLock()
+	defer cp.releaseRWMutexReadLock()
+
+	//
+	// Get lock for the given node ID. This is done so that multiple threads can create RPC
+	// clients for the different node IDs concurrently. Whereas only one thread can create RPC
+	// clients for the same node ID at a time.
+	//
+	nodeLock := cp.nodeLock.Get(nodeID)
+	nodeLock.Lock()
```
linuxsmiths: better to also wrap this inside
acquireNodeLock
releaseNodeLock
isNodeLocked

more readable code, better asserting, better commenting, and we can add useful debug logs

internal/dcache/rpc/client/client_pool.go
```diff
@@ -645,49 +806,76 @@ func (cp *clientPool) resetAllRPCClients(client *rpcClient) error {
...
 
+	//
+	// Take lock on the LRU node ID to ensure that other threads cannot create/get RPC clients
+	// for this node while we are closing and deleting its node client pool.
+	//
+	lruNodeLock := cp.nodeLock.Get(lruNodeID)
+	lruNodeLock.Lock()
+	defer lruNodeLock.Unlock()
+
 	err := lruNcPool.closeRPCClients()
```
linuxsmiths: after we decide to release this node client pool. following the successful check al L842 and before we acquire the node lock, some thread can call getRPCCLient() and get an rpc client to the node. In that case we should not delete the node client pool.

internal/dcache/rpc/client/client_pool.go
```diff
@@ -700,40 +976,55 @@ func (cp *clientPool) closeInactiveNodeClientPools() {
...
-	for key, ncPool := range cp.clients {
-		err := ncPool.closeRPCClients()
+	//
+	// Acquire write lock on the client pool to ensure that no other thread is accessing
+	// the client pool while we are closing all node client pools.
+	// This also waits till the read locks are released by the client pool operations
+	// like getRPCClient(), releaseRPCClient(), deleteAllRPCClients(), resetAllRPCClients(), etc.
+	//
+	cp.acquireRWMutexWriteLock()
+	defer cp.releaseRWMutexWriteLock()
```
linuxsmiths, line 996: It'll be better to have closeAllNodeClientPools() wait for any allocated RPC clients to be released.
For that it should check that ncPool.numActive should be 0 before proceeding to close the connections from the channel. If not 0, it should sleep for a second and then try again. While sleeping release the write lock so that relaseRPCClient() can release clients.

linuxsmiths (approved):

internal/dcache/rpc/client/client_pool.go
```diff
@@ -995,11 +995,43 @@ func (cp *clientPool) closeAllNodeClientPools() error {
...
+
+			// acquire write lock again and recheck if the active clients are released for this node
+			cp.acquireRWMutexWriteLock()
+		}
+
+		//
+		// Even after waiting for 60 seconds, if the active clients are not released back, return error to the caller.
+		//
+		if ncPool.numActive.Load() > 0 {
+			err = fmt.Errorf("Node %s has %d active clients (%d, %d), cannot close even after waiting for %v",
```
linuxsmiths, line 1029: we can set err at L1037 too.
also at L987, can we add the number of node client pools

	log.Debug("clientPool::closeAllNodeClientPools: Closing all %d node client pools")

souravgupta-msft: Updated

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Done, updated

internal/dcache/rpc/client/client_pool.go
```diff
@@ -1034,7 +1034,8 @@ func (cp *clientPool) closeAllNodeClientPools() error {
 
 		err = ncPool.closeRPCClients()
 		if err != nil {
-			log.Err("clientPool::closeAllNodeClientPools: Failed to close node client pool for node %s [%v]", key, err.Error())
+			err = fmt.Errorf("Failed to close RPC clients for node %s [%v]", key, err.Error())
```
linuxsmiths: nit, we can use nodeID  instead of key and err inplace if err.Error()

PR #1924, opened 2025-08-19
Write MV internal

souravgupta-msft:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description
In case of bad connection between two nodes, the PutChunkDC fails with BrokenChain error for some RVs. We cannot tell for sure which node has connection issue. So, retrying the WriteMV operation with OriginatorSendsToAll is safe here.

### DIFF
--- a/internal/dcache/replication_manager/replication_manager.go
+++ b/internal/dcache/replication_manager/replication_manager.go
@@ -351,52 +351,28 @@ retry:
 	return resp, nil
 }
 
-func WriteMV(req *WriteMvRequest) (*WriteMvResponse, error) {
-	common.Assert(req != nil)
-
-	if common.IsDebugBuild() {
-		startTime := time.Now()
-		defer func() {
-			timeTaken := time.Since(startTime).Microseconds()
-			log.Debug("ReplicationManager::WriteMV: WriteMV request took %d microseconds: %v", timeTaken, req.toString())
-		}()
-	}
-
-	log.Debug("ReplicationManager::WriteMV: Received WriteMV request (%v): %v", PutChunkStyle, req.toString())
-
-	//
-	// We don't expect the caller to pass invalid requests, so only verify in debug builds.
-	//
-	if common.IsDebugBuild() {
-		if err := req.isValid(); err != nil {
-			err = fmt.Errorf("invalid WriteMV request %s [%v]", req.toString(), err)
-			log.Err("ReplicationManager::WriteMV: %v", err)
-			common.Assert(false, err)
-			return nil, err
-		}
-	}
+func writeMVInternal(req *WriteMvRequest, putChunkStyle PutChunkStyleEnum) (*WriteMvResponse, error) {
+	log.Debug("ReplicationManager::WriteMV: Received WriteMV request (%v): %v", putChunkStyle, req.toString())
 
 	var rvsWritten []string
 	retryCnt := 0
 
 	//
-	// Flag to check if we have a BrokenChain error in the PutChunkDC response.
-	// If we have a BrokenChain error for an RV, it means that the PutChunkDC request was not
-	// forwarded to it as the nexthop RV was down/offline. We will get ThriftError for the nexthop RV
-	// and BrokenChain error for the subsequent RVs.
-	// In case of BrokenChain error, we retry the WriteMV() operation skipping the nexthop RV and refreshing
-	// the clustermap (if needed).
+	// If the putChunkStyle is OriginatorSendsToAll, it means that we are retrying after BrokenChain
+	// error in the previous attempt using DaisyChain mode.
 	//
-	brokenChain := false
+	if putChunkStyle == OriginatorSendsToAll {
+		retryCnt = 1
+	}
 
 	// TODO: TODO: hash validation will be done later
 	// get hash of the data in the request
 	// hash := getMD5Sum(req.Data)
 
 retry:
-	if retryCnt > 0 || brokenChain {
-		log.Info("ReplicationManager::WriteMV: [%d] Retrying WriteMV %v after clustermap refresh or broken chain (%v), RVs written in prev attempt: %v",
-			retryCnt, req.toString(), brokenChain, rvsWritten)
+	if retryCnt > 0 {
+		log.Info("ReplicationManager::WriteMV: [%d] Retrying WriteMV %v after clustermap refresh or due to OriginatorSendsToAll mode (%v), RVs written in prev attempt: %v",
+			retryCnt, req.toString(), putChunkStyle == OriginatorSendsToAll, rvsWritten)
 	}
 
 	//
@@ -451,7 +427,7 @@ retry:
 			Length:         int64(len(req.Data)),
 			SyncID:         "", // this is regular client write
 			ComponentRV:    componentRVs,
-			MaybeOverwrite: retryCnt > 0 || brokenChain,
+			MaybeOverwrite: retryCnt > 0,
 		},
 		NextRVs: make([]string, 0), // will be added later down, if needed
 	}
@@ -549,7 +525,7 @@ retry:
 	// Use PutChunk to write if PutChunkStyle is OriginatorSendsToAll or we have only the nexthop
 	// RV to send the request to.
 	//
-	if PutChunkStyle == OriginatorSendsToAll || len(putChunkDCReq.NextRVs) == 0 {
+	if putChunkStyle == OriginatorSendsToAll || len(putChunkDCReq.NextRVs) == 0 {
 		// TODO: Add rvName to Address to avoid potentially expensive search for RV name.
 		rvName := getRvNameFromRvID(putChunkDCReq.Request.Chunk.Address.RvID)
 
@@ -600,7 +576,7 @@ retry:
 				Length:         int64(len(req.Data)),
 				SyncID:         "", // this is regular client write
 				ComponentRV:    componentRVs,
-				MaybeOverwrite: retryCnt > 0 || brokenChain,
+				MaybeOverwrite: true, // in OriginatorSendsToAll we may be retrying to some nodes
 			}
 
 			log.Debug("ReplicationManager::WriteMV: Sending PutChunk request for %s/%s to node %s: %s",
@@ -615,7 +591,7 @@ retry:
 				respChannel:  responseChannel,
 			}, isLastComponentRV /* runInline */)
 		}
-	} else if PutChunkStyle == DaisyChain {
+	} else if putChunkStyle == DaisyChain {
 		rvName := getRvNameFromRvID(putChunkDCReq.Request.Chunk.Address.RvID)
 		targetNodeID := getNodeIDFromRVName(rvName)
 		common.Assert(common.IsValidUUID(targetNodeID))
@@ -665,7 +641,7 @@ retry:
 		//
 		addPutChunkDCResponseToChannel(putChunkDCResp, responseChannel)
 	} else {
-		common.Assert(false, "Unexpected PutChunkStyle", PutChunkStyle)
+		common.Assert(false, "Unexpected PutChunkStyle", putChunkStyle)
 	}
 
 processResponses:
@@ -690,9 +666,14 @@ processResponses:
 	clusterMapRefreshed := false
 
 	//
-	// Reset brokenChain flag to false before processing the responses.
+	// Flag to check if we have a BrokenChain error in the PutChunkDC response.
+	// If we have a BrokenChain error for an RV, it means that the PutChunkDC request was not
+	// forwarded to it as the nexthop RV was down/offline. We will get ThriftError for the nexthop RV
+	// and BrokenChain error for the subsequent RVs.
+	// In case of BrokenChain error, we return error to WriteMV() which retires the operation with
+	// OriginatorSendsToAll mode.
 	//
-	brokenChain = false
+	brokenChain := false
 
 	for i := 0; i < len(componentRVs); i++ {
 		respItem := <-responseChannel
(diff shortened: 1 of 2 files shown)


internal/dcache/clustermap/clustermap.go
```diff
@@ -175,6 +175,11 @@ func NodeIdToIP(nodeId string) string {
 	return clusterMap.nodeIdToIP(nodeId)
 }
 
+// It will return the RV name of the given nodeId as per local cache copy of cluster map.
+func NodeIdToRvName(nodeId string) string {
+	return clusterMap.nodeIdToRvName(nodeId)
+}
+
```
linuxsmiths: since a node can have multiple RVs, we should not use this

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -614,6 +616,18 @@ retry:
 		log.Debug("ReplicationManager::writeMVInternal: Sending PutChunkDC request for nexthop %s/%s to node %s: %s",
 			rvName, req.MvName, targetNodeID, rpc.PutChunkDCRequestToString(putChunkDCReq))
 
+		//
+		// Check if the next-hop RV and the next RVs in chain are present in the iffy RV map.
+		// If yes, we retry the operation using OriginatorSendsToAll.
+		//
```
linuxsmiths: pls explain why we need this here if we have it in functions.go

internal/dcache/replication_manager/replication_manager.go
```diff
@@ -636,6 +650,19 @@ retry:
...
 			common.Assert(putChunkDCResp == nil)
 
+			//
+			// If an RV is marked iffy, it means that either it is down or some other downstream
+			// connection issue is preventing the PutChunkDC call to succeed.
+			// So, if an RV is marked iffy, the RPC client will fail the PutChunkDC() call to prevent
+			// the timeout error from happening again. In this case, we will retry the WriteMV() operation
+			// with OriginatorSendsToAll mode.
+			//
+			if strings.Contains(err.Error(), "iffy") {
```
linuxsmiths: let's define a specific string error and return it as %w from functions.go and check here with Is

internal/dcache/replication_manager/utils.go
```diff
@@ -265,14 +266,25 @@ func addPutChunkDCResponseToChannel(response *models.PutChunkDCResponse, respons
...
+// Given the component RVs list, return the RVs which are marked iffy.
+func getIffyRVs(nextHopRV string, nextRVs []string) []string {
+	iffyRVs := make([]string, 0)
+
+	for _, rv := range append([]string{nextHopRV}, nextRVs...) {
+		common.Assert(cm.IsValidRVName(rv), rv)
+		if rpc_client.IsIffyRV(rv) {
+			iffyRVs = append(iffyRVs, rv)
+		}
+	}
```
linuxsmiths: do not use append, check separately

internal/dcache/replication_manager/utils.go
```diff
@@ -265,14 +266,25 @@ func addPutChunkDCResponseToChannel(response *models.PutChunkDCResponse, respons
 		len(responseChannel), getNumReplicas())
 }
 
-func init() {
-	common.Assert(MAX_SIMUL_SYNC_JOBS < cm.MAX_SIMUL_RV_STATE_UPDATES,
-		MAX_SIMUL_SYNC_JOBS, cm.MAX_SIMUL_RV_STATE_UPDATES)
+// Given the component RVs list, return the RVs which are marked iffy.
+func getIffyRVs(nextHopRV string, nextRVs []string) []string {
+	iffyRVs := make([]string, 0)
```
linuxsmiths: make the initial length len(nextRVs)+1
better to pass params larger than 8 bytes as pointers 
return pointer as well

internal/dcache/replication_manager/utils.go
```diff
@@ -265,14 +266,25 @@ func addPutChunkDCResponseToChannel(response *models.PutChunkDCResponse, respons
...
+	return iffyRVs
 }
 
 // Silence unused import errors for release builds.
 func init() {
 	common.IsValidUUID("00000000-0000-0000-0000-000000000000")
 	log.Info("")
 	fmt.Printf("")
+	common.Assert(MAX_SIMUL_SYNC_JOBS < cm.MAX_SIMUL_RV_STATE_UPDATES,
+		MAX_SIMUL_SYNC_JOBS, cm.MAX_SIMUL_RV_STATE_UPDATES)
```
linuxsmiths: revert

internal/dcache/rpc/client/client.go
```diff
@@ -100,6 +130,19 @@ func newRPCClient(nodeID string, nodeAddress string) (*rpcClient, error) {
...
+
+		//
+		// If the RPC client creation fails due to timeout error, it means some connection problem or
+		// the node is down. In this case we should prevent creating RPC clients to the same node by
+		// other threads, as each operation will fail with the timeout error.
+		// So, add this node ID to negativeNodes map to prevent creating new RPC clients to it by other
+		// threads till the negative timeout expires.
+		//
+		if rpc.IsTimedOut(err) {
+			log.Debug("rpcClient::newRPCClient: Adding node %s at %s to negative nodes map", nodeID, nodeAddress)
```
linuxsmiths: warn log

internal/dcache/rpc/client/client.go
```diff
@@ -73,6 +76,33 @@ var thriftCfg *thrift.TConfiguration
...
+			err := fmt.Errorf("not creating RPC clients for node %s, negative timeout not expired yet (%d seconds elapsed, %d seconds timeout)",
+				nodeID, timeElapsed, defaultNegativeTimeout)
+			log.Err("rpcClient::newRPCClient: %v", err)
+			return nil, err
+		} else {
+			log.Debug("rpcClient::newRPCClient: Negative timeout expired for node %s, removing from negative nodes map (%d seconds elapsed, %d seconds timeout)",
+				nodeID, timeElapsed, defaultNegativeTimeout)
+			cp.negativeNodes.Delete(nodeID)
+		}
+	}
```
linuxsmiths: instead of every caller repeating this, let's have a function which looks up negative node list and does the timeout check

internal/dcache/rpc/client/client.go
linuxsmiths: review that we need to add to negative nodes list in the case of actual commn timeout, not just connection creation timeut

internal/dcache/rpc/client/client_pool.go
```diff
@@ -1091,6 +1112,67 @@ func (cp *clientPool) deleteNodeClientPoolIfInactive(nodeID string) bool {
...
+//     RV is healthy.
+func (cp *clientPool) removeIffyRV(rvName string) {
+	common.Assert(cm.IsValidRVName(rvName), rvName)
+
+	if !cp.isIffyRV(rvName) {
+		return
+	}
+
+	cp.iffyRVMap.Delete(rvName)
+	log.Debug("clientPool::removeIffyRV: removed %s from IffyRVMap", rvName)
```
linuxsmiths: also add
, was added at %s

internal/dcache/rpc/client/client_pool.go
```diff
@@ -105,14 +122,20 @@ type clientPool struct {
 // TODO: Implement timeout support.
 func newClientPool(maxPerNode uint32, maxNodes uint32, timeout uint32) *clientPool {
 	log.Debug("clientPool::newClientPool: Creating new RPC client pool with maxPerNode: %d, maxNodes: %d, timeout: %d", maxPerNode, maxNodes, timeout)
-	return &clientPool{
+	cp := &clientPool{
 		nodeLock:   common.NewLockMap(),
 		maxPerNode: maxPerNode,
 		maxNodes:   maxNodes,
 		timeout:    timeout,
+		iffyTicker: time.NewTicker(defaultIffyRVTimeout),
```
linuxsmiths: this timeout should be much less to avoid delaying the removal too much

internal/dcache/rpc/client/client_pool.go
```diff
@@ -1091,6 +1112,67 @@ func (cp *clientPool) deleteNodeClientPoolIfInactive(nodeID string) bool {
...
+				_ = ok
+				common.Assert(ok, key)
+				common.Assert(cm.IsValidRVName(rvName), rvName)
+
+				addedTime, ok := value.(time.Time)
+				_ = ok
+				common.Assert(ok, addedTime)
+
+				if time.Since(addedTime) > defaultIffyRVTimeout {
+					cp.removeIffyRV(rvName)
```
linuxsmiths: add one more log here so that we know it's remove from here and not the successful rpc case

internal/dcache/rpc/client/functions.go
```diff
@@ -469,13 +516,28 @@ func PutChunkDC(ctx context.Context, targetNodeID string, req *models.PutChunkDC
...
+				}
+
+				//
+				// In PutChunkDC we can get timeout because of bad connection between the downstream nodes,
+				// and not between the client node and target node. In this case, we retry WriteMV using
+				// the OriginatorSendsToAll strategy. So, to be safe we don't delete the connections if we get
+				// timeout error in PutChunkDC. In the PutChunk using OriginatorSendsToAll strategy fails using
+				// timeout, we then delete the connections.
+				//
+				err1 := cp.resetRPCClientInternal(client, true)
```
linuxsmiths: resetall

internal/dcache/rpc/response_error.go
```diff
@@ -215,11 +215,13 @@ func IsTimedOut(err error) bool {
 	//
 	connectionTimedOut := "onnection timed out"
 	ioTimeout := "i/o timeout"
+	timeout := "timeout"
```
linuxsmiths: let's avoid string check in errors that we generate

internal/dcache/clustermap/clustermap.go
souravgupta-msft: Done, removed

internal/dcache/replication_manager/replication_manager.go
souravgupta-msft: Added

internal/dcache/replication_manager/replication_manager.go

internal/dcache/replication_manager/utils.go
souravgupta-msft: Updated

internal/dcache/replication_manager/utils.go
souravgupta-msft: Done

internal/dcache/rpc/client/client.go
souravgupta-msft: Done

internal/dcache/rpc/client/client.go
souravgupta-msft: Done

internal/dcache/rpc/client/client_pool.go

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Done, updated

internal/dcache/rpc/client/client_pool.go
souravgupta-msft: Done, added

internal/dcache/rpc/client/functions.go

internal/dcache/rpc/response_error.go
souravgupta-msft: Updated

internal/dcache/rpc/client/functions.go
```diff
@@ -135,6 +143,30 @@ func Hello(ctx context.Context, targetNodeID string, req *models.HelloRequest) (
...
+				// Assert, but not fail the Hello call.
+				common.Assert(false, err1)
+			}
+
+			err1 = fmt.Errorf("Failing Hello to node %s [%w]: %s",
+				targetNodeID, NegativeNodeError, reqStr)
+
+			log.Err("rpc_client::Hello: %v", err1)
+			return nil, err1
+		}
```
linuxsmiths: move this check to common place in getRPCClient().
Remove from here

internal/dcache/rpc/client/functions.go
```diff
@@ -193,6 +227,12 @@ func Hello(ctx context.Context, targetNodeID string, req *models.HelloRequest) (
 
 			// Fall through to release the RPC client.
 			resp = nil
+		} else {
+			//
+			// The RPC call to the target node succeeded.
+			// So, we can remove it from the negative nodes map.
+			//
+			cp.removeNegativeNode(targetNodeID)
```
linuxsmiths, line 212: todo here and for others, remove from iffy map also

internal/dcache/rpc/client/functions.go
```diff
@@ -437,6 +541,9 @@ func PutChunkDC(ctx context.Context, targetNodeID string, req *models.PutChunkDC
 	log.Debug("rpc_client::PutChunkDC: Sending PutChunkDC request to nexthop node %s and %d daisy chain RV(s): %v",
 		targetNodeID, len(req.NextRVs), reqStr)
 
+	rvName := cm.RvIdToName(req.Request.Chunk.Address.RvID)
+	common.Assert(cm.IsValidRVName(rvName), rvName)
```
linuxsmiths: this call could b expensive for clusters with lot of RVs,
see if we can index by rvid

internal/dcache/rpc/client/functions.go
```diff
@@ -465,24 +534,48 @@ func PutChunkDC(ctx context.Context, targetNodeID string, req *models.PutChunkDC
...
+				//
+				// In PutChunkDC we can get timeout because of bad connection between the downstream nodes,
+				// and not necessarily between the client node and target node. In this case, we retry WriteMV using
+				// the OriginatorSendsToAll strategy. So, to be safe we don't delete the connections if we get
+				// timeout error in PutChunkDC. In the PutChunk using OriginatorSendsToAll strategy fails using
+				// timeout, we then delete the connections.
+				//
+				// We do resetAllRPCClients() here because if the connection between client and target node is good,
+				// the response from target node will eventually return after the timeout error. In this case, we
+				// cannot reuse the same which was timed out as it will result in ambiguous behavior.
```
linuxsmiths: isn't it better to reset just this client since the error is received on this connection.
Others mostly will be fine or max one of the other connections would need to be reset but that we can do when we know for sure with OriginatorSendsToAll. Anyways, if it's bad the reset itself will timeout, so we don't save on time either.

internal/dcache/rpc/client/functions.go
```diff
@@ -391,6 +420,13 @@ func PutChunk(ctx context.Context, targetNodeID string, req *models.PutChunkRequ
 
 			// Fall through to release the RPC client.
 			resp = nil
+		} else {
+			//
+			// The RPC call to the target node succeeded. If the node or the RV is marked negative or iffy,
+			// clear it now.
+			//
+			cp.removeNegativeNode(targetNodeID)
+			cp.removeIffyRvId(req.Chunk.Address.RvID)
```
linuxsmiths, line 430: General question:
Btw, I'm curious, when is this actually helping.
If the node was negative then getRPCClient() would have failed.
So this will only help for the rare case where one of the other threads marked it negative after this thread got its client and made the RPC calls. Right?

internal/dcache/rpc/client/client_pool.go
```diff
@@ -299,28 +340,18 @@ func (cp *clientPool) getNodeClientPool(nodeID string) (*nodeClientPool, error)
...
-				err := fmt.Errorf("not creating RPC clients for node %s, negative timeout not expired yet (%d seconds elapsed, %d seconds timeout)",
-					nodeID, timeElapsed, defaultNegativeTimeout)
-				log.Err("clientPool::getNodeClientPool: %v", err)
-				return nil, err
-			} else {
-				log.Debug("clientPool::getNodeClientPool: Negative timeout expired for node %s, removing from negative nodes map (%d seconds elapsed, %d seconds timeout)",
-					nodeID, timeElapsed, defaultNegativeTimeout)
-				cp.negativeNodes.Delete(nodeID)
-			}
+		err := cp.checkNegativeNode(nodeID)
```
linuxsmiths, line 349: Let's just keep the logic of checking negative node expiry (and removing it on expiry) in the periodic thread. Other simply check if the node is in the negative nodes map or not.

linuxsmiths (approved):

PR #1933, opened 2025-08-24
Add support for parallel reads and writes to the same file from multi…

linuxsmiths:
…ple fuse threads

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

- **Feature / Bug Fix**:

## Checklist

## Related Links
- [Issues](<link>)

### DIFF
--- a/internal/dcache/file_manager/file_manager.go
+++ b/internal/dcache/file_manager/file_manager.go
@@ -348,7 +348,11 @@ func (file *DcacheFile) ReadFile(offset int64, buf *[]byte) (bytesRead int, err 
 				//
 				removed := file.removeChunk(chunk.Idx)
 				_ = removed
-				common.Assert(removed == true, chunk.Idx, file.FileMetadata.Filename)
+				//
+				// We cannot assert for the following as the chunk may be removed from the map by the time we
+				// reach here. e.g., one case I saw is some other thread called removeAllChunks() above.
+				//
+				//common.Assert(removed == true, chunk.Idx, file.FileMetadata.Filename)
 			}
 		}
 
@@ -762,7 +766,7 @@ func (file *DcacheFile) getChunk(chunkIdx, chunkOffset, length int64, noCache, a
 		file.FileMetadata.Filename, chunkIdx, chunkOffset, length)
 
 	common.Assert(chunkIdx >= 0, chunkIdx, chunkOffset, length, noCache, file.FileMetadata.Filename)
-	common.Assert(chunkOffset + length <= file.FileMetadata.FileLayout.ChunkSize,
+	common.Assert(chunkOffset+length <= file.FileMetadata.FileLayout.ChunkSize,
 		chunkIdx, chunkOffset, length, file.FileMetadata.FileLayout.ChunkSize, noCache, file.FileMetadata.Filename)
 
 	//
@@ -846,7 +850,7 @@ func (file *DcacheFile) getChunkForRead(chunkIdx, chunkOffset, length int64) (*S
 
 	common.Assert(chunkIdx >= 0, chunkIdx, chunkOffset, length)
 	common.Assert(chunkOffset >= 0, chunkIdx, chunkOffset, length)
-	common.Assert(length >= 0 && (chunkOffset + length) <= file.FileMetadata.FileLayout.ChunkSize,
+	common.Assert(length >= 0 && (chunkOffset+length) <= file.FileMetadata.FileLayout.ChunkSize,
 		chunkIdx, chunkOffset, length)
 
 	noCache := (length != 0)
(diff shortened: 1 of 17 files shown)


internal/dcache/file_manager/file_manager.go
```diff
@@ -614,95 +1085,201 @@ func (file *DcacheFile) releaseChunk(chunk *StagedChunk) {
...
+		// to application reads. In a healthy sequential read scenario, the chunks would soon qualify as
+		// "fully read" and would be removed from StagedChunks. If the chunks build up to more than the critical
+		// limit it means the pattern is not sequential and not even random. It's able to qualify as "cacheable"
+		// in ReadFile() but then application doesn't read the entire chunk or reads it in a way that we cannot
+		// track. We need to keep the cache usage under check.
+		//
+		// Note: numStagedChunksCritical should be large enough to accomodate 2*windowSize chunks, see
+		//       NewRPTracker.
+		//
+		numStagedChunksCritical := fileIOMgr.numReadAheadChunks * 8
```
syeleti-msft: this amounts to 2GB which is in higher side. could you share on which read pattern we hit this, I can't think of one

internal/dcache/file_manager/file_manager.go
```diff
@@ -614,95 +1085,201 @@ func (file *DcacheFile) releaseChunk(chunk *StagedChunk) {
...
+			// TODO: See if we should remove only the last few chunks instead of all.
+			log.Debug("DistributedCache::readChunkWithReadAhead: file: %s, too many staged chunks (%d > %d), removing all",
+				file.FileMetadata.Filename, len(file.StagedChunks), numStagedChunksCritical)
+			file.removeAllChunks(false /* needLock */)
+		}
+
+		// Conservative estimate of chunks in use (already staged + readahead scheduled but not yet issued).
+		chunksInUse := file.readaheadToBeIssued.Load() + int64(len(file.StagedChunks))
+		// How many more chunks are we allowed to readahead?
+		readAheadCount := max(int64(fileIOMgr.numReadAheadChunks-int(chunksInUse)), 0)
```
syeleti-msft, line 1307: I think it would be wiser to do exponential readahead based on the past total seq-read-size bytes done by the user application. 

Say with out direct-io enabled we were limited to the 128K reads and now when user tries to do the random read for the size of 512K we detect as seq pattern(as there are 4 * 124K req) and go readahead 64 chunks and when offset jumps we will clean all chunks(as we are unsure) and do readahead again for 64 chunks end up performing readaheads always .

internal/dcache/file_manager/file_manager.go
```diff
@@ -241,151 +387,272 @@ func (file *DcacheFile) WriteFile(offset int64, buf []byte) error {
...
+
+			chunkFullyWritten = chunk.IOTracker.MarkAccessed(chunkOffset, int64(copied), chunk.Idx)
+
+			// If chunkFullyWritten is true, chunk must have been fully written.
+			common.Assert(!chunkFullyWritten || (chunk.Len == file.FileMetadata.FileLayout.ChunkSize),
+				chunkFullyWritten, chunk.Len, file.FileMetadata.FileLayout.ChunkSize,
+				chunk.Idx, file.FileMetadata.Filename)
+		}
+
+		if chunkFullyWritten {
```
syeleti-msft, line 628: better to add a todo on improving the eviction policy where cache gets exploded and the chunks were never written

internal/dcache/file_manager/file_manager.go
linuxsmiths: The calculation is like this. 
How many chunks can be legitimately sitting in the cache?
All the readahead chunks + the entire window size (which will exhaust if user keeps so many outstanding IOs).
This some to 64 + 200.
I just picked the next multiple higher than this. We can make it a little lower, say 6.

internal/dcache/file_manager/file_manager.go
linuxsmiths: Since mostly we are expecting sequential reads, I prefer to keep it simple for now.
Later if access patterns demand more involved readahead algo we can do this.

internal/dcache/file_manager/file_manager.go
linuxsmiths: That can only happen when user is doing non-sequential reads which escape the random writes check.
In that corner case, we have protection in getChunkForWrite() and we will never allow cached write chunks to go beyond fileIOMgr.numStagingChunks

internal/dcache/file_manager/file_manager.go
```diff
@@ -141,22 +143,71 @@ func EndFileIOManager() {
...
+	// the window size supported by the read patter tracker. See NewRPTracker().
+	maxChunks int64
+
+	// Semaphore to limit number of in-use chunks.
+	// Note that in order to support parallel readers/writers, we need multiple partial chunks to be present
+	// at any time. As soon as chunk is "fully used" (fully read/written by application), it can be released.
+	// We limit the number of chunks in use to avoid using too much memory.
+	// NewStagedChunk() reads from this channel before allocating a new chunk.
+	// releaseChunk() signals on this channel when a chunk is freed.
+	freeChunkSemaphore chan struct{}
```
syeleti-msft: can we rename it to freeChunks? This would make the naming consistent

PR #1939, opened 2025-08-25
persist the stack trace, in the event of crash

syeleti-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

- **Feature / Bug Fix**:
persist the stack trace when blobfuse is mounted in background and crashed. in such a case user can get the stack trace from  /tmp/blobfuse2.pid file

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 **Bug Fixes**
 - Fail file open operation if the file being downloaded by file-cache can not fit in available disk space (either configured by user or computed implicitly by blobfuse). User application will receive ENOSPC (no space left on device) in response to file open call.
 - Mount will fail if FNS account is mounted as HNS account.
-- Redirect Stack trace to log file (WORK_DIR/mntpath.pid.trace>/) instead of console in case of panic for better debuggability.
+- Redirect Stack trace to log file (WORK_DIR/mount_path.pid.trace) instead of console in case of panic for better debuggability.
 
 ## 2.5.0 (2025-07-17)
 **Bug Fixes**


cmd/mount.go
```diff
@@ -496,13 +495,13 @@ var mountCmd = &cobra.Command{
...
 			pid := os.Getpid()
-			fname := fmt.Sprintf("/tmp/blobfuse2.%v", pid)
+			childStdErrFileName := fmt.Sprintf("/tmp/blobfuse2.%v", pid)
 
 			dmnCtx := &daemon.Context{
 				PidFileName: pidFileName,
 				PidFilePerm: 0644,
 				Umask:       022,
-				LogFileName: fname, // this will redirect stderr of child to given file
+				LogFileName: childStdErrFileName, // this will redirect stderr of child to given file
```
bhansalivikas: if stderr is redirected to this file than this shall have the stack trace as well ?

cmd/mount.go
```diff
@@ -561,16 +578,23 @@ var mountCmd = &cobra.Command{
...
 						log.Err("mount: failed to read child [%v] failure logs [%s]", child.Pid, err.Error())
-						return Destroy(fmt.Sprintf("failed to mount, please check logs [%s]", err.Error()))
+						err = fmt.Errorf("failed to mount, please check logs [%s]", err.Error())
 					} else {
-						return Destroy(string(buff))
+						err = fmt.Errorf(string(buff))
+					}
+
+					// Safe to delete the temp file.
+					rmErr := os.Remove(childStdErrFileName)
```
bhansalivikas: Parent shall not have the dump file created as our actual mount will go to child only. What are trying to save here?

cmd/mount.go
```diff
@@ -561,16 +578,23 @@ var mountCmd = &cobra.Command{
...
+					if rmErr != nil {
```
bhansalivikas, line 592: After starting the child process, parent will go down and delete this file. Ideally this file shall be created by child and deleted only when child exits. Only in case of foreground process may be parent can also have logic to use this file.

common/log/logger.go
```diff
@@ -184,8 +185,12 @@ func SetLogLevel(lvl common.LogLevel) {
 }
 
 // Destroy : DeInitialize the logging library
+// This should only be called from the main function.
 func Destroy() error {
-	return logObj.Destroy()
+	if logObj != nil {
+		return logObj.Destroy()
+	}
+	return fmt.Errorf("Logger is not initialized")
```
bhansalivikas, line 194: In case of any sort of early failure, we might hit this condition. Does not look like we shall emit a log here, as the failure condition itself shall be dumping the error out on the screen.

component/file_cache/lru_policy.go
```diff
@@ -275,6 +283,7 @@ func (p *lruPolicy) cacheValidate(name string) {
 // For all other timer based activities we check the stuff here
 func (p *lruPolicy) clearCache() {
 	log.Trace("lruPolicy::ClearCache")
+	defer p.wg.Done()
```
bhansalivikas, line 286: In reason to wait for this thread is to clear the cache before the unmount goes out ?

main.go
```diff
@@ -43,11 +43,16 @@ import (
...
 
 func main() {
-	_ = cmd.Execute()
+	defer log.Destroy()
+	// This recovers the panics only for the functions that run within this context. all the go-routine
+	// spawned by this function need to handle their panics seperately if required. Also the FUSE callbacks
+	// wouldn't run in this context, so the panics originated from the callbacks can't get recovered here.
 	defer func() {
 		if panicErr := recover(); panicErr != nil {
-			log.Err("PANIC: %v", panicErr)
```
bhansalivikas, line 48: If panics generated by other go routines can not be caught here, then it shall be safe to remove this code as we will not be catching anything here, other than cli parsing crashes.

cmd/mount.go
syeleti-msft: Yes, this will have it. Currently the parent process is deleting this file when it terminates

cmd/mount.go
syeleti-msft: when the mount fails to init we dont want this file to be there, so we delete it. Currently this file is used by the parent process to read the stderr logs given by the fuse library incase of any failures. It is safe to delete the file as we read the contents of it and displayed to the user why it failed.

cmd/mount.go
syeleti-msft: if child creates the file after it started, I see no way of redirecting the stderr to this newly created file (I mean we can't redirect the previous fd's pointing to this new file).

component/file_cache/lru_policy.go
syeleti-msft: This is added to synchronize the go routines properly. Currently we don't wait for them to complete before we yield the success for stopping.

main.go
syeleti-msft: Yes, you are correct. but this won't hurt anything

cmd/mount.go
```diff
@@ -529,22 +528,40 @@ var mountCmd = &cobra.Command{
...
 			}
 
 			log.Debug("mount: foreground disabled, child = %v", daemon.WasReborn())
 			if child == nil { // execute in child only
+
+				// defer the removal of this temp file. This file should only be removed when the mountpoint is
+				// gracefully unmounted. In case of any panic caused by go runtime/ by our application. This file
+				// would have the essential stack trace to debug the issue.
+				ppid := os.Getppid()
+				childStdErrFileName = fmt.Sprintf("/tmp/blobfuse2.%v", ppid)
```
jainakanksha-msft: Define the file name at one place and use it everywhere.
No need to calculate it every time.

jainakanksha-msft: Who will be responsible of cleaning up this file after panic situation?
Will it be a manual activity or another remount or still trying for unmount will take care of that.?

souravgupta-msft: Why to use /tmp directory for storing this file. Can we use the default working directory which is $HOME/.blobfuse2 or the one configured.

syeleti-msft: >Define the file name at one place and use it everywhere.
No need to calculate it every time.

The reason to do this is the file name would be different in both parent and child according to their pid. The file name we want to delete is created using parent pid. 

>Who will be responsible of cleaning up this file after panic situation?

Currently it is manual, the user must clean them. this can happen when user kill the process using SIGKILL as well

> Will it be a manual activity or another remount or still trying for unmount will take care of that.?

as we don't want to lose this file we don't delete it.

syeleti-msft: Yes we can use that. what do you say @vibhansa-msft @jainakanksha-msft @ashruti-msft ?

jainakanksha-msft: Good suggestion

common/log/logger.go
syeleti-msft: yes

cmd/mount.go
syeleti-msft: @souravgupta-msft , made the change. pls take a look at it now

PR #1941, opened 2025-09-01
Correct direct_io behaviour

vibhansa-msft:
## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description
Correcting direct_io behaviour
1. Disable kernel data and metadata caching
2. Set file-cache timeout to 1 second if not configured explicitly by user
3. Set attr-cache timeout to 1 second if not configured explicitly by user

## How Has This Been Tested?
Mabual

## Checklist
- [x] Documentation update required.

## Related Links
NA

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## 2.6.0 (Unreleased)
 **Bug Fixes**
 - Fail file open operation if the file being downloaded by file-cache can not fit in available disk space (either configured by user or computed implicitly by blobfuse). User application will receive ENOSPC (no space left on device) in response to file open call.
+- *direct_io* mode was disabling both kernel and blobfuse level caching post 2.4.0. With this release it will disable kernel data and metadata caching. Blobfuse level data and metadata caching will be valid for 1 second by default but user can tune these using file_cache and attr_cache timeouts.
+
 
 ## 2.5.0 (2025-07-17)
 **Bug Fixes**
--- a/cmd/mount.go
+++ b/cmd/mount.go
@@ -395,11 +395,6 @@ var mountCmd = &cobra.Command{
 			}
 		}
 
-		if config.IsSet("disable-kernel-cache") && directIO {
-			// Both flag shall not be enable together
-			return fmt.Errorf("direct-io and disable-kernel-cache cannot be enabled together")
-		}
-
 		if !config.IsSet("logging.file-path") {
 			options.Logging.LogFilePath = common.DefaultLogFilePath
 		}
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 ## 2.6.0 (Unreleased)
 **Bug Fixes**
 - Fail file open operation if the file being downloaded by file-cache can not fit in available disk space (either configured by user or computed implicitly by blobfuse). User application will receive ENOSPC (no space left on device) in response to file open call.
+- *direct_io* mode was disabling both kernel and blobfuse level caching post 2.4.0. With this release it will disable kernel data and metadata caching. Blobfuse level data and metadata caching will be valid for 1 second by default but user can tune these using file_cache and attr_cache timeouts.
```
jainakanksha-msft, line 4: We have mentioned metadata caching in disabled as well as valid for 1 sec.
I think we mean here is -
Starting with this release, direct_io will only disable kernel data. Blobfuse’s own data and metadata caches will remain active, with a default validity of 1 second. These timeouts can be adjusted by the user through the file_cache and attr_cache timeout configuration options.

cmd/mount.go
```diff
@@ -395,11 +395,6 @@ var mountCmd = &cobra.Command{
 			}
 		}
 
-		if config.IsSet("disable-kernel-cache") && directIO {
```
jainakanksha-msft, line 398: This check is still valid right.
User hasn't configure both of these config.

CHANGELOG.md
bhansalivikas: kernel metadata cache will be disabled but blobfuse level metadata cache will be valid for 1 second.

cmd/mount.go
bhansalivikas: now both the flags means the same, so we dont care which one is set.

jainakanksha-msft: Eventually, we will need a task to cleanup the duplicate configs also.

component/file_cache/file_cache.go
```diff
@@ -235,9 +235,9 @@ func (c *FileCache) Configure(_ bool) error {
 	directIO := false
 	_ = config.UnmarshalKey("direct-io", &directIO)
 
-	if directIO {
-		c.cacheTimeout = 0
-		log.Crit("FileCache::Configure : Direct IO mode enabled, cache timeout is set to 0")
+	if directIO && !config.IsSet(compName+".file-cache-timeout-in-seconds") && !config.IsSet(compName+".timeout-sec") {
```
jainakanksha-msft, line 238: I think we should also update gen-config correspondingly for attr_cache and file_cache.

PR #1954, opened 2025-09-11
Assert fixes

souravgupta-msft:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description
Fixing the assert failures seen while running test-cluster.sh script.

### DIFF
--- a/internal/dcache/rpc/client/client_pool.go
+++ b/internal/dcache/rpc/client/client_pool.go
@@ -1085,7 +1085,7 @@ func (cp *clientPool) resetRPCClientInternal(client *rpcClient, needLock bool) e
 		// This will happen when the target node is not running the blob service (but the node
 		// itself is up). When the node is no up we should get a connection timeout error here.
 		//
-		// Also seen connection reset, connection timeout or no route to host errors here.
+		// Also seen connection reset, connection timeout and no route to host errors here.
 		//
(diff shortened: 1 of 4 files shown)


internal/dcache/rpc/client/client.go
```diff
@@ -146,9 +145,15 @@ func newRPCClient(nodeID string, nodeAddress string) (*rpcClient, error) {
...
 
-		common.Assert(rpc.IsTimedOut(err) ||
-			rpc.IsConnectionRefused(err) ||
-			rpc.IsConnectionReset(err), err)
+		//
+		// TODO: We can also get other errors here like,
+		//   - "no route to host" - the node is on different subnet/vnet or the node is deallocated.
+		//   - "cannot assign requested address" - can be caused due to port exhaustion if we are
+		//     creating too many connections in a short period of time.
+		//
```
linuxsmiths: did you actually get the "not route" error?
also the "cannot assign..." error, as I said we need to understand.. if you are able to repro it, I'd like to see
let's discuss this.

And, if we are seeing other errors, we need to handle them .. 
Removing asserts puts us to a risk of not detecting some unhandled error that might be happening.
now we don't have enough runway left for the TODOs.

linuxsmiths (approved):

PR #1955, opened 2025-09-12
Auto fill the dcache

syeleti-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description

- **Feature / Bug Fix**:
When the reading the file from the unqualified namespace, if file exists in the azure, then bring the chunks to the dcache without blocking the open() call.

## How Has This Been Tested?

Test was done using 1 replica in D96VM

Throughput with this change: where the file would be downloaded in background and written to dcache parallelly. User application reads were served from the another dcache handle without blocking.
```
dcacheuser@dcache-vm-wu-10:~$ dd of=/dev/null if=~/mnt/100G.ooo bs=1M  iflag=direct status=progress
107260936192 bytes (107 GB, 100 GiB) copied, 53 s, 2.0 GB/s
102400+0 records in
102400+0 records out
107374182400 bytes (107 GB, 100 GiB) copied, 53.036 s, 2.0 GB/s
```

Through while reading directly from the Azure:
```
dcacheuser@dcache-vm-wu-10:~$ dd of=/dev/null if=~/mnt/fs=azure/100G.ooo bs=1M  iflag=direct status=progress
107298684928 bytes (107 GB, 100 GiB) copied, 36 s, 3.0 GB/s
102400+0 records in
102400+0 records out
107374182400 bytes (107 GB, 100 GiB) copied, 36.0244 s, 3.0 GB/s
```


syeleti-msft:
The approach for cache hydration is changed. Closing the PR

PR #1993, opened 2025-09-17
Prevent closing the RPC client again in PutChunkDC timeout

souravgupta-msft:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description
In PutChunkDC timeout error, we reset the client for the target node which first closes the client and then creates a new one. If the new client creation fails, we then try to delete all the clients for the target node. In delete, we again try to close the same client which results in assert failure.

### DIFF
--- a/internal/dcache/rpc/client/client_pool.go
+++ b/internal/dcache/rpc/client/client_pool.go
@@ -864,8 +864,9 @@ func (cp *clientPool) deleteAllRPCClients(client *rpcClient, confirmedBadNode bo
 	//
 	// Node client pool may not be present for the node in case of PutChunkDC timeout error,
 	// where we first reset the client, which closes the client and then creates a new client.
-	// If the new client creation fails, we come here to delete all clients. Meanwhile some other thread
-	// may have closed all the clients for the target node and deleted the node client pool.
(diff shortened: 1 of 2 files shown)


internal/dcache/rpc/client/client_pool.go
```diff
@@ -895,11 +919,23 @@ func (cp *clientPool) deleteAllRPCClients(client *rpcClient, confirmedBadNode bo
...
+	if isClientClosed {
+		log.Debug("clientPool::deleteAllRPCClients: client (%p) to %s node %s is already closed",
+			client, client.nodeAddress, client.nodeID)
+	} else {
+		//
+		// Delete this client. This closes this client and removes it from the pool.
+		// It can only fail if Thrift fails to close the connection.
+		// This should technically not happen, so we assert.
+		//
+		cp.deleteRPCClient(client)
```
linuxsmiths, line 936: move numConnDeleted++ inside this to make it more truthful.
then remove the assert at L969, it's not much useful now.

souravgupta-msft: Done

PR #2006, opened 2025-09-23
Fix when the source of the sync goes down

souravgupta-msft:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description
If the source RV of the sync operation goes down, the target RV remains in syncing state, and the sync job gets stuck. To fix this, we run a thread in target RV which checks its MVs where it is in syncing state. If the last write time via PutChunk sync call to an MV becomes greater than some threshold value, it marks itself as inband-offline, so that fix-mv workflow will then replace it wit some other RV.

### DIFF
--- a/internal/dcache/rpc/client/client_pool.go
+++ b/internal/dcache/rpc/client/client_pool.go
@@ -1922,6 +1922,7 @@ func (ncPool *nodeClientPool) createRPCClients(nodeID string, numClients uint32)
 	ncPool.clientChan = make(chan *rpcClient, numClients)
 	ncPool.lastUsed.Store(time.Now().Unix())
 
+	var err error
 	var wg sync.WaitGroup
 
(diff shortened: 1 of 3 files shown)


internal/dcache/rpc/client/client_pool.go
```diff
@@ -1957,7 +1956,7 @@ func (ncPool *nodeClientPool) createRPCClients(nodeID string, numClients uint32)
 	}
 
 	// Create RPC clients and add them to the channel.
-	for i := 0; i < int(numClients); i++ {
+	for range int(numClients) {
```
linuxsmiths: why?
let's have the more readable syntax.
note that numClients is an integer

linuxsmiths:
audit/update/merge

PR #2026, opened 2025-10-14
Dcache logs collection

souravgupta-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description
Simplify logs collection from all nodes in the cluster.

### DIFF
--- a/component/distributed_cache/distributed_cache.go
+++ b/component/distributed_cache/distributed_cache.go
@@ -1030,9 +1030,13 @@ func (dc *DistributedCache) OpenFile(options internal.OpenFileOptions) (*handlem
 		return nil, syscall.EINVAL
 	}
 
-	// todo: We should only support write if the file is only in Azure.
+	//
+	// TODO: We should support write for files which are only in Azure.
+	//
+	// Currently we support write only for fs=debug/logs.
+	//
 	if (options.Flags&os.O_WRONLY != 0 || options.Flags&os.O_RDWR != 0) &&
-		!(isDebugPath && rawPath == "logs") { // We support write only for fs=debug/logs
+		!(isDebugPath && rawPath == "logs") {
 		log.Err("DistributedCache::OpenFile: Dcache file cannot open with flags: %X, file : %s", options.Flags, options.Name)
 		return nil, syscall.EACCES
 	}
@@ -1160,8 +1164,7 @@ func (dc *DistributedCache) WriteFile(options *internal.WriteFileOptions) (int, 
 	// Allow writes only to fs=debug/logs; other debug proc files remain read-only.
 	if options.Handle.IsFsDebug() {
 		if options.Handle.Path == "logs" {
-			written, err := debug.WriteFile(options)
-			return written, err
+			return debug.WriteFile(options)
 		}
 		common.Assert(false, options.Handle.Path)
 	}
--- a/internal/dcache/debug/callbacks.go
+++ b/internal/dcache/debug/callbacks.go
@@ -44,6 +44,7 @@ import (
 	"github.com/Azure/azure-storage-fuse/v2/common/log"
 	cm "github.com/Azure/azure-storage-fuse/v2/internal/dcache/clustermap"
 	"github.com/Azure/azure-storage-fuse/v2/internal/dcache/debug/stats"
+	rpc_client "github.com/Azure/azure-storage-fuse/v2/internal/dcache/rpc/client"
 )
 
 //go:generate $ASSERT_REMOVER $GOFILE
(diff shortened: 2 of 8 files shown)


component/distributed_cache/distributed_cache.go
```diff
@@ -1031,7 +1031,8 @@ func (dc *DistributedCache) OpenFile(options internal.OpenFileOptions) (*handlem
 	}
 
 	// todo: We should only support write if the file is only in Azure.
-	if options.Flags&os.O_WRONLY != 0 || options.Flags&os.O_RDWR != 0 {
+	if (options.Flags&os.O_WRONLY != 0 || options.Flags&os.O_RDWR != 0) &&
+		isDebugPath && rawPath != "logs" { // We support write only for fs=debug/logs
```
linuxsmiths: this opens up write for non-debug files, correct check should be

    if (options.Flags&os.O_WRONLY != 0 || options.Flags&os.O_RDWR != 0) &&
        !(isDebugPath && (rawPath == "logs")) {

internal/dcache/debug/callbacks.go
```diff
@@ -90,6 +93,27 @@ func readStatsCallback(pFile *procFile) error {
...
 
+// proc file: logs
+// On first open, it triggers collection of logs from all nodes via RPC and returns a JSON summary.
+// Tarballs are stored on disk. This file only returns metadata mapping of node IDs to log tarball paths.
+//
+// By default, we use the default work dir for storing the collected logs and
+// we also collect only the most recent log file from each node by default.
+//
+// If users want to specify a different directory or number of logs, they can use the
+// fs=debug/logs file with a write call to specify those parameters.
```
linuxsmiths: the idea is not to allow collecting logs using write (and read).
write should purely only support configuring. Once onfigured it should remember the settings and subsequent reads of the logs file should honor those while collecting logs. So note that logs are only ever collected by reading fs=debug/logs file, never by writing to it.

internal/dcache/debug/debug_fs.go
```diff
@@ -159,25 +201,67 @@ func CloseFile(options internal.CloseFileOptions) error {
...
+		return 0, fmt.Errorf("output_dir must be an absolute path: %s", req.OutputDir)
+	}
+
+	pFile := options.Handle.IFObj.(*procFile)
+	common.Assert(atomic.LoadInt32(&pFile.openCnt) > 0)
+
+	err := collectLogs(pFile, req.OutputDir, req.NumLogs)
+	if err != nil {
+		return 0, err
+	}
```
linuxsmiths: logs should not be collected by write

internal/dcache/debug/debug_fs.go
```diff
@@ -159,25 +201,67 @@ func CloseFile(options internal.CloseFileOptions) error {
 	return nil
 }
 
-// Refresh the contents of the proc File if needed
-func openProcFile(path string) (*procFile, error) {
+func WriteFile(options *internal.WriteFileOptions) (int, error) {
```
linuxsmiths, line 243: like we have made the read pluggable, we should make write too.
I can see that we might soon need to support write for other debug files too.

linuxsmiths (commented):
still looking...

component/distributed_cache/distributed_cache.go
souravgupta-msft: Done

internal/dcache/debug/callbacks.go
souravgupta-msft: Updated

internal/dcache/debug/debug_fs.go
souravgupta-msft: Updated

internal/dcache/rpc/client/functions.go
```diff
@@ -1295,6 +1296,200 @@ func GetMVSize(ctx context.Context, targetNodeID string, req *models.GetMVSizeRe
...
+
+			log.Debug("rpc_client::GetLogs: Copying local log tar file from %s -> %s", srcPath, outPath)
+			err = copyFile(srcPath, outPath)
+			if err != nil {
+				log.Err("rpc_client::GetLogs: Failed to copy log tar file in local node %v [%v]", err)
+				return nil, err
+			}
+
+			break
+		}
```
linuxsmiths: no need to special case this. The advantage is miniscule and it ties the server and client other than the rpc call semantics.

internal/dcache/rpc/server/handler.go
```diff
@@ -3456,6 +3467,250 @@ func (h *ChunkServiceHandler) GetMVSize(ctx context.Context, req *models.GetMVSi
...
+	//
+	// Take logStateMu lock to prevent multiple threads from creating the log tarball.
+	//
+	h.logStateMu.Lock()
+
+	//
+	// Create the tarball on first chunk or,
+	// if the logTarPath is empty (client asked for non-zero chunk index without initiating)
+	//
+	if req.ChunkIndex == 0 || h.logTarPath == "" {
```
linuxsmiths, line 3509: don't need to support clients starting at non-zero index.
it's our clien and we control that

internal/dcache/rpc/server/handler.go
```diff
@@ -3456,6 +3472,258 @@ func (h *ChunkServiceHandler) GetMVSize(ctx context.Context, req *models.GetMVSi
...
+	}, nil
+}
+
+// Create tarball from atmost numLogs recent log files.
+// Caller must hold logStateMu lock.
+func (h *ChunkServiceHandler) createLogTarLocked(numLogs int64) error {
+	common.Assert(numLogs > 0, numLogs)
+
+	logFilePath := ""
+	err := config.UnmarshalKey("logging.file-path", &logFilePath)
```
linuxsmiths, line 3637: is this key guaranteed to be set, even if user doesn't override it?

internal/dcache/rpc/client/functions.go
```diff
@@ -1295,6 +1296,194 @@ func GetMVSize(ctx context.Context, targetNodeID string, req *models.GetMVSizeRe
...
+
+			// Release RPC client back to the pool.
+			err1 := cp.releaseRPCClient(client)
+			if err1 != nil {
+				log.Err("rpc_client::GetLogs: Failed to release RPC client for node %s %v: %v",
+					targetNodeID, reqStr, err1)
+				// Assert, but not fail the GetLogs call.
+				common.Assert(false, err1)
+			}
+		}
```
linuxsmiths, line 1429: we should break here

PR #2063, opened 2025-11-13
Add goroutine id in debug logs

souravgupta-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Other (describe):

## Description
Add goroutine id in debug logs for better debuggability.

## How Has This Been Tested?
Added unit tests and also tested via logs for both base and syslog types.

## Checklist

### DIFF
--- a/azure-pipeline-templates/e2e-tests.yml
+++ b/azure-pipeline-templates/e2e-tests.yml
@@ -74,7 +74,7 @@ steps:
   - script: |
       tail -n 200 blobfuse2-logs.txt
     displayName: 'View Logs'
-    condition: always()
+    condition: failed()
 
   - script: |
       > blobfuse2-logs.txt


common/log/base_logger.go
```diff
@@ -219,15 +219,25 @@ func (l *BaseLogger) logEvent(lvl string, format string, args ...any) {
 	// Only log if the log level matches the log request
 	_, fn, ln, _ := runtime.Caller(3)
 	msg := fmt.Sprintf(format, args...)
-	msg = fmt.Sprintf("%s : %s[%d] : [%s] %s [%s (%d)]: %s",
+
+	base := fmt.Sprintf("%s : %s[%d] : ",
 		time.Now().Format(time.UnixDate),
```
syeleti-msft: better to increase the verbosity of time to milliseconds here, very helpful for debugging

common/util.go
```diff
@@ -583,3 +584,32 @@ func UpdatePipeline(pipeline []string, component string) []string {
 
 	return pipeline
 }
+
+// Return the goroutine id of the current goroutine.
+// Example, goroutine 17 [running]: => return 17
+func GetGID() uint64 {
```
syeleti-msft: is it possible to rename this to something else, as the name is giving some different idea of file permission bits.
maybe getGoRoutineID?

common/log/base_logger.go
souravgupta-msft: Done

common/util.go
souravgupta-msft: Updated

common/util.go
```diff
@@ -583,3 +584,32 @@ func UpdatePipeline(pipeline []string, component string) []string {
...
+		return 0
+	}
+
+	// Parse those digits into a number
+	gid, err := strconv.ParseUint(string(b[:i]), 10, 64)
+	if err != nil {
+		return 0
+	}
+
+	return gid
```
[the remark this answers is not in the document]
souravgupta-msft: There is an alternative where we can use this dependency `github.com/petermattis/goid`, which reads the GID directly from the Go runtime's internal struct, instead of making `runtime.Stack` call.
The only con here is that it does not guarantee compatibility promise and can break with any new Go version. It does support our current Go version 1.25.

common/log/base_logger.go
```diff
@@ -219,15 +219,25 @@ func (l *BaseLogger) logEvent(lvl string, format string, args ...any) {
...
+
+	remaining := fmt.Sprintf("[%s] %s [%s (%d)]: %s",
 		common.MountPath,
 		lvl,
 		filepath.Base(fn), ln,
 		msg)
 
+	// If log level is debug in config, add goroutine id in the log
+	if l.fileConfig.LogLevel >= common.ELogLevel.LOG_DEBUG() {
+		msg = fmt.Sprintf("%s[%d]%s", base, common.GetGoroutineID(), remaining)
```
syeleti-msft, line 236: Any reason why we are restricting to put go-routine id in only debug logs and not in others? more info is always helpful right:)

syeleti-msft (approved):

common/log/base_logger.go
souravgupta-msft: Getting goroutine id from the runtime stack can cause perf degradation. This is why I kept it in just debug level logs.

vibhansa-msft (approved):

Issue #735, opened 2022-03-13
[blobfuse2] return code should be non-zero when mount parameter is wrong


andyzhangx:
### Which version of the blobfuse was used?
blobfuse v2 preview

### Which OS (please include version) are you using?
Ubuntu 18.04

### What problem was encountered?
It seems `-o` parameters are not supported any more in v2, while it still returns 0 which means mount is successful, and csi driver would regard it as successful mount.  return code should be non-zero when mount parameter is wrong

```
mkdir blob3
blobfuse2 mount blob3 --container-name=public --tmp-path=/tmp/blobfuse/ --config-file=/usr/share/blobfuse2/config.yaml -o allow_other --file-cache-timeout-in-seconds=120 --use-attr-cache=true --cancel-list-on-mount-seconds=10 -o attr_timeout=120 -o entry_timeout=120 -o negative_timeout=120 --log-level=LOG_WARNING --cache-size-mb=1000 -o ro --streaming=true# #
Error: unknown shorthand flag: 'o' in -o
Usage:
  blobfuse2 mount [path] [flags]
  blobfuse2 mount [command]

Available Commands:
  all         Mounts all azure blob container for a given account as a filesystem
  list        List all blobfuse2 mountpoints

Flags:
      --allow-other                Allow other users to access this mount point.
      --attr-cache-timeout int32   attribute cache timeout (default 120)
      --attr-timeout uint32         The attribute timeout in seconds
      --config-file string         Configures the path for the file where the account credentials are provided. Default is config.yaml (default "config.yaml")
      --container-name string      Configures the name of the container to be mounted
      --entry-timeout uint32       The entry timeout in seconds.
      --foreground                 Mount the system in foreground mode. Default value false.
  -h, --help                       help for mount
      --log-file-path string       Configures the path for log files. Default is $HOME/.blobfuse2/blobfuse2.log (default "$HOME/.blobfuse2/blobfuse2.log")
      --log-level string           Enables logs written to syslog. Set to LOG_WARNING by default. Allowed values are LOG_OFF|LOG_CRIT|LOG_ERR|LOG_WARNING|LOG_INFO|LOG_DEBUG (default "LOG_WARNING")
      --negative-timeout uint32    The negative entry timeout in seconds.
      --no-symlinks                whether or not symlinks should be supported
      --passphrase string          Key to decrypt config file. Can also be specified by env-variable BLOBFUSE2_SECURE_CONFIG_PASSPHRASE.
[...]

PR #2079, opened 2025-12-02
Print error msg on cmd line flag errors

syeleti-msft:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

- **Bug Fix**:
 when the incorrect flags are parsed, we are exiting with status 2 but not printing any message to the Stderr.
for example:
```
fantom@syeleti-vm:~/Software/azure-storage-fuse$ ./blobfuse2 mount ~/mnt --fasdfasdf
fantom@syeleti-vm:~/Software/azure-storage-fuse$ echo $?
2
```
This is mainly caused because of change in pflag [code](https://github.com/spf13/pflag/commit/5159cdaa32a65acb63296cbd45033696aa9c7e88) recently, where they removed printing the error to the StdErr and redirecting to the output stream when errorHandling is set to ExitOnError. But as Cobra doesn't support this errorHandling flags, we introduced this change in a private fork of the cobra [here](https://github.com/spf13/cobra/commit/fe9ade5260b2534e7eab30bdf472a96229f8692f). Now this doesn't handle output error streams properly. Now this is causing issue.

- the fork is not necessary as we return non-zero exit status when parsing fails, and will also help us stay on latest development on pkg.

the [issue ](https://github.com/Azure/azure-storage-fuse/issues/735)that originally introduced this fork is also resolved as we return non zero exit code when parsing fails, check [here](https://github.com/Azure/azure-storage-fuse/blob/a0590abf380af29afc9f050df470afb0f8b0a251/cmd/root.go#L285)

After Change: 
```
fantom@syeleti-vm:~/Software/azure-storage-fuse$ ./blobfuse2 mount ~/mnt --fasdfasdf
Error: unknown flag: --fasdfasdf
fantom@syeleti-vm:~/Software/azure-storage-fuse$ echo $?
1
```

## Checklist

## Related Links
- [Issues](<link>)


cmd/health-monitor_stop.go
```diff
@@ -45,11 +45,10 @@ import (
 var blobfuse2Pid string
 
 var healthMonStop = &cobra.Command{
-	Use:               "stop",
-	Short:             "Stops the health monitor binary associated with a given Blobfuse2 pid",
-	Long:              "Stops the health monitor binary associated with a given Blobfuse2 pid",
-	SuggestFor:        []string{"stp", "st"},
-	FlagErrorHandling: cobra.ExitOnError,
```
bhansalivikas, line 52: Are we removing this from all our commands ?

syeleti-msft: Yes, reason for removing is addressed in PR description

souravgupta-msft (approved):

PR #2100, opened 2026-01-13
Disable msi auth testing in Code coverage

syeleti-msft:
* Disable msi auth in unit tests to pass the code coverage, this is because the creating user assigned msi is disabled for 1es hosted pools.
* Add missing unit tests

### DIFF
--- a/blobfuse2-code-coverage.yaml
+++ b/blobfuse2-code-coverage.yaml
@@ -65,8 +65,9 @@ stages:
           # -------------------------------------------------------
           # Pull and build the code and create the containers.
           - template: 'azure-pipeline-templates/build.yml'
-            parameters:
-              skip_msi: "false"
+            # Disabling this for now, as user assigned MSI's are disabled for the 1es hosted pools.
+            # parameters:
+            #   skip_msi: "false"
 
           # -------------------------------------------------------
           # UT based code coverage test.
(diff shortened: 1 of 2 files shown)


common/util_64_test.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+func TestSetFrsize(t *testing.T) {
+	st := &syscall.Statfs_t{}
+	var val uint64 = 4096
+	SetFrsize(st, val)
+	if st.Frsize != int64(val) {
+		t.Errorf("SetFrsize failed: expected %d, got %d", val, st.Frsize)
+	}
+}
```
souravgupta-msft: What is this test for?

souravgupta-msft: I see this is for `SetFrsize` method. Can you add this in util_test.go?

common/util_64_test.go
```diff
@@ -0,0 +1,50 @@
...
+*/
+
+package common
+
+import (
+	"syscall"
+	"testing"
+)
+
+func TestSetFrsize(t *testing.T) {
```
bhansalivikas: Just for one small test why do we need a new file.

syeleti-msft: added as per convension, reverted back

common/util_64_test.go
syeleti-msft: Sure, reverted

PR #2125, opened 2026-02-17
Ensure logger is destroyed before returning from mount command.

syeleti-msft:
when the logger is destroyd in main function, when error occured and cobra exit the application with 1. then we are losing some logs to the end.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

- **Feature / Bug Fix**:
It's observed that some of the logs in the end were lost when there is any mount  failure. This is due to destroying the logger in main.go but cobra is exiting with error code 1 on any error. hence is logger is not getting destroyed and we are losing some logs to the end. 
made the change to destroy the logger before returning the control to cobra again. It is also better as logger init and destroy is happening at one place.

## Checklist

## Related Links
- [Issues](<link>)

### DIFF
--- a/main.go
+++ b/main.go
@@ -35,23 +35,11 @@ package main
 
 import (
 	"github.com/Azure/azure-storage-fuse/v2/cmd"
-	"github.com/Azure/azure-storage-fuse/v2/common/log"
 )
 
 //go:generate ./cmd/componentGenerator.sh $NAME
 //  To use go:generate run command   "NAME="component" go generate"
 
 func main() {
-	defer log.Destroy() // nolint:errcheck
-	// This recovers the panics only for the functions that run within this context. all the go-routine
-	// spawned by this function need to handle their panics separately if required. Also the FUSE callbacks
-	// wouldn't run in this context, so the panics originated from the callbacks can't get recovered here.
-	defer func() {
-		if panicErr := recover(); panicErr != nil {
-			log.Crit("PANIC: %v", panicErr)
-			panic(panicErr)
-		}
-	}()
-
 	_ = cmd.Execute()
 }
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -35,23 +35,11 @@ package main
...
-	// spawned by this function need to handle their panics separately if required. Also the FUSE callbacks
-	// wouldn't run in this context, so the panics originated from the callbacks can't get recovered here.
-	defer func() {
-		if panicErr := recover(); panicErr != nil {
-			log.Crit("PANIC: %v", panicErr)
-			panic(panicErr)
-		}
-	}()
-
 	_ = cmd.Execute()
```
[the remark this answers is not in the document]
syeleti-msft: These panics are very rare as this would only trigger in init process, but implementing this would need to complicate the logger destroy process.

main.go
```diff
@@ -35,23 +35,11 @@ package main
...
-	defer log.Destroy() // nolint:errcheck
-	// This recovers the panics only for the functions that run within this context. all the go-routine
-	// spawned by this function need to handle their panics separately if required. Also the FUSE callbacks
-	// wouldn't run in this context, so the panics originated from the callbacks can't get recovered here.
-	defer func() {
-		if panicErr := recover(); panicErr != nil {
-			log.Crit("PANIC: %v", panicErr)
-			panic(panicErr)
-		}
-	}()
```
souravgupta-msft, line 54: In case of panic, will the stack trace be also logged. Considering we are not running in foreground mode.

syeleti-msft: No, It will be not logged. Previously as well this could only log the stack trace while mount init steps have created some panic. Because this cannot catch the panic coming from the other go routines that were responsible for libfuse callbacks or any other go routines spawned by blobfuse. Hence this is not much helpful here, as we don't see any panics in init phase of mount and also removing it from here would make the logger init and destroy at the same place

souravgupta-msft (approved):

PR #2195, opened 2026-04-17
Data locality changes

syeleti-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

- **Feature / Bug Fix**:

## Checklist

## Related Links
- [Issues](<link>)


component/azstorage/azstorage.go
```diff
@@ -690,6 +690,9 @@ func init() {
 	preserveACL := config.AddBoolFlag("preserve-acl", false, "Preserve ACL and Permissions set on file during updates")
 	config.BindPFlag(compName+".preserve-acl", preserveACL)
 
+	blobLayoutAwareRouting := config.AddBoolFlag("blob-layout-aware-routing", false, "Uses GetBlobLayout API to route read requests to the optimal endpoint based on the layout of the blob. Default: False")
+	config.BindPFlag(compName+".blob-layout-aware-routing", blobLayoutAwareRouting)
```
souravgupta-msft, line 695: We should not expose this parameter to customer. Let's keep this as part of internal config option where can decide if we need to use this feature

syeleti-msft: this option will be removed once we merge this feature to main.

internal/attribute.go
```diff
@@ -35,11 +35,46 @@ package internal
...
+		if l.LayoutRanges[mid].End < offset {
+			left = mid + 1
+		} else {
+			right = mid
+		}
+	}
+
+	// Range is guaranteed to exist, return its endpoint
+	return l.LayoutRanges[left].Endpoint
+}
```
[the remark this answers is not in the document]
syeleti-msft: it's fair assumption that offset would always fall in some range, as blob is not getting modified externally.

component/azstorage/azauth_test.go
```diff
@@ -142,11 +143,18 @@ func generateEndpoint(useHttp bool, accountName string, accountType AccountType)
...
+		} else {
+			endpoint += ".dfs.core.windows.net/"
+		}
+	} else {
+		if storageTestConfigurationParameters.UsePreprod {
+			endpoint += ".blob.preprod.core.windows.net/"
+		} else {
+			endpoint += ".blob.core.windows.net/"
+		}
 	}
```
jainakanksha-msft, line 157: Can you please state the whole exact endpoint we want here? 
I feel something incorrect that's why checking with you.
abc.dfs.core.windows.net
&
abc.blob.core.windows.net

component/azstorage/layout.go
```diff
@@ -0,0 +1,80 @@
...
+			}
+			layoutRanges = append(layoutRanges, lr)
+		}
+	}
+
+	lr.layout = &internal.Layout{
+		LayoutRanges: layoutRanges,
+	}
+
+	return lr, nil
```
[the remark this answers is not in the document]
syeleti-msft: nice catch

component/azstorage/layout.go
```diff
@@ -0,0 +1,80 @@
...
+			endpoints[*ep.Index] = *ep.Value
+		}
+		for _, r := range resp.Ranges.Range {
+			lr := internal.LayoutRange{
+				Start:    *r.Start,
+				End:      *r.End,
+				Endpoint: endpoints[*r.EndpointIndex],
+			}
+			layoutRanges = append(layoutRanges, lr)
+		}
```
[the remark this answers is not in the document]
syeleti-msft: nice catch

component/azstorage/azstorage.go
souravgupta-msft: Let's add a TODO here

component/azstorage/layout_test.go
```diff
@@ -0,0 +1,348 @@
...
+			"azstorage:\n"+
+				"  account-name: %s\n"+
+				"  endpoint: https://%s.blob.preprod.core.windows.net/\n"+
+				"  type: block\n"+
+				"  account-key: %s\n"+
+				"  mode: key\n"+
+				"  container: %s\n"+
+				"  fail-unsupported-op: true\n"+
+				"  blob-layout-aware-routing: true",
+			storageTestConfigurationParameters.BlockAccount,
```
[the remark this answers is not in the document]
syeleti-msft: @copilot apply changes based on [this feedback](https://github.com/Azure/azure-storage-fuse/pull/2195#discussion_r3152033020)

component/azstorage/azauth_test.go
syeleti-msft: one possible endpoint may look like this 
https://acnt.blob.core.windows.net/
to run the unit tests on preprod acnts the endpoint must look something like this https://acnt.blob.preprod.core.windows.net/ , since our tests hardcode prod url's changed this behaviour in tests

component/azstorage/block_blob.go
```diff
@@ -499,6 +496,51 @@ func (bb *BlockBlob) getAttrUsingRest(name string) (attr *internal.ObjAttr, err
...
+			// Since block blob does not support acls, we set mode to 0 and FlagModeDefault to true so the fuse layer can return the default permission.
+			attr = &internal.ObjAttr{
+				Path:   name, // We don't need to strip the prefixPath here since we pass the input name
+				Name:   filepath.Base(name),
+				Size:   layoutResp.contentLength,
+				Mode:   0,
+				Mtime:  *layoutResp.lmt,
+				Atime:  *layoutResp.lmt,
+				Ctime:  *layoutResp.lmt,
+				Crtime: *layoutResp.lmt,
```
souravgupta-msft: GetLayout also returns creation time. This should point to that

component/azstorage/block_blob.go
```diff
@@ -593,6 +635,19 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
 		}
 	}
 
+	if bb.Config.isBlobLayoutAwareRoutingEnabled &&
+		attr != nil &&
+		attr.Layout == nil &&
+		attr.Mode&os.ModeDir == 0 {
```
souravgupta-msft, line 644: If layout is not enabled in the backend and virtual-directory is set to false, we first make call to GetLayout which fails and then we make GetProperties call to fetch blob's attributes. Now in this section we again make GetLayout call as attr.Layout is nil. So, we should prevent duplicate calls in cases like these

souravgupta-msft: For virtual-directory set to true, we are making additional GetLayout call below. Let's document this mentioning different cases and calls being made calling out that the best case comes when virtual-directory is set to false given that marker blobs exist for directories.

syeleti-msft: >If layout is not enabled in the backend and virtual-directory is set to false, we first make call to GetLayout which fails and then we make GetProperties call to fetch blob's attributes. Now in this section we again make GetLayout call as attr.Layout is nil. So, we should prevent duplicate calls in cases like these

Currently we don't need to optimize as the changes will only be tested on accounts that have data locality enabled. once the feature is GA then there wouldn't be this problem.

>For virtual-directory set to true, we are making additional GetLayout call below. Let's document this mentioning different cases and calls being made calling out that the best case comes when virtual-directory is set to false given that marker blobs exist for directories.

Sure

component/azstorage/layout.go
```diff
@@ -0,0 +1,92 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore/runtime"
+	"github.com/Azure/azure-sdk-for-go/sdk/storage/azblob/blob"
+	"github.com/Azure/azure-storage-fuse/v2/internal"
+)
+
+type layoutResp struct {
+	layout        *internal.Layout
+	contentLength int64
+	contentMD5    []byte
+	lmt           *time.Time
```
souravgupta-msft: Add creation time in layout response

component/azstorage/layout.go
```diff
@@ -0,0 +1,80 @@
...
+		endpoints := make([]string, len(resp.Endpoints.Endpoint))
+		for _, ep := range resp.Endpoints.Endpoint {
+			endpoints[*ep.Index] = *ep.Value
+		}
+		for _, r := range resp.Ranges.Range {
+			lr := internal.LayoutRange{
+				Start:    *r.Start,
+				End:      *r.End,
+				Endpoint: endpoints[*r.EndpointIndex],
+			}
```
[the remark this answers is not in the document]
syeleti-msft: @copilot apply changes based on [this feedback](https://github.com/Azure/azure-storage-fuse/pull/2195#discussion_r3152032948)

component/azstorage/layout.go
```diff
@@ -0,0 +1,80 @@
...
+	for pager.More() {
+		resp, err := pager.NextPage(ctx)
+		if err != nil {
+			return nil, err
+		}
+
+		contentLength = *resp.BlobContentLength
+		if eTag == nil {
+			eTag = resp.ETag
+		}
```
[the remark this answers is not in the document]
syeleti-msft: @copilot apply changes based on [this feedback](https://github.com/Azure/azure-storage-fuse/pull/2195#discussion_r3152033081)

component/azstorage/utils.go
```diff
@@ -109,7 +109,9 @@ func getAzStorageClientOptions(conf *AzStorageConfig) (azcore.ClientOptions, err
 		log.Err("utils::getAzStorageClientOptions : Failed to create transport client [%s]", err.Error())
 	}
 
-	perCallPolicies := []policy.Policy{telemetryPolicy}
+	layoutPolicy := NewLayoutPolicy()
+
+	perCallPolicies := []policy.Policy{telemetryPolicy, layoutPolicy}
```
souravgupta-msft: Shouldn't adding this layout policy be done behind that boolean flag

component/azstorage/layout.go
```diff
@@ -0,0 +1,111 @@
...
+				Endpoint: endpoints[epIdx],
+			})
+		}
+	}
+
+	if lr != nil {
+		// Sort the layout ranges by start offset to make sure they are in order.
+		sort.Slice(layoutRanges, func(i, j int) bool {
+			return layoutRanges[i].Start < layoutRanges[j].Start
+		})
```
souravgupta-msft, line 100: Is sorting needed here? We should check if the service returns the layout ranges in sorted order. If I recall correctly, in GO SDK sorting is not implemented for layouts.

syeleti-msft: In the spec document, I didn't find any where the ranges would be sorted. I confirmed with the ABFS team, they are sorting the results that they get

component/azstorage/block_blob.go
```diff
@@ -480,12 +482,7 @@ func (bb *BlockBlob) RenameDirectory(source string, target string) error {
 func (bb *BlockBlob) getAttrUsingRest(name string) (attr *internal.ObjAttr, err error) {
 	log.Trace("BlockBlob::getAttrUsingRest : name %s", name)
 
-	blobClient := bb.Container.NewBlockBlobClient(filepath.Join(bb.Config.prefixPath, name))
-	prop, err := blobClient.GetProperties(context.Background(), &blob.GetPropertiesOptions{
-		CPKInfo: bb.blobCPKOpt,
-	})
-
-	if err != nil {
+	parseError := func(err error) (attr *internal.ObjAttr, retErr error) {
```
jainakanksha-msft: ParseError function should return nil and actual error.
attr is not defined yet. So, no need to return attr I think.

component/azstorage/block_blob.go
```diff
@@ -499,6 +496,51 @@ func (bb *BlockBlob) getAttrUsingRest(name string) (attr *internal.ObjAttr, err
...
+				return parseError(err)
+			}
+		} else {
+			// Since block blob does not support acls, we set mode to 0 and FlagModeDefault to true so the fuse layer can return the default permission.
+			attr = &internal.ObjAttr{
+				Path:   name, // We don't need to strip the prefixPath here since we pass the input name
+				Name:   filepath.Base(name),
+				Size:   layoutResp.contentLength,
+				Mode:   0,
+				Mtime:  *layoutResp.lmt,
```
jainakanksha-msft: Please rename lmt to lastModifiedTime.
Avoid such abbreviation in code

component/azstorage/block_blob.go
```diff
@@ -951,7 +1021,7 @@ func (bb *BlockBlob) ReadBuffer(name string, offset int64, length int64) ([]byte
 
 // ReadInBuffer : Download specific range from a file to a user provided buffer
 // Specifying "0" len will download the entire blob.
-func (bb *BlockBlob) ReadInBuffer(name string, offset int64, length int64, data []byte, etag *string) error {
+func (bb *BlockBlob) ReadInBuffer(name string, offset int64, length int64, data []byte, etag *string, layout *internal.Layout) error {
```
jainakanksha-msft: Rather than adding 1 more parameter, shall we optimize this to pass options.

component/azstorage/block_blob.go
syeleti-msft: attr is defined in getAttr func return parameters and will be nil. but it would be better to return nil as you said.

component/azstorage/block_blob.go
syeleti-msft: Sure

component/azstorage/block_blob.go
syeleti-msft: leaving this for now, to make the change concise. we can do this later as there are other functions too which can be passed with an options bag

component/azstorage/block_blob.go
```diff
@@ -951,7 +1021,7 @@ func (bb *BlockBlob) ReadBuffer(name string, offset int64, length int64) ([]byte
 
 // ReadInBuffer : Download specific range from a file to a user provided buffer
 // Specifying "0" len will download the entire blob.
-func (bb *BlockBlob) ReadInBuffer(name string, offset int64, length int64, data []byte, etag *string) error {
+func (bb *BlockBlob) ReadInBuffer(name string, offset int64, length int64, data []byte, etag *string, layout *common.Layout) error {
```
jainakanksha-msft, line 1027: It would be better to pass options structure rather than adding 1 more parameter in the method.

common/types.go
```diff
@@ -359,3 +360,48 @@ var azureSpecialContainers = map[string]bool{
...
 	"changefeed": true,
 }
+
+const layoutValidityDuration = 5 * time.Minute
+
+type LayoutRange struct {
+	Start    int64
+	End      int64
+	Endpoint string
+}
```
souravgupta-msft: nit: would this be better to add in util.go?

component/azstorage/block_blob.go
```diff
@@ -480,25 +482,65 @@ func (bb *BlockBlob) RenameDirectory(source string, target string) error {
 func (bb *BlockBlob) getAttrUsingRest(name string) (attr *internal.ObjAttr, err error) {
 	log.Trace("BlockBlob::getAttrUsingRest : name %s", name)
 
-	blobClient := bb.Container.NewBlockBlobClient(filepath.Join(bb.Config.prefixPath, name))
-	prop, err := blobClient.GetProperties(context.Background(), &blob.GetPropertiesOptions{
-		CPKInfo: bb.blobCPKOpt,
-	})
-
-	if err != nil {
+	parseError := func(err error) (attr *internal.ObjAttr, retErr error) {
```
souravgupta-msft: If we are not using these variables in these functions, let's update this
```suggestion
	parseError := func(err error) (*internal.ObjAttr, error) {
```

component/azstorage/block_blob.go
```diff
@@ -480,25 +482,65 @@ func (bb *BlockBlob) RenameDirectory(source string, target string) error {
...
+		}
+	}
+
+	if bb.Config.isBlobLayoutAwareRoutingEnabled {
+		// Get the properties from layout call.
+		layoutResp, err := bb.getBlobLayout(name)
+		sc := bloberror.GetStatusCode(err)
+		if err != nil {
+			if sc == 400 || sc >= 500 { // fall back to old behavior if service doesn't support layout or layout wasn't fetched
+				log.Warn("BlockBlob::getAttrUsingRest : Layout call failed for %s with error [%v]. Falling back to get properties call", name, err)
```
souravgupta-msft, line 508: We should add a TODO to disable the `isBlobLayoutAwareRoutingEnabled` flag if layout is not supported by service.

component/azstorage/block_blob.go
```diff
@@ -480,25 +482,65 @@ func (bb *BlockBlob) RenameDirectory(source string, target string) error {
...
+			// Since block blob does not support acls, we set mode to 0 and FlagModeDefault to true so the fuse layer can return the default permission.
+			attr = &internal.ObjAttr{
+				Path:   name, // We don't need to strip the prefixPath here since we pass the input name
+				Name:   filepath.Base(name),
+				Size:   layoutResp.contentLength,
+				Mode:   0,
+				Mtime:  *layoutResp.lastModifiedTime,
+				Atime:  *layoutResp.lastModifiedTime,
+				Ctime:  *layoutResp.lastModifiedTime,
+				Crtime: bb.dereferenceTime(layoutResp.crtime, *layoutResp.lastModifiedTime),
```
souravgupta-msft, line 522: Do we see cases of creation time not returned by service?

component/azstorage/block_blob.go
```diff
@@ -593,6 +635,34 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
...
 
+	if bb.Config.isBlobLayoutAwareRoutingEnabled &&
+		attr != nil &&
+		attr.Layout == nil &&
+		attr.Mode&os.ModeDir == 0 {
+		// Layout-aware routing: GetAttr can be reached via two paths, depending on the virtual-directory setting:
+		//
+		// 1. virtual-directory=false (marker blobs exist for directories):
+		//    getAttrUsingRest() is called, which issues GetLayout first. If GetLayout succeeds, attr.Layout
+		//    is already populated and this block is skipped. If GetLayout fails (service unsupported / error),
```
souravgupta-msft:
```suggestion
		//    is already populated and this blob is skipped. If GetLayout fails (service unsupported / error),
```

component/block_cache/block_cache.go
```diff
@@ -1481,6 +1486,12 @@ func (bc *BlockCache) waitAndFreeUploadedBlocks(handle *handlemap.Handle, cnt in
 func (bc *BlockCache) upload(item *workItem) {
 	fileName := fmt.Sprintf("%s::%v", item.handle.Path, item.block.id)
 
+	// invalidate the layout of the blob before doing any updates to the blob.
+	layout := item.handle.Layout
+	if layout != nil {
+		layout.Invalidate()
+	}
```
souravgupta-msft, line 1493: Before update, here we are invalidating the layouts. Now in `ReadInBuffer` when we make this call,
`endpoint := layout.GetIdealEndpoint(offset)`
This will return empty string as layout is invalid. Are we validating the layouts anywhere again if attribute cache entry for this file is still present.

component/azstorage/block_blob.go
syeleti-msft: i dont exactly remember but this is how we are doing in other places. btw we don't use this creation time anywhere. I mean creation time is not a used while sending back the attributes to fuse.

component/azstorage/block_blob.go
syeleti-msft: this can be done in other PR. don't want to add more changes to the PR.

common/types.go
syeleti-msft: Sure

component/azstorage/block_blob.go
syeleti-msft: Sure

PR #2199, opened 2026-04-30
Fix duplicated prefix in CPK block blob attribute lookup for subdirectory mounts

imddanish:
## Summary
Fix a block blob path handling bug in the CPK plus subdirectory flow.

When blob names returned from the list API already include prefixPath, the CPK branch passed the full name into getAttrUsingRest, which then joined prefixPath again and built an incorrect blob path.

This change strips prefixPath before calling getAttrUsingRest, aligning the CPK path with the existing non-CPK handling.

## Root Cause
- List results return blobInfo.Name as the full path inside the container, including the configured prefix or subdirectory.
- The CPK branch in getBlobAttr forwarded that full path unchanged.
- getAttrUsingRest joins prefixPath with the incoming name, so the resulting lookup duplicated the prefix.

## Validation
- Full project build: `go build -tags fuse3 ./...`
- Static analysis: `go vet -tags fuse3 ./component/azstorage/...`
- Unit tests: `go test -v -timeout=2m -run TestUtils/TestRemovePrefixPath ./component/azstorage/ --tags=unittest,fuse3`
- Format check: `gofmt -s -l -d .`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 **Features**
 
 **Bug Fixes**
+- Fix CPK-encrypted blob attribute lookup duplicating the prefix path when subdirectory is configured ([PR #2199](https://github.com/Azure/azure-storage-fuse/pull/2199))
 
 ## 2.5.3 (2026-03-25)
 **Features**
--- a/component/azstorage/block_blob.go
+++ b/component/azstorage/block_blob.go
@@ -702,7 +702,7 @@ func (bb *BlockBlob) processBlobItems(blobItems []*container.BlobItem) ([]*inter
 func (bb *BlockBlob) getBlobAttr(blobInfo *container.BlobItem) (*internal.ObjAttr, error) {
 	if blobInfo.Properties.CustomerProvidedKeySHA256 != nil && *blobInfo.Properties.CustomerProvidedKeySHA256 != "" {
 		log.Trace("BlockBlob::List : blob is encrypted with customer provided key so fetching metadata explicitly using REST")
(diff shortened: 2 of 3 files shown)


syeleti-msft:
Hi @imddanish, thanks for the contribution. can you add unit test for the change?

imddanish:
@microsoft-github-policy-service agree company="Microsoft"

imddanish:
Added tests for the change:
 - `TestGetAttrFileWithCPKEnabledAndPrefixPath` — validates `GetAttr` resolves the correct path when CPK is enabled with a prefix path.
 - `TestReadDirWithCPKEnabledAndPrefixPath` — validates `ReadDir` (which goes through `getBlobAttr`) returns entries without duplicating the prefix path when CPK is enabled.
 
 These follow the existing test pattern in block_blob_test.go. I've verified the build, formatting, and `go vet` pass, but I couldn't run the tests locally as they require Azure Storage credentials. Please let me know if any adjustments are needed after CI runs.

syeleti-msft (approved):

souravgupta-msft (approved):

souravgupta-msft:
@imddanish, thanks for the bug fix. Can you please add an entry to the changelog for this fix.

imddanish:
> @imddanish, thanks for the bug fix. Can you please add an entry to the changelog for this fix.

Added the changelog @souravgupta-msft please reapprove the PR

syeleti-msft (approved):

souravgupta-msft (approved):

PR #2221, opened 2026-05-21
Return correct posix error ENAMETOOLONG for dir creation in ADLS

syeleti-msft:
## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

- **Feature / Bug Fix**:
correct the error from EIO to ENAMETOOLONG when the directory creation fails with pathTooDeep error in ADLS

### DIFF
--- a/component/azstorage/datalake.go
+++ b/component/azstorage/datalake.go
@@ -278,6 +278,9 @@ func (dl *Datalake) CreateDirectory(name string) error {
 		case ErrFileAlreadyExists:
 			log.Err("Datalake::CreateDirectory : Path already exists for %s [%s]", name, err.Error())
 			return syscall.EEXIST
+		case ErrPathTooDeep:
+			log.Err("Datalake::CreateDirectory : Path is too deep for %s [%s]", name, err.Error())
+			return syscall.ENAMETOOLONG
 		default:
 			log.Err("Datalake::CreateDirectory : Failed to create directory %s [%s]", name, err.Error())
 			return err
--- a/component/azstorage/utils.go
+++ b/component/azstorage/utils.go
@@ -240,6 +240,7 @@ const (
 	InvalidRange
 	BlobIsUnderLease
 	InvalidPermission
+	ErrPathTooDeep
 )
 
 // For detailed error list refer below link,
@@ -285,6 +286,8 @@ func storeDatalakeErrToErr(err error) uint16 {
 			return BlobIsUnderLease
 		case datalakeerror.AuthorizationPermissionMismatch:
 			return InvalidPermission
+		case datalakeerror.PathIsTooDeep:
+			return ErrPathTooDeep
 		default:
 			return ErrUnknown
 		}
--- a/test/e2e_tests/dir_test.go
+++ b/test/e2e_tests/dir_test.go
@@ -45,6 +45,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"syscall"
 	"testing"
 	"time"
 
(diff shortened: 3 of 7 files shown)


test/e2e_tests/dir_test.go
```diff
@@ -255,6 +256,30 @@ func (suite *dirTestSuite) TestDirDeleteNonEmpty() {
...
+		fmt.Println("Skipping TestDirCreateDeepPath: not an ADLS account")
+		return
+	}
+
+	// ADLS enforces a maximum path depth of 63 segments from the container root.
+	// Build 65 nested levels from testPath to guarantee the limit is exceeded.
+	const depth = 65
+	topDir := suite.testPath + "/deep"
+	deepPath := topDir
+	for i := range depth {
```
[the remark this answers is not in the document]
syeleti-msft: not true, valid example: https://go.dev/play/p/XBT7WLvQdQ7

souravgupta-msft (commented):
Good fix overall — clean, well-scoped, and the unit test coverage is thorough. A few inline notes below.

component/azstorage/utils.go
```diff
@@ -240,6 +240,7 @@ const (
 	InvalidRange
 	BlobIsUnderLease
 	InvalidPermission
+	ErrPathTooDeep
```
souravgupta-msft, line 243: Minor nit: sibling constants in this block use noun phrases without an `Err` prefix (`InvalidRange`, `BlobIsUnderLease`, `InvalidPermission`). Consider renaming to `PathTooDeep` for consistency.

component/azstorage/datalake.go
```diff
@@ -278,6 +278,9 @@ func (dl *Datalake) CreateDirectory(name string) error {
 		case ErrFileAlreadyExists:
 			log.Err("Datalake::CreateDirectory : Path already exists for %s [%s]", name, err.Error())
 			return syscall.EEXIST
+		case ErrPathTooDeep:
```
souravgupta-msft, line 281: The ADLS SDK will also return `PathIsTooDeep` when creating a *file* at a too-deep path. Should `CreateFile` (and potentially `RenameFile`/`CreateLink`) get the same treatment so they surface `ENAMETOOLONG` instead of falling through to the generic `EIO`?

test/e2e_tests/dir_test.go
```diff
@@ -255,6 +256,30 @@ func (suite *dirTestSuite) TestDirDeleteNonEmpty() {
...
 
+// # Create directory path exceeding ADLS depth limit
+func (suite *dirTestSuite) TestDirCreateDeepPath() {
+	if !suite.adlsTest {
+		fmt.Println("Skipping TestDirCreateDeepPath: not an ADLS account")
+		return
+	}
+
+	// ADLS enforces a maximum path depth of 63 segments from the container root.
+	// Build 65 nested levels from testPath to guarantee the limit is exceeded.
```
souravgupta-msft, line 267: The comment says the limit is 63 segments *from the container root*, but the test builds 65 levels relative to `suite.testPath` — which may itself already be several levels deep within the container. The test still works correctly (65 >> 63 - depth(testPath)), but the comment slightly overstates the precision. Consider noting that 65 was chosen to comfortably exceed the limit regardless of `testPath`'s own depth.

component/azstorage/utils.go
syeleti-msft: err as a prefix is more meaningful imo for const errors naming.

test/e2e_tests/dir_test.go
syeleti-msft: MkdirAll function in go returns the first failure, so the test wouldn't fail.

component/azstorage/datalake.go
syeleti-msft: this must be investigated more. lets address this in another PR

souravgupta-msft: You can add a TODO for this (if not done already), so that we don't forget about it later

souravgupta-msft (approved):

component/azstorage/datalake.go
syeleti-msft: filed an issue https://github.com/Azure/azure-storage-fuse/issues/2241

PR #2240, opened 2026-06-03
feat: expose attr_cache max-files as a CLI flag (--attr-cache-max-files)

### LINKED ISSUE #2218: GetBlobProperties API calls continue even after all data is cached in blobfuse2 cache mode
Hello,

I am running model training in an AKS environment. To minimize Blob Storage API calls, I use blobfuse2 cache mode so that the data required for training is cached on local NVMe storage.

However, even after all training data has been cached, `GetBlobProperties` API calls continue to occur during training.

Both `file cache` and `attr cache` are enabled, and their timeout values are set to sufficiently long durations. The local storage size is also large enough compared to the dataset size.

The scenario is as follows:

1. Before training, all training data is accessed during the data loading/warm-up phase, and all required training data is cached. During this phase, both `GetBlob` and `GetBlobProperties` API calls occur. I believe this is expected behavior. The total number of data files is approximately 13 million.
2. During training, around 20,000 `GetBlobProperties` API calls occur per minute. This is the issue I am facing. Fortunately, `GetBlob` API calls do not occur during this phase.

`GetBlobProperties` API calls are charged as other operations, and around 20,000 calls per minute can result in a significant cost for us.

Therefore, I would like to prevent `GetBlobProperties` API calls from occurring during training as well. Is there any way to achieve this?

## Configuration

- Azure Blob Storage: Standard Storage V2
- HNS enabled: Yes
- Blobfuse2 version: v2.5.2
- PersistentVolume mountOptions:

```yaml
- -o allow_other
- --read-only=true
- --use-adls=true # Use ADLS Gen2 for Hierarchical Namespace structure (HNS)
- --tmp-path=/mnt/nvme # NVMe temp disk cache path
- --use-attr-cache=true
- --attr-cache-timeout=1209600 # 14 days
- --file-cache-timeout=1209600
- -o attr_timeout=1209600
- -o entry_timeout=1209600
- -o negative_timeout=1209600
- --cancel-list-on-mount-seconds=0
```

Please let me know if you need any additional information.

I look forward to your response.

Thank you.

mvanhorn:
## Summary
Adds an `--attr-cache-max-files` CLI flag to the attribute cache component. The flag is bound to the existing `attr_cache.max-files` config key in `init()`, mirroring the existing `attr-cache-timeout` flag registration. It defaults to `defaultMaxFiles` (5,000,000), so behavior is identical when the flag is not passed.

## Why this matters
Issue #2218 reports that `GetBlobProperties` calls continue (~20,000/min) even after a large dataset is fully cached, during AKS model training over ~13M files. The cause is that the attribute cache is capped at `defaultMaxFiles = 5,000,000`; files beyond that cap are never cached, so every access re-issues a GetProperties call.

The cap is already configurable through the YAML key `attr_cache.max-files`, but there was no command-line flag. PersistentVolume / CSI `mountOptions` users pass flags rather than a config file and therefore could not raise the limit. The maintainer (@syeleti-msft) committed to exposing this as a CLI option in the next release; this change does exactly that, keeping scope to the flag binding only (no change to the default or to cache/eviction logic).

## Testing
- `GOOS=linux GOARCH=amd64 go build ./component/attr_cache/...` — clean.
- `GOOS=linux GOARCH=amd64 go vet ./component/attr_cache/...` — clean.
- `GOOS=linux GOARCH=amd64 go test -c ./component/attr_cache/` — test binary compiles (existing config tests assert `max-files` via YAML and are unaffected; the flag default preserves current behavior).
- The change is additive flag registration in `init()`; `Configure()` already keys off `config.IsSet(compName + ".max-files")`.

Fixes #2218

AI was used for assistance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 ## 2.5.4 (Unreleased)
 **Features**
 - Make Blobfuse2 binary FIPS compliant by building with the Microsoft Go toolchain (`systemcrypto` GOEXPERIMENT) and `CGO_ENABLED=1`, routing all `crypto/*` calls through the system OpenSSL FIPS provider ([PR #2226](https://github.com/Azure/azure-storage-fuse/pull/2226))
+- Expose the attribute cache max-files limit as a `--attr-cache-max-files` CLI flag so mountOptions/CSI users can raise the cache cap without a config file ([#2218](https://github.com/Azure/azure-storage-fuse/issues/2218))
 
 **Bug Fixes**
 - Fix CPK-encrypted blob attribute lookup duplicating the prefix path when subdirectory is configured ([PR #2199](https://github.com/Azure/azure-storage-fuse/pull/2199))
--- a/component/attr_cache/attr_cache.go
+++ b/component/attr_cache/attr_cache.go
@@ -688,6 +688,9 @@ func init() {
 	attrCacheTimeout := config.AddUint32Flag("attr-cache-timeout", defaultAttrCacheTimeout, "attribute cache timeout")
 	config.BindPFlag(compName+".timeout-sec", attrCacheTimeout)
 
+	attrCacheMaxFiles := config.AddIntFlag("attr-cache-max-files", defaultMaxFiles, "Maximum number of file attributes to cache in the attribute cache.")
+	config.BindPFlag(compName+".max-files", attrCacheMaxFiles)
+
 	noSymlinks := config.AddBoolFlag("no-symlinks", false, "whether or not symlinks should be supported")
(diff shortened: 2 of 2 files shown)


syeleti-msft:
@mvanhorn, Thanks for the contribution. After the careful review of the attribute cache component, we think that exposing `max-files` parameter in attribute cache would indeed become a foot gun for the users as it doesn't give a sense on how much underlying memory the attribute cache would consume and may lead to OOM killing of the blobfuse2 process in some cases if it's misused. So, we are going ahead with exposing new `attr-cache-max-size-mb` parameter where you would set upper bound for attribute cache memory. (we'll provide some formula on how to calculate this memory based on the number of attributes you want to cache).

PR #2242, opened 2026-06-03
Mirror Go runtime crash dumps into the Blobfuse2 log file

souravgupta-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [x] Other (describe): Diagnostics / observability — capture runtime crash dumps in the Blobfuse2 log file

## Description

- **Feature / Bug Fix**:

Until now, when the Go runtime aborted a Blobfuse2 process due to an unhandled panic or a fatal runtime error, the resulting stack trace was only written to the per-mount `.trace` file under `$HOME/.blobfuse2/<mount_path>.<pid>.trace` (via the daemon library's stderr redirection). The Blobfuse2 log file (whether the configured `base`-logger file or the rsyslog-managed `/var/log/blobfuse2.log` in `syslog` mode) had no record of the crash, which made post-mortem diagnostics on field deployments — particularly in AKS via the Blob CSI driver — significantly harder because the `.trace` file is per-process and is not co-located with the Blobfuse2 log stream.

This PR mirrors Go runtime crash dumps (panics and fatal errors from **any** goroutine, including those spawned by libfuse callbacks) into the Blobfuse2 log file in addition to the existing `.trace` file, and keeps that crash output reliably attached to the live log file across log rotations.

**What changes:**

1. **`common/log` package — new exported entry point `log.SetupCrashOutput(loggerType, logFilePath)`** that:
   - Uses `runtime/debug.SetCrashOutput` (Go 1.23+) to dup the target log file's fd into the runtime, so crash dumps land in the log file in addition to stderr. Because this operates at the runtime level, it captures panics from any goroutine — including the goroutines libfuse spawns to dispatch FUSE callbacks (which `recover()` in `mount.go` could never reach).
   - For `base` logging, targets the configured log file path.
   - For `syslog` logging (the default, and the mode used by the AKS Blob CSI driver), targets `/var/log/blobfuse2.log` — the rsyslog sink for blobfuse2-tagged messages — so crash dumps land alongside the rest of the Blobfuse2 log stream rather than in an unrelated location.
   - Opens the file without `O_CREATE` so blobfuse2 never races against rsyslog for file ownership; if the file is missing the call logs a warning and gracefully no-ops (the `.trace` file still captures the panic via stderr).
   - Is a no-op for the silent logger and unknown logger types.

2. **Log rotation hook plumbing.** `SetCrashOutput` dups the file descriptor internally, so once a rotator moves the live log file aside the pinned fd still points at the rotated copy — meaning any subsequent crash dump would land in the wrong file. To prevent this, `SetupCrashOutput` re-attaches the fd after rotation through two mechanisms:
   - A new internal `registerLogRotateHook` API in `common/log` invoked from `BaseLogger.LogRotate` immediately after a successful in-process size-based rotation. This covers the `base` logger.
   - A `SIGHUP` handler installed via `signal.Notify` (guarded by `sync.Once` so only one listener is ever started). This handles external rotators that follow the standard `postrotate` convention — `logrotate` on dev boxes, and the AKS Blob CSI driver's host-side rotation hook.

3. **Wiring in `cmd/mount.go`.** The mount command now invokes `log.SetupCrashOutput(...)` once, right after the first startup `log.Crit` block. The ordering matters in `syslog` mode: those initial log lines are what trigger rsyslog to create `/var/log/blobfuse2.log` for the first time, so opening the file before they are emitted would race on a fresh install.

4. **Supporting bits:**
   - New `common.SyslogFilePath` constant pointing at `/var/log/blobfuse2.log`, documented as the rsyslog sink declared in `setup/11-blobfuse2.conf`.
   - `.gitignore` updated to ignore rotated log files (`blobfuse2.log.*`) produced by local dev testing.
   - `CHANGELOG.md` updated under the 2.5.4 **Other Changes** section.

## How Has This Been Tested?

**Unit tests** (`common/log/logger_test.go`, all passing under `go test --tags=unittest,fuse3 ./common/log/...`):

- `TestRegisterLogRotateHook` — exercises the rotate-hook registry: nil hooks are ignored, registered hooks fire in registration order, and re-invocation re-runs all hooks (they are not one-shot).
- `TestBaseLoggerRotateInvokesHook` — end-to-end: registers a hook, calls `LogRotate()` on a `BaseLogger`, and asserts the hook counter increments — proving the hook is invoked from inside `BaseLogger.LogRotate`.
- `TestSetCrashOutput` — covers every branch of the helper: success path for a writable `base` file, early no-op for empty path / `stdout` / `silent` / unknown logger type, graceful Warn-and-return for a non-existent base path, and the `""` / `default` / `syslog` paths that target the rsyslog sink (typically unwritable in test environments — verified to not panic).
- `TestSetupCrashOutputRegistersHookAndHandler` — verifies `SetupCrashOutput` registers exactly one rotate hook, that invoking the registered hooks does not panic, that a second call is safe (`sync.Once` guards the SIGHUP listener), and that `SIGHUP` is actually delivered to the process within 2 s using a sentinel listener.

**Manual / smoke tests:**

- Verified clean build (`./build.sh`).
- Triggered a deliberate panic inside a libfuse callback goroutine via a test mount and confirmed:
  - The per-mount `.trace` file under `$HOME/.blobfuse2/` contains the stack trace (existing behaviour, unchanged).
  - The configured log file (in `base` mode) and `/var/log/blobfuse2.log` (in `syslog` mode) **also** contain the stack trace — the new behaviour.
- Forced a `BaseLogger` in-process rotation and confirmed a subsequent panic landed in the new `blobfuse2.log`, not in the rotated `blobfuse2.log.1`.
- Forced an external `logrotate` rotation and sent `SIGHUP` to the running blobfuse2 process; confirmed the next crash dump landed in the freshly-opened file.

## Checklist

- [x] Updates to module CHANGELOG.md are included.

## Related Links
- [Issues](<link>)


common/log/logger.go
```diff
@@ -232,6 +237,115 @@ func LogRotate() error {
...
+		setCrashOutput(loggerType, logFilePath)
+	})
+
+	installCrashSighupHandler(loggerType, logFilePath)
+}
+
+func setCrashOutput(loggerType, logFilePath string) {
+	var crashFilePath string
+	switch loggerType {
+	case "base":
```
syeleti-msft, line 373: can't this function SetCrashOutput can be a method to interface logger? and each logger implements its own method for its type

common/log/logger.go
```diff
@@ -232,6 +237,115 @@ func LogRotate() error {
...
+	//   1. BaseLogger's in-process size-based rotation invokes the registered rotate hook directly. The rename
+	//      moves the original file aside while our dup'd fd stays pinned to the old inode, so we must re-open
+	//      the live file.
+	//   2. External rotators (e.g. logrotate, AKS CSI driver host) that signal the process via SIGHUP after
+	//      rotating /var/log/blobfuse2.log out of the way -- handled below by installCrashSighupHandler.
+	registerLogRotateHook(func() {
+		setCrashOutput(loggerType, logFilePath)
+	})
+
+	installCrashSighupHandler(loggerType, logFilePath)
```
syeleti-msft: from the comments i understand that this is only necessary for syslogger, but here it seems like we are initializing for every logger. is this necessary for other loggers too? if not can we call it under a condition?

common/log/logger.go
```diff
@@ -232,6 +237,115 @@ func LogRotate() error {
...
+
+	// Re-attach the crash output fd after log rotation, so dumps land in the live log file rather than the
+	// rotated copy. Covers two rotation paths:
+	//   1. BaseLogger's in-process size-based rotation invokes the registered rotate hook directly. The rename
+	//      moves the original file aside while our dup'd fd stays pinned to the old inode, so we must re-open
+	//      the live file.
+	//   2. External rotators (e.g. logrotate, AKS CSI driver host) that signal the process via SIGHUP after
+	//      rotating /var/log/blobfuse2.log out of the way -- handled below by installCrashSighupHandler.
+	registerLogRotateHook(func() {
+		setCrashOutput(loggerType, logFilePath)
```
anmolanmol1234, line 337: Every log rotation will invoke every hook.
Although the current code path probably calls it once during mount, the API itself isn't idempotent. Better to use :-   crashHookOnce.Do(func() {
        registerLogRotateHook(...)
    })

common/log/logger.go
```diff
@@ -232,6 +237,115 @@ func LogRotate() error {
...
+//
+// It also keeps the crash output fd attached to the live log file across rotations:
+//  1. BaseLogger's in-process size-based rotation invokes the registered rotate hook directly.
+//  2. External rotators (logrotate's postrotate, the AKS Blob CSI driver, ...) signal the process
+//     via SIGHUP after rotating the file aside.
+//
+// A no-op for the silent logger or when the target path is not writable. The .trace file still
+// captures the panic in that case.
+func SetupCrashOutput(loggerType, logFilePath string) {
+	setCrashOutput(loggerType, logFilePath)
```
anmolanmol1234: **Concurrency: `setCrashOutput` needs a mutex.**

After `SetupCrashOutput` runs, this function has two independent callers:

1. The rotate hook (fired from `BaseLogger.LogRotate` → `invokeRotateHooks`), running on the log-write goroutine.
2. The SIGHUP handler goroutine installed by `installCrashSighupHandler`.

On a busy node these can execute concurrently — e.g. logrotate's `postrotate: kill -HUP $pid` firing near an in-process size-based rotation.

Inside `setCrashOutput` we open a new fd and hand it to `debug.SetCrashOutput`. The runtime dups the fd, stores it in its internal crash-fd slot, and closes the previously-installed one. That sequence is not atomic across concurrent callers, so two racing goroutines can end up closing the same fd number. Linux recycles fd numbers aggressively, so the second `close` may end up closing an unrelated file descriptor elsewhere in the process. Symptom: sporadic `EBADF` far from the rotation code, effectively impossible to reproduce.

**Suggested fix** — package-level mutex, serializing the whole body:

```go
var crashOutputMu sync.Mutex

func setCrashOutput(loggerType, logFilePath string) {
    crashOutputMu.Lock()
    defer crashOutputMu.Unlock()
    // ... existing body unchanged
}

File: common/log/logger.go [hunk 5]
```diff
@@ -232,6 +237,115 @@ func LogRotate() error {
...
+var sighupOnce sync.Once
+
+// installCrashSighupHandler arranges for the crash output fd to be re-attached when the process receives
+// SIGHUP. This is the conventional signal external log rotators (logrotate's postrotate, the AKS Blob CSI
+// driver's rotation hook, etc.) send after moving the live log file aside, so the next panic dumps into
+// the freshly-opened file rather than the rotated copy. The listener is installed at most once per process.
+func installCrashSighupHandler(loggerType, logFilePath string) {
+	sighupOnce.Do(func() {
+		ch := make(chan os.Signal, 1)
+		signal.Notify(ch, syscall.SIGHUP)
```
anmolanmol1234: do we need to call signal.Stop() once finished ?

common/log/logger_test.go
```diff
@@ -144,6 +152,144 @@ func (lts *LoggerTestSuite) TestNegative() {
...
+// resetCrashOutputState clears process-global state used by the crash-output / rotation-hook
+// machinery so each test starts from a known baseline. logObj is swapped for a fresh silent logger
+// because earlier tests may have called Destroy() on a base logger, which closes its channel and
+// would cause subsequent Warn/Info calls (e.g. from setCrashOutput) to panic. The previously
+// installed SIGHUP listener goroutine (if any) is orphaned -- harmless for a unit-test process.
+func resetCrashOutputState() {
+	rotateHooksMu.Lock()
+	rotateHooks = nil
+	rotateHooksMu.Unlock()
+	sighupOnce = sync.Once{}
```
anmolanmol1234, line 175: Here it resets sighupOnce but leaves the previously installed SIGHUP listener goroutine running. Subsequent tests can install additional listeners, causing multiple handlers to receive the same SIGHUP and making future tests potentially flaky.

common/log/logger_test.go
```diff
@@ -144,6 +152,144 @@ func (lts *LoggerTestSuite) TestNegative() {
...
+
+	// "syslog"/""/"default" branches all target common.SyslogFilePath. In a test environment that
+	// file is usually not writable; the call must still not panic and must not return an error.
+	assert.NotPanics(func() { setCrashOutput("", "") })
+	assert.NotPanics(func() { setCrashOutput("default", "") })
+	assert.NotPanics(func() { setCrashOutput("syslog", "") })
+}
+
+func (lts *LoggerTestSuite) TestSetupCrashOutputRegistersHookAndHandler() {
+	assert := assert.New(lts.T())
```
anmolanmol1234, line 382: This test only verifies that the process received SIGHUP, it doesn't verify that the installed crash-output handler actually executed. Can that check also be added ?

common/log/logger.go
```diff
@@ -232,6 +237,115 @@ func LogRotate() error {
...
+	if err != nil {
+		Warn("log: failed to open %s for crash output [%s]", crashFilePath, err.Error())
+		return
+	}
+	// SetCrashOutput dups the fd, so the file handle can be closed immediately.
+	defer f.Close()
+
+	if err := debug.SetCrashOutput(f, debug.CrashOptions{}); err != nil {
+		Warn("log: failed to set crash output to %s [%s]", crashFilePath, err.Error())
+	}
```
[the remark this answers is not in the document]
souravgupta-msft: Added `recover()` method in the starting to prevent this

common/log/logger.go
souravgupta-msft: Added a condition around this method so that is called only in case of syslog

common/log/logger.go
souravgupta-msft: This is just called once during mount and on each log rotation, the callback method is invoked which is `setCrashOutput` which updates the file descriptor to point to the latest log file which is referred by the log file path.

common/log/logger.go
souravgupta-msft: Updated the code so that `setCrashOutput` method is called from these paths in these modes,
- `base`: called from `invokeRotateHooks` which calls this method in the callback registered
- `syslog`: called from `installCrashSighupHandler` only. `registerLogRotateHook` is no op here

Also added condition around `installCrashSighupHandler` so that it is called only in syslog mode.

common/log/logger.go
souravgupta-msft: No, it's not needed in production path.

The handler installed by `installCrashSighupHandler` is intentionally process-lifetime:
- Guarded by sync.Once - installed once, never re-installed, never torn down.
- Its whole purpose is to receive every SIGHUP for as long as blobfuse2 runs, so each rotation event re-attaches the crash fd. Calling `signal.Stop` would defeat this — the very next logrotate cycle would silently stop working.
- The channel + goroutine are singleton resources roughly the size of a slice header + a scheduler stack; the OS reclaims them at process exit.
- If we ever wanted a "graceful teardown" story (e.g. during unmount before exit), we'd need to close the channel and call `signal.Stop` together — but there's no benefit today because the process exits immediately after `Destroy()` returns.

common/log/logger_test.go
souravgupta-msft: Addressed

common/log/logger_test.go
souravgupta-msft: Addressed

common/log/logger.go
```diff
@@ -232,6 +238,164 @@ func LogRotate() error {
...
+
+	// Open without O_CREATE: in base mode BaseLogger has already created the file; in syslog mode rsyslog owns
+	// the file and must have created it first. If the file is missing (e.g. rsyslog not restarted yet on a fresh
+	// install), let it fail -- the per-mount .trace file still captures the panic via stderr redirection.
+	f, err := os.OpenFile(crashFilePath, os.O_WRONLY|os.O_APPEND, 0)
+	if err != nil {
+		Warn("log: failed to open %s for crash output [%s]", crashFilePath, err.Error())
+		return
+	}
```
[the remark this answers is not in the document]
souravgupta-msft: Done

common/log/logger.go
```diff
@@ -232,6 +238,164 @@ func LogRotate() error {
...
+	// SetCrashOutput dups the fd, so the file handle can be closed immediately.
+	defer f.Close()
+
+	if err := debug.SetCrashOutput(f, debug.CrashOptions{}); err != nil {
+		Warn("log: failed to set crash output to %s [%s]", crashFilePath, err.Error())
+	}
```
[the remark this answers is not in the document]
souravgupta-msft: Done

common/log/logger.go
```diff
@@ -232,6 +238,164 @@ func LogRotate() error {
...
+// invokeRotateHooks runs all registered post-rotation hooks. Called by logger implementations
+// after a successful rotation.
+func invokeRotateHooks() {
+	rotateHooksMu.Lock()
+	hooks := append([]func(){}, rotateHooks...)
+	rotateHooksMu.Unlock()
+	for _, fn := range hooks {
+		fn()
+	}
+}
```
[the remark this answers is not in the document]
souravgupta-msft: Done

common/log/logger.go
```diff
@@ -232,6 +238,174 @@ func LogRotate() error {
...
+
+// invokeRotateHooks runs all registered post-rotation hooks. Called by logger implementations
+// after a successful rotation.
+func invokeRotateHooks() {
+	rotateHooksMu.Lock()
+	hooks := append([]func(){}, rotateHooks...)
+	rotateHooksMu.Unlock()
+	for _, fn := range hooks {
+		fn()
+	}
```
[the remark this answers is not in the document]
souravgupta-msft: Done

common/log/logger.go
```diff
@@ -232,6 +238,174 @@ func LogRotate() error {
...
+
+	setCrashOutput(loggerType, logFilePath)
+
+	// Re-attach the crash output fd after BaseLogger's in-process size-based rotation. The rename
+	// moves the original file aside while our dup'd fd stays pinned to the old inode, so we must
+	// re-open the live file. Registered unconditionally: it is a no-op for logger types whose
+	// LogRotate() implementations never invoke the hook (SysLogger, SilentLogger).
+	registerLogRotateHook(func() {
+		setCrashOutput(loggerType, logFilePath)
+	})
```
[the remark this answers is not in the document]
souravgupta-msft: Added `setupCrashOutputOnce sync.Once` wrapping the whole body of `SetupCrashOutput`. First successful call wins, subsequent calls are no-ops.

common/log/logger.go
```diff
@@ -232,6 +238,174 @@ func LogRotate() error {
 	return logObj.LogRotate()
 }
 
+// rotateHooksMu guards rotateHooks against concurrent registration and invocation.
+var rotateHooksMu sync.Mutex
+var rotateHooks []func()
```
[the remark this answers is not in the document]
souravgupta-msft: Added `setupCrashOutputOnce` to make the `SetupCrashOutput` method idempotent.

common/log/logger_test.go
```diff
@@ -144,6 +152,268 @@ func (lts *LoggerTestSuite) TestNegative() {
...
+//
+// The previously-installed SIGHUP listener goroutine (if any) is stopped: signal.Stop unregisters
+// the channel from the runtime signal dispatch, and closing the channel lets the "for range"
+// goroutine drain the buffer and exit. Without this, tests that call SetupCrashOutput repeatedly
+// would leak goroutines still registered for SIGHUP, and future SIGHUPs would fan out to all of
+// them, making later tests flaky.
+func resetCrashOutputState() {
+	rotateHooksMu.Lock()
+	rotateHooks = nil
+	rotateHooksMu.Unlock()
```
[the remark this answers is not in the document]
souravgupta-msft: Done, resetting the `setupCrashOutputOnce` here

common/log/logger.go
souravgupta-msft: Considered it but decided against making `SetupCrashOutput` an interface method:
- The crash-output plumbing is not logger behaviour — it manipulates runtime-level state (`runtime/debug.SetCrashOutput` dups an fd into the Go runtime, `signal.Notify` registers with the process-wide signal handler, and the rotate-hook registry is a package-level list). None of that is per-logger-instance state.
- SysLogger doesn't own a file at all; its crash target is the rsyslog sink at a well-known path — that's a policy decision outside the logger implementation, not something the logger interface should encode.
- SilentLogger would need a no-op method purely to satisfy the interface, and every future logger implementation would have to remember to implement this cross-cutting concern.
- The Logger interface today is a clean "how to write log lines" abstraction. Adding a lifecycle method for runtime crash routing muddles it.

Keeping it as a free `log.SetupCrashOutput(loggerType, logFilePath)` that dispatches internally based on loggerType (with crashOutputTarget + isSyslogFamily helpers). Happy to reconsider if there's a case I'm missing.

syeleti-msft (approved):

PR #2245, opened 2026-06-07
attr_cache: replace entry-count limit with memory-bounded LRU and background TTL sweeper

### LINKED ISSUE #2218: GetBlobProperties API calls continue even after all data is cached in blobfuse2 cache mode

syeleti-msft:
### Summary

- **Memory-bounded attr-cache** — the attribute cache now uses an LRU eviction policy with a configurable memory limit, replacing the previous unbounded map that could cause OOM kills on large workloads
- **Auto-tuned default** — when `max-size-mb` is unset (or set to `0`), the limit is auto-tuned to 1% of total system RAM, clamped to [64 MB, 1 GB], giving deterministic per-machine sizing without operator intervention
- **Background TTL sweeper** — a goroutine reclaims memory from expired entries when the cache is idle, with a fixed idle gate to avoid holding the write lock under active traffic
- **Hot reload removed** — `OnConfigChange` is now a no-op; config changes require a restart, eliminating a class of lifecycle races

### Configuration

**Config file:**
```yaml
attr_cache:
  timeout-sec: 120
  max-size-mb: 256   # omit or set to 0 for auto-tuning
```

**CLI flag:**
```
--attr-cache-max-size-mb=256
```

**Behaviour by value:**

| Value | Behaviour |
|---|---|
| omitted | auto-tune: 1% of total RAM, clamped [64 MB, 1 GB] |
| `0` | same as omitted — auto-tune |
| `N > 0` | hard limit of N MB |

**Cache capacity estimates** (52-byte path, 20-byte ETag, 16-byte MD5):

| Limit | Positive entries (~964 B each) | Negative/tombstone entries (~320 B each) |
|---|---|---|
| 64 MB (floor) | ~70 K | ~210 K |
| 1 GB (ceiling) | ~1.1 M | ~3.4 M |

Capacity scales linearly; recalculation instructions are in `cache_item_test.go`.

### LRU changes (`common/cache_policy`)

- Generic `LRU[K, V]` with caller-provided `sizeOf` for accurate two-layer memory accounting (structural overhead + user data)
- `Put` returns `bool` and rejects oversized entries upfront — existing entries are never displaced by a key that can never fit
- `ReplaceIf` is single-pass with in-place update and `MoveToFront`, replacing the previous unsafe two-pass approach
- `SetMaxSize` removed — `maxSize` is immutable after construction, stored as plain `int64` (no atomics needed)
- New ops: `DeleteIf`, `ReplaceIf`, `Peek`, `Range`, `Purge`

### attr_cache changes (`component/attr_cache`)

- All `Put` call sites check the return value and emit `log.Err` on rejection
- `no-cache-on-list` / `cache-on-list` (v1 compat) now correctly honoured in `ReadDir` and `StreamDir`
- `sweepExpired` idle gate: fixed integer division truncation bug (`cacheTimeout/(2*time.Second)` → `cacheTimeout/2`) and added debug log showing size reclaimed
- `max-size-mb` added to `GenConfig` output
- `updateCacheEntry` normalises path with `TruncateDirName` before cache lookup

### New utility (`common/sysinfo.go`)

- `TotalMemoryBytes() uint64` — reads total physical RAM via `syscall.Sysinfo`, with tests

### Config / docs

- Updated CHANGELOG, `baseConfig.yaml`, `advancedConfig.yaml`, and all three sample config files to reflect auto-tuned default and correct entry capacity estimates


component/attr_cache/attr_cache.go
```diff
@@ -143,13 +187,10 @@ func (ac *AttrCache) GenConfig() string {
 	return sb.String()
 }
```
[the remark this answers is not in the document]
syeleti-msft: Updated

component/attr_cache/attr_cache.go
```diff
@@ -158,313 +199,151 @@ func (ac *AttrCache) Configure(_ bool) error {
 	}
 
 	if config.IsSet(compName + ".timeout-sec") {
-		ac.cacheTimeout = conf.Timeout
+		ac.cacheTimeout = time.Duration(conf.Timeout) * time.Second
 	} else {
-		ac.cacheTimeout = defaultAttrCacheTimeout
+		ac.cacheTimeout = time.Duration(defaultAttrCacheTimeout) * time.Second
 	}
```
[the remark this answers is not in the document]
syeleti-msft: Updated the configure function so that hot reload will restart the  background sweeper with new timeout.

component/attr_cache/attr_cache.go
```diff
@@ -158,313 +199,151 @@ func (ac *AttrCache) Configure(_ bool) error {
...
-// ReadDir : Optionally cache attributes of paths returned by next component
+// ReadDir caches attributes of all paths returned by the next component.
 func (ac *AttrCache) ReadDir(options internal.ReadDirOptions) (pathList []*internal.ObjAttr, err error) {
 	log.Trace("AttrCache::ReadDir : %s", options.Name)
 
 	pathList, err = ac.NextComponent().ReadDir(options)
 	if err == nil {
-		ac.cacheAttributes(pathList)
+		ac.lru.cacheAttributes(pathList)
 	}
```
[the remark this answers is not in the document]
syeleti-msft: Added it to use no-cache-on-list option which was ignored previously in the code

component/attr_cache/attr_cache_integration_test.go
```diff
@@ -0,0 +1,914 @@
...
+	dir, err := os.MkdirTemp("", "attr_cache_integration_*")
+	suite.Require().NoError(err)
+	suite.loopbackDir = dir
+
+	cfg := fmt.Sprintf(
+		"attr_cache:\n  timeout-sec: 30\n  max-size-mb: 32\n\nloopbackfs:\n  path: %s\n",
+		dir,
+	)
+	_ = config.ReadConfigFromReader(strings.NewReader(cfg))
+
```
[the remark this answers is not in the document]
syeleti-msft: Updated

component/attr_cache/attr_cache_test.go
```diff
@@ -710,24 +719,24 @@ func (suite *attrCacheTestSuite) TestRenameFile() {
...
 	assertAttributesTransferred(suite, options.SrcAttr, modifiedDstAttr)
 
 	// Src Entry Exist and Dst Entry Don't Exist
 	addPathToCache(suite.assert, suite.attrCache, src, false)
 	// Add negative entry to cache for Dst
-	suite.attrCache.cacheMap[dst] = newAttrCacheItem(&internal.ObjAttr{}, false, time.Now())
-	options.SrcAttr = suite.attrCache.cacheMap[src].attr
-	options.DstAttr = suite.attrCache.cacheMap[dst].attr
+	suite.attrCache.lru.Put(dst, &attrCacheItem{attr: &internal.ObjAttr{}, cachedAt: time.Now()})
+	options.SrcAttr = getCacheItem(suite.attrCache, src).attr
```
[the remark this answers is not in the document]
syeleti-msft: Updated

component/attr_cache/attr_cache.go
```diff
@@ -158,313 +199,151 @@ func (ac *AttrCache) Configure(_ bool) error {
...
-		ac.noSymlinks = conf.NoSymlinks
+	// Propagate the new limit to a running cache (hot-reload support).
+	if ac.lru != nil {
+		ac.lru.SetMaxSize(ac.maxSizeBytes)
 	}
 
-	log.Crit("AttrCache::Configure : cache-timeout %d, symlink %t, max-files %d",
-		ac.cacheTimeout, ac.noSymlinks, ac.maxFiles)
+	log.Crit("AttrCache::Configure : cache-timeout %v, symlink %t, max-size-mb %d",
+		ac.cacheTimeout, ac.noSymlinks, conf.MaxSizeMB)
```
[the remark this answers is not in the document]
syeleti-msft: updated

component/attr_cache/attr_cache.go
```diff
@@ -523,131 +386,115 @@ func (ac *AttrCache) CopyFromFile(options internal.CopyFromFileOptions) error {
...
 	return err
 }
 
-// GetAttr : Try to serve the request from the attribute cache, otherwise cache attributes of the path returned by next component
+// GetAttr serves from cache on hit (promoting the entry to MRU), or fetches from the
+// next component and caches the result on miss.
 func (ac *AttrCache) GetAttr(options internal.GetAttrOptions) (*internal.ObjAttr, error) {
 	log.Trace("AttrCache::GetAttr : %s", options.Name)
+	t := time.Now()
+	ac.lastOp.Store(&t)
```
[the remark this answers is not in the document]
syeleti-msft: Updated the lastOp to store unix seconds, so that it would not cause this excessive garbage on each getAttr or cache update calls. as the precision we work is on is seconds its okay to not store it in time.Time type.

common/cache_policy/lru_cache.go
```diff
@@ -0,0 +1,285 @@
...
+	for elem := l.list.Front(); elem != nil; elem = elem.Next() {
+		item := elem.Value.(*lruItem[K, V])
+		if pred(item.key, item.val) {
+			keys = append(keys, item.key)
+		}
+	}
+	for _, k := range keys {
+		l.put(k, newVal(k))
+	}
+}
```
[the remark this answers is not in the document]
syeleti-msft: Updated, good catch

common/cache_policy/lru_cache_test.go
```diff
@@ -0,0 +1,639 @@
...
+func (s *lruCacheTestSuite) TestReplaceIfPromotesToMRU() {
+	lru := NewLRU[string, int](0, zeroSizeOf[string, int])
+	lru.Put("a", 1)
+	lru.Put("b", 2) // b is MRU
+
+	lru.ReplaceIf(func(k string, _ int) bool { return k == "a" }, func(_ string) int { return 99 })
+
+	// "a" should have been promoted to MRU after replacement
+	v, _ := lru.Get("a")
+	s.assert.Equal(99, v)
```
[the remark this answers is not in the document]
syeleti-msft: updated

component/attr_cache/cache_item.go
```diff
@@ -0,0 +1,201 @@
...
+
+// Delete removes an entry and records cache activity for the idle gate.
+func (l *attrCacheLRU) Delete(key string) {
+	l.touch()
+	l.LRU.Delete(key)
+}
+
+// ReplaceIf replaces matching entries and records cache activity for the idle gate.
+// DeleteIf is intentionally not overridden — it is called only by the TTL sweeper
+// and must not reset the idle gate, otherwise the sweeper would suppress itself.
```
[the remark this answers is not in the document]
syeleti-msft: Updated

common/cache_policy/lru_cache.go
```diff
@@ -0,0 +1,286 @@
...
+		if pred(item.key, item.val) {
+			v := newVal(item.key)
+			newSize := l.sizeOf(item.key, v)
+			l.currSize += newSize - item.userSize
+			item.val = v
+			item.userSize = newSize
+		}
+	}
+	l.evictIfNeeded()
+}
```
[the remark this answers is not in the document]
syeleti-msft: Updated

component/attr_cache/cache_item.go
```diff
@@ -0,0 +1,201 @@
...
+func (l *attrCacheLRU) invalidateDirectory(path string) {
+	// Invalidate all the child entries
+	prefix := internal.ExtendDirName(path)
+	l.DeleteIf(func(key string, _ *attrCacheItem) bool {
+		return strings.HasPrefix(key, prefix)
+	})
+
+	// Invalidate directory entry itself
+	l.invalidatePath(path)
+}
```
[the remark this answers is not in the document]
syeleti-msft: Updated

component/attr_cache/attr_cache.go
```diff
@@ -48,41 +50,50 @@ import (
...
-	cleanupStop  context.CancelFunc
+	cacheTimeout  time.Duration
+	noSymlinks    bool
+	maxSizeBytes  int64
+	lru           *attrCacheLRU
+	stopCh        chan struct{}
+	sweepWg       sync.WaitGroup
+	lastOp        atomic.Int64 // Unix seconds of last LRU operation; 0 = no activity yet
+	noCacheOnList bool
 }
```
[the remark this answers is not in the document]
syeleti-msft: Updated

component/attr_cache/attr_cache_integration_test.go
```diff
@@ -0,0 +1,914 @@
...
+	// Remove from disk then re-create via the cache.
+	os.Remove(filepath.Join(suite.loopbackDir, "newdir"))
+
+	err := suite.attrCache.CreateDir(internal.CreateDirOptions{Name: "newdir2", Mode: 0755})
+	suite.assert.NoError(err)
+
+	// The created path should not be in the cache (invalidated by CreateDir).
+	_, ok := suite.attrCache.lru.Peek("newdir2")
+	suite.assert.False(ok)
```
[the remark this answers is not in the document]
syeleti-msft: Updated

component/attr_cache/attr_cache.go
```diff
@@ -102,54 +113,95 @@ func (ac *AttrCache) Priority() internal.ComponentPriority {
...
+// Memory held by expired entries is reclaimed by the GC once the component is released.
 func (ac *AttrCache) Stop() error {
 	log.Trace("AttrCache::Stop : Stopping component %s", ac.Name())
+	if ac.stopCh != nil {
+		close(ac.stopCh)
+		ac.sweepWg.Wait()
+		ac.stopCh = nil
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
syeleti-msft: Updated with synchronization using mutex

common/cache_policy/lru_cache.go
```diff
@@ -0,0 +1,289 @@
...
+// Go's map uses a bucket-based design; 64 B covers the amortised cost of key/value copies
+// and tophash bookkeeping per entry.
+const mapBucketOverhead = 64
+
+// lruItem is the value stored inside each list.Element in an LRU[K,V].
+type lruItem[K comparable, V any] struct {
+	key      K
+	val      V
+	userSize int64 // sizeOf(key, val) frozen at insertion time
+}
```
souravgupta-msft, line 52: Any reason for going with generics here as we know that the key will always be of string type and value will be *attrCacheItem?

common/cache_policy/lru_cache.go
```diff
@@ -0,0 +1,289 @@
...
+	l.mu.RLock()
+	defer l.mu.RUnlock()
+	return l.currSize
+}
+
+// MaxSize returns the configured memory limit (0 = no limit).
+func (l *LRU[K, V]) MaxSize() int64 {
+	l.mu.RLock()
+	defer l.mu.RUnlock()
+	return l.maxSize
```
souravgupta-msft: We should not take lock for maxSize as it is configured once during init and should not change later

common/cache_policy/lru_cache.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// PerEntryOverhead returns the structural bytes added per entry (useful for testing).
+// This value is immutable after construction and requires no locking.
+func (l *LRU[K, V]) PerEntryOverhead() int64 {
+	return l.perEntryOverhead
+}
+
+// SetMaxSize changes the memory limit and immediately evicts LRU entries if over the new limit.
+func (l *LRU[K, V]) SetMaxSize(newMax int64) {
```
souravgupta-msft: This should not be needed as we set max size during init.
We should not support OnConfigChange because it applies only to config files (not cli) and is something which should not be used in production setups.

component/attr_cache/cache_item.go
```diff
@@ -0,0 +1,202 @@
...
+	if item != nil {
+		sz += int64(unsafe.Sizeof(*item))
+
+		if item.attr != nil {
+			sz += int64(unsafe.Sizeof(*item.attr))
+
+			sz += int64(len(item.attr.Path))
+			sz += int64(len(item.attr.Name))
+			sz += int64(len(item.attr.ETag))
+			sz += int64(len(item.attr.MD5))
```
souravgupta-msft, line 88: Let's add some comment here describing that sizeof method does not account for the size of string and slice values

component/attr_cache/cache_item.go
```diff
@@ -0,0 +1,202 @@
...
+
+			for k, v := range item.attr.Metadata {
+				sz += 32 + int64(len(k))
```
souravgupta-msft, line 93: A comment will be better describing why we are using 32 here

component/attr_cache/attr_cache.go
```diff
@@ -48,41 +50,51 @@ import (
...
+// AttrCache is the pipeline component that caches file/directory attributes.
+// The LRU is thread-safe; no additional locking is needed around individual operations.
+// Expired entries are rejected lazily at GetAttr time; the memory-bounded LRU evicts
+// stale entries by LRU order as new entries arrive.
+// A background sweeper goroutine reclaims memory from TTL-expired entries when the cache is idle.
 type AttrCache struct {
 	internal.BaseComponent
-	cacheTimeout uint32
-	noSymlinks   bool
-	maxFiles     int
```
souravgupta-msft, line 59: Instead of completely removing maxFiles, should we translate it to max size?
If someone explicitly sets maxFiles in config as some value, we can honour that and internally convert it to max size as per calculations. This way it won't break existing users.
If not set in config, then it's fine to not consider it.

common/cache_policy/lru_cache.go
```diff
@@ -0,0 +1,289 @@
...
+	item := &lruItem[K, V]{key: key, val: val, userSize: userSize}
+	elem := l.list.PushFront(item)
+	l.items[key] = elem
+	l.currSize += l.perEntryOverhead + userSize
+	l.evictIfNeeded()
+}
+
+// Get retrieves a value and promotes it to MRU.  Returns (zero, false) on miss.
+func (l *LRU[K, V]) Get(key K) (V, bool) {
+	l.mu.Lock()
```
anmolanmol1234, line 129: Get() currently takes a write lock since it mutates the LRU list. This makes all cache hits mutually exclusive and could become a bottleneck under high read concurrency. Do we expect this cache to be read-heavy, and if so, have we considered approaches that reduce promotion-related lock contention?

common/cache_policy/lru_cache.go
```diff
@@ -0,0 +1,289 @@
...
+}
+
+// Range calls fn for every entry in MRU→LRU order.
+//
+// The cache is read-locked for the duration of the call. fn must not call any
+// LRU methods (Put, Get, Delete, etc.) — doing so will deadlock. Mutating a
+// value pointed to by V is safe as long as the caller manages value-level
+// concurrency separately.
+func (l *LRU[K, V]) Range(fn func(K, V) bool) {
+	l.mu.RLock()
```
anmolanmol1234, line 208: Here is a possible deadlock scenario -> 
Range() acquires RLock()
Callback executes
Callback calls Delete()
Delete() tries to acquire Lock()
Write lock waits for all readers to exit
Current goroutine is the reader and won't release RLock() until callback returns
Callback is waiting for Delete() → deadlock. Would it be safer to avoid executing callbacks under the lock (e.g. iterate over a snapshot), or should we make this restriction more explicit in the API?

common/cache_policy/lru_cache.go
```diff
@@ -0,0 +1,289 @@
...
+		l.currSize += userSize
+		l.list.MoveToFront(elem)
+		l.evictIfNeeded()
+		return
+	}
+	item := &lruItem[K, V]{key: key, val: val, userSize: userSize}
+	elem := l.list.PushFront(item)
+	l.items[key] = elem
+	l.currSize += l.perEntryOverhead + userSize
+	l.evictIfNeeded()
```
anmolanmol1234, line 123: What is the expected behavior when userSize + perEntryOverhead > maxSize? In that case the entry is inserted and evictIfNeeded() will immediately evict it, so Put() succeeds but the entry is not actually retained in the cache. Is this intentional, or should oversized entries be rejected explicitly?

component/attr_cache/cache_item.go
```diff
@@ -0,0 +1,202 @@
...
+	l.DeleteIf(func(key string, _ *attrCacheItem) bool {
+		return strings.HasPrefix(key, prefix)
+	})
+
+	// Invalidate directory entry itself
+	l.invalidatePath(path)
+}
+
+func (l *attrCacheLRU) updateCacheEntry(path string, attr *internal.ObjAttr) {
+	if l.Has(path) {
```
anmolanmol1234, line 229: Everywhere else you have normalized path but here you are using it directly, any specific reason for that ?

anmolanmol1234, line 229: Has() bypasses idle-gate tracking. Is that intentional, or should it update lastOp like other cache operations?

component/attr_cache/cache_item.go
```diff
@@ -0,0 +1,202 @@
...
+// ReplaceIf replaces matching entries and records cache activity for the idle gate.
+// DeleteIf is intentionally not overridden because the TTL sweeper relies on it without
+// resetting the idle gate. Call l.touch() explicitly before using DeleteIf elsewhere.
+func (l *attrCacheLRU) ReplaceIf(pred func(string, *attrCacheItem) bool, newVal func(string) *attrCacheItem) {
+	l.touch()
+	l.LRU.ReplaceIf(pred, newVal)
+}
+
+func (l *attrCacheLRU) cachePositiveEntry(path string, attr *internal.ObjAttr) {
+	l.Put(path, &attrCacheItem{attr: attr, exists: true, cachedAt: time.Now()})
```
anmolanmol1234: The cache stores the caller's *ObjAttr directly. Could a later modification to attr by the caller unintentionally affect the cached entry?

component/attr_cache/attr_cache.go
```diff
@@ -102,54 +114,99 @@ func (ac *AttrCache) Priority() internal.ComponentPriority {
...
-	ac.cleanupDone = make(chan bool)
-	go ac.backgroundCleanup()
+// startSweeper stops any running sweeper goroutine and starts a new one using
+// the current cacheTimeout. If cacheTimeout is zero, no sweeper is started.
+func (ac *AttrCache) startSweeper() {
+	ac.lifecycleMu.Lock()
+	defer ac.lifecycleMu.Unlock()
+	if ac.stopCh != nil {
+		close(ac.stopCh)
+		ac.sweepWg.Wait()
```
anmolanmol1234: Potential deadlock: sweepWg.Wait() is called while lifecycleMu is held. If ttlSweeper() needs lifecycleMu during shutdown, the goroutine may be unable to exit, causing Wait() to block indefinitely. Can we verify the sweeper never acquires lifecycleMu or avoid waiting while holding the mutex?

common/cache_policy/lru_cache.go
syeleti-msft: to make the cache used across the codebase, f.e., my upcoming PR for supporting the kernel list caching also uses the same cache https://github.com/Azure/azure-storage-fuse/pull/2250/changes#diff-26c4c73256464591dea220cc70c33427d245cf2969072965e546c858fc4210eaR62

common/cache_policy/lru_cache.go
syeleti-msft: SetMaxSize method modifies the variable maxSize, that's the reason we take lock here.

common/cache_policy/lru_cache.go
syeleti-msft: I agree that we should not support the hot reload in production systems, but i added the support to align with the existing implementation of the other components.

common/cache_policy/lru_cache.go
syeleti-msft: this is addressed in the comment of the Range func. the main reason to use this func is to print the list for debugging.

```
// Range calls fn for every entry in MRU→LRU order.
//
// The cache is read-locked for the duration of the call. fn must not call any
// LRU methods (Put, Get, Delete, etc.) — doing so will deadlock. Mutating a
// value pointed to by V is safe as long as the caller manages value-level
// concurrency separately.
```

component/attr_cache/attr_cache.go
syeleti-msft: as this is the hidden parameter, which is not exposed to the user side front, it is safe to assume that there are no users relying on it.

common/cache_policy/lru_cache.go
syeleti-msft: usually `maxSize >> userSize + perEntryOverhead`, but we can return additional non-nil error parameter when something like this happens as that would set the right expectation to the user.

component/attr_cache/cache_item.go
syeleti-msft: Yes it will, but attributes are not changed inside the code after they were created. making a copy every time is costlier and also creates lot of garbage as the attribute struct is big.

component/attr_cache/cache_item.go
syeleti-msft: nice catch, missed to normalize the path. will update it in my new commit.

syeleti-msft: No, this method should also update the lastOp, I missed this part. will address this in a new commit

component/attr_cache/cache_item.go
syeleti-msft: added the comment https://github.com/Azure/azure-storage-fuse/pull/2245/commits/a14df81fd8aaa9efc14f377b248d23fa033c4c89

component/attr_cache/cache_item.go
syeleti-msft: added https://github.com/Azure/azure-storage-fuse/pull/2245/commits/a14df81fd8aaa9efc14f377b248d23fa033c4c89

common/cache_policy/lru_cache.go
syeleti-msft: updated in commit: https://github.com/Azure/azure-storage-fuse/pull/2245/commits/f46029f86eb7d3c17ff1e11f6c64417413f17222

component/attr_cache/cache_item.go
syeleti-msft: updated in commit: https://github.com/Azure/azure-storage-fuse/pull/2245/commits/a14df81fd8aaa9efc14f377b248d23fa033c4c89

syeleti-msft: updated in commit: https://github.com/Azure/azure-storage-fuse/pull/2245/commits/a14df81fd8aaa9efc14f377b248d23fa033c4c89

component/attr_cache/attr_cache.go
syeleti-msft: the lifecycleMu is completly dropped in the new change https://github.com/Azure/azure-storage-fuse/pull/2245/commits/ed871494dbe473b21e874deacf53fce6b9715e5a as we decided to no-op the hot reload support for attr-cache

common/cache_policy/lru_cache.go
syeleti-msft: removed the hot reload support and made it a no-op https://github.com/Azure/azure-storage-fuse/pull/2245/commits/ed871494dbe473b21e874deacf53fce6b9715e5a

common/cache_policy/lru_cache.go
syeleti-msft: removed the lock here https://github.com/Azure/azure-storage-fuse/pull/2245/commits/ed871494dbe473b21e874deacf53fce6b9715e5a

common/cache_policy/lru_cache.go
syeleti-msft: Yes, Get() takes a write lock to move the element to the front, making all cache hits serialize. In our case though the parallelism is bounded by libfuse's worker thread pool — 10 threads by default. With only 10 threads actually dispatching requests, write-lock contention on the LRU is unlikely to be a bottleneck in practice.

If profiling ever shows this as a hotspot, the standard fix is lazy/skipped promotions under a read lock, but that complexity isn't needed yet.

component/attr_cache/attr_cache.go
```diff
@@ -102,54 +114,99 @@ func (ac *AttrCache) Priority() internal.ComponentPriority {
...
+// to avoid holding the write lock during active traffic.
+func (ac *AttrCache) sweepExpired() {
+	if ac.cacheTimeout <= 0 {
+		return
+	}
+	if last := ac.lastOp.Load(); last != 0 {
+		if time.Now().Unix()-last < int64(ac.cacheTimeout/(2*time.Second)) {
+			return
+		}
+	}
```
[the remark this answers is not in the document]
syeleti-msft: that makesense, it is addressd in https://github.com/Azure/azure-storage-fuse/pull/2245/commits/c76e7a94a53b898a2b7b23b187037eaf36129994

component/attr_cache/attr_cache.go
```diff
@@ -102,54 +113,96 @@ func (ac *AttrCache) Priority() internal.ComponentPriority {
...
+			if time.Since(time.Unix(last, 0)) < idleThreshold {
+				return
+			}
+		}
+	}
+	timeout := ac.cacheTimeout
+	before := ac.lru.Size()
+	ac.lru.DeleteIf(func(_ string, item *attrCacheItem) bool {
+		return time.Since(item.cachedAt) >= timeout
+	})
```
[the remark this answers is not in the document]
syeleti-msft: this is resolved in this https://github.com/Azure/azure-storage-fuse/pull/2245/commits/9dcf7c56e8ebe8c1441651789661641a0f620a59

component/attr_cache/attr_cache.go
```diff
@@ -158,313 +213,153 @@ func (ac *AttrCache) Configure(_ bool) error {
...
-
-	for key, value := range ac.cacheMap {
-		if strings.HasPrefix(key, prefix) {
-			value.invalidate()
-		}
+	if config.IsSet(compName + ".max-size-mb") {
+		ac.maxSizeBytes = int64(conf.MaxSizeMB) * 1024 * 1024
+	} else {
+		ac.maxSizeBytes = int64(defaultMaxSizeMB) * 1024 * 1024
 	}
```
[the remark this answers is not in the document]
syeleti-msft: yes, this warn should be present

component/attr_cache/attr_cache_test.go
```diff
@@ -1027,30 +1030,70 @@ func (suite *attrCacheTestSuite) TestGetAttrEnonetError() {
...
-			suite.assert.Contains(suite.attrCache.cacheMap, truncatedPath)
-			suite.assert.Equal(&internal.ObjAttr{}, suite.attrCache.cacheMap[truncatedPath].attr)
-			suite.assert.True(suite.attrCache.cacheMap[truncatedPath].valid())
-			suite.assert.False(suite.attrCache.cacheMap[truncatedPath].exists())
-			suite.assert.NotNil(suite.attrCache.cacheMap[truncatedPath].cachedAt)
+			item := getCacheItem(suite.attrCache, truncatedPath)
+			suite.assert.NotNil(item)
+			suite.assert.Nil(item.attr)
+			suite.assert.False(item.exists)
+			suite.assert.NotNil(item.cachedAt)
```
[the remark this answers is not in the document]
syeleti-msft: fixed

setup/baseConfig.yaml
```diff
@@ -29,6 +29,11 @@ block_cache:
   prefetch: <number of blocks to be prefetched in serial read case. Min - 11, Default - 2 times number of CPU cores>
   parallelism: <number of parallel threads downloading the data and writing to disk cache. Default - 3 times number of CPU cores> 
 
+# Attribute cache related configuration
+attr_cache:
+  timeout-sec: <time attributes can be cached (in sec). Default - 120 sec>
+  max-size-mb: <maximum memory the attribute cache may use (in MB). Default - auto-tuned to 1% of total RAM, clamped to [64 MB, 1 GB]. Set to 0 to use the auto-tuned value. Increase for workloads with large directory trees or many distinct file paths to avoid frequent cache misses.>
```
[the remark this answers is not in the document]
syeleti-msft: fixed

setup/advancedConfig.yaml
```diff
@@ -113,6 +113,7 @@ file_cache:
 attr_cache:
   timeout-sec: <time attributes can be cached (in sec). Default - 120 sec>
   no-symlinks: true|false <to improve performance disable symlink support. symlinks will be treated like regular files.>
+  max-size-mb: <maximum memory the attribute cache may use (in MB). Default - auto-tuned to 1% of total RAM, clamped to [64 MB, 1 GB]. Set to 0 to use the auto-tuned value. Increase for workloads with large directory trees or many distinct file paths to avoid frequent cache misses.>
```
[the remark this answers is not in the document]
syeleti-msft: fixed

component/attr_cache/attr_cache.go
```diff
@@ -158,313 +231,157 @@ func (ac *AttrCache) Configure(_ bool) error {
...
-
-	for key, value := range ac.cacheMap {
-		if strings.HasPrefix(key, prefix) {
-			value.invalidate()
-		}
+	if config.IsSet(compName+".max-size-mb") && conf.MaxSizeMB > 0 {
+		ac.maxSizeBytes = int64(conf.MaxSizeMB) * 1024 * 1024
+	} else {
+		ac.maxSizeBytes = defaultMaxSizeBytes()
 	}
```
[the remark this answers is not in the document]
syeleti-msft: PR description needs to be updated.

anmolanmol1234 (approved):

common/cache_policy/lru_cache.go
souravgupta-msft: Any reason for having the callback method in the parameter? I don't see it being used in code except the unit tests

common/cache_policy/lru_cache.go
```diff
@@ -0,0 +1,289 @@
...
+	for elem := l.list.Front(); elem != nil; {
+		next := elem.Next()
+		item := elem.Value.(*lruItem[K, V])
+		if pred(item.key, item.val) {
+			v := newVal(item.key)
+			newSize := l.sizeOf(item.key, v)
+			l.currSize += newSize - item.userSize
+			item.val = v
+			item.userSize = newSize
+			l.list.MoveToFront(elem)
```
souravgupta-msft, line 260: I see that `ReplaceIf` is used only in delete/rename dir to invalidate the dir entries. In this case, we can avoid moving the replaced entries to MRU as these are negative entries.

common/cache_policy/lru_cache.go
```diff
@@ -0,0 +1,289 @@
...
+		}
+		elem = next
+	}
+	l.evictIfNeeded()
```
souravgupta-msft, line 264: Same goes for this. Is this needed as the entries are marked negative where the size of the new negative entries will be less than the previous entries.

common/sysinfo.go
```diff
@@ -0,0 +1,46 @@
...
+
+// TotalMemoryBytes returns the total physical RAM of the system in bytes.
+// Returns 0 if the syscall fails.
+func TotalMemoryBytes() uint64 {
+	var info syscall.Sysinfo_t
+	if err := syscall.Sysinfo(&info); err != nil {
+		return 0
+	}
+	return uint64(info.Totalram) * uint64(info.Unit)
+}
```
souravgupta-msft: Can we move this method to util.go where we have other utility methods. Same applies for its test file which can be moved to util_test.go

component/attr_cache/cache_item.go
```diff
@@ -0,0 +1,232 @@
...
+
+func newAttrCacheLRU(maxSizeBytes int64, lastOp *atomic.Int64) *attrCacheLRU {
+	return &attrCacheLRU{cachepolicy.NewLRU(maxSizeBytes, estimateAttrCacheEntrySize), lastOp}
+}
+
+func (l *attrCacheLRU) touch() {
+	if l.lastOp != nil {
+		l.lastOp.Store(time.Now().Unix())
+	}
+}
```
souravgupta-msft, line 122: Since this method is a fast path, can we remove this condition here

component/attr_cache/attr_cache.go
souravgupta-msft: This can be passed in the config file via `max-files` parameter. If someone is using this parameter in their config file, then it will be a breaking change for them.

component/attr_cache/attr_cache.go
```diff
@@ -158,313 +231,157 @@ func (ac *AttrCache) Configure(_ bool) error {
...
-}
-
-// deletePath: removes a path from cache
-func (ac *AttrCache) deletePath(path string, time time.Time) {
-	// Keys in the cache map do not contain trailing /, truncate the path before referencing a key in the map.
-	value, found := ac.cacheMap[internal.TruncateDirName(path)]
-	if found {
-		value.markDeleted(time)
+	if config.IsSet(compName + ".max-files") {
+		log.Warn("AttrCache::Configure : 'max-files' is deprecated/ignored; use 'max-size-mb' instead")
```
souravgupta-msft: Should we translate this to max-size-mb based on rough estimates, else it would be a breaking change for customers using this flag in config file

component/attr_cache/attr_cache.go
```diff
@@ -102,54 +131,98 @@ func (ac *AttrCache) Priority() internal.ComponentPriority {
...
+	// Only apply the idle gate when the cache is memory-bounded. If MaxSize()==0 (unlimited),
+	// skipping sweeps under continuous traffic can let TTL-expired entries accumulate forever.
+	idleThreshold := ac.cacheTimeout / 2
+	if idleThreshold > 0 && ac.lru.MaxSize() > 0 {
+		if last := ac.lastOp.Load(); last != 0 {
+			if time.Since(time.Unix(last, 0)) < idleThreshold {
+				return
+			}
+		}
+	}
```
souravgupta-msft, line 191: For cases where there are large number of read/write operations, this ttlSweeper will never evict the LRU elements from the list. In this case, are we relying on the Put() method to evict the LRU element if the size goes beyond the configured size?

component/attr_cache/attr_cache.go
```diff
@@ -158,313 +231,157 @@ func (ac *AttrCache) Configure(_ bool) error {
...
+// CreateDir marks the directory invalid in the cache.
 func (ac *AttrCache) CreateDir(options internal.CreateDirOptions) error {
 	log.Trace("AttrCache::CreateDir : %s", options.Name)
 	err := ac.NextComponent().CreateDir(options)
 
 	if err == nil || err == syscall.EEXIST {
-		ac.cacheLock.RLock()
-		defer ac.cacheLock.RUnlock()
-		ac.invalidatePath(options.Name)
+		ac.lru.invalidatePath(options.Name)
```
souravgupta-msft, line 288: Previously for all the operations like invalidate path/dir and delete path/dir we were taking read lock as the delete and invalidate methods were internally handled by the attr flag in the attr cache item. Now, since we are taking write lock for all these cases, can this lead to perf regression?

component/attr_cache/cache_item.go
```diff
@@ -0,0 +1,232 @@
...
+	prefix := internal.ExtendDirName(path)
+	l.DeleteIf(func(key string, _ *attrCacheItem) bool {
+		return strings.HasPrefix(key, prefix)
+	})
+
+	// Invalidate directory entry itself
+	l.invalidatePath(path)
+}
+
+func (l *attrCacheLRU) updateCacheEntry(path string, attr *internal.ObjAttr) {
```
souravgupta-msft: In `updateCacheEntry` initial Has() and then Put() is not atomic. A concurrent GetAttr refresh or Delete between the read and the write can result in race condition. Atomic update on the LRU would be cleaner in this case.

souravgupta-msft: Same applies to `Chmod` operation where we first do Peek() and then Put().

component/attr_cache/attr_cache.go
```diff
@@ -523,131 +424,115 @@ func (ac *AttrCache) CopyFromFile(options internal.CopyFromFileOptions) error {
...
-
-	// Try to serve the request from the attribute cache
-	if found && value.valid() && time.Since(value.cachedAt).Seconds() < float64(ac.cacheTimeout) {
-		if value.isDeleted() {
-			log.Debug("AttrCache::GetAttr : %s served from cache", options.Name)
-			// no entry if path does not exist
-			return &internal.ObjAttr{}, syscall.ENOENT
-		} else {
+	if item, ok := ac.lru.Get(truncatedPath); ok {
+		if time.Since(item.cachedAt) < ac.cacheTimeout {
```
souravgupta-msft, line 469: If cache timeout is set to 0, this condition will always be false. Meaning we will be making REST call everytime and storing it in the attr cache as well via `cachePositiveEntry` or `cacheNegativeEntry`.
How are we evicting the attr cache items in this case since the ttl sweeper thread won't run for 0 timeout value.

component/attr_cache/cache_item.go
```diff
@@ -0,0 +1,232 @@
...
+}
+
+func (l *attrCacheLRU) deleteDirectory(path string, t time.Time) {
+	// Mark all the child entries as negative entries
+	prefix := internal.ExtendDirName(path)
+	l.ReplaceIf(func(key string, _ *attrCacheItem) bool {
+		return strings.HasPrefix(key, prefix)
+	}, func(_ string) *attrCacheItem {
+		return &attrCacheItem{cachedAt: t}
+	})
```
souravgupta-msft, line 203: `deleteDirectory` promotes tombstones to MRU. ReplaceIf calls MoveToFront on every matched child, so negative entries become the most recently used entries and can outlive real entries under memory pressure. We should not promote tombstones as MRU.

common/cache_policy/lru_cache.go
syeleti-msft: generally when you range the LRU, most of the times you want eval a condition on key/value to select some entries from the LRU which satisfies the condition. that's the main reason to have callback.

common/cache_policy/lru_cache.go
syeleti-msft: as we are refreshing the timeout field, we're treating as new negative entry. that's the main motive behind converting to MRU. 

we treat negative/positive entries in the same way, as they were give equal importance.

common/cache_policy/lru_cache.go
syeleti-msft: we can't take change the implementation of the LRU based on the caller's assumptions(attr_cache here). 

> entries are marked negative where the size of the new negative entries will be less than the previous entries.

If this is the case the function evictIfNeeded will be no-op

component/attr_cache/attr_cache.go
syeleti-msft: Yes, the assumption that we are bounded by the configured memory size gives the freedom for us to use this hack

component/attr_cache/attr_cache.go
syeleti-msft: Okay, then we should convert the max-files to memory size based on our calculation and warn the user regarding the same

component/attr_cache/cache_item.go
syeleti-msft: addressed in this commit https://github.com/Azure/azure-storage-fuse/pull/2245/commits/d9051aa0c716c7393de63fb0dc3978222f1abb4c

component/attr_cache/attr_cache.go
syeleti-msft: that makes sense, that we should be backward compatible for the users who use it anyways. addressed it in this commit https://github.com/Azure/azure-storage-fuse/pull/2245/commits/f56bce4c5449a7eba3051f5425d6bd4a3e8ece68

component/attr_cache/attr_cache.go
syeleti-msft: that's because previously we are not deleting the entry when we are invalidating the entry. and for the delete we are wasting the space by preserving the entry. Write lock wouldn't cause the contention as the operation were O(1) in LRU for invalidating/delete path

component/attr_cache/attr_cache.go
syeleti-msft: this is the corner case, where user shouldn't actually use. In this case the cache would be memory bounded and eviction happens based on the upper bound of memory configured.

component/attr_cache/cache_item.go
syeleti-msft: I think we don't need to special case these scenarios. as negative and positive entries were given equal waitage. 

one more thing to remember here, while removing the directory the user application always list the dir and remove each entry seperately.(as rmdir syscall expects the dir must be empty to success). so it would be safe to assume that all the entries for the directory would anyway be MRU entries.

common/sysinfo.go
syeleti-msft: addressed https://github.com/Azure/azure-storage-fuse/pull/2245/commits/184626bbe3c656e7841862c3cfbf11705a9711c6

component/attr_cache/cache_item.go
syeleti-msft: Yes, there is a TOCTOU race, but it will not cause any correctness issue. If getattr comes before delete we will return positive entry and if it comes after deleting we return negative entry. we can't guard on the atomicity of the ordering of these two calls anyways as the call comes from libfuse and we don't gaurantee which one to serve first(as we don't serialize these requests). 

supporting this atomicity in attr_cache would make the LRU to expose a PutIf func to check condition before putting which is a non-trivial method and unnecessary imo

component/attr_cache/cache_item.go
```diff
@@ -0,0 +1,233 @@
...
+		if attr != nil {
+			copied := *attr // copy so we don't mutate the caller's struct
+			copied.Path = path
+			attr = &copied
+		}
+		if !l.Put(path, &attrCacheItem{attr: attr, exists: attr != nil, cachedAt: time.Now()}) {
+			log.Err("attrCacheLRU::updateCacheEntry : entry too large for cache, skipping path %s", path)
+		}
+	}
+}
```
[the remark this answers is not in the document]
syeleti-msft: this wouldn't happen as the src attr would always be filled by downstream components (libfuse here) but its always safe to add.
addressed this https://github.com/Azure/azure-storage-fuse/pull/2245/commits/71480722ba459e6e093156a3a1ec604683c433e5

component/attr_cache/cache_item.go
```diff
@@ -0,0 +1,236 @@
...
+	lastOp atomic.Int64
+}
+
+func newAttrCacheLRU(maxSizeBytes int64) *attrCacheLRU {
+	return &attrCacheLRU{LRU: cachepolicy.NewLRU(maxSizeBytes, estimateAttrCacheEntrySize)}
+}
+
+func (l *attrCacheLRU) touch() {
+	l.lastOp.Store(time.Now().Unix())
+}
```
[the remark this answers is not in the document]
syeleti-msft: It is expected that TTLs would be greater than few seconds. generally(>60s) in most cases. so its not required.

component/attr_cache/attr_cache.go
```diff
@@ -102,54 +130,98 @@ func (ac *AttrCache) Priority() internal.ComponentPriority {
...
+	}
+
+	// Only apply the idle gate when the cache is memory-bounded. If MaxSize()==0 (unlimited),
+	// skipping sweeps under continuous traffic can let TTL-expired entries accumulate forever.
+	idleThreshold := ac.cacheTimeout / 2
+	if idleThreshold > 0 && ac.lru.MaxSize() > 0 {
+		if last := ac.lru.lastOp.Load(); last != 0 {
+			if time.Since(time.Unix(last, 0)) < idleThreshold {
+				return
+			}
```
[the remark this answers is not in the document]
syeleti-msft: small cache timeouts <=1s were out of scope of the change. it would always be few seconds.

souravgupta-msft (approved):

souravgupta-msft (approved):

PR #2251, opened 2026-06-12
Return ENAMETOOLONG when renaming to a too-deep ADLS path

### LINKED ISSUE #2241: Investigate error paths for ENAMETOOLONG in datalake
Should CreateFile (and potentially RenameFile/CreateLink) get the same treatment so they surface ENAMETOOLONG instead of falling through to the generic EIO?

currently we return the ENAMETOOLONG for only createDir path

related PR: https://github.com/Azure/azure-storage-fuse/pull/2221

CynicDog:
## What

ADLS rejects paths with more than 63 segments. `CreateDirectory` already returns `ENAMETOOLONG` for this case (PR #2221), but `RenameDirectory` still returned a generic `EIO`. This extends the same handling to directory rename.

## Changes

- Map the existing `ErrPathTooDeep` error to `ENAMETOOLONG` in `Datalake.RenameDirectory`.
- Surface `ENAMETOOLONG` through the fuse2 and fuse3 directory-rename handlers instead of returning `EIO`.
- Add libfuse handler unit tests for the new mapping (run in CI).
- Add an end-to-end test that renames a directory to a path exceeding the 63-segment limit and expects `ENAMETOOLONG` (skipped on non-ADLS accounts).

## Notes

`RenameFile` is left unchanged. Renaming a file does not add directory depth, so the ADLS depth limit does not apply and `ErrPathTooDeep` cannot be returned for file renames.

`CreateFile` and `CreateLink` are also left unchanged. They go through the Block Blob API, which does not report the depth limit, so the error cannot occur there.

The end-to-end test needs a live HNS account to confirm the surfaced error code.

Fixes #2241

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 **Bug Fixes**
 - Fix CPK-encrypted blob attribute lookup duplicating the prefix path when subdirectory is configured ([PR #2199](https://github.com/Azure/azure-storage-fuse/pull/2199))
 - Return `ENAMETOOLONG` instead of `EIO` when creating a directory path that exceeds the ADLS depth limit of 63 segments ([PR #2221](https://github.com/Azure/azure-storage-fuse/pull/2221))
-- Return `ENAMETOOLONG` instead of `EIO` when renaming a file or directory to an ADLS path that exceeds the depth limit of 63 segments ([Issue #2241](https://github.com/Azure/azure-storage-fuse/issues/2241))
+- Return `ENAMETOOLONG` instead of `EIO` when renaming a directory to an ADLS path that exceeds the depth limit of 63 segments ([PR #2251](https://github.com/Azure/azure-storage-fuse/pull/2251))
 
 **Other Changes**
 - CBL-Mariner 2.0 has reached end-of-life; Blobfuse2 packages will no longer be published to Mariner 2.0 repositories.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 **Bug Fixes**
 - Fix CPK-encrypted blob attribute lookup duplicating the prefix path when subdirectory is configured ([PR #2199](https://github.com/Azure/azure-storage-fuse/pull/2199))
 - Return `ENAMETOOLONG` instead of `EIO` when creating a directory path that exceeds the ADLS depth limit of 63 segments ([PR #2221](https://github.com/Azure/azure-storage-fuse/pull/2221))
+- Return `ENAMETOOLONG` instead of `EIO` when renaming a file or directory to an ADLS path that exceeds the depth limit of 63 segments ([Issue #2241](https://github.com/Azure/azure-storage-fuse/issues/2241))
 
 **Other Changes**
 - CBL-Mariner 2.0 has reached end-of-life; Blobfuse2 packages will no longer be published to Mariner 2.0 repositories.
--- a/component/azstorage/datalake.go
+++ b/component/azstorage/datalake.go
@@ -356,10 +356,14 @@ func (dl *Datalake) RenameFile(source string, target string, srcAttr *internal.O
 	})
 	if err != nil {
 		serr := storeDatalakeErrToErr(err)
-		if serr == ErrFileNotFound {
+		switch serr {
+		case ErrFileNotFound:
 			log.Err("Datalake::RenameFile : %s does not exist", source)
 			return syscall.ENOENT
-		} else {
+		case ErrPathTooDeep:
+			log.Err("Datalake::RenameFile : Path is too deep for %s -> %s [%s]", source, target, err.Error())
+			return syscall.ENAMETOOLONG
+		default:
 			log.Err("Datalake::RenameFile : Failed to rename file %s to %s [%s]", source, target, err.Error())
 			return err
 		}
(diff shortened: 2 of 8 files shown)


test/e2e_tests/dir_test.go
```diff
@@ -280,6 +280,35 @@ func (suite *dirTestSuite) TestDirCreateDeepPath() {
...
+	// ADLS enforces a maximum path depth of 63 segments from the container root.
+	// Renaming into a 65-level destination guarantees the limit is exceeded; the
+	// service validates the destination path depth and the rename must surface
+	// ENAMETOOLONG rather than a generic I/O error.
+	const depth = 65
+	topDir := suite.testPath + "/renamedeep"
+	deepPath := topDir
+	for i := range depth {
+		deepPath = filepath.Join(deepPath, fmt.Sprintf("l%d", i))
+	}
```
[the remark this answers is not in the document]
CynicDog: resolved.

test/e2e_tests/dir_test.go
```diff
@@ -280,6 +280,35 @@ func (suite *dirTestSuite) TestDirCreateDeepPath() {
...
+
+	err = os.Rename(srcDir, deepPath)
+	suite.Error(err)
+	suite.ErrorIs(err, syscall.ENAMETOOLONG)
```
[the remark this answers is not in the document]
CynicDog: Valid, fixed. The test now creates directories one level at a time until ADLS rejects the next one, leaving deepest at the maximum allowed depth, then renames a source into a child of it.

syeleti-msft (approved):
LGTM, thanks for the contribution

CynicDog:
Hi @syeleti-msft, pushed some upstream sync commits with no functional changes. Could you re-comment /mirror to refresh the CI branch?

Also, this might happen again while waiting. Is there a preferred way to handle it? Many Thanks for the code review!

syeleti-msft:
> Hi @syeleti-msft, pushed some upstream sync commits with no functional changes. Could you re-comment /mirror to refresh the CI branch?
> 
> Also, this might happen again while waiting. Is there a preferred way to handle it? Many Thanks for the code review!

as the CI for the external contributors is not enabled automatically on the repo, a member from the repo must trigger the CI. but we will explore ways to simplify this process in the future:)

syeleti-msft:
@CynicDog RenameFile doesn't return the `ENAMETOOLONG` error. pls revert the change for this callback

component/azstorage/datalake.go
```diff
@@ -356,10 +356,14 @@ func (dl *Datalake) RenameFile(source string, target string, srcAttr *internal.O
...
 		serr := storeDatalakeErrToErr(err)
-		if serr == ErrFileNotFound {
+		switch serr {
+		case ErrFileNotFound:
 			log.Err("Datalake::RenameFile : %s does not exist", source)
 			return syscall.ENOENT
-		} else {
+		case ErrPathTooDeep:
+			log.Err("Datalake::RenameFile : Path is too deep for %s -> %s [%s]", source, target, err.Error())
+			return syscall.ENAMETOOLONG
```
souravgupta-msft: ADLS path depth limit applies to directory namespace depth. A file is a leaf, so renaming a file under max depth (63) parent doesn't deepen the namespace, it only adds a leaf. So, this error does not apply to rename file and makes this code unreachable.
Can you please validate this case and if you are seeing this error in rename file operation, please add a e2e test in `test/e2e_tests/file_test.go`

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 **Bug Fixes**
 - Fix CPK-encrypted blob attribute lookup duplicating the prefix path when subdirectory is configured ([PR #2199](https://github.com/Azure/azure-storage-fuse/pull/2199))
 - Return `ENAMETOOLONG` instead of `EIO` when creating a directory path that exceeds the ADLS depth limit of 63 segments ([PR #2221](https://github.com/Azure/azure-storage-fuse/pull/2221))
+- Return `ENAMETOOLONG` instead of `EIO` when renaming a file or directory to an ADLS path that exceeds the depth limit of 63 segments ([Issue #2241](https://github.com/Azure/azure-storage-fuse/issues/2241))
```
souravgupta-msft: Update the changelog entry to point it to PR instead of issue as per our convention. Also, if the rename file case is removed as per previous comment, the changelog can be updated as below.
```suggestion
- Return `ENAMETOOLONG` instead of `EIO` when renaming a directory to an ADLS path that exceeds the depth limit of 63 segments ([PR #2251](https://github.com/Azure/azure-storage-fuse/pull/2251))
```

CynicDog:
> @CynicDog RenameFile doesn't return the `ENAMETOOLONG` error. pls revert the change for this callback

Reverted the if/else → switch restructure in `RenameFile`. Makes total sense since files don't add directory depth.

component/azstorage/datalake.go
CynicDog: Fixed! Validated that the error doesn't come back for file renames, so removed the unreachable code. No e2e test added for the file case since it can't actually happen. Thanks for the clear explanation!

CHANGELOG.md
CynicDog: Applied!

souravgupta-msft (approved):
@CynicDog, thanks for the quick follow up. We will merge the changes, and it will be part of the next release.

syeleti-msft (approved):

PR #2261, opened 2026-06-22
Fix #2249: support blob index tag filtering on Block Blob, ADLS accounts

### LINKED ISSUE #2249: Add support for filtering on blob index tags
### Summary

The `--filter` option (config `azstorage.filter`) uses the `github.com/vibhansa-msft/blobfilter` library, which includes a `tag=key:value` filter type. However, blobfuse2 does not currently wire up blob index tags into that library, so `tag=` filters effectively match nothing and the mount appears empty. I'd like blobfuse2 to fully support filtering blobs by their blob index tags.

### Use case

I want to volume-mount an Azure Blob container into Kubernetes pods (via the blob CSI driver) but expose only blobs matching a blob index tag, e.g. `--filter=tag=domain:optical`. This would let different workloads share a container while each pod sees only the subset of blobs relevant to it.

Example mount option:

```
mountOptions:
  - --filter=tag=domain:optical
```

### Current behavior

With a `tag=` filter on a read-only mount, `ls` returns nothing even though matching tagged blobs exist.

The filter library's tag check reads `BlobAttr.Tags`:

```go
// vibhansa-msft/blobfilter/blobTagFilter.go
func (btf *blobTagFilter) isAcceptable(fileInfo *BlobAttr) bool {
    if val, ok := fileInfo.Tags[btf.key]; ok {
        return (btf.opr == (btf.value == strings.ToLower(val)))
    }
    return false 
}
```

blobfuse2 never supplies that data, so the tag filter always rejects.

It looks like this is supportable based on what the sdks expose.

- The list request could ask the service for tags. Today `listDetails` sets `Metadata: true` but not `Tags: true`, so the `ListBlobs` response carries no tag data:
  https://github.com/Azure/azure-storage-fuse/blob/bb288d475/component/azstorage/block_blob.go#L106-L111

- `getBlobAttr` could read tags off the list item. The SDK `BlobItem` exposes `BlobTags`, but the constructed `ObjAttr` only copies metadata today; tags are not carried through:
  https://github.com/Azure/azure-storage-fuse/blob/bb288d475/component/azstorage/block_blob.go#L702-L734

- `processBlobItems` could set `Tags` on the filter input. Today only `Name`, `Mtime`, and `Size` are populated before `IsAcceptable`, so `filterAttr.Tags` is always `nil`:
  https://github.com/Azure/azure-storage-fuse/blob/bb288d475/component/azstorage/block_blob.go#L684-L693

```go
// block_blob.go:684-693
if bb.Config.filter != nil && !blobAttr.IsDir() {
    filterAttr.Name = blobAttr.Name
    filterAttr.Mtime = blobAttr.Mtime
    filterAttr.Size = blobAttr.Size
    // no filterAttr.Tags assignment

    if bb.Config.filter.IsAcceptable(&filterAttr) {
        blobList = append(blobList, blobAttr)
    } else {
        log.Debug("BlockBlob::List : Filtered out blob %s", blobAttr.Name)
    }
}
```

The `GetAttr` paths similarly don't supply tags:
- `BlockBlob.GetAttr` — https://github.com/Azure/azure-storage-fuse/blob/bb288d475/component/azstorage/block_blob.go#L585-L594
- `Datalake.GetAttr` — https://github.com/Azure/azure-storage-fuse/blob/bb288d475/component/azstorage/datalake.go#L455-L464

There's already a commented-out `parseBlobTags` helper in `component/azstorage/utils.go` (~L646) that looks intended for this.

syeleti-msft:
## Summary

Fixes #2249. The `--filter=tag=key:value` option (config `azstorage.filter`)
silently rejected every blob because blobfuse never populated blob index tags
into the `BlobAttr` passed to `blobfilter.IsAcceptable`.

## Changes

- **List path (`block_blob.go`)**: request `Tags: true` in `listDetails`; parse
  tags from `BlobItem.BlobTags` into the local filter `BlobAttr` so list
  results are filtered correctly.
- **GetAttr path (`block_blob.go`)**: build a local `blobfilter.BlobAttr` and,
  only when the configured filter references a `tag=` clause, issue a separate
  `GetTags` call (the blob `GetProperties` response doesn't include tags).
- **`filterHasTag` flag (`connection.go`, `config.go`, `block_blob.go`)**:
  computed once from the raw filter string via `filterReferencesTag`, so the
  hot GetAttr path skips the extra round-trip when no tag clause is configured.
- **ADLS Gen2 (`config.go`)**: blob index tags are not generally available on
  hierarchical-namespace accounts. `configureBlobFilter` but preview is available
  behind a flag. hence it is supported. listBlob APi in blob endpoint is currently
  is not supported to include tags for HNS accounts, so we call getTags API
  for every blob returned by listBlob to apply the filter
- **`parseBlobTags` helper (`utils.go`)**: converts `*container.BlobTags` to
  `map[string]string`, tolerating nil entries.

## Test plan

- [ ] `TestParseBlobTags` and `TestFilterReferencesTag` (unit) pass.
- [ ] `TestBlobTagFilter` (block_blob integration) passes against a real
      Block Blob account: list and GetAttr respect `tag=domain:optical`,
      `filterHasTag` toggles correctly across `SetFilter` calls.
- [ ] Configuring `filter: tag=...` against an ADLS Gen2 account fails mount
      with a clear error.
- [ ] Existing non-tag filters (`name=`, `size=`, …) still work on both
      Block Blob and ADLS Gen2 accounts.
- [ ] `go build ./...` and `go vet ./...` clean.

### DIFF
--- a/component/azstorage/block_blob.go
+++ b/component/azstorage/block_blob.go
@@ -590,7 +590,6 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
 
 	attr, err = bb.getAttrUsingRest(name)
 	if err != nil {
-		log.Err("BlockBlob::GetAttr : Failed to get attributes for %s [%s]", name, err.Error())
 		return attr, err
 	}
 	if bb.Config.filter != nil && attr != nil {
--- a/component/azstorage/block_blob.go
+++ b/component/azstorage/block_blob.go
@@ -105,6 +105,7 @@ func (bb *BlockBlob) Configure(cfg AzStorageConfig) error {
 
 	bb.listDetails = container.ListBlobsInclude{
 		Metadata:    true,
+		Tags:        bb.Config.filterHasTag,
 		Deleted:     false,
 		Snapshots:   false,
 		Permissions: false, //Added to get permissions, acl, group, owner for HNS accounts
@@ -577,23 +578,48 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
 
 	// To support virtual directories with no marker blob, we call list instead of get properties since list will not return a 404
 	if bb.Config.virtualDirectory {
-		attr, err = bb.getAttrUsingList(name)
-	} else {
-		attr, err = bb.getAttrUsingRest(name)
+		// getAttrUsingList -> List -> processBlobItems already evaluates the
+		// filter (including tag filters), so a successful return means the blob
+		// passed the filter.
+		return bb.getAttrUsingList(name)
 	}
 
+	attr, err = bb.getAttrUsingRest(name)
 	if bb.Config.filter != nil && attr != nil {
-		if !bb.Config.filter.IsAcceptable(&blobfilter.BlobAttr{
+		filterAttr := blobfilter.BlobAttr{
 			Name:  attr.Name,
 			Mtime: attr.Mtime,
 			Size:  attr.Size,
-		}) {
+		}
+		// GetProperties does not return blob index tags. When the configured
+		// filter references tags, fetch them explicitly before deciding.
+		if bb.Config.filterHasTag && !attr.IsDir() {
+			tags, terr := bb.getBlobTags(name)
+			if terr != nil {
+				log.Warn("BlockBlob::GetAttr : Failed to get tags for %s [%s]", name, terr.Error())
+			} else {
+				filterAttr.Tags = tags
+			}
+		}
+
+		if !bb.Config.filter.IsAcceptable(&filterAttr) {
 			log.Debug("BlockBlob::GetAttr : Filtered out %s", name)
 			return nil, syscall.ENOENT
 		}
 	}
 
 	return attr, err
+}
+
+// getBlobTags fetches blob index tags for the given blob name and returns them
+// as a flat map. Returns nil if the blob has no tags.
+func (bb *BlockBlob) getBlobTags(name string) (map[string]string, error) {
+	blobClient := bb.Container.NewBlockBlobClient(filepath.Join(bb.Config.prefixPath, name))
+	resp, err := blobClient.GetTags(context.Background(), nil)
+	if err != nil {
+		return nil, err
+	}
+	return parseBlobTags(&resp.BlobTags), nil
 }
 
 // List : Get a list of blobs matching the given prefix
@@ -685,6 +711,7 @@ func (bb *BlockBlob) processBlobItems(blobItems []*container.BlobItem) ([]*inter
 			filterAttr.Name = blobAttr.Name
 			filterAttr.Mtime = blobAttr.Mtime
 			filterAttr.Size = blobAttr.Size
+			filterAttr.Tags = parseBlobTags(blobInfo.BlobTags)
 
 			if bb.Config.filter.IsAcceptable(&filterAttr) {
 				blobList = append(blobList, blobAttr)
@@ -1846,9 +1873,16 @@ func (bb *BlockBlob) CommitBlocks(name string, blockList []string, newEtag *stri
 func (bb *BlockBlob) SetFilter(filter string) error {
 	if filter == "" {
 		bb.Config.filter = nil
+		bb.Config.filterHasTag = false
+		bb.listDetails.Tags = false
 		return nil
 	}
 
 	bb.Config.filter = &blobfilter.BlobFilter{}
-	return bb.Config.filter.Configure(filter)
+	if err := bb.Config.filter.Configure(filter); err != nil {
+		return err
+	}
+	bb.Config.filterHasTag = filterReferencesTag(filter)
+	bb.listDetails.Tags = bb.Config.filterHasTag
+	return nil
 }
--- a/component/azstorage/block_blob.go
+++ b/component/azstorage/block_blob.go
@@ -600,7 +600,7 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
 		if bb.Config.filterHasTag && !attr.IsDir() {
 			tags, terr := bb.getBlobTags(name)
 			if terr != nil {
-				log.Warn("BlockBlob::GetAttr : Failed to get tags for %s [%s]", name, terr.Error())
+				log.Err("BlockBlob::GetAttr : Failed to get tags for %s [%s]", name, terr.Error())
 			} else {
 				filterAttr.Tags = tags
 			}
--- a/component/azstorage/block_blob.go
+++ b/component/azstorage/block_blob.go
@@ -601,9 +601,9 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
 			tags, terr := bb.getBlobTags(name)
 			if terr != nil {
 				log.Err("BlockBlob::GetAttr : Failed to get tags for %s [%s]", name, terr.Error())
-			} else {
-				filterAttr.Tags = tags
+				return attr, syscall.EACCES
 			}
+			filterAttr.Tags = tags
 		}
 
 		if !bb.Config.filter.IsAcceptable(&filterAttr) {
@@ -725,10 +725,10 @@ func (bb *BlockBlob) processBlobItems(blobItems []*container.BlobItem) ([]*inter
 				if bb.Config.isHNS {
 					tagResp, err := bb.Container.NewBlockBlobClient(*blobInfo.Name).GetTags(context.Background(), nil)
 					if err != nil {
-						log.Warn("BlockBlob::processBlobItems : Failed to get tags for %s [%s]", *blobInfo.Name, err.Error())
-					} else {
-						filterAttr.Tags = parseBlobTags(&tagResp.BlobTags)
+						log.Err("BlockBlob::processBlobItems : Failed to get tags for %s [%s]", *blobInfo.Name, err.Error())
+						return nil, nil, syscall.EACCES
 					}
+					filterAttr.Tags = parseBlobTags(&tagResp.BlobTags)
 				} else {
 					filterAttr.Tags = parseBlobTags(blobInfo.BlobTags)
 				}
--- a/component/azstorage/block_blob.go
+++ b/component/azstorage/block_blob.go
@@ -104,8 +104,12 @@ func (bb *BlockBlob) Configure(cfg AzStorageConfig) error {
 	}
 
 	bb.listDetails = container.ListBlobsInclude{
-		Metadata:    true,
-		Tags:        bb.Config.filterHasTag,
+		Metadata: true,
+		// Include.Tags is only honored on Block Blob (flat namespace) accounts.
+		// On HNS (ADLS Gen2), the list endpoint does not populate BlobTags in
+		// the response; tags must be fetched per-blob via GetTags — handled in
+		// processBlobItems.
+		Tags:        bb.Config.filterHasTag && !bb.Config.isHNS,
 		Deleted:     false,
 		Snapshots:   false,
 		Permissions: false, //Added to get permissions, acl, group, owner for HNS accounts
@@ -711,7 +715,24 @@ func (bb *BlockBlob) processBlobItems(blobItems []*container.BlobItem) ([]*inter
 			filterAttr.Name = blobAttr.Name
 			filterAttr.Mtime = blobAttr.Mtime
 			filterAttr.Size = blobAttr.Size
-			filterAttr.Tags = parseBlobTags(blobInfo.BlobTags)
+			filterAttr.Tags = nil
+			if bb.Config.filterHasTag {
+				// HNS (ADLS Gen2) does not return blob index tags in the list response
+				// (BlobTags is nil and BlobTagCount is not populated on list items even
+				// when Include.Tags is requested). Fetch tags per-blob via GetTags so
+				// the tag filter can be evaluated. This N+1 cost only applies when a
+				// tag= filter is configured on an HNS account.
+				if bb.Config.isHNS {
+					tagResp, err := bb.Container.NewBlockBlobClient(*blobInfo.Name).GetTags(context.Background(), nil)
+					if err != nil {
+						log.Warn("BlockBlob::processBlobItems : Failed to get tags for %s [%s]", *blobInfo.Name, err.Error())
+					} else {
+						filterAttr.Tags = parseBlobTags(&tagResp.BlobTags)
+					}
+				} else {
+					filterAttr.Tags = parseBlobTags(blobInfo.BlobTags)
+				}
+			}
 
 			if bb.Config.filter.IsAcceptable(&filterAttr) {
 				blobList = append(blobList, blobAttr)
@@ -1883,6 +1904,8 @@ func (bb *BlockBlob) SetFilter(filter string) error {
 		return err
 	}
 	bb.Config.filterHasTag = filterReferencesTag(filter)
-	bb.listDetails.Tags = bb.Config.filterHasTag
+	// HNS accounts do not populate tags in the list response; leave Include.Tags off
+	// and let processBlobItems fetch tags per-blob when needed.
+	bb.listDetails.Tags = bb.Config.filterHasTag && !bb.Config.isHNS
 	return nil
 }
--- a/component/azstorage/block_blob.go
+++ b/component/azstorage/block_blob.go
@@ -589,6 +589,10 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
 	}
 
 	attr, err = bb.getAttrUsingRest(name)
+	if err != nil {
+		log.Err("BlockBlob::GetAttr : Failed to get attributes for %s [%s]", name, err.Error())
+		return attr, err
+	}
 	if bb.Config.filter != nil && attr != nil {
 		filterAttr := blobfilter.BlobAttr{
 			Name:  attr.Name,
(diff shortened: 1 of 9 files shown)


component/azstorage/config.go
```diff
@@ -559,11 +559,39 @@ func configureBlobFilter(azStorage *AzStorage, opt AzStorageOptions) error {
 		log.Err("configureBlobFilter : Failed to configure blob filter %s", err.Error())
 		return errors.New("failed to configure blob filter")
 	}
+	azStorage.stConfig.filterHasTag = filterReferencesTag(opt.Filter)
+
+	if azStorage.stConfig.filterHasTag && azStorage.stConfig.authConfig.AccountType == EAccountType.ADLS() {
```
gleason-m: There is a public preview feature to support blob index tags for HNS storage accounts. See https://learn.microsoft.com/en-us/rest/api/storageservices/set-blob-tags?tabs=microsoft-entra-id#remarks

> Note: Blob tags for hierarchical namespace enabled accounts is in public preview, and accessible via the Microsoft.Storage/BlobIndexForHns AFEC flag. For more information, see [Set up preview features in Azure subscription](https://learn.microsoft.com/en-us/azure/azure-resource-manager/management/preview-features).

Possible to support accounts that have onboarded to the preview feature?

syeleti-msft: usually, GA releases of blobfuse2 doesn't support preview features in the storage. hence, we can't procced to add this feature in blobfuse yet.

syeleti-msft: Seems like this feature is already supported in go SDK, we can continue supporting this feature.

component/azstorage/config.go
```diff
@@ -559,11 +559,34 @@ func configureBlobFilter(azStorage *AzStorage, opt AzStorageOptions) error {
...
 	return nil
 }
 
+// filterReferencesTag returns true when the raw filter expression includes a
+// `tag=` clause. The blobfilter package does not expose its parsed filter set,
+// so we inspect the input string ourselves to know whether GetAttr paths must
+// fetch blob index tags before evaluating the filter.
+func filterReferencesTag(filter string) bool {
+	for _, clause := range strings.Split(filter, "||") {
+		for _, sub := range strings.Split(clause, "&&") {
```
souravgupta-msft: IDE was suggesting this: Ranging over SplitSeq is more efficient
```suggestion
	for clause := range strings.SplitSeq(filter, "||") {
		for sub := range strings.SplitSeq(clause, "&&") {
```

component/azstorage/config.go
souravgupta-msft, line 548: Can we update the method signature here to only take the filter string and not the entire AzStorageOptions, as we are only using the filter parameter of this struct.
```
func configureBlobFilter(azStorage *AzStorage, filter string) error {
```

component/azstorage/block_blob.go
```diff
@@ -577,17 +578,31 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
...
 			Mtime: attr.Mtime,
 			Size:  attr.Size,
-		}) {
+		}
+		// GetProperties does not return blob index tags. When the configured
+		// filter references tags, fetch them explicitly before deciding.
+		if bb.Config.filterHasTag && !attr.IsDir() {
+			tags, terr := bb.getBlobTags(name)
+			if terr != nil {
+				log.Warn("BlockBlob::GetAttr : Failed to get tags for %s [%s]", name, terr.Error())
```
souravgupta-msft: This should be error level log

component/azstorage/block_blob.go
```diff
@@ -577,17 +578,31 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
...
+		filterAttr := blobfilter.BlobAttr{
 			Name:  attr.Name,
 			Mtime: attr.Mtime,
 			Size:  attr.Size,
-		}) {
+		}
+		// GetProperties does not return blob index tags. When the configured
+		// filter references tags, fetch them explicitly before deciding.
+		if bb.Config.filterHasTag && !attr.IsDir() {
+			tags, terr := bb.getBlobTags(name)
```
souravgupta-msft, line 604: If GetTags call fails for a blob and filter is based on tags, then should we return EIO error for this blob?

component/azstorage/datalake.go
```diff
@@ -457,11 +457,20 @@ func (dl *Datalake) GetAttr(name string) (blobAttr *internal.ObjAttr, err error)
 	}
 
 	if dl.Config.filter != nil {
-		if !dl.Config.filter.IsAcceptable(&blobfilter.BlobAttr{
+		filterAttr := &blobfilter.BlobAttr{
```
souravgupta-msft, line 461: Let's figure out what we need to do for the list call with tags filter

component/azstorage/config.go
```diff
@@ -559,11 +559,34 @@ func configureBlobFilter(azStorage *AzStorage, opt AzStorageOptions) error {
...
+			token := strings.TrimSpace(sub)
+			lowered := strings.ToLower(strings.Map(func(r rune) rune {
+				if r == ' ' || r == '\t' {
+					return -1
+				}
+				return r
+			}, token))
+			if strings.HasPrefix(lowered, "tag=") {
```
souravgupta-msft: Tags filter also supports `!=` operator. So, it should be added here

```suggestion
			if strings.HasPrefix(lowered, "tag=") || strings.HasPrefix(lowered, "tag!=") {
```

souravgupta-msft: Also, add tests for `!=` operator as well

component/azstorage/datalake.go
```diff
@@ -457,11 +457,20 @@ func (dl *Datalake) GetAttr(name string) (blobAttr *internal.ObjAttr, err error)
...
-		}) {
+		}
+		if dl.Config.filterHasTag && !blobAttr.IsDir() {
+			tagResp, err := fileClient.GetTags(context.Background(), nil)
+			if err != nil {
+				log.Err("Datalake::GetAttr : Failed to get tags for %s [%s]", name, err.Error())
+			} else {
+				filterAttr.Tags = parseBlobTags(&tagResp.BlobTags)
+			}
+		}
```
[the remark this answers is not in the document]
souravgupta-msft: Same as this comment, should we return EIO for this case when GetTags fails for the blob

component/azstorage/block_blob.go
```diff
@@ -105,6 +105,7 @@ func (bb *BlockBlob) Configure(cfg AzStorageConfig) error {
 
 	bb.listDetails = container.ListBlobsInclude{
 		Metadata:    true,
+		Tags:        bb.Config.filterHasTag,
```
souravgupta-msft: Add an entry in the changelog for this bug fix

component/azstorage/config.go
syeleti-msft: addressed https://github.com/Azure/azure-storage-fuse/pull/2261/changes/d42c68fb5d9823d7b660f06ebb4876a4c2640790

component/azstorage/block_blob.go
syeleti-msft: addressed https://github.com/Azure/azure-storage-fuse/pull/2261/changes/d42c68fb5d9823d7b660f06ebb4876a4c2640790

component/azstorage/datalake.go
syeleti-msft: calling getTags API is only way i can see to get the tags for the listed blobs. also the tagCounter is also not getting returned for listBlob API in HNS acnts to optimize. so added a change for calling getTags on every blob to filter.

component/azstorage/block_blob.go
syeleti-msft: EIO is not much relavant in stat calls. i think EACCES suits better for this kind of errors

component/azstorage/config.go
syeleti-msft: addressed https://github.com/Azure/azure-storage-fuse/pull/2261/changes/d42c68fb5d9823d7b660f06ebb4876a4c2640790

component/azstorage/block_blob.go
syeleti-msft: added

component/azstorage/config.go
syeleti-msft: we are also using the `azStorage.stConfig.filterHasTag` to see if the tag is configured in the filter.

component/azstorage/block_blob.go
```diff
@@ -577,17 +578,31 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
...
+		// GetProperties does not return blob index tags. When the configured
+		// filter references tags, fetch them explicitly before deciding.
+		if bb.Config.filterHasTag && !attr.IsDir() {
+			tags, terr := bb.getBlobTags(name)
+			if terr != nil {
+				log.Warn("BlockBlob::GetAttr : Failed to get tags for %s [%s]", name, terr.Error())
+			} else {
+				filterAttr.Tags = tags
+			}
+		}
```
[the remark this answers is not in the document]
syeleti-msft: Sure, EACCES sounds good

component/azstorage/config.go
```diff
@@ -559,11 +559,34 @@ func configureBlobFilter(azStorage *AzStorage, opt AzStorageOptions) error {
 		log.Err("configureBlobFilter : Failed to configure blob filter %s", err.Error())
 		return errors.New("failed to configure blob filter")
 	}
+	azStorage.stConfig.filterHasTag = filterReferencesTag(opt.Filter)
 
 	log.Crit("configureBlobFilter : Blob filter configured %s", opt.Filter)
 	return nil
 }
```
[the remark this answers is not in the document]
syeleti-msft: added support for ADLS accounts, forget to change the description

component/azstorage/block_blob.go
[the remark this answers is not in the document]
syeleti-msft: returning EACCES makes more sense here

component/azstorage/block_blob.go
```diff
@@ -685,6 +715,24 @@ func (bb *BlockBlob) processBlobItems(blobItems []*container.BlobItem) ([]*inter
...
+				if bb.Config.isHNS {
+					tagResp, err := bb.Container.NewBlockBlobClient(*blobInfo.Name).GetTags(context.Background(), nil)
+					if err != nil {
+						log.Warn("BlockBlob::processBlobItems : Failed to get tags for %s [%s]", *blobInfo.Name, err.Error())
+					} else {
+						filterAttr.Tags = parseBlobTags(&tagResp.BlobTags)
+					}
+				} else {
+					filterAttr.Tags = parseBlobTags(blobInfo.BlobTags)
+				}
```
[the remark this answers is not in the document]
syeleti-msft: will address

component/azstorage/config.go
```diff
@@ -559,11 +559,34 @@ func configureBlobFilter(azStorage *AzStorage, opt AzStorageOptions) error {
...
+				}
+				return r
+			}, token))
+			if strings.HasPrefix(lowered, "tag=") || strings.HasPrefix(lowered, "tag!=") {
+				return true
+			}
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
syeleti-msft: we pin the latest go release, so wouldn't be a problem

component/azstorage/block_blob.go
```diff
@@ -596,6 +615,17 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
...
+// getBlobTags fetches blob index tags for the given blob name and returns them
+// as a flat map. Returns nil if the blob has no tags.
+func (bb *BlockBlob) getBlobTags(name string) (map[string]string, error) {
+	blobClient := bb.Container.NewBlockBlobClient(filepath.Join(bb.Config.prefixPath, name))
+	resp, err := blobClient.GetTags(context.Background(), nil)
+	if err != nil {
+		return nil, err
+	}
+	return parseBlobTags(&resp.BlobTags), nil
+}
```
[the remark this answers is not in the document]
syeleti-msft: not applicable

component/azstorage/block_blob.go
```diff
@@ -577,17 +582,31 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
...
+				log.Err("BlockBlob::GetAttr : Failed to get tags for %s [%s]", name, terr.Error())
+			} else {
+				filterAttr.Tags = tags
+			}
+		}
+
+		if !bb.Config.filter.IsAcceptable(&filterAttr) {
 			log.Debug("BlockBlob::GetAttr : Filtered out %s", name)
 			return nil, syscall.ENOENT
 		}
```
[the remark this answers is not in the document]
syeleti-msft: fixed in https://github.com/Azure/azure-storage-fuse/pull/2261/commits/90808d95d75a57c908f92b77e369b546d1222ff0

component/azstorage/block_blob.go
```diff
@@ -685,6 +715,24 @@ func (bb *BlockBlob) processBlobItems(blobItems []*container.BlobItem) ([]*inter
...
+					tagResp, err := bb.Container.NewBlockBlobClient(*blobInfo.Name).GetTags(context.Background(), nil)
+					if err != nil {
+						log.Warn("BlockBlob::processBlobItems : Failed to get tags for %s [%s]", *blobInfo.Name, err.Error())
+					} else {
+						filterAttr.Tags = parseBlobTags(&tagResp.BlobTags)
+					}
+				} else {
+					filterAttr.Tags = parseBlobTags(blobInfo.BlobTags)
+				}
+			}
```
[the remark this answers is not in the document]
syeleti-msft: fixed in https://github.com/Azure/azure-storage-fuse/pull/2261/commits/90808d95d75a57c908f92b77e369b546d1222ff0

component/azstorage/block_blob.go
syeleti-msft: fixed in https://github.com/Azure/azure-storage-fuse/pull/2261/commits/90808d95d75a57c908f92b77e369b546d1222ff0

component/azstorage/block_blob.go
```diff
@@ -577,17 +582,31 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
...
-	} else {
-		attr, err = bb.getAttrUsingRest(name)
+		// getAttrUsingList -> List -> processBlobItems already evaluates the
+		// filter (including tag filters), so a successful return means the blob
+		// passed the filter.
+		return bb.getAttrUsingList(name)
 	}
 
+	attr, err = bb.getAttrUsingRest(name)
 	if bb.Config.filter != nil && attr != nil {
```
souravgupta-msft, line 595: We should check the error here first and if error is not nil, log and return it

component/azstorage/block_blob.go
```diff
@@ -596,6 +615,17 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
...
 }
 
+// getBlobTags fetches blob index tags for the given blob name and returns them
+// as a flat map. Returns nil if the blob has no tags.
+func (bb *BlockBlob) getBlobTags(name string) (map[string]string, error) {
+	blobClient := bb.Container.NewBlockBlobClient(filepath.Join(bb.Config.prefixPath, name))
+	resp, err := blobClient.GetTags(context.Background(), nil)
+	if err != nil {
+		return nil, err
+	}
```
souravgupta-msft, line 629: Add an error log here

component/azstorage/block_blob.go
```diff
@@ -685,6 +715,24 @@ func (bb *BlockBlob) processBlobItems(blobItems []*container.BlobItem) ([]*inter
...
 			filterAttr.Size = blobAttr.Size
+			filterAttr.Tags = nil
+			if bb.Config.filterHasTag {
+				// HNS (ADLS Gen2) does not return blob index tags in the list response
+				// (BlobTags is nil and BlobTagCount is not populated on list items even
+				// when Include.Tags is requested). Fetch tags per-blob via GetTags so
+				// the tag filter can be evaluated. This N+1 cost only applies when a
+				// tag= filter is configured on an HNS account.
+				if bb.Config.isHNS {
+					tagResp, err := bb.Container.NewBlockBlobClient(*blobInfo.Name).GetTags(context.Background(), nil)
```
souravgupta-msft: We should use the method `bb.getBlobTags()` here so that GetTags API is called from one place only. Please check if you have to remove the prefix path from the blob name

component/azstorage/block_blob.go
syeleti-msft: resolved https://github.com/Azure/azure-storage-fuse/pull/2261/commits/541faf68c04c3a7f9d2e4107118b1964c36dc6d0

component/azstorage/block_blob.go
syeleti-msft: resolved https://github.com/Azure/azure-storage-fuse/pull/2261/commits/541faf68c04c3a7f9d2e4107118b1964c36dc6d0

component/azstorage/block_blob.go
syeleti-msft: resolved https://github.com/Azure/azure-storage-fuse/pull/2261/commits/541faf68c04c3a7f9d2e4107118b1964c36dc6d0

souravgupta-msft (approved):

component/azstorage/block_blob.go
```diff
@@ -577,17 +582,35 @@ func (bb *BlockBlob) GetAttr(name string) (attr *internal.ObjAttr, err error) {
...
+		// filter (including tag filters), so a successful return means the blob
+		// passed the filter.
+		return bb.getAttrUsingList(name)
 	}
 
+	attr, err = bb.getAttrUsingRest(name)
+	if err != nil {
+		log.Err("BlockBlob::GetAttr : Failed to get attributes for %s [%s]", name, err.Error())
+		return attr, err
+	}
```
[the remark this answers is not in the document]
syeleti-msft: resolved https://github.com/Azure/azure-storage-fuse/pull/2261/commits/92fba9f65cfb5e15a8a2518f9206d7f1458f588f

souravgupta-msft (approved):

PR #2265, opened 2026-06-26
Blobfuse integration with distributed cache

nearora-msft:
## Type of Change

- [ ] Bug fix
- [ x] New feature
- [ ] Other (describe):

## Description

Adds a new optional pipeline component, dist_cache, that sits between the local cache (file_cache or block_cache, L1) and azstorage (L3) and uses a shared cluster of cache servers (L2) to reduce Azure Blob egress and read latency for multi-node workloads.

The dist-cache client itself has been factored out into a separate, versioned Go module — github.com/nearora-msft/dist-cache-client-go (pinned at v0.1.0) — so this PR contains only the blobfuse-side integration: a new component, a few internal API additions for ETag plumbing, pipeline wiring, validation, and configuration/docs.

A running Tachyon cache server cluster is a prerequisite to use the feature; without it, the dist_cache component is simply not included in the pipeline.

- **Feature**:

**New component:**

 dist_cache: new pipeline component implementing both the file-cache path (CopyToFile / CopyFromFile) and the block-cache path (ReadInBuffer / StageData / CommitData), plus invalidation hooks (DeleteFile, RenameFile, TruncateFile). Key behaviors:

1. Write-through to azstorage first; L2 populated asynchronously after commit succeeds.
2. ETag-versioned group IDs so a DeleteGroup for an old revision never affects chunks uploaded under a new revision.
3. Per-file "dirty" TTL window after an invalidation so other nodes bypass stale L2 data while the new version is being repopulated.
4. Bounded parallel chunk-recovery on cache misses; per-chunk fallthrough to azstorage with optional L2 backfill.
5. Lock-on-miss coordination (ErrNotFoundGotLock / ErrNotFoundAlreadyLocked) with polling so multiple nodes don't race to populate the same chunk.
6. Pending-write buffer with size cap and idle-TTL eviction for abandoned writes.
7. Cancellation of in-flight flush/populate/read-upload goroutines on new writes or invalidations to prevent stale overwrites.
8. bypass-on-error for graceful degradation to direct azstorage. 
10. dist_cache_test.go — comprehensive unit tests against a dcacheClient interface mock covering happy paths, all miss paths, invalidation races, write versioning, dirty-window behavior, pending-write eviction, and error fallthrough.

**Pipeline wiring & validation**
1. imports.go— register the dist_cache component.
2. mount.go— auto-include dist_cache in the pipeline when a dist_cache config section is present.
3. util.go— ValidatePipeline rejects the unsupported dist_cache + xload combination.
4. config_parser.go — fix a latent nil-pointer panic in IsSet when traversing intermediate tree nodes (surfaced by the new config.IsSet("dist_cache") check).

**ETag plumbing through azstorage and file_cache**

The L2 cache key is versioned by the blob's Azure ETag, so the integration needs the ETag to flow into reads and out of writes:

1. component_options.go — add CopyToFileOptions.Etag (pinned at open time for versioned lookups) and CopyFromFileOptions.NewETag *string(populated by azstorage after upload).
2. connection.go— AzConnection.WriteFromFile signature gains a [newEtag *string](vscode-file://vscode-app/c:/Users/nearora/AppData/Local/Programs/Microsoft%20VS%20Code/fcf604774b/resources/app/out/vs/code/electron-browser/workbench/workbench.html) out-parameter.
3. block_blob.go— capture the upload response and write the sanitized [resp.ETag](vscode-file://vscode-app/c:/Users/nearora/AppData/Local/Programs/Microsoft%20VS%20Code/fcf604774b/resources/app/out/vs/code/electron-browser/workbench/workbench.html) into *newEtag when non-nil.
4. datalake.go— pass newEtag through to the underlying BlockBlob.WriteFromFile.
5. azstorage.go— forward options.NewETag from CopyFromFile into WriteFromFile.
6. file_cache.go — read the current attr in OpenFile and propagate attr.ETag into the downstream CopyToFile call.
7. block_blob_test.go and datalake_test.go— existing tests updated to pass nil for the new newEtag parameter.

**Dependencies**
go.mod / go.sum— add github.com/nearora-msft/dist-cache-client-go v0.1.0, golang.org/x/sync v0.21.0, and google.golang.org/protobuf v1.36.11 (indirect). No other direct dependency changes.

**Documentation & configuration**
sampleDistCacheConfig.yaml — new end-to-end sample showing the libfuse → block_cache → dist_cache → attr_cache → azstorage pipeline with all three discovery options (discovery URL, Kubernetes DNS, static server list) and the Tachyon prerequisite called out.
baseConfig.yaml— documents the full set of dist_cache options (discovery, port, bypass-on-error, ttl-seconds, chunk-size-mband the L1 requirements.

## How Has This Been Tested?

1. Unit tests: go test -timeout=5m ./component/dist_cache/... --tags=unittest,fuse3 — passes (mock-based; covers read/write paths, invalidation, pending writes, dirty window, miss-recovery, error fallthrough).
2. Existing e2e test suites are passing.
3. Adding integration tests to test with a loopbackfs and local API server
4. Local build: build.sh produces a working blobfuse2 binary; ./blobfuse2 --version and gen-config exercised.
5. End-to-end against a Tachyon AKS cluster.
6. More e2e tests will be updated in the upcoming PRs.

## Checklist

## Related Links
- [Issues](<link>)


cmd/mount.go
```diff
@@ -298,6 +298,11 @@ var mountCmd = &cobra.Command{
 				pipeline = append(pipeline, "file_cache")
 			}
 
+			// dist_cache sits between local cache (L1) and azstorage (L3)
+			if config.IsSet("dist_cache") {
```
souravgupta-msft, line 302: Where are you defining this CLI param?
This case is for when the mount is done via CLI parameters and not config file. See how it is configured for `--block-cache` and `--preload` above.

setup/baseConfig.yaml
```diff
@@ -41,6 +42,28 @@ file_cache:
...
+# Requires a running Tachyon cache server cluster (https://github.com/Azure/Tachyon).
+# Requires block_cache or file_cache in the pipeline. Mutually exclusive with xload.
+dist_cache:
+  # Server discovery — at least one method must be configured
+  discovery-url: <discovery endpoint URL for auto-detecting cache servers. Recommended for production>
+  # OR
+  k8s-service: <Kubernetes headless service name for DNS-based discovery>
+  k8s-namespace: <Kubernetes namespace where the cache service runs>
+  # OR
+  server-list: <comma-separated list of cache server addresses, e.g., "host1:9065,host2:9065,host3:9065">
```
souravgupta-msft, line 57: Can we add a comment here stating that at least one discovery method must be configured, either discovery-url, k8s-service or server-list.

setup/baseConfig.yaml
```diff
@@ -41,6 +42,28 @@ file_cache:
   # Optional 
   timeout-sec: <default cache eviction timeout (in sec). Default - 120 sec>
   max-size-mb: <maximum cache size allowed. Default - 80% of free disk space>
+
+# Distributed cache configuration (optional, for multi-node shared caching)
+# Requires a running Tachyon cache server cluster (https://github.com/Azure/Tachyon).
+# Requires block_cache or file_cache in the pipeline. Mutually exclusive with xload.
+dist_cache:
+  # Server discovery — at least one method must be configured
```
souravgupta-msft, line 50: How are we dealing with the case when none of these 3 discovery methods are set?
Are we failing the mount or falling back to single node file cache or block cache mode?

setup/baseConfig.yaml
souravgupta-msft: We must callout that specifying one of these methods are mandatory

souravgupta-msft, line 57: What happens when two or more of these discovery methods are specified in the config. Do we give preference to one or fail the mount?

setup/baseConfig.yaml
```diff
@@ -41,6 +42,28 @@ file_cache:
...
+  # Common options
+  port: <cache server port. Default - 9065>
+  bypass-on-error: <if true, cache errors fall through to Azure Blob Storage. Default - false>
+  ttl-seconds: <TTL for cached data in seconds. Default - 0 (no expiry)>
+  chunk-size-mb: <chunk size in MB, used when block_cache is not present. Default - 16>
+  cache-prefix: <prefix for cache keys. Default - derived from azstorage account/container>
+  max-conns-per-server: <maximum TCP connections per cache server. Default - 8>
+  request-timeout-sec: <timeout for individual cache requests in seconds. Default - 30>
+  auth-account-name: <cache auth account name, if auth is enabled on server>
+  auth-account-key: <cache auth account key, if auth is enabled on server>
```
souravgupta-msft:
```suggestion
  auth-account-name: <cache server auth account name, if auth is enabled on server>
  auth-account-key: <cache server auth account key, if auth is enabled on server>
```

go.mod
```diff
@@ -22,12 +22,18 @@ require (
...
 	go.uber.org/atomic v1.11.0
+	golang.org/x/sync v0.21.0
 	golang.org/x/time v0.15.0
 	gopkg.in/ini.v1 v1.67.3
 	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.1
 )
 
+require (
+	github.com/nearora-msft/dist-cache-client-go v0.1.0
```
souravgupta-msft: Any plan to make this repo official under github.com/Azure

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1108 @@
...
+func init() {
+	internal.AddComponent(compName, NewDistCacheComponent)
+
+	discoveryFlag := config.AddStringFlag("dist-cache-discovery-url", "",
+		"distributed cache discovery endpoint (recommended)")
+	config.BindPFlag(compName+".discovery-url", discoveryFlag)
+
+	serverListFlag := config.AddStringFlag("dist-cache-server-list", "",
+		"comma-separated list of distributed cache server addresses (fallback)")
+	config.BindPFlag(compName+".server-list", serverListFlag)
```
souravgupta-msft, line 171: Should we also export `k8s-service` and `k8s-namespace` in the CLI parameters so that AKS customer can use these options. Limiting them to config only will not directly allow then to use these params.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1108 @@
...
+	if dc.conf.K8sService != "" && dc.conf.K8sNamespace != "" {
+		opts = append(opts, dcache.WithK8sDiscovery(dc.conf.K8sService, dc.conf.K8sNamespace))
+	}
+	if dc.conf.ServerList != "" {
+		servers := strings.Split(dc.conf.ServerList, ",")
+		for i := range servers {
+			servers[i] = strings.TrimSpace(servers[i])
+		}
+		opts = append(opts, dcache.WithServerList(servers))
+	}
```
souravgupta-msft, line 293: This should be if-elseif, otherwise it will configure all the discovery modes.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1108 @@
...
+	// Common options
+	Port            int    `config:"port"              yaml:"port,omitempty"`
+	TTLSeconds      uint32 `config:"ttl-seconds"       yaml:"ttl-seconds,omitempty"`
+	MaxFileSizeMB   int    `config:"max-file-size-mb"  yaml:"max-file-size-mb,omitempty"`
+	AuthAccountName string `config:"auth-account-name" yaml:"auth-account-name,omitempty"`
+	AuthAccountKey  string `config:"auth-account-key"  yaml:"auth-account-key,omitempty"`
+	HashType        string `config:"hash-type"         yaml:"hash-type,omitempty"`
+	BypassOnError   bool   `config:"bypass-on-error"   yaml:"bypass-on-error,omitempty"`
+	CachePrefix     string `config:"cache-prefix"      yaml:"cache-prefix,omitempty"`
+	MaxConnsPerSvr  int    `config:"max-conns-per-server" yaml:"max-conns-per-server,omitempty"`
```
souravgupta-msft: Where are we defining the default values for these params?
Is it defined in the dist-cache-client-go?

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1085 @@
...
+	if maxSize > 0 && pf != nil && pf.totalSize+dataLen > maxSize {
+		log.Debug("DistCache::StageData : pending size would exceed %dMB for %s, skipping L2 write-warming",
+			dc.conf.MaxFileSizeMB, options.Name)
+		delete(dc.pendingWrites, options.Name)
+		dc.pendingMu.Unlock()
+		return nil
+	}
+
+	// Buffer the chunk for deferred L2 population at commit time.
+	dataCopy := make([]byte, dataLen)
```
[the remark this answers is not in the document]
nearora-msft: Resolving because it compiles.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1085 @@
...
+// and writes it to the file at the correct offset. If populateCache is true,
+// the chunk is also uploaded to the distributed cache asynchronously.
+func (dc *DistCache) fetchChunkFromRemote(ctx context.Context, options internal.CopyToFileOptions, offset, size int64, populateCache bool) error {
+	select {
+	case <-ctx.Done():
+		return ctx.Err()
+	default:
+	}
+
+	buf := make([]byte, size)
```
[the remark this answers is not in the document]
nearora-msft: Resolving because it compiles.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1085 @@
...
+		uploadCtx := dc.getReadUploadCtx(options.Name)
+		go dc.uploadChunkAsync(uploadCtx, options.Name, options.Etag, offset, dataCopy)
+	}
+	return nil
+}
+
+// pollUntilChunkCached waits for a single chunk to become available in the
+// distributed cache and writes it to the file. Returns nil on success.
+func (dc *DistCache) pollUntilChunkCached(ctx context.Context, options internal.CopyToFileOptions, offset, size int64) error {
+	buf := make([]byte, size)
```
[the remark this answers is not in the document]
nearora-msft: Resolving because it compiles.

cmd/mount.go
```diff
@@ -298,6 +298,11 @@ var mountCmd = &cobra.Command{
...
+				pipeline = append(pipeline, "dist_cache")
+			}
```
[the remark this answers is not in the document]
nearora-msft: I will take this up in the UX PR.

common/util.go
```diff
@@ -572,6 +572,12 @@ func ValidatePipeline(pipeline []string) error {
 		return fmt.Errorf("mount: block-cache and xload cannot be used together")
 	}
 
+	// dist_cache requires file_cache or block_cache as upstream L1
+	if ComponentInPipeline(pipeline, "dist_cache") &&
+		ComponentInPipeline(pipeline, "xload") {
+		return fmt.Errorf("mount: dist-cache and xload cannot be used together")
+	}
```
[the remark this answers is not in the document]
nearora-msft: This will be taken up in the UX related changes where only block_cache will be allowed as an L1 for distributed cache.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1108 @@
...
```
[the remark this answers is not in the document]
nearora-msft: It compiles.

common/util.go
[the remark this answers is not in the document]
nearora-msft: This will be taken up when a UX is finalized.

component/dist_cache/dist_cache_integration_test.go
```diff
@@ -0,0 +1,1744 @@
...
+		}
+
+		// Read upload data if present
+		var uploadData []byte
+		if upload := req.GetUploadrequest(); upload != nil && upload.Filesize > 0 {
+			uploadData = make([]byte, upload.Filesize)
+			if _, err := io.ReadFull(nc, uploadData); err != nil {
+				return
+			}
+		}
```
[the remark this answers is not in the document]
nearora-msft: It compiles and the file size check is an over optimization because this is a test file.

component/dist_cache/dist_cache_integration_test.go
```diff
@@ -0,0 +1,1744 @@
...
+		offset := p.Downloadrequest.Offset
+		length := p.Downloadrequest.Length
+		if offset >= uint64(len(data)) {
+			return &pb.DownloadResponse{Result: pb.DownloadResponse_NOT_FOUND}, nil
+		}
+		end := offset + length
+		if end > uint64(len(data)) || length == 0 {
+			end = uint64(len(data))
+		}
+		slice := data[offset:end]
```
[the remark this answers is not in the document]
nearora-msft: It compiles.

cmd/mount.go
nearora-msft: The CLI params aren't defined yet and will come in after the UX design proposal is finalized. I will create another PR with all the relevant UX changes.

setup/baseConfig.yaml
nearora-msft: The comments already say that:   # Server discovery — at least one method must be configured

setup/baseConfig.yaml
nearora-msft: If none of them are set, mount will fail. See the validations here:
https://github.com/Azure/azure-storage-fuse/blob/103ea675ddef27b2a1903cfd8c5d78a12c071b3b/component/dist_cache/dist_cache.go#L184

setup/baseConfig.yaml
nearora-msft: When more than one are set, the priority order is

discoveryURL > K8s DNS > Static server list.

See here:
https://github.com/nearora-msft/dist-cache-client-go/blob/473b2a7262c6f6893ae3c48e3a41faf12e908fcb/discovery.go#L73-L105

go.mod
nearora-msft: This client will most likely be placed under the Tachyon repo. Open sourcing of Tachyon repo is under progress.

component/dist_cache/dist_cache.go
nearora-msft: We would like the customers to prefer a discovery endpoint URL instead. K8s-service and k8s-namespace should only be used as a fallback. This is why the plan is to just have the discovery url as the CLI param. The UX is more clearly defined in the UX proposal document.

setup/baseConfig.yaml
nearora-msft: Updating the comments.

component/dist_cache/dist_cache.go
nearora-msft: Instead of doing an if-elseif, logging a warning if multiple discovery methods are set. This will ensure that if one if them isn't reachable, we can fall through to the next one.

component/dist_cache/dist_cache.go
nearora-msft: Yes, the defaults are defined in the client:
https://github.com/nearora-msft/dist-cache-client-go/blob/main/options.go#L8-L18

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1145 @@
...
+	// Common options
+	Port            int    `config:"port"              yaml:"port,omitempty"`
+	TTLSeconds      uint32 `config:"ttl-seconds"       yaml:"ttl-seconds,omitempty"`
+	MaxFileSizeMB   int    `config:"max-file-size-mb"  yaml:"max-file-size-mb,omitempty"`
+	AuthAccountName string `config:"auth-account-name" yaml:"auth-account-name,omitempty"`
+	AuthAccountKey  string `config:"auth-account-key"  yaml:"auth-account-key,omitempty"`
+	HashType        string `config:"hash-type"         yaml:"hash-type,omitempty"`
+	BypassOnError   bool   `config:"bypass-on-error"   yaml:"bypass-on-error,omitempty"`
+	CachePrefix     string `config:"cache-prefix"      yaml:"cache-prefix,omitempty"`
+	MaxConnsPerSvr  int    `config:"max-conns-per-server" yaml:"max-conns-per-server,omitempty"`
```
souravgupta-msft: Should we decide what all parameters we should expose in the CLI or config. We should try to make the public surface area as clean and minimal as possible. So, parameters like `ttl-seconds`, `hash-type`, etc. can be decided based on some heuristics and should not be exposed to customers.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1139 @@
...
+		err = config.UnmarshalKey("block_cache.block-size-mb", &blockSizeMB)
+		if err != nil {
+			log.Warn("DistCache::Configure : Failed to read block-size-mb, using default %d MB", defaultBlockSizeMB)
+			blockSizeMB = defaultBlockSizeMB
+		}
+	} else if config.IsSet("stream.block-size-mb") {
+		err = config.UnmarshalKey("stream.block-size-mb", &blockSizeMB)
+		if err != nil {
+			blockSizeMB = defaultBlockSizeMB
+		}
```
souravgupta-msft, line 252: Error log missing here

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1139 @@
...
+	} else if conf.ChunkSizeMB > 0 {
+		blockSizeMB = conf.ChunkSizeMB
+	}
+	dc.chunkSize = int64(blockSizeMB * 1024 * 1024)
+
+	log.Info("DistCache::Configure : chunk-size=%d, bypass-on-error=%v",
```
souravgupta-msft, line 267: Let's log all the config parameters here like port, TTLSeconds, MaxFileSizeMB, MaxConnsPerSvr, etc.

component/dist_cache/dist_cache.go
nearora-msft: Resolving that as this will be taken up in the UX PR.

component/dist_cache/dist_cache.go
nearora-msft: This code will be changed with the UX PR, will take it up then.

component/dist_cache/dist_cache.go
nearora-msft: This code will be changed with the UX PR, will take it up then.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1139 @@
...
+		// We own this chunk's miss — download from Azure and populate cache
+		log.Debug("DistCache::ReadInBuffer : L2 miss (got lock) %s offset=%d", name, options.Offset)
+		n, err = dc.NextComponent().ReadInBuffer(options)
+		if err != nil {
+			return n, err
+		}
+		dataCopy := make([]byte, n)
+		copy(dataCopy, options.Data[:n])
+		uploadCtx := dc.getReadUploadCtx(name)
+		go dc.uploadChunkAsync(uploadCtx, name, etag, options.Offset, dataCopy)
```
syeleti-msft, line 564: should we need to implement worker pool to limit the number of go rountines spawned by this comp?

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1139 @@
...
+	}
+
+	if err == dcache.ErrNotFoundGotLock {
+		// We own this chunk's miss — download from Azure and populate cache
+		log.Debug("DistCache::ReadInBuffer : L2 miss (got lock) %s offset=%d", name, options.Offset)
+		n, err = dc.NextComponent().ReadInBuffer(options)
+		if err != nil {
+			return n, err
+		}
+		dataCopy := make([]byte, n)
```
syeleti-msft: where are we managing the memory? if we are making the data copies for all the calls coming from the block cache, then wouldn't we be using more memory than necessary? any reason why we are dynamically allocating always rather than having some memory management of the buffers using sync.pool or some other approach.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1139 @@
...
+		copy(dataCopy, options.Data[:n])
+		uploadCtx := dc.getReadUploadCtx(name)
+		go dc.uploadChunkAsync(uploadCtx, name, etag, options.Offset, dataCopy)
+		return n, nil
+	}
+
+	if err == dcache.ErrNotFound {
```
syeleti-msft, line 588: I see there is a similar code for all the 3 errors (i.e., getting data from azure and upload it to dcache), can't we create a anonymous function and call it from the all 3 error if conditions? and also I see `ErrorNotFound` and `ErrNotFoundGotLock ` does the same thing, why have 2 different branches?

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1139 @@
...
+		log.Debug("DistCache::StageData : pending size would exceed %dMB for %s, skipping L2 write-warming",
+			dc.conf.MaxFileSizeMB, options.Name)
+		delete(dc.pendingWrites, options.Name)
+		dc.pendingMu.Unlock()
+		return nil
+	}
+
+	// Buffer the chunk for deferred L2 population at commit time.
+	dataCopy := make([]byte, dataLen)
+	copy(dataCopy, options.Data)
```
syeleti-msft: as we are asking the user to configure the memory pool size for dcache. which we internally split between dcache and block cache is what i believe. then we should implement some kind buffer manager inside dcache to adhere to that limit. in the past we see OOM killing of blobfuse gives very bad experience to the user.

syeleti-msft: we are doing the buffer copy inside the `dc.pendingMu.Lock()` which can cause contention as the buffer copy of std 16MB buffer can take some time. can't we move the copy of the buffer before the lock?

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1139 @@
...
+	// Cancel any in-flight read-path uploads that may overwrite our fresh data
+	dc.cancelReadUploads(options.Name)
+
+	dc.markDirty(options.Name)
+
+	// Delete old version chunks if we had a previous ETag
+	if oldETag != "" {
+		oldGID := fileGroupID(options.Name, oldETag)
+		log.Debug("DistCache::CommitData : deleting old group %q for %s", string(oldGID), options.Name)
+		if err := dc.client.DeleteGroup(context.Background(), oldGID); err != nil {
```
syeleti-msft, line 707: if `dc.NextComponent().CommitData(options)` this call is success that means that the file that was written by our blobfuse process was success. although eTag changes due to the putBlockList REST call, that doesn't mean the data got changed. 

I see that we are using eTag in chunk naming that would conflict with the upcoming reads. but is that the only reason why we are deleting the chunks here? if that's the reason, can't we make the UUID as filename_blockID as blockID is also 24 byte UUID

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1139 @@
...
+
+	// Buffer the chunk for deferred L2 population at commit time.
+	dataCopy := make([]byte, dataLen)
+	copy(dataCopy, options.Data)
+
+	if pf == nil {
+		pf = &pendingFile{}
+		dc.pendingWrites[options.Name] = pf
+	}
+	pf.chunks = append(pf.chunks, pendingChunk{
```
syeleti-msft, line 654: I see we are not doing async upload to L2 while staging the data. so are we limiting the size of the slice `pf.chunks` somewhere? otherwise it would easily exhaust all the memory we have if we write a sufficiently big size of a file.

I see that evictStalePending wouldn't guard this as while writing the `lastActivity` time would always be recent and we wouldn't evict that file

syeleti-msft: also did we done a test of writing a 100GB file using this code in real world use case? i feel this would crash the process. but correct me if i'm wrong

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1139 @@
...
+	}
+	_, err = options.File.WriteAt(buf[:n], offset)
+	return err
+}
+
+// pollChunkIntoBuffer waits for a single chunk to become available in the
+// distributed cache and copies it into buf. Returns the number of bytes read.
+func (dc *DistCache) pollChunkIntoBuffer(ctx context.Context, name, etag string, offset int64, buf []byte) (int, error) {
+	const (
+		maxPollDuration = 30 * time.Second
```
syeleti-msft: given reading the same file from 2 nodes is common scenario. and when this happens concurrently, doesn't this 30 second duration causes large delay in IO / thorughput? did we test reading the file (say 100GB) from more than 2 nodes concurrently to exercise this path and how does the perf numbers compare to single node read.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,1139 @@
...
+	return nil
+}
+
+func (dc *DistCache) Priority() internal.ComponentPriority {
+	return internal.EComponentPriority.LevelMid()
+}
+
+// --- Read path (file_cache) ---
+
+func (dc *DistCache) CopyToFile(options internal.CopyToFileOptions) error {
```
syeleti-msft, line 339: lets move out the file cache changes from this PR if we are not sure we want to use them for now and let's keep it in separate branch. that would reduce some testing load.

nearora-msft: I don't have any strong opinions regarding this.
@souravgupta-msft suggested that we keep it for now and can remove that later if not needed.

component/dist_cache/dist_cache.go
nearora-msft: Yes, there is an upper cap on the max size per file see here: 

https://github.com/Azure/azure-storage-fuse/blob/5e05b9b2c9b9871534207b324fbf3a9a783f5099/component/dist_cache/dist_cache.go#L621

 This is configurable by MaxFileSizeMB : 
 
 https://github.com/Azure/azure-storage-fuse/blob/5e05b9b2c9b9871534207b324fbf3a9a783f5099/component/dist_cache/dist_cache.go#L607

nearora-msft: Yes, I did run 100G data_validation_test and truncate_test with my changes. Blobfuse didn't crash.

component/dist_cache/dist_cache.go
nearora-msft: We are deleting chunks here because we are unable to partially overwrite a file in dist_cache. So what we do instead is:
1. Delete all the existing chunks with the older etag.
2. Upload the partially modfied chunks.
3. The missing chunks will be populated on the next read. The next read will treat the chunks as a cache miss and will populate the cache.

AFAIk blockId is per block, here we need an ID that is the same for the entire file.

component/dist_cache/dist_cache.go
nearora-msft: This is a good catch. Will update.

nearora-msft: There's no memory pool configuration for dcache as of now. There's a per file configurable memory cap though.
I have a task to create a memory cap at the node level just like block_cache but the details will be hashed out when we plan to enable write path.

https://github.com/Azure/azure-storage-fuse/blob/5e05b9b2c9b9871534207b324fbf3a9a783f5099/component/dist_cache/dist_cache.go#L607

component/dist_cache/dist_cache.go
nearora-msft: Good catch. 

Technically, ErrNotFound wouldn't surface if we are downloading with lock enabled.
https://github.com/Azure/azure-storage-fuse/blob/5e05b9b2c9b9871534207b324fbf3a9a783f5099/component/dist_cache/dist_cache.go#L516

But if it does occur for some reason, we shouldn't populate dist_cache on a cache miss to prevent thundering herd.

Will Push a fix.

component/dist_cache/dist_cache.go
nearora-msft: That's a good suggestion. Will implement a fix for this.

component/dist_cache/dist_cache.go
syeleti-msft: `if maxSize > 0 && pendingSize > maxSize { `

how can you assume that maxSize of a file would fit always in their physical memory? often times `fileSize >>> total memory in vm`. am i missing something here?

nearora-msft: Yes, no assumptions that the file would fit the VM's memory. We are dropping all the pending chunks as soon as the per file cap is reached.

https://github.com/Azure/azure-storage-fuse/blob/5e05b9b2c9b9871534207b324fbf3a9a783f5099/component/dist_cache/dist_cache.go#L612

nearora-msft: Long term plan is to have a memory cap per node in addition to the per file memory cap and I have an item that tracks that.

component/dist_cache/dist_cache.go
nearora-msft: If we limit the number of goroutines the below mentioned issue could happen:

One client holds the lock for warming up the chunk with the cache in case of a cache miss and the other clients wait for the chunk to be available in the cache. If that client is blocked because we have limited the number of goroutines, other clients end up wasting cycles waiting on the cache.

I would avoid putting a limit here currently and would prefer to do it in case we see any issues later on.

component/dist_cache/dist_cache.go
nearora-msft: I've reduced the poll duration and maxBackoff to reasonable values.

component/dist_cache/dist_cache.go
syeleti-msft: its a bit confusing that maxFileSizeMB is actually max memory used by a file descriptor in dcache but what i thought from the naming is how much max file size a user can write.

nearora-msft:
Closing this pull request in favour of another PR : https://github.com/Azure/azure-storage-fuse/pull/2297

PR #2295, opened 2026-07-24
Optimize REST calls for single-page virtual directory listings

syeleti-msft:
## Summary

Optimize Block Blob directory listings by avoiding per-directory marker lookups when the hierarchical listing completes in a single page.

- Detect a single-page listing when both the incoming and outgoing continuation markers are empty.
- Create attributes for markerless virtual directories directly from `BlobPrefix` entries.
- Preserve the existing `dirList` marker-deduplication logic.
- Continue using REST marker checks for paginated listings, where the marker blob may appear on another page.
- Add unit tests and update the changelog.

## Motivation

Previously, listing a directory containing multiple markerless virtual directories triggered one `GetProperties` request for every `BlobPrefix`.

For example, 200 markerless virtual directories could require:

```text
1 List request + 200 GetProperties requests

### CODE REVIEW

File: component/azstorage/block_blob.go [hunk 1]
```diff
@@ -697,6 +699,10 @@ func (bb *BlockBlob) getListPath(prefix string) string {
 	return listPath
 }
 
+func isSinglePageListing(marker, nextMarker *string) bool {
+	return (marker == nil || *marker == "") && (nextMarker == nil || *nextMarker == "")
+}
```
souravgupta-msft, line 704: This can also be extended to last page when the next marker is nil, and we haven't yet found the marker blobs for the virtual directories. Since we have already done the full iteration, it will be safe to assume that the marker blobs for these virtual directories does not exist, and we can avoid making GetProperties calls for them.
Do you see any risks for this?

syeleti-msft: say, last page is n.

when blob prefix comes in (n)th page, and its marker came in (n - 1)th page, then it is possible that we return the same dirctory entry twice.

souravgupta-msft (approved):

PR #2297, opened 2026-07-24
feat: Initial changes to integrate distributed cache

nearora-msft:
## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] Other (describe):

## Description

**Distributed cache integration**:

Adds a new dist_cache component that provides a shared L2 distributed cache (backed by Azure/Tachyon sitting between block_cach (L1) and azstorage (L3). This PR is read-only: reads consult L2 first and, on miss, populate L2 asynchronously from Azure. The write path (staging, commit, dirty tracking, invalidation) will follow in a separate PR.

What's included

**New component: dist_cache**

- dist_cache.go— component wiring (Configure / Start / Stop), ReadInBuffer implementation, memory-bounded async populate.
- dist_cache_test.go — unit tests (interface-mocked client).
- dist_cache_integration_test.go — end-to-end tests against a real dcache client with a mock server.

**Read path (ReadInBuffer)**

- L2 hit → return.
- L2 zero-byte hit → treat as miss, read from Azure, populate.
- ErrNotFoundGotLock → we own the miss; read from Azure and populate L2.
- ErrNotFoundAlreadyLocked → poll L2 with exponential backoff (200 ms → 2 s cap, 10 s deadline); on timeout fall through to Azure.
- ErrNotFound (no lock semantics) → read from Azure but do not populate (avoids thundering-herd writes on cold chunks).
- Any other error → honors bypass-on-error (fall through to next component) or returns the error.

**Bounded async upload pool**
Populates are fire-and-forget goroutines using a preallocated, bounded buffer pool — schedulePopulate never blocks, and drops the populate if the pool is exhausted so read latency never depends on dcache cluster health. Stop() waits for all in-flight uploads via sync.WaitGroup before closing the client.

## Checklist

## Related Links
- [Issues](<link>)


common/util.go
```diff
@@ -572,6 +577,22 @@ func ValidatePipeline(pipeline []string) error {
...
+		return fmt.Errorf("mount: dist-cache and xload cannot be used together")
+	}
+
+	if ComponentInPipeline(pipeline, "dist_cache") &&
+		ComponentInPipeline(pipeline, "file_cache") {
+		return fmt.Errorf("mount: file-cache and dist-cache cannot be used together (block_cache is required as L1)")
+	}
+
+	if ComponentInPipeline(pipeline, "dist_cache") &&
+		!ComponentInPipeline(pipeline, "block_cache") {
```
syeleti-msft: are we expecting block cache component to be present in the config file for dist_cache to work?

component/block_cache/block_cache.go
```diff
@@ -107,14 +107,20 @@ type BlockCacheOptions struct {
...
 	MIN_WRITE_BLOCK         = 3
 	MIN_RANDREAD            = 10
 	MAX_FAIL_CNT            = 3
 	MAX_BLOCKS              = 50000
+
+	// MemShareFraction is the fraction of free RAM block_cache reserves for
+	// its block pool when block_cache.mem-size-mb is not explicitly set.
+	// Exported so other components (e.g. dist_cache) can compute their share
+	// relative to block_cache's actual footprint rather than raw free RAM.
+	MemShareFraction = 0.8
```
syeleti-msft: rebase your branch to the main, this has been made to 50% of available memory

common/util.go
```diff
@@ -659,3 +680,25 @@ func TotalMemoryBytes() uint64 {
...
 }
+
+// FreeMemoryBytes returns the currently free RAM of the system in bytes.
+// Returns 0 if the syscall fails.
+func FreeMemoryBytes() uint64 {
+	var info syscall.Sysinfo_t
+	if err := syscall.Sysinfo(&info); err != nil {
+		return 0
+	}
+	return uint64(info.Freeram) * uint64(info.Unit)
```
syeleti-msft: info.Freeram doesn't give the ram that was used by the page cache which can be reclaimed by the user processes. So the better estimate could be to use available memory which also includes the memory used for buffers in the page cache. 
https://github.com/Azure/azure-storage-fuse/blob/d302f13072d7af5d6777ebe5e06f46d66bb5b316/common/util.go#L666
rebase your branch to main to get the recent upstream changes

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,599 @@
...
+	"runtime"
+	"strings"
+	"sync"
+	"syscall"
+	"time"
+
+	"github.com/Azure/azure-storage-fuse/v2/common"
+	"github.com/Azure/azure-storage-fuse/v2/common/config"
+	"github.com/Azure/azure-storage-fuse/v2/common/log"
+	"github.com/Azure/azure-storage-fuse/v2/component/block_cache"
```
syeleti-msft: general rule we follow, components must not import each other, pls define the common variables in the common package and use it from their

component/block_cache/block_cache.go
```diff
@@ -107,14 +107,20 @@ type BlockCacheOptions struct {
 const (
 	compName                = "block_cache"
 	defaultTimeout          = 120
-	defaultBlockSize        = 16
+	DefaultBlockSize        = 16
```
syeleti-msft: define this in the common/types.go

setup/baseConfig.yaml
```diff
@@ -41,6 +42,27 @@ file_cache:
   # Optional 
   timeout-sec: <default cache eviction timeout (in sec). Default - 120 sec>
   max-size-mb: <maximum cache size allowed. Default - 80% of free disk space>
+
+# Distributed cache configuration (optional, for multi-node shared caching)
+# Requires a running Tachyon cache server cluster (https://github.com/Azure/Tachyon).
+# Requires block_cache in the pipeline. Mutually exclusive with xload and file_cache.
+dist_cache:
```
syeleti-msft, line 49: keep the minimum config params that were necessary to config in base config and rest of the config params in setup/advancedConfig.yaml

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,599 @@
...
+	// distCacheMinBuffers is the smallest buffer count the pool will use when
+	// enabled. schedulePopulate never blocks — it drops the populate if no
+	// buffer is free — so a too-small pool doesn't hurt read latency, but it
+	// does collapse the L2 hit rate: any burst of concurrent misses, or a
+	// single stuck upload holding a buffer for up to the dcache client's
+	// per-request timeout, leaves nothing for the next populate and the
+	// feature stops paying for itself. Four buffers absorb a small burst and
+	// tolerate one stuck upload while still being tiny in absolute terms
+	// (e.g. 4 × 16 MiB = 64 MiB at the default block size).
+	distCacheMinBuffers = 4
```
syeleti-msft, line 54: isn't this too low for feature to work properly? 

in block cache the read_ahead of the blocks are scheduled using prefetch count. if we do agressive prefetch/ with default prefetch being 11 blocks this can be bottleneck.

What is the cache hit rate you are seeing when it is set to 4?

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,599 @@
...
+		"distributed cache discovery endpoint (recommended)")
+	config.BindPFlag(compName+".discovery-url", discoveryFlag)
+
+	serverListFlag := config.AddStringFlag("dist-cache-server-list", "",
+		"comma-separated list of distributed cache server addresses (fallback)")
+	config.BindPFlag(compName+".server-list", serverListFlag)
+
+	// Support DIST_CACHE_SERVER_LIST env var
+	config.BindEnv(compName+".server-list", "DIST_CACHE_SERVER_LIST")
+}
```
syeleti-msft: nit: keep this functions to the end of the file

cmd/mount.go
```diff
@@ -298,6 +298,11 @@ var mountCmd = &cobra.Command{
 				pipeline = append(pipeline, "file_cache")
 			}
 
+			// dist_cache sits between local cache (L1) and azstorage (L3)
+			if config.IsSet("dist_cache") {
+				pipeline = append(pipeline, "dist_cache")
+			}
```
souravgupta-msft: Since the user will only provide the flags for distributed cache as per the ux doc. So, here we will have to configure the pipeline accordingly something like below. This should also move in the if-elseif clause above.

```suggestion
          // dist_cache sits between block cache (L1) and azstorage (L3)
          if config.IsSet("dist_cache") {
              pipeline = append(pipeline, "block_cache") // L1 cache
          	  pipeline = append(pipeline, "dist_cache") // L2 cache
          }
```
```

File: sampleDistCacheConfig.yaml [hunk 10]
```diff
@@ -0,0 +1,65 @@
...
+# See https://github.com/Azure/Tachyon for installation instructions.
+
+logging:
+  type: syslog
+  level: log_warning
+
+components:
+  - libfuse
+  - block_cache
+  - dist_cache
```
souravgupta-msft: As per the UX doc, we should only have dist_cache here, right?

setup/baseConfig.yaml
```diff
@@ -41,6 +42,27 @@ file_cache:
...
+  # OR
+  server-list: <comma-separated list of cache server addresses, e.g., "host1:9065,host2:9065,host3:9065">
+  # Common options
+  port: <cache server port. Default - 9065>
+  bypass-on-error: <if true, cache errors fall through to Azure Blob Storage. Default - false>
+  ttl-seconds: <TTL for cached data in seconds. Default - 0 (no expiry)>
+  chunk-size-mb: <chunk size in MB, aligned with block_cache block-size-mb. Default - 16>
+  cache-prefix: <prefix for cache keys. Default - derived from azstorage account/container>
+  max-conns-per-server: <maximum TCP connections per cache server. Default - 8>
+  request-timeout-sec: <timeout for individual cache requests in seconds. Default - 30>
```
souravgupta-msft: Add please add these config parameters also in the advanced config
```
  block-size-mb: 16 
  memory-mb: 4096 
  prefetch: 32 
  parallelism: 128
```

component/block_cache/block_cache.go
souravgupta-msft: Its 60%
https://github.com/Azure/azure-storage-fuse/blob/main/component/block_cache/block_cache.go#L358-L366

common/util.go
nearora-msft: No @syeleti-msft, but for this PR purpose the configuration works like block_cache as L1 and dist_cache as L2. I will remove this validation for now since this is confusing. 
When the UX PR comes in and we only allow dist_cache related config, we can add a validation saying that dist_cache doesn't work with block_cache.

cmd/mount.go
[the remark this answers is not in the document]
nearora-msft: That's ok, the CLI flags should only work with the --distributed-cache CLI param. These changes will be made in the UX PR.

sampleDistCacheConfig.yaml
nearora-msft: Yes @souravgupta-msft, but those changes aren't in this PR yet, so keeping this for correctness of this commit. It will be modified in the UX related PR.

component/dist_cache/dist_cache.go
nearora-msft: Honestly, we will only end up using this value when the memory available is <1GiB - which means this path isn't going to be used for the targeted model loading cases.

Let's say the memory available is 1GiB. block+cache + dist_cache get ~600MiB and dist_cache gets ~60MiB.

If we increase this value, and the memory is <1GiB , block_cache will keep getting smaller portion of the memory and hence that will become the bottleneck.

This is a sane default to ensure that when the available memory is too low, block_cache gets enough room, while dist_cache also works at a sensible pace. This will not become a bottleneck because if the memory is not available, we are just going to drop populating the cache. The reads won't get blcoked on this.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,599 @@
...
+	// A nil pool means async populate is disabled (caller memory budget is
+	// smaller than a single chunk); reads still work as pure passthrough.
+	budget := dc.resolveMemBudget()
+	if budget >= dc.chunkSize {
+		numBuffers := int(budget / dc.chunkSize)
+		dc.bufs = make(chan *[]byte, numBuffers)
+		for i := 0; i < numBuffers; i++ {
+			b := make([]byte, dc.chunkSize)
+			dc.bufs <- &b
+		}
```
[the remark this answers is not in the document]
nearora-msft: This compiles.

setup/baseConfig.yaml
nearora-msft: @souravgupta-msft, these will come in the UX PR. For this PR, we are still assuming block_cache will be configured as L1. Will update this in the UX PR.

cmd/mount.go
nearora-msft: Done!

component/block_cache/block_cache.go
nearora-msft: Done!

component/dist_cache/dist_cache.go
nearora-msft: Done!

setup/baseConfig.yaml
nearora-msft: Done!

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+	if config.IsSet("block_cache.block-size-mb") {
+		err = config.UnmarshalKey("block_cache.block-size-mb", &blockSizeMB)
+		if err != nil {
+			log.Warn("DistCache::Configure : Failed to read block-size-mb, using default %d MB", common.DefaultBlockSize)
+			blockSizeMB = common.DefaultBlockSize
+		}
+	} else if config.IsSet("stream.block-size-mb") {
+		err = config.UnmarshalKey("stream.block-size-mb", &blockSizeMB)
+		if err != nil {
+			blockSizeMB = common.DefaultBlockSize
```
croomes, line 211: nit: do you want to log a warning when falling back to the default here too?

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+		maxPollDuration = 10 * time.Second
+		maxBackoff      = 2 * time.Second
+	)
+
+	deadline := time.Now().Add(maxPollDuration)
+	backoff := 200 * time.Millisecond
+
+	for time.Now().Before(deadline) {
+		select {
+		case <-ctx.Done():
```
croomes, line 423: Harmless, but nothing ever cancels the ctx created on L325 so this will never fire.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+//  2. Floor at distCacheMinBuffers × chunkSize so async populate is functional
+//     when enabled. Then re-cap at the demand ceiling so the floor never
+//     inflates a naturally-tiny budget.
+func (dc *DistCache) resolveMemBudget() int64 {
+	if config.IsSet("block_cache.mem-size-mb") {
+		var n uint64
+		if err := config.UnmarshalKey("block_cache.mem-size-mb", &n); err == nil && n > 0 {
+			return int64(n) * _1MiB
+		}
+	}
```
croomes, line 547: An explicit mem-size-mb=8192 gives dist_cache up to 8 GiB more on top of block_cache's own 8 GiB additive, not shared, which seems to contradict the doc comment above it ("we reuse that value verbatim so the two caches stay coordinated under one memory budget").

Was this intentional (e.g., explicit sizing means the user is opting into a larger footprint deliberately), or should this branch apply the same  distCacheSharePct  / demand-ceiling treatment as the auto-computed path?

Also, I didn't see unit tests for `resolveMemBudget()`?

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+	log.Trace("Stopping component : %s", dc.Name())
+	// Wait for any in-flight populates to finish before closing the client so
+	// they don't observe a nil/closed client mid-upload.
+	dc.inflight.Wait()
+	if dc.client != nil {
+		return dc.client.Close()
+	}
+	return nil
+}
+
```
croomes, line 299: `Stop()` isn't called in any of the tests, so the wait ordering is never tested. Since Wait() doesn't have a timeout, wondering if this could block shutdown?

cmd/mount.go
```diff
@@ -298,6 +298,12 @@ var mountCmd = &cobra.Command{
 				pipeline = append(pipeline, "file_cache")
 			}
 
+			// dist_cache sits between local cache (L1) and azstorage (L3)
+			if config.IsSet("dist_cache") {
+				pipeline = append(pipeline, "block_cache") // L1 cache
+				pipeline = append(pipeline, "dist_cache") // L2 cache
+			}
```
croomes, line 308: Do you need to check if block_cache is already enabled, so it doesn't get appended twice? There doesn't seem to be unit tests that would catch this.

common/util.go
```diff
@@ -573,6 +573,15 @@ func ValidatePipeline(pipeline []string) error {
...
 
+	if ComponentInPipeline(pipeline, "dist_cache") &&
+		ComponentInPipeline(pipeline, "xload") {
+		return fmt.Errorf("mount: dist-cache and xload cannot be used together")
+	}
+
+	if ComponentInPipeline(pipeline, "dist_cache") &&
+		ComponentInPipeline(pipeline, "file_cache") {
+		return fmt.Errorf("mount: file-cache and dist-cache cannot be used together")
+	}
```
croomes, line 584: Should this also check that if dist_cache is enabled, so is block_cache? Arguably it could also check for duplicates, but probably beyond PR scope & not sure it even matters.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+	}
+	if dc.conf.TTLSeconds > 0 {
+		opts = append(opts, dcache.WithTTL(dc.conf.TTLSeconds))
+	}
+
+	data := (*buf)[:length]
+	if err := dc.client.UploadChunk(context.Background(), name, etag, offset, data, opts...); err != nil {
+		log.Warn("DistCache::doUpload : upload failed: %v", err)
+	}
+}
```
croomes: `doUpload()` is only being tested indirectly via `TestSchedulePopulate_BufferReturnedAfterUpload`, and only the happy path.

It wouldn't hurt to validate the error cases too, along with opts passing. I don't see how it could fail but it might help protect against regressions.

component/dist_cache/dist_cache_test.go
```diff
@@ -0,0 +1,556 @@
...
+	}
+	require.Eventually(t, func() bool { return started.Load() == 4 },
+		time.Second, 5*time.Millisecond, "expected all 4 initial uploads to start")
+
+	// Schedule many more populates while the pool is empty. Each must return
+	// immediately without blocking and without incrementing started.
+	const extras = 50
+	for i := 0; i < extras; i++ {
+		dc.schedulePopulate("f", "e", int64(1000+i), []byte("data"))
+	}
```
croomes, line 832: Just a suggestion, but would scheduling the work in concurrent goroutines instead of serially be more representative of fuse goroutines, where they all race on dc.bufs? I wouldn't expect it to find anything different, but you'd have a better chance of detecting problems, especially if `-race` was enabled (it isn't).

croomes (commented):
Looks great! Good integration test coverage too. Just a few minor comments.

cmd/mount.go
souravgupta-msft: This should go under the above if-elseif checks before the file cache else block

setup/advancedConfig.yaml
```diff
@@ -115,7 +115,16 @@ attr_cache:
...
-  
+
+# Distributed cache advanced configuration
+# See setup/baseConfig.yaml for the base dist_cache options (e.g. discovery-url).
+dist_cache:
+  k8s-service: <Kubernetes headless service name for DNS-based discovery>
+  k8s-namespace: <Kubernetes namespace where the cache service runs>
+  # OR
+  server-list: <comma-separated list of cache server addresses, e.g., "host1:9065,host2:9065,host3:9065">
+  ttl-seconds: <TTL for cached data in seconds. Default - 0 (no expiry)>
```
souravgupta-msft, line 127: In the advanced config file, we can mention all the config parameters that are exposed

component/dist_cache/dist_cache.go
nearora-msft: Thanks for pointing this out, this was unintentional and it should follow the same fainess/demand model. Will update this and will also add the unit tests.

component/dist_cache/dist_cache.go
nearora-msft: Wait() doesn't have a timeout but the upload function is bound by client side timeout of 30s. So, Stop() will hang for 30s in the worst case. Lmk if you think we should still have a separate timeout for this.
I'll add a test that ensures this path is exercised.

cmd/mount.go
nearora-msft: The actual UX implementation would enusure through validations that dist_cache and block_cache are mutually exclusive and cannot be configured at the same time. But, I plan to do that in the UX PR.
For current PR, I'm assuming that the yaml configures both block_cache and dist_cache. I should actually revert this change where I add block_cache in the pipeline when dist_cache is present for this PR.

common/util.go
nearora-msft: It should Simon, but again, I would make this change in the UX PR.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+	discoveryFlag := config.AddStringFlag("dist-cache-discovery-url", "",
+		"distributed cache discovery endpoint (recommended)")
+	config.BindPFlag(compName+".discovery-url", discoveryFlag)
+
+	serverListFlag := config.AddStringFlag("dist-cache-server-list", "",
+		"comma-separated list of distributed cache server addresses (fallback)")
+	config.BindPFlag(compName+".server-list", serverListFlag)
+
+	// Support DIST_CACHE_SERVER_LIST env var
+	config.BindEnv(compName+".server-list", "DIST_CACHE_SERVER_LIST")
```
souravgupta-msft: Can you create a constant for this "DIST_CACHE_SERVER_LIST"
Also do we need environment variable for discovery url as well?

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+	if err != nil {
+		log.Err("DistCache: config error [invalid config attributes]")
+		return fmt.Errorf("dist_cache: config error: %w", err)
+	}
+
+	// Validate that at least one server discovery method is configured
+	if conf.DiscoveryURL == "" && conf.K8sService == "" && conf.ServerList == "" {
+		if os.Getenv("DIST_CACHE_SERVER_LIST") == "" {
+			return fmt.Errorf("dist_cache: no server discovery configured (set discovery-url, k8s-service, server-list, or DIST_CACHE_SERVER_LIST)")
+		}
```
souravgupta-msft: This might not be needed as we are already binding the `DIST_CACHE_SERVER_LIST` environment variable in the init() method to the respective config parameter. Please verify this once.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+	// applies them in precedence order: discovery-url > k8s DNS > server-list;
+	// lower-precedence entries are effectively ignored.
+	var configured []string
+	if conf.DiscoveryURL != "" {
+		configured = append(configured, "discovery-url")
+	}
+	if conf.K8sService != "" {
+		configured = append(configured, "k8s-service")
+	}
+	if conf.ServerList != "" || os.Getenv("DIST_CACHE_SERVER_LIST") != "" {
```
souravgupta-msft: Binding of the environment variable to `dist_cache.server-list` is already done in the init() method
```suggestion
	if conf.ServerList != "" {
```

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+	// Common options
+	Port           int    `config:"port"              yaml:"port,omitempty"`
+	TTLSeconds     uint32 `config:"ttl-seconds"       yaml:"ttl-seconds,omitempty"`
+	BypassOnError  bool   `config:"bypass-on-error"   yaml:"bypass-on-error,omitempty"`
+	CachePrefix    string `config:"cache-prefix"      yaml:"cache-prefix,omitempty"`
+	MaxConnsPerSvr int    `config:"max-conns-per-server" yaml:"max-conns-per-server,omitempty"`
+
+	// Chunk size for distributed cache operations. When block_cache is present,
+	// this is overridden by block_cache.block-size-mb to keep alignment consistent.
+	ChunkSizeMB float64 `config:"chunk-size-mb" yaml:"chunk-size-mb,omitempty"`
```
souravgupta-msft: Can we add a comment mentioning the default values of these common options.

souravgupta-msft: Also, I think few of these options will be removed in the UX PR as we will go with the default values for these, right?

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+			blockSizeMB = common.DefaultBlockSize
+		}
+	} else if config.IsSet("stream.block-size-mb") {
+		err = config.UnmarshalKey("stream.block-size-mb", &blockSizeMB)
+		if err != nil {
+			blockSizeMB = common.DefaultBlockSize
+		}
+	} else if conf.ChunkSizeMB > 0 {
+		blockSizeMB = conf.ChunkSizeMB
+	}
```
souravgupta-msft, line 213: I believe this section will be updated in the UX PR as we will be exposing the block size parameter in dist_cache options.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+	dc.chunkSize = int64(blockSizeMB * 1024 * 1024)
```
souravgupta-msft: nit
```suggestion
	dc.chunkSize = int64(blockSizeMB * _1MiB)
```

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+		dc.bufs = make(chan *[]byte, numBuffers)
+		for i := 0; i < numBuffers; i++ {
+			b := make([]byte, dc.chunkSize)
+			dc.bufs <- &b
+		}
+		log.Info("DistCache::Configure : async upload pool = %d buffers × %d bytes = %d MiB",
+			numBuffers, dc.chunkSize, int64(numBuffers)*dc.chunkSize/_1MiB)
+	} else {
+		log.Warn("DistCache::Configure : async upload disabled (budget %d < chunk size %d); L2 populate skipped",
+			budget, dc.chunkSize)
```
souravgupta-msft: When can we hit this case?
If this is a no-op, please remove it

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+	}
+
+	log.Info("DistCache::Configure : chunk-size=%d, bypass-on-error=%v",
+		dc.chunkSize, dc.bypassOnError)
```
souravgupta-msft: Let's log all the configured options here

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+		opts = append(opts, dcache.WithDiscoveryRefresh(
+			time.Duration(dc.conf.DiscoveryRefreshSec)*time.Second))
+	}
+
+	client, err := dcache.New(opts...)
+	if err != nil {
+		if dc.bypassOnError {
+			log.Warn("DistCache::Start : Failed to connect to distributed cache, bypassing: %v", err)
+			return nil
+		}
```
souravgupta-msft: Why would we need this? If we are unable to create the client, then we should ideally fail the mount.
Enabling dist_cache in the pipeline and then bypassing it sets wrong expectation.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+	return options.Path
+}
+
+func (dc *DistCache) ReadInBuffer(options *internal.ReadInBufferOptions) (int, error) {
+	if dc.client == nil {
+		if !dc.bypassOnError {
+			dc.nilClientLogOnce.Do(func() {
+				log.Err("DistCache::ReadInBuffer : client not initialized and bypass-on-error=false; failing reads")
+			})
+			return 0, syscall.EIO
```
souravgupta-msft: Wouldn't it be better to log this error every time instead of just logging it once in the beginning?

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+		}
+
+		n, err := dc.client.DownloadChunk(ctx, name, etag, offset, buf)
+		if err == nil {
+			return n, nil
+		}
+
+		if err != dcache.ErrNotFoundAlreadyLocked && err != dcache.ErrNotFound {
+			return 0, err
+		}
```
souravgupta-msft: Can there be a case of `ErrNotFoundGotLock` in this?
Suppose the previous node which initially got the lock failed to upload the chunk to dist_cache. Now this node gets the lock after calling the DownloadChunk() method.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+		select {
+		case <-ctx.Done():
+			return 0, ctx.Err()
+		case <-time.After(backoff):
+		}
+
+		n, err := dc.client.DownloadChunk(ctx, name, etag, offset, buf)
+		if err == nil {
+			return n, nil
+		}
```
souravgupta-msft: We should also handle the case when `n == 0` because of previous node which got the lock was unable to upload the chunk in dist_cache

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+	return 0, err
+}
+
+// --- Internal helpers ---
+
+// pollChunkIntoBuffer waits for a single chunk to become available in the
+// distributed cache and copies it into buf. Returns the number of bytes read.
+func (dc *DistCache) pollChunkIntoBuffer(ctx context.Context, name, etag string, offset int64, buf []byte) (int, error) {
+	const (
+		maxPollDuration = 10 * time.Second
```
souravgupta-msft: Isn't 10 seconds too large for waiting for single chunk download and then upload to dist_cache from some other node?

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+	}
+
+	if err == dcache.ErrNotFound {
+		// L2 miss without lock semantics — read from Azure but do NOT populate.
+		// Populating here would cause a thundering herd: every reader that
+		// races on a cold chunk would also write. Cache population is reserved
+		// for the client that explicitly holds the miss-lock (ErrNotFoundGotLock).
+		log.Debug("DistCache::ReadInBuffer : L2 miss %s offset=%d", name, options.Offset)
+		return dc.NextComponent().ReadInBuffer(options)
+	}
```
souravgupta-msft, line 381: Curious when will the cache server return this error?
Ideally either this node or some other node will take the responsibility of uploading the chunk to dist_cache by acquiring the lock. So, it should get either the got_lock or already_locked error.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+
+// doUpload runs a single populate. It always returns the borrowed buffer to
+// the pool, even on error. The per-request timeout is enforced by the dcache
+// client (see WithRequestTimeout; default 30s), which applies a socket-level
+// deadline covering both send and receive.
+func (dc *DistCache) doUpload(name, etag string, offset int64, buf *[]byte, length int) {
+	defer dc.inflight.Done()
+	defer func() { dc.bufs <- buf }()
+
+	gid := fileGroupID(name, etag)
```
souravgupta-msft, line 620: Are we ensuring anywhere that the etag is not empty?
Also, what happens if etag is empty

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,597 @@
...
+	default:
+		log.Debug("DistCache::schedulePopulate : buffer pool exhausted, dropping populate for %s offset=%d", name, offset)
+		return
+	}
+
+	*buf = (*buf)[:len(src)]
+	copy(*buf, src)
+
+	dc.inflight.Add(1)
+	go dc.doUpload(name, etag, offset, buf, len(src))
```
souravgupta-msft, line 609: Are we controlling anywhere how many threads are spawned

cmd/mount.go
nearora-msft: I've made some UX related improvements and also added unit tests for that.
They ensure that block_cache isn't getting appended twice.

common/util.go
nearora-msft: I've made UX related improvements. PTAL.

setup/baseConfig.yaml
nearora-msft: Updated in this PR.

component/dist_cache/dist_cache.go
nearora-msft: I've removed the stream.block-size-mb branch altogether.

component/dist_cache/dist_cache.go
nearora-msft: Updated to use a context that cancels on timeout.

common/util.go
nearora-msft: I've added these validations in the latest UX related improvements.

component/dist_cache/dist_cache.go
nearora-msft: Added failure tests for doUpload.

component/dist_cache/dist_cache_test.go
nearora-msft: Scheduling work concurrently now.

setup/advancedConfig.yaml
nearora-msft: Done.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,593 @@
...
+	name := resolveReadPath(options)
+
+	// Resolve ETag from handle (pinned at open time by block_cache)
+	etag := resolveETag(options)
+	log.Debug("DistCache::ReadInBuffer : %s offset=%d etag=%q", name, options.Offset, etag)
+
+	ctx := context.Background()
+
+	n, err := dc.client.DownloadChunk(ctx, name, etag, options.Offset, options.Data,
+		dcache.WithLock(true))
```
souravgupta-msft, line 325: How are we ensuring data integrity in case of `DownloadChunk` and `UploadChunk`?
- What will happen when the chunk uploaded to the cache server by some node gets corrupted during transfer or because of some other reason. Is the cache server in this case taking care of invalidating/deleting that inconsistent chunk so that a read request for that chunk will issue a fresh download for that chunk.
- In the below happy path, we are just checking `if err == nil && n > 0`. How are we accounting for cases when partial or inconsistent chunk is returned.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,593 @@
...
+		dcache.WithMetadata(map[string][]byte{"gid": gid}),
+	}
+	if dc.conf.TTLSeconds > 0 {
+		opts = append(opts, dcache.WithTTL(dc.conf.TTLSeconds))
+	}
+
+	data := (*buf)[:length]
+	if err := dc.client.UploadChunk(context.Background(), name, etag, offset, data, opts...); err != nil {
+		log.Warn("DistCache::doUpload : upload failed: %v", err)
+	}
```
souravgupta-msft, line 634: Should we use some sort of optional validation (like using md5sum) to ensure integrity in case of chunks upload?

cmd/mount.go
```diff
@@ -294,10 +301,17 @@ var mountCmd = &cobra.Command{
 				pipeline = append(pipeline, "block_cache")
 			} else if options.Preload {
 				pipeline = append(pipeline, "xload")
-			} else {
+			} else if !config.IsSet("dist_cache") {
+				// dist_cache brings its own L1 (block_cache), so skip
+				// the file_cache default.
 				pipeline = append(pipeline, "file_cache")
 			}
```
souravgupta-msft: For cases when Blobfuse is mounted using CLI params like AKS, we use file cache as the default caching mode unless explicitly specified in the CLI flags. So, this check will break this use case for customers who are not passing any flag for caching and internally expect file cache to be used.

We can rather have an else-if clause above this to check for dist_cache that we are doing below.

cmd/mount.go
```diff
@@ -784,6 +802,72 @@ func cleanupCachePath(componentName string, globalCleanupFlag bool) error {
...
+	if !userWantsDistCache {
+		return nil
+	}
+
+	// Reject any sibling L1 cache signal, whether in components: or as a
+	// top-level section.
+	incompatible := []string{"block_cache", "file_cache", "xload", "stream"}
+	for _, name := range incompatible {
+		if common.ComponentInPipeline(userComponents, name) || config.IsSet(name) {
+			return fmt.Errorf("mount: dist_cache is incompatible with %s; dist_cache uses block_cache as its L1 and cannot coexist with another L1 cache. Remove %s from components: and any %s: section", name, name, name)
```
souravgupta-msft, line 822: There can be a misleading error statement here when block_cache is present in the userComponents slice.

> mount: dist_cache is incompatible with **block_cache**; dist_cache uses block_cache as its L1 and cannot coexist with another L1 cache. Remove **block_cache** from components: and any **block_cache**: section

cmd/mount.go
```diff
@@ -784,6 +802,72 @@ func cleanupCachePath(componentName string, globalCleanupFlag bool) error {
...
+// dist_cache. Idempotent: no-op if block_cache is already present or
+// dist_cache is absent.
+func injectBlockCacheForDistCache(components []string) []string {
+	distIdx := slices.Index(components, "dist_cache")
+	if distIdx < 0 {
+		return components
+	}
+	if slices.Contains(components, "block_cache") {
+		return components
+	}
```
souravgupta-msft: This will be a no-op, right?

component/dist_cache/dist_cache.go
nearora-msft: I have made the UX related changes and updated this section.

component/dist_cache/dist_cache.go
nearora-msft: ErrNotFound occurs when a client calls DownloadChunk without the lock and 
a. The chunk is not found.
b. The chunk is not found but another reader has the lock to download and populate this chunk.

We are calling DownloadChunk with dcache.WithLock(true), so ideally we shouldn't receive this error. But in case there's a server-side issue where it unexpectedly sends ErrNotFound for some reason, we should download from azure. We should still avoid populating the cache since we don't hold the lock.

component/dist_cache/dist_cache.go
nearora-msft: It is not bounded by the number of threads, but by the amount of memory that we can allocate. 
If there are no buffers available uploads will be dropped.

component/dist_cache/dist_cache.go
nearora-msft: The etag is set the the block_cache on the handle on file open.
What could be the reason that it is nil? If we know that, we can handle the missing etag cases better.
https://github.com/Azure/azure-storage-fuse/blob/db7ab778bc7d28e3fdddd3c6b397e5fbc63ed8e4/component/block_cache/block_cache.go#L403

component/dist_cache/dist_cache.go
nearora-msft: That's a great point @souravgupta-msft, this is something that needs to be enabled on the go-client side. I can create a task to do that. Once that support comes in, I can update the go-client version in this PR or the next PR if this merges by then.

component/dist_cache/dist_cache.go
nearora-msft: I agree, the changes need to be made on the go client side. It would calculate a checksum and upload that as metadata in the UploadChunk call. On download, it would verify the checksum.

cmd/mount.go
nearora-msft: This isn't misleading. We don't want the users to set block_cache with dist_cache explicitly. That is something that the code should do implicitly.

cmd/mount.go
nearora-msft: Fixed.

cmd/mount.go
nearora-msft: Changed.

component/dist_cache/dist_cache.go
nearora-msft: Changed to 3s.

component/dist_cache/dist_cache.go
nearora-msft: Added a branch to handle this case.

component/dist_cache/dist_cache.go
nearora-msft: Added handling for ErrNotFoundGotLock.

component/dist_cache/dist_cache.go
nearora-msft: Yes, logging it every time.

component/dist_cache/dist_cache.go
nearora-msft: Agreed. Failing the mount when client is nil.

component/dist_cache/dist_cache.go
nearora-msft: Updated the memory allocation logic a little bit. With the new changes, if we don't even have the budget to allocate 4 chunks of memory, we would bypass the upload entirely.

component/dist_cache/dist_cache.go
nearora-msft: Updated.

component/dist_cache/dist_cache.go
nearora-msft: Updated.

component/dist_cache/dist_cache.go
nearora-msft: Done.

component/dist_cache/dist_cache.go
nearora-msft: I'm planning to removing others, but thinking of keeping Port and TTLSeconds.
I hadn't mentioned port in the UX doc but it might be worth keeping in the advanced config given that the users might be running some other applications on the default cache-server pod and may need to modify that. 
What do you think? Would it be ok to keep in the advanced yaml?

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,633 @@
...
+// sets Handle but not Path.
+func resolveReadPath(options *internal.ReadInBufferOptions) string {
+	if options.Handle != nil {
+		return options.Handle.Path
+	}
+	return options.Path
+}
+
+func (dc *DistCache) ReadInBuffer(options *internal.ReadInBufferOptions) (int, error) {
+	if dc.client == nil {
```
syeleti-msft: when can be this true? we are failing the mount when client can't be created right https://github.com/Azure/azure-storage-fuse/pull/2297/changes#diff-e6acfd1e94e33df86a019f56ca59935f52bfb5e897b1e02661645bbdd11e186dR285

nearora-msft: That's correct. Removed this part.

component/dist_cache/dist_cache.go
nearora-msft: Done!

component/dist_cache/dist_cache.go
nearora-msft: With the latest changes, we don't populate the L2 cache if etag is empty.

component/dist_cache/dist_cache.go
nearora-msft: I've enabled checksum on the go-client and updated this code to use that : https://github.com/nearora-msft/dist-cache-client-go/pull/1

component/dist_cache/dist_cache.go
souravgupta-msft: Yes, let's keep these as part of advanced config paramaters

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,646 @@
...
+
+	chunkSize     int64
+	cachePrefix   string
+	bypassOnError bool
+
+	// Bounded async-upload buffer pool. Preallocated in Configure and never
+	// grown. cap(bufs) * chunkSize is the hard ceiling on memory dist_cache
+	// holds for in-flight L2 populates. Nil means async populate was disabled
+	// (resolveMemBudget returned 0); ReadInBuffer then runs as passthrough.
+	bufs chan *[]byte
```
syeleti-msft: @copilot , any advantage in storing it as type *[]byte instead of []byte?

syeleti-msft: @copilot, pls remove the latest commit, who told you to add the changes to the PR

syeleti-msft: @nearora-msft , sry for that, pls take a look at the original comment

nearora-msft: Removed the pointer.

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,645 @@
...
+
+// --- Internal helpers ---
+
+// pollChunkIntoBuffer waits for a chunk to land in the cache. Polls with
+// WithLock(true) so we can inherit the miss-lock if the peer's expires.
+// Returns (n, nil) on L2 hit, (0, ErrNotFoundGotLock) if we inherited the
+// lock (caller must fetch and populate), (0, timeout err) if the peer still
+// holds the lock at deadline (caller should fetch without populating), or
+// (0, err) on other client errors.
+func (dc *DistCache) pollChunkIntoBuffer(ctx context.Context, name, etag string, offset int64, buf []byte) (int, error) {
```
syeleti-msft, line 410: curious to know if there is any real-world workflow that would exercise this path?  

I feel 2 nodes/file descriptors reading the same block in the same file at the same time is something rare. anyways did we exercise this path in manual testing somehow with multiple readers reading the same cold file in parallel.

nearora-msft: It can happen if there's a scale out event. Multiple clients may try to read the same chunk at the same time.
It is covered in my e2e test PR : https://github.com/Azure/azure-storage-fuse/pull/2308

component/dist_cache/dist_cache.go
```diff
@@ -0,0 +1,645 @@
...
+	if err == dcache.ErrNotFound {
+		// L2 miss without lock semantics — read from Azure but do NOT populate.
+		// Populating here would cause a thundering herd: every reader that
+		// races on a cold chunk would also write. Cache population is reserved
+		// for the client that explicitly holds the miss-lock (ErrNotFoundGotLock).
+		log.Debug("DistCache::ReadInBuffer : L2 miss %s offset=%d", name, options.Offset)
+		return dc.NextComponent().ReadInBuffer(options)
+	}
+
+	if errors.Is(err, dcache.ErrChecksumMismatch) {
```
souravgupta-msft, line 383: How is the cache server returning this error? Are we passing the checksum to cache server from Blobfuse while uploading the chunks?

souravgupta-msft (approved):

syeleti-msft (approved):

croomes (approved):
Very nice - thank you @nearora-msft!

component/dist_cache/dist_cache.go
nearora-msft: This error isn't coming from the cache server. The go-client is checking the checksum on download and will return an error on mismatch. See here: https://github.com/nearora-msft/dist-cache-client-go/pull/1/changes

PR #2316, opened 2026-08-14
UX for blobfuse-tachyon integration

MahashriSaini:
UX for Blobfuse Tachyon Integration

This PR improves the user experience for integrating Blobfuse with Tachyon's distributed cache by exposing configuration options through the CLI and YAML configuration.

Changes
Added CLI support for distributed cache configuration:

--dist-cache-discovery-endpoint
--distributed-cache-block-size
--distributed-cache-node-memory
--distributed-cache-parallelism
--distributed-cache-prefetch
--distributed-cache-node-ttl

Added advanced distributed cache configuration options through YAML:

discovery-endpoint
servers
k8s-service
k8s-namespace
block-size-mb
mem-size-mb
prefetch
parallelism
ttl-seconds
port
verify-checksum
Ensured that mounts using the distributed cache are allowed only in read-only mode by adding and enforcing the read-only mount requirement through --read-only flag and -o ro flag.
Standardized the terminology and naming conventions for distributed_cache across all user-facing configuration surfaces, including CLI flags and YAML configuration.

Fixed a Bug: In distributed cache, the isSet check only detects CLI/YAML keys, causing storage account and container values provided through environment variables to be ignored which was causing mount to fail via env parameters.
Solution: Remove the isSet check in distributed cache and allow Viper unmarshalling to resolve configuration from CLI, YAML, and environment variables.

Testing

Added/updated/verified unit tests, integrated tests and test suites.
Performed manual testing using a local setup with a running Tachyon cluster.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

- **Feature / Bug Fix**:

## Checklist

## Related Links
- [Issues](<link>)


cmd/gen-config.go
```diff
@@ -69,8 +71,22 @@ var generatedConfig = &cobra.Command{
...
+				return fmt.Errorf("--distributed-cache cannot be combined with --block-cache")
+			}
+			if optsGenCfg.tmpPath != "" {
+				return fmt.Errorf("--distributed-cache cannot be combined with file-cache mode; --distributed-cache does not use --tmp-path")
+			}
+			if optsGenCfg.distributedCacheDiscoveryEndpoint == "" {
+				return fmt.Errorf("--distributed-cache requires --distributed-cache-discovery-endpoint to be set")
+			}
+			// Distributed cache is read-only only.
+			optsGenCfg.readOnly = true
```
nearora-msft: Instead of automatically setting this to readOnly, we should instead expect that from the customer and fail if ro isn't set to true.

cmd/gen-config.go
```diff
@@ -69,8 +71,22 @@ var generatedConfig = &cobra.Command{
 			return cmd.Help()
 		}
 
-		// Check if configTmp is not provided when component is fc
-		if (!optsGenCfg.blockCache) && optsGenCfg.tmpPath == "" {
+		// distributed-cache is mutually exclusive with the L1-only modes and
+		// implies read-only. It does not require a tmp-path.
+		if optsGenCfg.distributedCache {
+			if optsGenCfg.blockCache {
```
nearora-msft, line 77: Should we also reject file_cache here?

cmd/mount.go
```diff
@@ -797,40 +798,49 @@ func cleanupCachePath(componentName string, globalCleanupFlag bool) error {
...
-		(len(userComponents) == 0 && config.IsSet("dist_cache"))
+	userWantsDistCache := common.ComponentInPipeline(userComponents, "distributed_cache") ||
+		(len(userComponents) == 0 && config.IsSet("distributed-cache"))
 	if !userWantsDistCache {
 		return nil
 	}
 
+	//readonly check
+	var readOnly bool
+	if err := config.UnmarshalKey("read-only", &readOnly); err != nil {
```
nearora-msft, line 857: This would throw an error if read-only is set using "-o ro" option. We should support adding read-only, both through --read-only and "-o ro" flags.

nearora-msft:
It would be great if you could add some distributed-cache CLI parameters related tests in mount_test.go.

cmd/mount.go
MahashriSaini: -o ro is already supported in cmd/mount.go in libfuseOptions [https://github.com/MahashriSaini/azure-storage-fuse2/blob/386575524e89f161931423ccd52a7ef221436489/cmd/mount.go#L365](url)

cmd/gen-config.go
MahashriSaini: In the next few lines tmpPath is checked and rejected, tmpPath is required for file_cache, it means that we are indirectly rejecting file_cache. Or should we have a explicit file_cache check?

cmd/gen-config.go
MahashriSaini: Done

cmd/mount.go
nearora-msft: That's true, but I think mounting with -o ro with distributed cache will fail. Can you run a command like below and try:

blobfuse2 mount /mnt/blob --distributed-cache -o ro --discovery-endpoint sample-endpoint

cmd/mount_test.go
```diff
@@ -771,6 +821,268 @@ components:
...
+	_, err = executeCommandC(rootCmd, "mount", mntDir, fmt.Sprintf("--config-file=%s", confFileName),
+		"--distributed-cache", "--distributed-cache-parallelism=24")
+	suite.assert.Error(err)
+
+	var parallelism uint32
+	err = config.UnmarshalKey("block_cache.parallelism", &parallelism)
+	suite.assert.NoError(err)
+	suite.assert.Equal(uint32(24), parallelism)
+}
+
```
nearora-msft, line 1211: The tests look really good. Could you also add a test for the "-o ro" case?

cmd/mount.go
MahashriSaini: Yes, you are right. support foe -o ro is added now.

cmd/mount_test.go
MahashriSaini: done

nearora-msft (approved):

cmd/mount_distributed_cache_test.go
```diff
@@ -190,64 +269,68 @@ components:
...
 block_cache:
   block-size-mb: 8
 `)
 	defer viper.Reset()
 
-	err := normalizeDistCacheConfig([]string{"libfuse", "dist_cache", "azstorage"})
+	err := normalizeDistCacheConfig([]string{"libfuse", "distributed_cache", "azstorage"})
 	assert.Error(t, err)
-	assert.Contains(t, err.Error(), "incompatible with block_cache")
+	assert.Contains(t, err.Error(), "block_cache")
```
croomes, line 345: Here, we fail if there is a block_cache config section but block_cache is not enabled on the CLI. In TestNormalizeDistCacheConfig_StaleSectionNoL1IsNoop above, distributed_cache config section is ignored when not enabled on the CLI. 

Is this difference expected?

cmd/mount_test.go
```diff
@@ -771,6 +821,315 @@ components:
...
+	defer os.Remove(confFileName)
+
+	_, err = confFile.WriteString(configDistCacheCLIBase)
+	suite.assert.NoError(err)
+	confFile.Close()
+
+	// Users should be able to satisfy the distributed_cache read-only gate
+	// with the FUSE-style `-o ro` option, not just `--read-only`.
+	op, err := executeCommandC(rootCmd, "mount", mntDir, fmt.Sprintf("--config-file=%s", confFileName), "--distributed-cache", "-o", "ro")
+	suite.assert.Error(err)
```
croomes, line 956: What error are you expecting here?

cmd/mount_distributed_cache_test.go
```diff
@@ -121,25 +121,27 @@ components:
...
 
 func TestNormalizeDistCacheConfig_DistCacheSectionWithoutComponents(t *testing.T) {
-	// dist_cache: set but components: omitted — the synthesis-path case.
+	// distributed_cache: set but components: omitted — the synthesis-path case.
 	setDistCacheYAML(t, `
-dist_cache:
+read-only: true
+distributed_cache:
   discovery-endpoint: d
 `)
```
croomes, line 148: Can you try setting `block-size-mb` here, then verify normalization runs, e.g.

```
distributed_cache:
   discovery-endpoint: d
   block-size-mb: 32
...
var blockSize uint32
assert.NoError(t, config.UnmarshalKey("block_cache.block-size-mb", &blockSize))
assert.Equal(t, uint32(32), blockSize)
```
I think this should pass, but I get a failure. 

Is there a bug in `mount.go` on L835 where it should use `config.IsSet("distributed_cache")` instead of `distributed-cache`? It's a bit confusing what the key should be, as the `config.IsSet("distributed-cache") in `mount.go` on L312 seems to work.

cmd/mount_test.go
```diff
@@ -771,6 +821,315 @@ components:
...
+	_, err = confFile.WriteString(configDistCacheCLIBase)
+	suite.assert.NoError(err)
+	confFile.Close()
+
+	op, err := executeCommandC(rootCmd, "mount", mntDir, fmt.Sprintf("--config-file=%s", confFileName), "--distributed-cache", "--read-only")
+	suite.assert.Error(err)
+	suite.assert.NotContains(op, "distributed cache is allowed only for read-only mounts")
+	suite.assert.True(options.DistributedCache)
+}
+
```
croomes, line 887: Can you try adding a test where the config has an explicit `components` pipeline containing `file_cache`? 

Currently all CLI tests use `configDistCacheCLIBase`, which omits `components`, so they do not detect that `--distributed-cache` is silently ignored when a pipeline is already configured. The test should assert either a conflict error or that the resulting pipeline contains `block_cache` and `distributed_cache`, but never `file_cache`.

```
func (suite *mountTestSuite) TestDistributedCacheCLIOverridesExplicitFileCachePipeline() {
	defer suite.cleanupTest()

	mntDir, err := os.MkdirTemp("", "mntdir")
	suite.assert.NoError(err)
	defer os.RemoveAll(mntDir)

	confFile, err := os.CreateTemp("", "conf*.yaml")
	suite.assert.NoError(err)
	confFileName := confFile.Name()
	defer os.Remove(confFileName)

	_, err = confFile.WriteString(`
read-only: true
components:
  - libfuse
  - file_cache
  - attr_cache
  - azstorage
azstorage:
  account-name: myAccountName
  account-key: myAccountKey
  mode: key
  endpoint: myEndpoint
  container: myContainer
`)
	suite.assert.NoError(err)
	confFile.Close()

	_, _ = executeCommandC(rootCmd, "mount", mntDir, fmt.Sprintf("--config-file=%s", confFileName),
		"--distributed-cache", "--distributed-cache-discovery-endpoint=example.com:9065")

	suite.assert.Contains(options.Components, "block_cache")
	suite.assert.Contains(options.Components, "distributed_cache")
	suite.assert.NotContains(options.Components, "file_cache")
}
`````

This currently fails because the CLI flag is being ignored.

**Review by croomes — COMMENTED** [review]
Thanks @MahashriSaini, looks very good overall, just a few comments.

File: cmd/mount_distributed_cache_test.go (hunk 5 above)
Comment (MahashriSaini) [reply cmd/mount_distributed_cache_test.go:345]: This is expected because, in case of distributed_cache, we want user to explicitly specify either in CLI or config yaml in components section to be able to use distributed_cache.
now when user have specified distributed_cache in (CLI/ config yaml) then neither block_cache nor file_cache section should be their because to user it (file cache/ block cache/ distributed cache) looks as exclusive single surface configuration model.
Though block_cache is used at L1 for distributed cache implementation but that is hidden from user.
So, now if user adds block_cache section then it would be failed. Even if user want to configure the block_cache parameters(used in distributed_cache), they are provided in distributed_cache section, so user can do there.
If it is not failing in case block_cache config section, then this config will affect the injected block _cache configuration for distributed_cache at L1.

In my understanding, In any case there is no need of user to specify block_cache/ file_cache section when using distributed_cache, even if user does then it is failed because, the block_cache config section would otherwise affect the injected block _cache for distributed_cache at L1.
Does this make sense to you, or is there anything I might be missing?

File: cmd/mount_test.go (hunk 6 above)
Comment (MahashriSaini) [reply cmd/mount_test.go:956]: The mount is expected to fail later because this unit test uses placeholder Azure Storage configuration. The error itself is not the behavior under test. The relevant assertion is NotContains, which verifies that the failure is not the distributed-cache read-only validation error. In other words, -o ro was successfully recognized as read-only and the command proceeded beyond that validation gate.

**Review by syeleti-msft — APPROVED** [review]

File: cmd/mount_distributed_cache_test.go (hunk 7 above)
Comment (MahashriSaini) [reply cmd/mount_distributed_cache_test.go:148]: Yes, its failing because distributed-cache flag is not set true. So distributed_cache should be specified either in cli flag or in components in config yaml. I will fix it.

About distributed_cache and distributed-cache in mount.go
It should not use config.IsSet("distributed_cache"). These keys represent different things:
1. distributed-cache is the boolean enable option bound to --distributed-cache CLI flag
2. distributed_cache is the component name and the namespace for its settings in config yaml.

Line 312 checks distributed-cache (CLI flag) because it decides whether to synthesize distributed_cache into the pipeline when no component section is given by user in config yaml.

At lines 834–835
In normalizeDistCacheConfig, the two checks cover different ways distributed cache can be enabled. ComponentInPipeline(userComponents, "distributed_cache") detects when the component is explicitly listed under components. The second condition detects when components is omitted and the pipeline will be generated from the distributed-cache enable option. Therefore, the hyphenated key is intentional

So using config.IsSet("distributed_cache") at L835 will just check the distributed_cache section in config yaml and not the flag value from CLI.

File: cmd/mount_test.go (hunk 8 above)
Comment (MahashriSaini) [reply cmd/mount_test.go:887]: the CLI flag  --distributed-cache  is silently ignored because  normalizeDistCacheConfig  gates the CLI-flag branch behind  len(userComponents) == 0 :
```
userWantsDistCache := common.ComponentInPipeline(userComponents, "distributed_cache") ||
    (len(userComponents) == 0 && config.IsSet("distributed-cache"))
```
When the config file supplies its own  components:  list, the CLI flag never triggers  userWantsDistCache , so the incompatible-L1 check is skipped and the mount proceeds with the sibling L1 (e.g.  file_cache ) and no distributed_cache — no error either.

So, I will:
1. Drop the  len(userComponents) == 0  guard so either surface (CLI flag or  components: ) is authoritative. The existing incompatible-L1 check will then surface the conflict with the clear "distributed_cache is a single configuration surface" error.
2. Add a test ( TestDistributedCacheCLIOverridesExplicitFileCachePipeline ) with  components: [libfuse, file_cache, attr_cache, azstorage]  +  --distributed-cache , asserting the mutual-exclusion error fires (or, if the CLI path wins, that the resulting pipeline contains  distributed_cache  +  block_cache  and no  file_cache )

File: cmd/mount_distributed_cache_test.go [hunk 9]
```diff
@@ -121,40 +121,125 @@ components:
...
+block_cache:
+  block-size-mb: 8
+`)
+	defer viper.Reset()
+
+	err := normalizeDistCacheConfig([]string{"libfuse", "distributed_cache", "azstorage"})
+	assert.Error(t, err)
+	assert.Contains(t, err.Error(), "read-only")
+}
+
```
croomes, line 235: I think there is still a bug where `config.IsSet("distributed-cache")` checks whether the option was provided, not whether it is true . Therefore this valid configuration is incorrectly treated as enabling distributed cache:

```yaml
distributed-cache: false
components:
  - libfuse
  - file_cache
  - azstorage
```

Can you add a test for:

```golang
func TestNormalizeDistCacheConfig_DisabledFlagDoesNotConflictWithFileCache(t *testing.T) {
	setDistCacheYAML(t, `
read-only: true
distributed-cache: false
components:
  - libfuse
  - file_cache
  - azstorage
`)
	defer viper.Reset()

	err := normalizeDistCacheConfig(
		[]string{"libfuse", "file_cache", "azstorage"},
	)

	assert.NoError(t, err)
}
```
An end-to-end CLI variant using `--distributed-cache=false` would also be useful.

croomes (commented):
Thanks, just one other potential issue

cmd/mount.go
```diff
@@ -797,40 +805,67 @@ func cleanupCachePath(componentName string, globalCleanupFlag bool) error {
...
+	// distributed_cache is "wanted" whenever either surface signals it:
+	// either it is listed in components:, or --distributed-cache /
+	// distributed-cache: is set. Both surfaces are authoritative, so the
+	// incompatible-L1 check below runs whichever one the user chose. This
+	// prevents the case where --distributed-cache is silently ignored
+	// because a components: pipeline was also configured (e.g. with
+	// file_cache), which would otherwise run the sibling L1 with no
+	// dist-cache and no error.
+	userWantsDistCache := common.ComponentInPipeline(userComponents, "distributed_cache") ||
+		config.IsSet("distributed-cache")
```
croomes: `config.IsSet("distributed-cache")` checks whether the option was supplied, not whether its value is true. As a result, `distributed-cache: false` and `--distributed-cache=false` still enter distributed-cache normalization and can incorrectly conflict with a normal `file_cache` pipeline. 

Could we base this on the unmarshalled boolean value instead? A regression test with `distributed-cache: false` plus `components: [libfuse, file_cache, azstorage]` currently fails with the mutual-exclusion error.

I must admit I find the `distributed-cache` and `distributed_cache` confusing and error prone. Would it be better to use constants instead? e.g.

```golang
const (
    distributedCacheEnableKey = "distributed-cache"
    distributedCacheComponent = "distributed_cache"
)
```
Using these in the `IsSet` , `BindPFlag`, pipeline, and component checks would make the enable-option versus component/config-namespace distinction explicit and reduce the chance of using the wrong spelling.

cmd/mount_distributed_cache_test.go
MahashriSaini: Done

cmd/mount.go
MahashriSaini: Yes, I’ve introduced a variable for unmarshalling and then used the flag value accordingly. Thank you for pointing this out and for the observation.

Regarding the suggestion to introduce constants, since block_cache and file_cache already use their respective string values when adding components to the pipeline, I feel it would be more consistent to follow the same approach for distributed_cache for consistency. 
With your suggestion, I’ll keep this PR consistent with the existing implementation and continue using the strings. We can discuss Blobfuse team regarding introducing variables for flags and components for block cache and file cache as well, and if they agree, consider taking it up as a follow-up PR across the components.

cmd/mount_distributed_cache_test.go
```diff
@@ -0,0 +1,582 @@
...
+	err := normalizeDistCacheConfig([]string{"libfuse", "distributed_cache", "azstorage"})
+	assert.NoError(t, err)
+	assert.False(t, viper.IsSet("block_cache.block-size-mb"))
+}
+
+func TestNormalizeDistCacheConfig_DistCacheSectionWithoutComponents(t *testing.T) {
+	// distributed_cache: set but components: omitted — the synthesis-path case.
+	setDistCacheYAML(t, `
+read-only: true
+distributed-cache: true
```
nearora-msft, line 144: We could add another test to cover the behaviour where distributed-cache: true but the components mention some other L1 cache like block_cache, file_cache.

cmd/mount_distributed_cache_test.go
nearora-msft: @MahashriSaini , the inconsistency is that yaml doesn't have a distributed_cache component and yet we end up passing that to the normalize function. normalizeDistCacheConfig doesn't really re-read that part from the config. We could update the yaml by adding distributed_cache as a component for consistency purposes.

cmd/mount_test.go
```diff
@@ -771,6 +821,364 @@ components:
...
+	confFile.Close()
+
+	// `-o ro=true` is the equivalent long form of `-o ro`.
+	op, err := executeCommandC(rootCmd, "mount", mntDir, fmt.Sprintf("--config-file=%s", confFileName), "--distributed-cache", "-o", "ro=true")
+	suite.assert.Error(err)
+	suite.assert.NotContains(op, "distributed cache is allowed only for read-only mounts")
+	suite.assert.True(options.DistributedCache)
+}
+
+func (suite *mountTestSuite) TestDistributedCacheCLIWithBlockCache() {
```
nearora-msft, line 984: Could you also add tests for:
1. When both distributed-cache and any other L1 cache like block_cache/file_cache are set using CLI parameters.
2. When distributed-cache is set as a CLI parameter and block_cache/file_cache are set in the config yaml
3. When distributed-cache is set as a CLI parameter but the block_cache knobs are provided. e.g. "--distributed-cache --block-cache-block-size=80"

cmd/gen-config.go
```diff
@@ -83,11 +101,20 @@ var generatedConfig = &cobra.Command{
 			config.Set("direct-io", "true")
 		}
 
+		if optsGenCfg.distributedCache {
+			config.Set("distributed_cache.discovery-endpoint", optsGenCfg.distributedCacheDiscoveryEndpoint)
+		}
```
souravgupta-msft: We can move this in the `if optsGenCfg.distributedCache` block above at the end after all validations are done, so that it's all in one place.

MahashriSaini: Done

cmd/mount_test.go
MahashriSaini: Done

cmd/mount_distributed_cache_test.go
MahashriSaini: Done

cmd/mount.go
```diff
@@ -301,8 +309,8 @@ var mountCmd = &cobra.Command{
 				pipeline = append(pipeline, "block_cache")
 			} else if options.Preload {
 				pipeline = append(pipeline, "xload")
-			} else if config.IsSet("dist_cache") {
-				pipeline = append(pipeline, "dist_cache") // L2 cache
+			} else if config.IsSet("distributed-cache") && options.DistributedCache {
+				pipeline = append(pipeline, "distributed_cache") // L2 cache
```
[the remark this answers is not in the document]
MahashriSaini: Fixed in latest commit by adding checks in normalizeDistCacheConfig in cmd/mount.go

souravgupta-msft (approved):

cmd/gen-config.go
```diff
@@ -69,8 +70,23 @@ var generatedConfig = &cobra.Command{
...
 		}
 
-		// Check if configTmp is not provided when component is fc
-		if (!optsGenCfg.blockCache) && optsGenCfg.tmpPath == "" {
+		// distributed-cache is mutually exclusive with the L1-only modes and
+		// requires read-only to be set explicitly. It does not require a tmp-path.
+		// It is enabled by providing --distributed-cache-discovery-endpoint.
+		if optsGenCfg.distributedCacheDiscoveryEndpoint != "" {
+			if optsGenCfg.blockCache {
+				return fmt.Errorf("--distributed-cache cannot be combined with --block-cache")
```
wolfgang-desalvador, line 78: Is this error handling still required?

MahashriSaini: In my knowledge, this is required as distributed cache, block cache and file cache are each mutually explicit L1 modes at same level for user, and user should be able to pass only one at a time. Any of the combinations like(distributed cache and block cache, block cache and file cache, file cache and distributed cache) must be discarded.
If that's aligned with the intended behavior, perhaps we could simplify the message to:
 "distributed-cache cannot be combined with block-cache"

croomes (approved):

cmd/mount.go
```diff
@@ -301,8 +308,8 @@ var mountCmd = &cobra.Command{
 				pipeline = append(pipeline, "block_cache")
 			} else if options.Preload {
 				pipeline = append(pipeline, "xload")
-			} else if config.IsSet("dist_cache") {
-				pipeline = append(pipeline, "dist_cache") // L2 cache
+			} else if config.IsSet("distributed_cache.discovery-endpoint") {
```
kapilupadhayay: •  config.IsSet  (common/config/config_parser.go:290) only consults viper + the pflag tree.  BindEnv  (line 201) writes to a custom  envTree  that  IsSet  never looks at. So  DISTRIBUTED_CACHE_DISCOVERY_ENDPOINT=...  with no  components:  silently builds a file_cache pipeline — no distributed cache, no error, and the read-only gate is skipped. This is the same class of bug that commit 8def381bc fixed inside  Configure()  by dropping  config.IsSet  in favour of  UnmarshalKey ; the fix wasn't carried into mount.go.
• Same gate ignores  k8s-service  /  server-list . A YAML with only  distributed_cache: {k8s-service: ...}  and no  components:  used to be caught by the old section-level  config.IsSet("dist_cache")  check. Now it silently falls through to file_cache. That's a regression against the branch base.

Fix: resolve the three discovery keys with  config.UnmarshalKey("distributed_cache", &conf)  (or check all three keys) and derive  wanted  from the values, not  IsSet .

cmd/mount.go
```diff
@@ -797,40 +804,84 @@ func cleanupCachePath(componentName string, globalCleanupFlag bool) error {
...
+	// either it is listed in components:, or a discovery-endpoint has been
+	// provided (via --distributed-cache-discovery-endpoint on the CLI or a
+	// distributed_cache.discovery-endpoint entry in the YAML). Both surfaces
+	// are authoritative, so the incompatible-L1 check below runs whichever
+	// one the user chose. This prevents the case where the CLI/YAML
+	// distributed_cache signal is silently ignored because a components:
+	// pipeline was also configured (e.g. with file_cache), which would
+	// otherwise run the sibling L1 with no dist-cache and no error.
+	userWantsDistCache := common.ComponentInPipeline(userComponents, "distributed_cache") ||
+		config.IsSet("distributed_cache.discovery-endpoint")
```
kapilupadhayay: Some comment as on line 311

cmd/mount.go
```diff
@@ -797,40 +804,84 @@ func cleanupCachePath(componentName string, globalCleanupFlag bool) error {
 	return nil
 }
 
-// normalizeDistCacheConfig enforces the single-surface UX for dist_cache:
+// applyLibfuseReadOnlyOption translates the FUSE-style `-o ro` (or `-o ro=true`)
+// libfuse option into the blobfuse2 `read-only` config key. This runs before the
+// main libfuse-options loop so early checks (e.g. the distributed_cache
+// read-only gate in normalizeDistCacheConfig) treat `-o ro` equivalently to
+// `--read-only`. It is a no-op when neither variant is present.
+func applyLibfuseReadOnlyOption(libfuseOpts []string) {
```
kapilupadhayay, line 828: -o ro  logic duplicated —  applyLibfuseReadOnlyOption   re-implements the  v == "ro" || v == "ro=true"  branch that already exists in the libfuse-options loop (mount.go:369). Two copies will drift. Hoist the shared match into one helper called by both, and consider also accepting  ro=1 / rw  handling for symmetry.

cmd/gen-config.go
```diff
@@ -69,8 +70,23 @@ var generatedConfig = &cobra.Command{
...
+			}
+			// Distributed cache is read-only only; user must opt in explicitly.
+			if !optsGenCfg.readOnly {
+				return fmt.Errorf("--distributed-cache requires --ro to be set explicitly; distributed cache is supported only for read-only mounts")
+			}
+			config.Set("distributed_cache.discovery-endpoint", optsGenCfg.distributedCacheDiscoveryEndpoint)
+		} else if (!optsGenCfg.blockCache) && optsGenCfg.tmpPath == "" {
+			// tmp-path is required for file-cache mode.
 			return fmt.Errorf("temp path is required for file cache mode. Use flag --tmp-path to provide the path")
 		}
```
kapilupadhayay, line 91: Just check: 
 if we allow  --direct-io  + distributed-cache not rejected in gen-config, while  --block-cache  and  --tmp-path  are. direct-io bypasses caching entirely, so the combination is meaningless.

cmd/mount.go
MahashriSaini: Thank You Kapil for pointing this and suggesting fix.

cmd/mount.go
MahashriSaini: I agree that the recognition logic itself should not be duplicated.

cmd/gen-config.go
MahashriSaini: direct-io  bypasses the kernel cache and disables FUSE/attribute caching, but it does not bypass Blobfuse’s userspace  block_cache  or  distributed_cache  components.  gen-config  also supports  --block-cache --direct-io . 
 Here are few tests I ran to check direct-io with 3 caches-
 
Screenshot 2026-09-09 135552
Screenshot 2026-09-09 140031
Screenshot 2026-09-09 140056

syeleti-msft: Yes, direct-io is supported along with the caching modes present in blobfuse like block-cache, file-cache, now dist_cache.

setup/baseConfig.yaml
```diff
@@ -45,8 +45,11 @@ file_cache:
 
 # Distributed cache configuration (optional, for multi-node shared caching)
 # Requires a running Tachyon cache server cluster (https://github.com/Azure/Tachyon).
-# When dist_cache is configured, block_cache is auto-added as the L1 layer.
-dist_cache:
+# When distributed_cache is configured, block_cache is auto-added as the L1 layer.
```
syeleti-msft: > When distributed_cache is configured, block_cache is auto-added as the L1 layer.

I don't think user need to know regarding block_cache or L1 layer when using distributed cache. as it will create more confusion on block_cache related params. if you think the same, we can remove this sentence from the config.

sampleDistCacheConfig.yaml
```diff
@@ -1,24 +1,30 @@
 # Refer ./setup/baseConfig.yaml for full set of config parameters
 #
-# This sample shows dist_cache (distributed cache) configuration.
+# This sample shows distributed_cache (distributed cache) configuration.
 # The distributed cache provides a shared L2 cache between a local in-memory
 # L1 (block_cache) and Azure Blob Storage (L3), reducing Azure egress and
 # read latency for multi-node workloads.
 #
 # block_cache is added to the pipeline automatically as the L1 layer;
```
syeleti-msft: likewise https://github.com/Azure/azure-storage-fuse/pull/2316/changes#r3986448638

setup/baseConfig.yaml
MahashriSaini: I agree. Done

sampleDistCacheConfig.yaml
MahashriSaini: Done

syeleti-msft (approved):

PR #2326, opened 2026-08-21
Switch on data-locality and sesion API's by default

syeleti-msft:
[switch on data-locality and sesion API's by default](https://github.com/Azure/azure-storage-fuse/commit/9c6cf4c7a387474f59609be1d029f22a61429663) 

also pass endpoint for getBlob in options bag directly as the SDK supports it now, removed the old code for custom policy to do the same earlier.


component/azstorage/azstorage.go
```diff
@@ -62,6 +62,13 @@ type AzStorage struct {
 
 const compName = "azstorage"
 
+func defaultAzStorageOptions() AzStorageOptions {
+	return AzStorageOptions{
+		BlobLayoutAwareRouting: true,
+		UseSession:             true,
+	}
+}
```
souravgupta-msft, line 70: If we change the default to true, then will this cause breaking change for accounts where these features are not enabled?

souravgupta-msft (approved):

PR #2308, opened 2026-08-07
test: e2e tests and infra setup for dist_cache

nearora-msft:
## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Code quality improvement
- [ ] Other (describe):

## Description

- Adds a kind-based E2E framework for read-only distributed caching with blobfuse2 and a three-node Tachyon cluster.
- Tests L2 miss/population/hit behavior, concurrent-read stampede prevention, warm-cache access from late-joining pods, and Azure fallback after cache-node failure.
- Verifies data integrity using MD5, cache behavior through Prometheus metrics, and Azure requests through blobfuse2 logs.
- Adds deployment scripts, Kubernetes templates, isolated per-test deployments, failure-log artifacts, documentation, and an opt-in nightly Azure DevOps stage.

## How Has This Been Tested?

All E2E tests pass locally.

## Checklist

- [ ] The purpose of this PR is explained in this or a referenced issue.
- [ ] Tests are included and/or updated for code changes.
- [ ] Documentation update required.
- [ ] Updates to module CHANGELOG.md are included.
- [ ] License headers are included in each file.

## Related Links
- [Issues](<link>)

PR #2349, opened 2026-09-13
E2E Tests : CLI, stale/invalid, ranged reads

MahashriSaini:
Added few more E2E tests on top of [https://github.com/Azure/azure-storage-fuse/pull/2308#pullrequestreview-4917002105](url).

Tests include:
1. CLI tests
Supplies Azure settings via environment variables, and configures L2 entirely through  --distributed-cache-*  flags. It verifies the full Cobra → Viper → pipeline wiring with a cold read/populate, remount and warm L2 hit.
2. Stale/ Invalid tests
Covers out-of-band blob overwrite and deletion, both on the same live mount and after restart. Same-mount reads intentionally document current cached behavior stale v1 remains readable. After restart, an overwrite must return and repopulate v2 using the new ETag-keyed cache group, while a deleted blob must return  ENOENT rather than orphaned L2 data.
3. Ranged/ Partial Reads
Coverage includes exactly one populate for the last chunk, an L2 hit when rereading that chunk after remount, an independent miss when reading a different chunk and exactly two populates for a range crossing the final chunk boundary. End-of-file ranges make populate counts deterministic by preventing forward prefetch beyond EOF.

## Type of Change

- [ ] Bug fix
- [ ] New feature
- [ ] Other (describe):

## Description

- **Feature / Bug Fix**:

## Checklist

## Related Links
- [Issues](<link>)


cmd/mount.go
kapilupadhayay, line 316: Discovery configuration is treated as an authoritative distributed-cache signal by  normalizeDistCacheConfig , but  distributed_cache  is only added while synthesizing an empty components pipeline.
With an explicit pipeline that omits  distributed_cache , validation succeeds but neither  distributed_cache  nor its injected  block_cache  L1 is added. The mount therefore succeeds without enabling the requested cache.
Please either insert  distributed_cache  at the correct pipeline position or reject this configuration and require it to be listed explicitly. Please add a read-only test with an explicit compatible pipeline and a discovery endpoint.

test/dcache_e2e/staleness_test.go
```diff
@@ -0,0 +1,316 @@
...
+
+	// Confirm v2 landed in L2 under its new group.
+	waitAndAssertUploadSuccess(t, beforeV2, "restart-rewrite/v2-populate", 1)
+}
+
+// TestStaleness_Delete_SameMount_ServesCached documents the same-mount contract
+// for an out-of-band delete. Within the caches' TTLs the kernel serves the
+// pages it already has without asking FUSE, so the read succeeds with v1 even
+// though the blob no longer exists in Azure.
+func TestStaleness_Delete_SameMount_ServesCached(t *testing.T) {
```
kapilupadhayay, line 183: This test uses the best-effort cleanup helper for the deletion that establishes the test condition. Since  deleteBlob  only logs deletion errors, the test passes if the Azure deletion fails and the subsequent read returns the still-existing blob.
Please add a strict deletion helper that fails the test for required test actions, while retaining the best-effort helper for  t.Cleanup . Ideally, verify that the blob is absent through the Azure SDK before reading through the mount.

test/dcache_e2e/cli_flags_test.go
```diff
@@ -0,0 +1,60 @@
...
+   LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+   OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+   SOFTWARE
+*/
+
+package dcache_e2e
+
+import "testing"
+
+// TestCLIFlags_ReadPath_NoYAML runs the same L2 miss -> populate -> remount
```
kapilupadhayay, line 41: This test proves that the CLI discovery endpoint enables distributed caching, but it does not prove that the tuning flags are correctly bound and consumed.
Please use distinguishable non-default values and verify the effective startup configuration from BlobFuse logs. At minimum, use a non-default block size and assert the corresponding number of populated chunks. Memory, TTL, prefetch and parallelism should also be checked against the configuration logged by  DistCache::Configure  and  BlockCache::Configure .

kapilupadhayay (changes requested):

test/dcache_e2e/staleness_test.go
MahashriSaini: Done

test/dcache_e2e/cli_flags_test.go
MahashriSaini: Done

